PR #28, opened 2025-01-22
[Doc] Create a workflow to host docs using GitHub Pages.

xwhzz:


LeiWang1999:
thanks @xwhzz for the workflow, but I think the deprecated docs should be removed as content in those docs are legacy

LeiWang1999:
Merged, thanks @xwhzz

Issue #36, opened 2025-01-23
[Bug] Potential Issue in LoopVectorize when encounter Insufficient Buffer Liveness Chain


LeiWang1999:
ref to issue #35 , and a simple test case to reproduce is:

```python
from tilelang import tvm as tvm
from tvm.script import ir as I
from tvm.script import tir as T

@I.ir_module
class Module:
    @T.prim_func
    def main(Q: T.Buffer((1, 4096, 32, 128), "float16")):
        T.func_attr({"target": T.target({"arch": "sm_89", "host": {"keys": ["cpu"], "kind": "llvm", "mtriple": "x86_64-unknown-linux-gnu", "tag": ""}, "keys": ["cuda", "gpu"], "kind": "cuda", "max_num_threads": 1024, "tag": "", "thread_warp_size": 32})})
        by = T.launch_thread("blockIdx.y", 32)
        v = T.launch_thread("threadIdx.x", 256)
        if v < 16:
            i = T.int32()
            T.attr(i, "pragma_unroll_explicit", T.bool(False))
            for i in T.vectorized(8):
                Q_shared = T.allocate([16384], "float16", "shared.dyn")
                Q_shared_1 = T.Buffer((16384,), "float16", data=Q_shared, scope="shared.dyn")
                Q_1 = T.Buffer((16777216,), "float16", data=Q.data)
                Q_shared_1[v * 8 + i] = Q_1[by * 128 + v * 8 + i]

mod = Module
from tvm import tir
transformed_mod = tir.transform.VectorizeLoop()(mod)

print(transformed_mod)
```

LeiWang1999:
closed as has been resolved :)

Issue #35, opened 2025-01-23
Copy issue when tensor dim is 1


DD-DuDa:
Assume we have Q tensor shape with [bs, 1, head, dim].
And we allocate a shared memory Q_shared [block_M, dim].

how to copy Q_shared[0, :] = Q[bid, 0, hid, :]?

```
# type: ignore

import torch
import torch.nn.functional as F
import tilelang
from tilelang import Profiler
from tilelang.autotuner import *
import tilelang.language as T
import itertools
import argparse
from functools import partial

def flashdecoding(batch, heads, seqlen_q, seqlen_kv, dim, is_casual, num_split, tune=False):
    scale = (1.0 / dim) ** 0.5 * 1.44269504  # log2(e)
    shape_q = [batch, seqlen_q, heads, dim]
    shape_kv = [batch, seqlen_kv, heads, dim]
    part_shape = [batch, seqlen_q, heads, num_split, dim]
    dtype = "float16"
    accum_dtype = "float"

    def kernel_func(block_M, block_N):
        
        @T.macro
        def flash_attn_split(
            Q: T.Buffer(shape_q, dtype),
            K: T.Buffer(shape_kv, dtype),
            V: T.Buffer(shape_kv, dtype),
            glse: T.Buffer([batch, heads, num_split, seqlen_q], dtype),
            Output_partial: T.Buffer(part_shape, dtype),
        ):
            print("flash_attn_split")
            with T.Kernel(T.ceildiv(seqlen_q, block_M), heads * batch, num_split, threads=128 * 2) as (bx, by, bz):
                Q_shared = T.alloc_shared([block_M, dim], dtype)
                K_shared = T.alloc_shared([block_N, dim], dtype)
                V_shared = T.alloc_shared([block_N, dim], dtype)
                O_shared = T.alloc_shared([block_M, dim], dtype)
                acc_s = T.alloc_fragment([block_M, block_N], accum_dtype)
                acc_s_cast = T.alloc_fragment([block_M, block_N], dtype)
                acc_o = T.alloc_fragment([block_M, dim], accum_dtype)
                scores_max = T.alloc_fragment([block_M], accum_dtype)
                scores_max_prev = T.alloc_fragment([block_M], accum_dtype)
                scores_scale = T.alloc_fragment([block_M], accum_dtype)
                scores_sum = T.alloc_fragment([block_M], accum_dtype)
                logsum = T.alloc_fragment([block_M], accum_dtype)

                mid = bx
[...]

PR #37, opened 2025-01-23
[Bugfix] Reorder Passes: Place Vectorize Loop Before StorageFlatten and FlattenBuffer to Prevent Redundant Allocations

LeiWang1999:
When intermediate buffers are used only once (e.g., `Q_shared` in the code below), the storage flatten pass can incorrectly lower the `allocate` statement into the `for` loop. For example:

```python
def kernel_func(block_M, block_N):
        
    @T.prim_func
    def main(
        Q: T.Buffer(shape_q, dtype),
    ):
        with T.Kernel(T.ceildiv(seqlen_q, block_M), heads * batch, num_split, threads=128 * 2) as (bx, by, bz):
            Q_shared = T.alloc_shared([block_M, dim], dtype)

            hid = by % heads
            bid = by // heads

            T.copy(Q[bid, 0, hid, :], Q_shared[0, :])
            # Alternatively:
            # for d in T.serial(dim):
            #     Q_shared[0, d] = Q[bid, 0, hid, d]

    return main
```

After the **storage flatten pass**, the `allocate` statement is incorrectly lowered into the loop, as seen in this output:

```python
i = T.int32()
T.attr(i, "pragma_unroll_explicit", T.bool(False))
for i in T.vectorized(8):
    Q_shared = T.decl_buffer((128, 128), "float16", scope="shared.dyn")
    Q_shared[0, v * 8 + i] = Q[0, 0, by, v * 8 + i]
```

This behavior is problematic because it causes the `vectorize` pass to attempt vectorization on the `allocate` node. Instead, we want the `allocate` statement (`T.decl_buffer`) to be moved outside the loop, as shown below:

```python
Q_shared = T.alloc_buffer((128, 128), "float16", scope="shared.dyn")
if v < 16:
    for i in T.vectorized(8, annotations={"pragma_unroll_explicit": T.bool(False)}):
        Q_shared[0, v * 8 + i] = Q[0, 0, by, v * 8 + i]
```

This pull request reorders the pass sequence, applying the **vectorize schedule** before the **storage flatten** and **buffer flatten** passes. By doing so, the `allocate` statement remains outside the loop, preventing unintended vectorization behavior and ensuring correct code generation.


LeiWang1999:
This is a fix for issue #35 and #36

PR #154, opened 2025-03-06
Add cpu jit with backend ctypes

xs-keju:
A simple JIT compilation for a CPU has been implemented, with ctypes as the backend. It successfully passes the matmul operator's compile test using the command:
**tilelang.compile(cpu_func, -1, execution_backend="ctypes", target="c")**

### DIFF
--- a/src/tl_templates/cpp/common.h
+++ b/src/tl_templates/cpp/common.h
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+// Copyright (c) Tile-AI Corporation.
 // Licensed under the MIT License.
 #pragma once
 
--- a/src/tl_templates/cpp/common.h
+++ b/src/tl_templates/cpp/common.h
@@ -1,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+// #include <arm_neon.h>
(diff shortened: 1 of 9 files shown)


src/tl_templates/cpp/common.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+// #include <arm_neon.h>
+#pragma once
```
LeiWang1999: Why do we need two `#pragma once` directives there? I think // #include <arm_neon.h> should be removed as well.

testing/python/cpu/test_tilelang_cpu_gemm.py
LeiWang1999, line 1: I think we should keep the original version of the matmul kernel and create a separate, simpler kernel for JIT. The current kernel is a bit messy, so we should focus on improving it and fixing any bugs in future.

LeiWang1999:
CC @xs-keju , I left some comments

LeiWang1999:
Please fix the lint @xs-keju :)

xxw-keju:
> Please fix the lint @xs-keju :)
Okay, it's done.

LeiWang1999 (approved):

PR #165, opened 2025-03-07
Use SS-GEMM for PV in mla

YouJiacheng:
It's slightly faster than T.copy then RS-GEMM, and simpler.


LeiWang1999:
CC @chengyupku If you have time.

chengyupku:
LGTM, Merged

Issue #143, opened 2025-03-05
[Feature Request] Upgrade cutlass to v3.5.0+ for fp8 `T.gemm`


LeiWang1999:
Now we only support fp8 with mma instructions instead of `T.gemm`, because the cutlass version of tilelang is a bit oudated, we should update the cutlass version in future.

Test case:

```python
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

import tilelang
from tilelang import Profiler
import tilelang.language as T


def matmul(M, N, K, block_M, block_N, block_K, dtype="e5m2_float8", accum_dtype="float"):

    @T.prim_func
    def main(
            A: T.Buffer((M, K), dtype),
            B: T.Buffer((K, N), dtype),
            C: T.Buffer((M, N), dtype),
    ):
        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128) as (bx, by):
            A_shared = T.alloc_shared((block_M, block_K), dtype)
            B_shared = T.alloc_shared((block_K, block_N), dtype)
            C_local = T.alloc_fragment((block_M, block_N), accum_dtype)

            T.clear(C_local)
            for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=3):
                T.copy(A[by * block_M, k * block_K], A_shared)
                T.copy(B[k * block_K, bx * block_N], B_shared)
                T.gemm(A_shared, B_shared, C_local)

            T.copy(C_local, C[by * block_M, bx * block_N])

    return main


func = matmul(1024, 1024, 1024, 128, 128, 32)

print(func)

rt_mod, params = tilelang.lower(func)

profiler = Profiler(rt_mod, params, result_idx=[2])

import torch

a = torch.randn(1024, 1024).cuda().half()
b = torch.randn(1024, 1024).cuda().half()

c = profiler(a, b)

ref_c = a @ b

print(c)
print(ref_c)

torch.testing.assert_close(c, ref_c, rtol=1e-2, atol=1e-2)

# Get CUDA Source
print(rt_mod.imported_modules[0].get_source())
```

LeiWang1999:
Need a cutlass expert to help us upgrade :)

jeromeku:
@LeiWang1999 

I'm fairly familiar with Cutlass abstractions (e.g., CuTE), though by no means an expert.

Eager to help with this -- can you briefly outline how tilelang lowers to Cutlass and / or point to a few examples of such code?

Thanks!

[...]

PR #181, opened 2025-03-10
[Examples] Implement DeepGEMM

zcnrex:
Modified the example fp8 gemm to be equivalent to [DeepGEMM](https://github.com/deepseek-ai/DeepGEMM/blob/main/deep_gemm/include/deep_gemm/fp8_gemm.cuh), which has scales.


LeiWang1999:
Thanks for your contribution @zcnrex , but for deepseek deepgemm example, I think we should create another directory under examples.

zcnrex:
@LeiWang1999 Thanks for the feedback! Will create another directory. Is this the right direction for implementation?

examples/gemm_fp8/example_tilelang_gemm_fp8_intrinsic.py
```diff
@@ -75,7 +75,9 @@ def tl_matmul(
     block_K = chunk
 
     A_shape = (M, K)
+    Scales_A_shape = (M, T.cdiv(K / block_K))
```
hebiao064: now block_K is a hardcode number, but we should use the same param of `num_groups` from  `per_token_group_quant` kernel

@LeiWang1999 Is it correct?

LeiWang1999: true dude :)

zcnrex: Updated `Scales_A_shape = (M, T.ceildiv(K, block_K) * num_groups)`

LeiWang1999:
Thanks @zcnrex for your contributions! I've left some comments, but I think it will be much better if we can still use `T.gemm` for this case, would love to see if you have interest in working on this task #143

zcnrex:
Thanks @LeiWang1999 ! Couldn't see the comments, are they posted?

Sorry not very familiar with cutlass, can't help with the issue..

LeiWang1999:
nevermind, thanks for your contribution @zcnrex :)

LeiWang1999:
Thank you for your contribution! I noticed that a later PR #217  has made similar improvements. Since it covers the same updates, I’ll go ahead and close this PR. Looking forward to your future contributions!

PR #188, opened 2025-03-11
[Dev][Bugfix] Add RMS Normalization Kernels and Fix Reduce Bug

chengyupku:
- Implement two RMS normalization implementations in TileLang:
  * `rms_norm_splitk`: Split-K reduction approach for large matrices
  * `rms_norm`: Full reduction kernel with simplified implementation
- Add reference implementation using PyTorch for validation
- Include performance benchmarking for both kernel variants
- Demonstrate flexible block size and matrix size configurations

### DIFF
--- a/examples/norm/rms_norm.py
+++ b/examples/norm/rms_norm.py
@@ -1,3 +1,6 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
+
 import torch
 import tilelang
 import tilelang.language as T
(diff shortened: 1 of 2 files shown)


LeiWang1999 (commented):
We should add license to this file.

examples/norm/rms_norm.py
```diff
@@ -0,0 +1,80 @@
+import torch
```
LeiWang1999, line 4: license required.

examples/norm/rms_norm.py
```diff
@@ -0,0 +1,80 @@
...
+if __name__ == "__main__":
+    M, N, blk_m, blk_k = 8192, 8192, 1, 512
+    # program = rms_norm_splitk(M, N, blk_m, blk_k)
+    program = rms_norm(M, N, blk_m)
+    kernel = tilelang.compile(
+        program,
+        out_idx=-1,
+        target="cuda",
+        execution_backend="cython",
+        pass_configs={"tl.disable_tma_lower": True})
```
LeiWang1999: Why we need `"tl.disable_tma_lower"` ?

examples/norm/rms_norm.py
```diff
@@ -0,0 +1,80 @@
...
+    return main
+
+
+def ref_program(x):
+    return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + 1e-12)
+
+
+if __name__ == "__main__":
+    M, N, blk_m, blk_k = 8192, 8192, 1, 512
+    # program = rms_norm_splitk(M, N, blk_m, blk_k)
```
LeiWang1999: comments should be removed :)

chengyupku: OK, done.

examples/norm/rms_norm.py
chengyupku: Looks unnecessary, removed.

examples/norm/rms_norm.py
chengyupku: done

PR #192, opened 2025-03-12
[Language] Support clamp in language

hyx1999:
Added the clamp function to T.language

### DIFF
--- a/tilelang/language/customize.py
+++ b/tilelang/language/customize.py
@@ -15,3 +15,15 @@ def atomic_addx2(dst, value):
 
 def dp4a(A, B, C):
     return T.call_extern("handle", "DP4A", T.address_of(A), T.address_of(B), T.address_of(C))
+
+def clamp(dst, min=None, max=None):
+    if min is None and max is None:
+        raise ValueError("min and max can't both be None.")
+    elif min is not None and max is None:
+        return T.max(dst, min)
+    elif min is None and max is not None:
+        return T.min(dst, max)
+    else:
+        if min > max:
+            raise ValueError("min must be less than or equal to max.")
+        return T.min(T.max(dst, min), max)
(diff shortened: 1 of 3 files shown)


tilelang/language/customize.py
```diff
@@ -15,3 +15,15 @@ def atomic_addx2(dst, value):
...
+    if min is None and max is None:
+        raise ValueError("min and max can't both be None.")
+    elif min is not None and max is None:
+        return T.max(dst, min)
+    elif min is None and max is not None:
+        return T.min(dst, max)
+    else:
+        if min > max:
+            raise ValueError("min must be less than or equal to max.")
+        return T.min(T.max(dst, min), max)
```
LeiWang1999, line 29: would be better to add typing
```suggestion
 def clamp(dst, min_val:Optional[PrimExpr]=None, max_val:Optional[PrimExpr]=None):
    if min_val is None and max_val is None:
        raise ValueError("min_val and max_val can't both be None.")

    if min_val is not None and max_val is not None and min_val > max_val:
        raise ValueError("min_val must be less than or equal to max_val.")

    if isinstance(dst, (torch.Tensor, np.ndarray)):
        if isinstance(dst, torch.Tensor):
            return dst.clamp(min=min_val, max=max_val)
        return np.clip(dst, min_val, max_val)

    # Fallback for scalars
    if min_val is not None:
        dst = max(dst, min_val)
    if max_val is not None:
        dst = min(dst, max_val)
    return dst
```

LeiWang1999 (changes requested):

LeiWang1999:
Thanks @hyx1999 for your contribution! I left some comments and I thought that was some linting issues. You can execute `format.sh` to detect and fix the linting issues.

LeiWang1999:
@hyx1999 We should resolve the linting, execute `format.sh` would help, and you can install necessary packages via `pip install -r requirements-lint.txt`

LeiWang1999:
lgtm, merged :)

PR #196, opened 2025-03-12
Update expired example code.

66RING:
Expired example code, update readme.


LeiWang1999:
LGTM, Thanks for your contributon :) @66RING

Issue #101, opened 2025-02-21
[Bug] compilation error for mma on NVIDIA Hopper GPU


smallscientist1:
Hi, I am running matmul with `blockM=64, blockN=192, threads=128*2, policy=GemmWarpPolicy.FullCol`, but it encounters compilation error on NVIDIA H100. For each warpgroup(128 threads), the mma size for this warpgroup is `M64,N96,K32` , which should be supported because it is a multiple of ptx instruction `wgmma.mma_async.m64n16k16`. 
```
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

import tilelang
from tilelang import Profiler
import tilelang.language as T


def matmul(M, N, K, block_M, block_N, block_K, dtype="float16", accum_dtype="float"):

    @T.prim_func
    def main(
            A: T.Buffer((M, K), dtype),
            B: T.Buffer((K, N), dtype),
            C: T.Buffer((M, N), dtype),
    ):
        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128*2) as (bx, by):
            A_shared = T.alloc_shared((block_M, block_K), dtype)
            B_shared = T.alloc_shared((block_K, block_N), dtype)
            C_local = T.alloc_fragment((block_M, block_N), accum_dtype)

            T.clear(C_local)
            for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=3):
                T.copy(A[by * block_M, k * block_K], A_shared)
                T.copy(B[k * block_K, bx * block_N], B_shared)
                T.gemm(A_shared, B_shared, C_local, policy=T.GemmWarpPolicy.FullCol)

            T.copy(C_local, C[by * block_M, bx * block_N])

    return main


func = matmul(1024, 192, 1024, 64, 192, 32)

print(func)

rt_mod, params = tilelang.lower(func)

profiler = Profiler(rt_mod, params, result_idx=[2])

import torch

a = torch.randn(1024, 1024).cuda().half()
b = torch.randn(1024, 192).cuda().half()

c = profiler(a, b)

ref_c = a @ b

print(c)
print(ref_c)

torch.testing.assert_close(c, ref_c, rtol=1e-2, atol=1e-2)

# Get CUDA Source
print(rt_mod.imported_modules[0].get_source())

```
The compilation error is:
```
Compilation error:
[...]

PR #224, opened 2025-03-16
[Bugfix] Fix mismatch of shared memory layout and mma atom on Hopper

zqh-wz:
This pull request fixes issue #101.

The issue comes from the mismatch between the smem layout and the mma atom layout in this example.
Since the user specifies `GemmWarpPolicy.FullCol`, `num_warp_n` now equals 2 because of the 2 warp groups in total (256 threads).
This results in the selected mma atom being `wgmma.mma_async.m64n96k16`. However, the selection of the shared memory layout is not aware of `num_warp_n`, producing a 128B (64 x fp16) swizzle mode. As 64 cannot be divided by 96 from the mma atom, the issue occurs.

To fix this, we should also select the shared memory's layout with regard to the warp policy (see comment below).
We also need to adjust the code of layout inference correspondingly. We add a `continuity` parameter in `makeGemmABLayout` because now the matrix shape cannot be directly used to determine the swizzle mode.

We only fix the issue for sm90 code to avoid causing other issues not considered. However, we suspect the same problem also exists in sm80/70 codes.

### DIFF
--- a/src/tl_templates/cuda/gemm_sm90.h
+++ b/src/tl_templates/cuda/gemm_sm90.h
@@ -36,9 +36,11 @@ public:
       trans_B ? GMMA::Major::K : GMMA::Major::MN;
 
   using SmemLayoutAtomA =
-      decltype(ss_smem_selector<GmmaMajorA, A_type, Int<M>, Int<K>>());
+      decltype(ss_smem_selector<GmmaMajorA, A_type, Int<M / (num_warp_m / 4)>,
+                                Int<K>>());
   using SmemLayoutAtomB =
-      decltype(ss_smem_selector<GmmaMajorB, B_type, Int<N>, Int<K>>());
+      decltype(ss_smem_selector<GmmaMajorB, B_type, Int<N / num_warp_n>,
+                                Int<K>>());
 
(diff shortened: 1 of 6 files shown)


src/tl_templates/cuda/gemm_sm90.h
```diff
@@ -92,9 +36,11 @@ class GemmTensorOp {
       trans_B ? GMMA::Major::K : GMMA::Major::MN;
 
   using SmemLayoutAtomA =
-      decltype(ss_smem_selector<GmmaMajorA, A_type, Int<M>, Int<K>>());
+      decltype(ss_smem_selector<GmmaMajorA, A_type, Int<M / (num_warp_m / 4)>,
+                                Int<K>>());
   using SmemLayoutAtomB =
-      decltype(ss_smem_selector<GmmaMajorB, B_type, Int<N>, Int<K>>());
```
zqh-wz, line 97: This line is the source of the problem.

LeiWang1999: Awesome :)

LeiWang1999 (approved):
LGTM! Let’s merge the PR once the tests pass. :)

PR #239, opened 2025-03-17
[Refactor] Align torch_assert_close tensor comparison with torch.testing.assert_close

xwhzz:


tilelang/testing/__init__.py
```diff
@@ -8,98 +8,15 @@
 import numpy as np
 from tvm.testing.utils import *
 
+from ..utils.tensor import torch_assert_close as torch_assert_close
```
LeiWang1999:
```suggestion
from tilelang.utils.tensor import torch_assert_close as torch_assert_close
```

LeiWang1999 (approved):

PR #245, opened 2025-03-18
[BugFix] Fix bug of mismatching dtype in testing

xwhzz:


testing/python/jit/test_tilelang_jit_gemm_ctypes.py
```diff
@@ -382,7 +382,8 @@ def run_ctypes_dynamic_shape(M,
 
     matmul_kernel(tensor_a, tensor_b, tensor_c)
 
-    tensor_ref_c = torch.matmul(tensor_a.to(torch.float), tensor_b.to(torch.float))
+    tensor_ref_c = torch.matmul(tensor_a.to(torch.float),
+                                tensor_b.to(torch.float)).to(torch.__getattribute__(out_dtype))
```
LeiWang1999: would be great if we can use `map_torch_dtype` from tilelang.utils , that will be more general and precise.

LeiWang1999 (changes requested):

LeiWang1999 (approved):

PR #246, opened 2025-03-19
[Enhancement] Add zero initialization option to GEMM operations

chengyupku:
- Introduced a new `zero_init` parameter to the GEMM function, allowing for optional zero initialization of the accumulator.
- Updated the GEMM implementation across various CUDA architectures to support the new parameter.
- Modified the Python interface for GEMM to include the `zero_init` argument, enhancing flexibility in kernel execution.
- Ensured compatibility with existing functionality while improving initialization control for performance optimization.


LeiWang1999 (commented):
Would be great to rename `zero_init` into `clear_accum` or `keep_accum`?

chengyupku:
> Would be great to rename `zero_init` into `clear_accum` or `keep_accum`?

ok

PR #248, opened 2025-03-19
[Enhancement][CUDA] Avoid C7508 for CUDA backend via assigning default value to `minBlocksPerMultiprocesor `

cherichy:
We should set the second parametr minBlockperSM for __launch_bounds__ to avoid C7508, just set it to 1 to make compiler happy.


LeiWang1999:
LGTM, as cutlass always set 1 to `minBlocksPerMultiprocesor`.

LeiWang1999:
Merged, as likely the failing CI was not caused by this PR.

PR #299, opened 2025-03-28
[Bugfix] Correct method call for block reduction check when analyzing memory footprint

NaOHCC:
The `is_reduction` method was not correctly invoked, resulting in the buffer cache eligibility check always passing, which may lead to unnecessary cache statistics.


LeiWang1999:
Thanks for the fix, Merged

PR #300, opened 2025-03-28
[Bugfix] Resolve autotuner bugs for blocksparse GEMM example

tth37:
This PR addresses several issues identified in the `tilelang.autotuner` module, primarily discovered while running `example_blocksparse_gemm.py`.

### Key Fixes:
- **Logging Configuration:** Reworked the autotuner's logging setup to correctly route `DEBUG` messages to `autotuner.log` and `INFO+` messages to the console, resolving problems where debug logs were missed.
- **Boolean Tensor Generation:** Corrected `get_tensor_supply` to properly generate tensors with `torch.bool` dtype, avoiding out-of-bounds errors.
- **Input Tensor Regeneration:** Modified the autotuner to always regenerate JIT input tensors for each configuration trial, preventing errors caused by reusing potentially incompatible cached inputs from previous trials.
- **Example Update:** Updated `example_blocksparse_gemm.py` to align with these autotuner fixes.

### Limitations and Possible Enhancements:
- **Sparsity Control (Important):** Currently the block mask generation for sparsity is random (~50%). An enhancement could allow passing a *custom tensor generator* to the autotuner to control sparsity during tuning.
- **Config Access:** The `AutotuneResult.config` should be improved to support dictionary-style key access (e.g., `result.config['block_M']`) for better usability.
- **Testing `get_tensor_supply`:** Should we add a test case for `get_tensor_supply` for various dtypes and supply types?

### DIFF
--- a/tilelang/autotuner/__init__.py
+++ b/tilelang/autotuner/__init__.py
@@ -186,9 +186,7 @@ class AutoTuner:
             atol = jit_context.atol
             max_mismatched_ratio = jit_context.max_mismatched_ratio
 
-            self.jit_input_tensors = profiler._get_inputs(
-                with_output=profiler ==
-                "tvm") if self.jit_input_tensors is None else self.jit_input_tensors
+            self.jit_input_tensors = profiler._get_inputs(with_output=(profiler == "tvm"))
 
             if (not skip_check) and (ref_prog is not None):
                 profiler.assert_allclose(
--- a/tilelang/autotuner/__init__.py
+++ b/tilelang/autotuner/__init__.py
@@ -318,6 +318,13 @@ class AutoTuner:
             tqdm.write(f"Tuned Latency {latency} with config {config} at index {i}")
 
         pool.shutdown()
+
+        if best_jit_context is None:
+            error_msg = ("Auto-tuning failed: No configuration successfully "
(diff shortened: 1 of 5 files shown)


LeiWang1999:
Thanks @yyttt6 , and would you mind help review this pull request?

tilelang/autotuner/__init__.py
```diff
@@ -174,9 +186,7 @@ def target_fn(jit_context):
             atol = jit_context.atol
             max_mismatched_ratio = jit_context.max_mismatched_ratio
 
-            self.jit_input_tensors = profiler._get_inputs(
-                with_output=profiler ==
-                "tvm") if self.jit_input_tensors is None else self.jit_input_tensors
+            self.jit_input_tensors = profiler._get_inputs(with_output=(profiler == "tvm"))
```
LeiWang1999: I think we should keep the original logic, because otherwise, we’d have to recreate the tensor every time we run the benchmark, which would be quite costly.

tth37: I've updated the logic in latest commit, using `cache_input_tensors` flag (default True) to control input tensor caching. For those kernels where input_tensors' shape vary between configurations, we have to compromise.

LeiWang1999:
LGTM, thanks for your contributions!

Issue #303, opened 2025-03-30
Encounter Error When Running FP8 GEMM on 4090


shadowpa0327:
## Problem Description
When running the FP8 GEMM example on an RTX 4090, compilation fails with incomplete-type errors related to gemm_sm80.h.

## Reproduce
+ Environments
    + `tilelang`: `0.1.3+8a5daad.ubuntu.18.4.cu121`
    + `g++`: `12.1.0`
    + `nvcc`: `12.1`

+ Install nightly build. 
    ```
    python -m pip install tilelang -f https://tile-ai.github.io/whl/nightly/cu121/
    ```
+ Run fp8 example
    ```
    python examples/gemm_fp8/example_tilelang_gemm_fp8.py 
    ```

## Detail Error Message
```
/home/cc2869/.conda/envs/tilelang/lib/python3.10/site-packages/tilelang/src/tl_templates/cuda/gemm_sm80.h(210): error: incomplete type is not allowed
    using TileMma = TiledMMA<typename Instruction::MMA,
                                      ^
          detected during:
            instantiation of class "cute::GemmTensorOp<M, N, K, num_warp_m, num_warp_n, trans_A, trans_B, clear_accum, A_type_raw, B_type_raw, C_type_raw> [with M=128, N=128, K=64, num_warp_m=2, num_warp_n=2, trans_A=false, trans_B=true, clear_accum=false, A_type_raw=fp8_e4_t, B_type_raw=fp8_e4_t, C_type_raw=float]" at line 352
            instantiation of "void tl::gemm_ss<M,N,K,num_warp_m,num_warp_n,trans_A,trans_B,clear_accum,A_type,B_type,C_type>(A_type *, B_type *, C_type *) [with M=128, N=128, K=64, num_warp_m=2, num_warp_n=2, trans_A=false, trans_B=true, clear_accum=false, A_type=fp8_e4_t, B_type=fp8_e4_t, C_type=float]" at line 48 of /tmp/tmpm6ijak8t.cu

/home/cc2869/.conda/envs/tilelang/lib/python3.10/site-packages/tilelang/src/tl_templates/cuda/gemm_sm80.h(212): error: incomplete type is not allowed
                             typename Instruction::MMA_Group>;
                                      ^
          detected during:
[...]

PR #304, opened 2025-03-30
[Enhancement] Add support for CUDA architecture 8.9 in GEMM template

LeiWang1999:
- Introduced conditional inclusion of "gemm_sm89.h" for CUDA architectures 8.9 and above, enhancing compatibility with newer hardware.
- This change ensures that the GEMM template can leverage optimizations specific to the 8.9 architecture, improving performance for users with compatible GPUs.


LeiWang1999:
Also a fix for issue #303

PR #305, opened 2025-03-30
[Bugfix] Replace profiler.mod with profiler.adapter to fix AttributeError

LeslinD:
The use of profiler.mod in the example scripts elementwise_add.py and rms_norm.py causes an error when passed to profiler.do_bench(): `AttributeError: 'Profiler' object has no attribute 'mod'`

Replaced profiler.mod with profiler.adapter, which correctly references the backend module for benchmarking. This change resolves the error and allows the examples to run as expected.


LeiWang1999:
LGTM. I removed the target functions since kernel.adapter should be used when the position is set to null.

Issue #290, opened 2025-03-26
[BUG]  Incorrect result when using inputs with dynamic shapes


chengyupku:
You can find the faulty kernel for reproducing the issue [here](https://github.com/xiayuqing0622/tilelang/blob/5ce04c9478f745f5187af03c37d690af7c896310/examples/flash_decoding/example_sparse_gqa_decode_varlen_indice.py).
Issue: Incorrect result when using dynamic input.
I’ve identified that the error originates from the following tilelang statement: 
```
if k == 0:
    for i, j in T.Parallel(block_H, block_N):
        acc_s[i, j] = T.if_then_else(i_s * block_N + j >= cache_seqlens[bid], -T.infinity(accum_dtype), acc_s[i, j])
```
It seems that the error occurs during vectorization when handling dynamic input.

chengyupku:
Could you help with this issue? @LeiWang1999 @tzj-fxz

LeiWang1999:
Do we need a Hopper device to reproduce this issue?

chengyupku:
> Do we need a Hopper device to reproduce this issue?

I think we can reproduce on 4090 or A100.

tzj-fxz:
> You can find the faulty kernel for reproducing the issue [here](https://github.com/xiayuqing0622/tilelang/blob/5ce04c9478f745f5187af03c37d690af7c896310/examples/flash_decoding/example_sparse_gqa_decode_varlen_indice.py). Issue: Incorrect result when using dynamic input. I’ve identified that the error originates from the following tilelang statement:
> 
> ```
> if k == 0:
>     for i, j in T.Parallel(block_H, block_N):
>         acc_s[i, j] = T.if_then_else(i_s * block_N + j >= cache_seqlens[bid], -T.infinity(accum_dtype), acc_s[i, j])
> ```
> 
> It seems that the error occurs during vectorization when handling dynamic input.

I explored the generated CUDA code and found that the bug may be in the `VectorizeDynamicCallRemover` pass. This pass will eliminate the dynamic bound check when vectorizing data movement and make `IfThenElse` get some wrong elements.

Adding this pass before is because TVM does not expect us to vectorize the `IfThenElse` statements. The related code is in `vectorize_loop.cc`:

```C++
  PrimExpr MutateIfThenElseExpr_(const CallNode* op) {
    PrimExpr cond = this->VisitExpr(op->args[0]);
    if (cond.dtype().is_scalable_or_fixed_length_vector()) {
[...]

PR #326, opened 2025-04-03
[Dynamic Symbolic] Adaptively vectorize with different condition expressions

tzj-fxz:
[Dynamic Symbolic] Adaptively vectorize with different condition expressions


LeiWang1999:
CC @xiayuqing0622 , also for #290

LeiWang1999 (approved):

PR #361, opened 2025-04-09
[Example] Handle Scenarios in Which a Threadblock is Assigned Only Invalid Block Indices for Sparse Attention 

xiayuqing0622:


LeiWang1999 (changes requested):
LGTM

LeiWang1999 (approved):

PR #368, opened 2025-04-10
[Bugfix] Adjust Autotuner threadpool `max_workers` limit to available CPUs

tth37:
Currently the autotuner calculates the `max_workers` for its `ThreadPoolExecutor` (used for concurrent kernel compilation) based on `os.cpu_count()`.

However, `os.cpu_count()` reflects the total number of CPUs available on the *system*, which doesn't necessarily reflect the number of cores the *current process* is actually allowed to use. The mismatch is common in containerized environments or when CPU affinity is set using tools like `taskset`.

Attempting to use more worker threads than available CPUs can lead to system thrashing and slow down the compilation process rather than accelerating it.

I think a better approach is to use `len(os.sched_getaffinity(0))` to get the number of CPUs that the process is actually available to use ([reference](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count)).

### DIFF
--- a/examples/blocksparse_gemm/example_blocksparse_gemm.py
+++ b/examples/blocksparse_gemm/example_blocksparse_gemm.py
@@ -7,7 +7,7 @@ import tilelang
 import tilelang.language as T
 from tilelang.autotuner import AutoTuner
 from tilelang.engine.param import KernelParam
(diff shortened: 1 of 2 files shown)


LeiWang1999:
LGTM, Merged

PR #437, opened 2025-04-25
[Enhancement] Reduce CPU overhead during kernel execution

Cunxiao2002:


LeiWang1999:
we should fix the test `language/test_tilelang_language_ptr.py`, to resolve other failed tests should merge the latest code.

PR #459, opened 2025-05-08
docs: add llvm version info to installation.md.

AsakusaRinne:
When I built tile-lang from source, I found that compiling issues will raise if I use LLVM-20 due to the version of tvm submodule. Finally I fixed it by using lower version LLVM, therefore I'm adding a small hint to the doc in this PR.

Below is one of the errors:

```
/mnt/<user>/code/community/tilelang/3rdparty/tvm/src/target/llvm/codegen_amdgpu.cc:274:9: error: invalid use of incomplete type 'class llvm::Module'
  274 |     mlib->setTargetTriple(llvm_target->GetTargetTriple());
```

### DIFF
--- a/docs/get_started/Installation.md
+++ b/docs/get_started/Installation.md
@@ -39,6 +39,7 @@ python -c "import tilelang; print(tilelang.__version__)"
 - **Operating System**: Linux
 - **Python Version**: >= 3.7
 - **CUDA Version**: >= 10.0
+- **LLVM**: < 20 if you are using the bundled TVM submodule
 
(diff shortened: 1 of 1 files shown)


LeiWang1999:
Thanks @AsakusaRinne , but this is strange because we’ve had llvm disabled for quite some time.

SanftMonster:
@LeiWang1999 When I followed the instructions to build, I found that LLVM is used by setting the cmake option: https://github.com/tile-ai/tilelang/blob/main/docs/get_started/Installation.md?plain=1#L136. Does this DOC have been outdated?

LeiWang1999:
yes @AsakusaRinne , thanks for your report, now de can set the USE_LLVM(OFF)

PR #461, opened 2025-05-09
feat: add nvrtc execution backend

lucifer1004:


LeiWang1999:
likely all test passed, I'll dive deep to benchmark this backend in my device later. Thanks @lucifer1004 for your huge contribution!

LeiWang1999:
Hi @lucifer1004, I've reviewed the proposal and overall LGTM. While nvrtc is promising, it will likely face GIL-related limitations in mainstream cuda environments (<=12.9) that prevent thread parallelization during autotuning. Therefore, I think we should default to Cython for now and enable it as a default backend when we used proc parallel or cuda toolkits >= 13.0.

If users prioritize faster compilation speeds, we can recommend nvrtc as an alternative. 

Would you be able to share some benchmark data from your earlier tests showing the compilation speed improvements?

PR #518, opened 2025-05-25
[Enhancement] Add warp specialization attribute handling in IR and rewriter

chengyupku:
* Introduced an `AttrFrame` for warp specialization in the IR, enhancing the handling of warp-specific optimizations.
* Refactored the `VisitStmt_` method in `warp_specialized_rewriter.cc` to check for the new warp specialization attribute, improving the detection of warp specialization conditions.
* Removed outdated code related to condition checks in `IfThenElseNode`, streamlining the specialization logic.

### DIFF
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -288,8 +288,10 @@ WarpSpecializeFrame WarpSpecialize(Array<IntImm> warp_group_ids,
     }
   }
   IfFrame if_frame = If(condition);
+  AttrFrame attr_frame = Attr(Integer(0), "warp_specialize", Integer(1));
   n->frames.push_back(if_frame);
   n->frames.push_back(Then());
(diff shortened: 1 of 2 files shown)


src/ir.cc
```diff
@@ -288,8 +288,10 @@ WarpSpecializeFrame WarpSpecialize(Array<IntImm> warp_group_ids,
     }
   }
   IfFrame if_frame = If(condition);
+  AttrFrame attr_frame = Attr(Integer(0), "warp_specialize", Integer(1));
```
LeiWang1999, line 291: What's the first argument `Integer(0)` means in this part

LeiWang1999 (approved):

src/ir.cc
chengyupku: just a placeholder :)

PR #525, opened 2025-05-26
Add carver's amd driver support and minor modifications to carver's cuda info

Hamerlate:
In the `carver/driver` folder, `hip_driver.py` was developed based on the `hipDeviceProp_t` structure from `hip_runtime_api.h`.
Accordingly, `cdna.py` was modified as well.
In `/carver/arch/cuda.py`, `self.max_smem_usage` and `self.smem_cap` were updated from the `prop.sharedMemPerBlock` to `prop.sharedMemPerMultiprocessor` method.

### DIFF
--- a/tilelang/carver/arch/cuda.py
+++ b/tilelang/carver/arch/cuda.py
@@ -120,13 +120,11 @@ class CUDA(TileDevice):
         self.device: tvm.runtime.Device = device
         self.platform: str = "CUDA"
         # TODO(lei): maybe static shared memory, can be improved in future
-        # self.smem_cap = cuda_driver.get_shared_memory_per_block()
         self.smem_cap = cuda_driver.get_max_dynamic_shared_size_bytes()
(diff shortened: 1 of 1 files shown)


tilelang/carver/arch/cuda.py
```diff
@@ -120,12 +120,12 @@ def __init__(self, target: Union[Target, str]):
         self.device: tvm.runtime.Device = device
         self.platform: str = "CUDA"
         # TODO(lei): maybe static shared memory, can be improved in future
-        self.smem_cap = cuda_driver.get_shared_memory_per_block()
+        self.smem_cap = cuda_driver.get_max_dynamic_shared_size_bytes()
```
LeiWang1999, line 123: We should likely continue using static shared memory to ensure Roller functions correctly

LeiWang1999:
Closed as we will move this part into project tilesight.

PR #537, opened 2025-05-28
[AMD] Support float8 matrix core

LeiWang1999:
let gemini generate the details.


LeiWang1999:
Phase out `e4m3_float8` is a bit hacky, we can move this part in future when we rebase our tvm into main

PR #538, opened 2025-05-29
[CI] Add hadamard example to CI

Rachmanino:


Rachmanino (commented):
Use format.sh to fix lint

PR #566, opened 2025-06-11
[Feature] Implement Swizzle 32B

LeiWang1999:
- Introduced a new `makeQuarterBankSwizzleLayout` function for layout swizzling of 32 bytes.
- Updated `makeGemmABLayout` to include an `enable_padding` parameter, allowing for conditional layout selection between padded and quarter bank swizzle layouts.
- Adjusted layout inference in GEMM operations to utilize the new quarter bank swizzle layout when appropriate.
- Enhanced bulk copy operations to recognize and handle the new layout type, improving memory access patterns.

TODO Items:

- [x] Check if we can have better performance on 4090 if we replace padding with swizzl 32b.
- [x] If so, phase out padding layout.

### DIFF
--- a/src/layout/gemm_layouts.cc
+++ b/src/layout/gemm_layouts.cc
@@ -318,6 +318,26 @@ PrimExpr xor8x8(const PrimExpr &i, const PrimExpr j) {
   return 2 * xor4x4(i1, j1) + xor2x2(i0, j0);
 }
 
+// Layout swizzling for 32 bytes
+Layout makeQuarterBankSwizzleLayout(int stride, int continuous, int element_size) {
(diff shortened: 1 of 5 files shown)


LeiWang1999:
Also a related pr for #464

LeiWang1999:
M16384N16384K16 on 4090
    - pad performance: 0.8273919820785522
    - sw32 performance: 0.8765439987182617

likely the sw32 will lead to a bad memory traffic when compared with padding if we use this  without tma, so we have to separate a new MakeHopperGEMMABLayout.

LeiWang1999:
Padding is also conflict free for this case, so sw32 will introduce extra offset computing.

PR #587, opened 2025-06-20
[Bugfix] fix missing node in ws role maker

botbw:
Triggered https://github.com/tile-ai/tilelang/blob/1406a113804623d3a08cfe2301d34316361b107b/src/transform/warp_specialized_rewriter.cc#L97 when using while loop


LeiWang1999:
LGTM, Merged, thanks @botbw

PR #589, opened 2025-06-21
[Bugfix] fix identation

botbw:
https://github.com/tile-ai/tilelang/actions/runs/15793120882/job/44521638472

### DIFF
--- a/tilelang/autotuner/__init__.py
+++ b/tilelang/autotuner/__init__.py
@@ -338,25 +338,25 @@ class AutoTuner:
                             continue
 
                         # Check tensor compatibility using generator expression
-                    if len(params) == len(self.jit_input_tensors):
-                        def shape_equal(a, b):
-                            if len(a.shape) != len(b.shape):
-                                return False
-                            return all(a_dim == b_dim or isinstance(a_dim, Var) or isinstance(b_dim, Var) for a_dim, b_dim in zip(a.shape, b.shape))
+                        if len(params) == len(self.jit_input_tensors):
+                            def shape_equal(a, b):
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@botbw Thank you for your reports. The linting issue and bug were introduced by Gemini Copilot's suggested commit, but have now been resolved via PR #588."

src/transform/layout_inference.cc
```diff
@@ -145,7 +145,7 @@ class ParallelLoopTransformer : public IRMutatorWithAnalyzer {
                 {Range::FromMinExtent(0, IntImm(index.dtype(), upper_bound))},
                 analyzer_);
 
-            loop_extents.Set(i, IntImm(index.dtype(), shape));
+            loop_extents.Set(j, IntImm(index.dtype(), shape));
```
LeiWang1999, line 148: Would you mind explaining why we should change this part?"

botbw: > Would you mind explaining why we should change this part?"

@LeiWang1999 I guess `j` here iterates over `loop_vars` and corresponding `loop_extents` should be updated? Not 100% sure though.

https://github.com/tile-ai/tilelang/blob/37ace3c816c1c39fc0ac60738c3c2c82022646fa/src/transform/layout_inference.cc#L120-L121

LeiWang1999: make sense, btw how did you find this bug

botbw: > make sense, btw how did you find this bug

By reading through the codebase :p

PR #600, opened 2025-06-26
[Enhancement] Add tma bulk copy.

cherichy:
Add wrapper for "cp.async.bulk". 
@LeiWang1999 You can use it to solve the issue when store unswizzled data with "cp.async.bulk.tensor".


LeiWang1999:
LGTM, Merged :)

PR #604, opened 2025-06-29
Fixed mha_bwd shape inconsistency error

Nathancgy:
dv and dk both have shapes [block_M, dim], but the current dv_shared and dk_shared have shapes [block_N, dim]. This commit fixed their shape inconsistency.

In the previous version, the reason this file still works is because block_N and block_M are the same (64), but this file will run into issue once you change their sizes.


LeiWang1999:
LGTM, Thanks for your contribution! @Nathancgy  Merged.

PR #618, opened 2025-07-08
[Refactor] Add parallel loop transform pass for condition extraction

xs-keju:
Add parallel loop transform pass as a common util to extract conditions and refact the parallel for-loops, and use the pass both in  layoutInference(layourInference.cc) and copy op lowering(in elem.cc).By adding the pass in LayoutInference, we can directly write
```python
                for i in T.Parallel(valid_block_H:):
                        glse[bid, hid * valid_block_H + i, sid] = logsum[i]
```
rather than
```python
                for i in T.Parallel(block_H):
                    if i < valid_block_H:
                        glse[bid, hid * valid_block_H + i, sid] = logsum[i]
```
By adding the pass in copy op lowering,the use of partial copy like
```python
                T.copy(logsum[:valid_block_H], glse[bid, hid * valid_block_H:(hid + 1) * valid_block_H, sid]) 
                # logsum = T.alloc_fragment([block_H]) && valid_block_H < block_H
``` 
works.

### DIFF
--- a/src/op/elem.cc
+++ b/src/op/elem.cc
@@ -15,6 +15,7 @@
 
 #include "../target/utils.h"
 #include "../transform/common/loop_fusion_utils.h"
+#include "../transform/common/loop_parallel_transform_utils.h"
 #include "../transform/loop_partition.h"
 #include "../transform/loop_vectorize.h"
 #include "builtin.h"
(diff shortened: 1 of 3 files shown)


LeiWang1999 (dismissed):

LeiWang1999 (approved):
changed and related test cases made, thanks for your contributions!

PR #653, opened 2025-07-21
[BugFix] Do not modify strict layout in common or relax level of layout inference. More conditions on layout checking

tzj-fxz:
- Do not modify strict layout in common or relax level of layout inference
- More conditions on layout checking

### DIFF
--- a/examples/deepseek_mla/example_mla_decode.py
+++ b/examples/deepseek_mla/example_mla_decode.py
@@ -277,12 +277,12 @@ def ref_program(q, q_pe, kv, k_pe, glse, Output_partial):
 
 
 def main(
-    batch = 1,
-    heads = 128,
-    kv_heads = 1,
-    kv_ctx = 8192,
-    dim = 512,
-    pe_dim = 64,
+    batch=1,
+    heads=128,
+    kv_heads=1,
+    kv_ctx=8192,
+    dim=512,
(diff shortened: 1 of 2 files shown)


src/transform/layout_inference.cc
```diff
@@ -153,10 +153,17 @@ class BufferUseDefCollector : public IRVisitorWithAnalyzer {
...
-              << "Get different layout for " << buffer
-              << "\n current layout: " << layout->DebugOutput()
-              << "\n previous layout: " << layout_map[buffer]->DebugOutput();
+          // (zhengju) We can not modify the strict layout map when current
+          // level is not strict. This check should be done in certain
+          // conditions, since the strict layout map is not updated in the
+          // above code when current level is not strict
+          if (level == InferLevel::kStrict ||
+              !strict_layout_map.count(buffer)) {
+            ICHECK(StructuralEqual()(layout, layout_map[buffer]))
```
LeiWang1999, line 162: It is likely that we skipped the layout update under this condition. Is this behavior expected?

tzj-fxz: Yes. The strict layouts are not modified in relax or common inference process, so the condition check should be skipped to avoid unexpected compile errors.

LeiWang1999 (approved):

PR #656, opened 2025-07-22
[Enhancement] Add compile_flags parameter to JIT kernel and adapter classes for improved compilation control

xwhzz:


LeiWang1999 (commented):
Would be better if we can provide an examples to help us know how we can utilize this feature.

LeiWang1999 (approved):

PR #668, opened 2025-07-28
[Feat] Introduce an offset option in threadblock swizzle

Rachmanino:
Add an `offset` option in threadblock swizzle API. 

This is useful in some distributed scenarios, e.g. overlapped AllGather+GEMM, where an offset could be applied so that SMs can be dispatched to compute blocks of **local** data first, alleviating the overhead of waiting.


Rachmanino:
@LeiWang1999 @chengyupku

PR #678, opened 2025-07-30
[Refactor] Phaseout Frontend Legalize Pass

LeiWang1999:
as title, as let binding will be handled during simplification pass currently.


LeiWang1999:
closed as some pass analysis may rely on let bindings.

PR #682, opened 2025-07-30
Add Flash Attn example on amd mi300 series

Alex4210987:


examples/amd/example_amd_flash_attn_fwd.py
```diff
@@ -0,0 +1,240 @@
+# Copyright (c) Tile-AI Corporation.
+# Licensed under the MIT License.
```
LeiWang1999: lisence should be removed.

LeiWang1999 (changes requested):

PR #712, opened 2025-08-13
[Enhancement][Bugfix] Fix bug in warp specialized pass and add gemm_sr fallback support for Hopper

xwhzz:


LeiWang1999:
@Alex4210987 likely build failed on amd

PR #716, opened 2025-08-14
[CUDA] Init support for sm_120

oraluben:
This is a poc for sm_120 support. Current it passes all tests under `testing/python/kernel/`.

### DIFF
--- a/src/tl_templates/cuda/gemm_mma.h
+++ b/src/tl_templates/cuda/gemm_mma.h
@@ -1,14 +1,14 @@
 #pragma once
 
 #include <cute/algorithm/clear.hpp>
+#include <cute/arch/mma_sm120.hpp>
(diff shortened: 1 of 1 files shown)


src/target/utils.h
```diff
@@ -19,6 +19,7 @@ bool TargetIsVolta(Target target);
 bool TargetIsTuring(Target target);
 bool TargetIsAmpere(Target target);
 bool TargetIsHopper(Target target);
+bool TargetIsBlackwellWmma(Target target);
```
LeiWang1999: Why did we use the name `BlackwellWmma` for this architecture? TargetIsSM120 looks good to me, by the way.

LeiWang1999 (approved):
Awesome contributions! left some comments.

PR #753, opened 2025-08-24
[MXFP4] Add bias to MXFP4 GEMM kernel

tzj-fxz:
As Title

## Summary by CodeRabbit

- **New Features**
  - Optional bias support added to the quantized GEMM example so outputs can include a bias term.
  - Reference implementations include bias-aware variants and accept an optional bias input for parity.

- **Public API**
  - Public interfaces extended with a with_bias option; default behavior unchanged when not used.

- **Tests**
  - Test coverage expanded to exercise four configurations (fast/slow dequantization × bias on/off).


LeiWang1999:
@coderabbitai rename bias into Bias in kernel argument

LeiWang1999 (approved):

PR #798, opened 2025-09-10
kurisu refactor vectorize patch 1

kurisu6912:
- **Fix bug 0905: vectorize with broadcasted value**
- **fix lint error**


yyttt6:
Thank you for submitting this PR!  We have performed some initial testing to evaluate the performance changes.
As shown in the results below, we observed slight performance improvements in the gemm,  gemv , conv, and mha operations. 
result
The following are the specific commands I executed：
python example_gemm_autotune.py 
python example_gemv.py
python example_convolution_autotune.py
 python example_mha_sink_fwd_bhsd.py   --batch 4   --heads 8   --seq_q 512   --seq_kv 512   --dim 64   --window_size 256  --tune

LeiWang1999:
Awesome, let it in.

Issue #820, opened 2025-09-16
[Feature request] Support PyTorch scalar type as tensor dtypes


LyricZhao:
Besides `t: T.Tensor[(n, ), "float32"]`, is there any plan to support `t: T.Tensor[(n, ), torch.float]`?

As users may put kernels of different dtypes into one, using hardcoded `str` may introduce TVM compatibility bugs and force users to maintain a PyTorch-dtype-to-string map.

SiriusNEO:
Addressed in: https://github.com/tile-ai/tilelang/pull/916

PR #832, opened 2025-09-17
WIP feat: support torch.dtype in T.tensor

LiuYuHui:
Relate to #820


LeiWang1999:
thanks @LiuYuHui , but closed for now we this will be merged into the design of #916

Issue #839, opened 2025-09-18
[Bug] Fast-math is dangerous to enable by default


LyricZhao:
As titled, fast-math in some cases is harmful for pretraining/RL, and costs weeks to debug, or give wrong experiment results. In my understanding, fast-math **must** be disabled by default.

Also I suggest all the passes must ensure binary identical for results, e.g. `a / b` can not be optimized into `a * (1 / b)`.

LeiWang1999:
These two issues that mentioned have now been addressed:
- PR #846 turned off fast‑math optimizations by default.
- PR #849 disabled the rewrite rule that simplified a / b into a * (1 / b).

But after double checking the [rewrite simplification rules](https://github.com/tile-ai/tvm/blob/tilelang_main/src/arith/rewrite_simplify.cc) and think a bit more, I think this may need further discussions (also see see https://discuss.tvm.apache.org/t/discuss-is-constant-division-to-multiplication-rewrite-in-tvm-necessary/18615/4).

Because some of the constant‑folding and algebraic rewrites may also introduce similar issues. Examples include:

- `(x * c1) * c2 → x * (c1 * c2)`
- `(x / c1) / c2 → x / (c1 * c2)` (with correct division semantics)  
- `(x * c1) / c2` simplified if `c1` divisible by `c2`, or vice versa.
- `(x / c1) + c2` normalized into a single division when possible.
- `(y + z) * x → y * x + z * x`  
- `(x + c1) * c2 → x * c2 + c1 * c2`
- `y*x + z*x → (y+z) * x`


However, these rewrites are important for optimization and performance, I think we should consider introducing a deterministic switch or similar option that can disable these transformations when strict, this pr should be closed.

LeiWang1999:
Below rules are safe, as those transformations only applied on IndexType (which are all integers).

Closed as now resolved.

LyricZhao:
Thanks!

PR #875, opened 2025-09-25
[FastMath] Disable default TVM fastmath intrinsic dispatch and add explicit fastmath op to invoke

LeiWang1999:
This PR addresses issue #839. It makes certain math intrinsics use the fast‑math versions of operators through TVM’s intrinsic rules.

After some consideration, we decided **not** to expose `__exp` at the base operator level and keep `exp` definition in tvm. The idea is that `__exp` is a CUDA‑specific intrinsic, whereas `exp` is a standard mathematical operation available on almost all hardware. Therefore, `exp` should remain defined at the fundamental level, making it easier to adapt to different operators across backends.

Once this PR is merged, all calls like `T.exp`, `T.sin`, etc. will be code‑generated into `expf` / `sinf` or the corresponding precision math function. Fast‑math intrinsics such as `__exp` will still be accessible internally in TileLang.

Example usage:

```python
# Enable fast math explicitly
T.__exp()

# Or enable globally via pass configs
pass_configs = {
    tilelang.PassConfigKey.ENABLE_FAST_MATH: True
}

T.exp  # Will use fast-math version if enabled
```

and we also provide a test for this component.

## Summary by CodeRabbit

- New Features
  - Fast-math ops added (exp, exp10, log, log2, log10, sin, cos, tan) with language-level fastmath wrappers and CUDA codegen emitting optimized fast-math intrinsics; new CUDA math dispatch for type-specific variants.
  - New CUDA operator backend offering precise, double-precision, and fast execution modes.

- Tests
  - Added CUDA-targeted tests validating fast-math vs non-fast-math codegen and numeric correctness.

- Documentation
  - Added comprehensive precision/accuracy comparison report.

- Chores
  - Updated third-party submodule reference.

### DIFF
--- a/testing/python/fastmath/test_mathops_fastmath.py
+++ b/testing/python/fastmath/test_mathops_fastmath.py
@@ -79,8 +79,6 @@ def run_single_arg_mathop_test(mathop_name,
         out_idx=[1],
         target="cuda",
         pass_configs={
-            tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
-            tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
             tilelang.PassConfigKey.TL_ENABLE_FAST_MATH: False,
         })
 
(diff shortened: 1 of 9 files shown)


testing/python/fastmath/test_mathops_fastmath.py
```diff
@@ -0,0 +1,336 @@
...
+        expected = b_fastmath  # Just check compilation works
+
+    torch.testing.assert_close(b_fastmath, expected, rtol=1e-3, atol=1e-3)
+    print(f"✓ {mathop_name} numerical test passed")
+
+
+@tilelang.testing.requires_cuda
+def test_mathops_generate_no_fastmath():
+    """Test that our tl.* mathops generate fastmath CUDA code (__expf etc.)"""
+    # Based on test results, our tl.* intrinsics actually generate fastmath versions
```
LeiWang1999: comments should be fixed/

PR #883, opened 2025-09-26
[Bugfix]:Fix atomicadd auto vectorize identify var error

yyttt6:


src/transform/atomicadd_vectorize.cc
```diff
@@ -319,6 +319,33 @@ static int GetVectorizeSizeMax(int compute_capability, DataType dtype) {
 For VectorizeAtomicAdd(const For &for_node, const Var &thread_var,
                        const Range &thread_bounds, int compute_capability) {
 
+  auto ParseIndex = [](const PrimExpr &idx, PrimExpr &var_out,
```
LeiWang1999: I think we should add some comments for this function.

LeiWang1999 (changes requested):

yyttt6:
I refactored part of the AtomicAdd auto vectorization code.
This change fixes the previous error of incorrect variable matching, and makes the AtomicAdd auto vectorization more robust and adaptable to a wider range of cases.

PR #888, opened 2025-09-28
[Bugfix] Fix CopyNode Lower method to include disable_tma flag in GetCopyInst

Rachmanino:
This pull request makes a targeted change to the `CopyNode::Lower` function in `src/op/copy.cc` to enhance the logic for selecting the copy instruction. The update ensures that the `disable_tma` flag is now considered in addition to the configuration option when determining whether to disable TMA lowering.

Key change:

* Improved the call to `GetCopyInst` by combining the `disable_tma_lower` configuration and the `disable_tma` flag, ensuring TMA lowering is properly disabled when either is set.

## Summary by CodeRabbit

* **Bug Fixes**
  * Per-operation and global disable flags now consistently prevent bulk-transfer fast-paths, fixing incorrect path selection and improving correctness.

* **Performance**
  * More predictable behavior when bulk transfers are disabled; slight performance variations where fast-paths are intentionally bypassed.

* **Examples**
  * Updated examples to use default lowering and enable per-copy SIMT fallbacks with clarifying comments.

### DIFF
--- a/examples/flash_decoding/example_mha_inference.py
+++ b/examples/flash_decoding/example_mha_inference.py
@@ -8,7 +8,7 @@ from functools import partial
 num_split = 4
 
 
-@tilelang.jit(out_idx=[5], pass_configs={tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True})
+@tilelang.jit(out_idx=[5])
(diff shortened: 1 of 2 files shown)


Rachmanino:
cc @LeiWang1999 @xwhzz

PR #900, opened 2025-09-29
[Bugfix] Fix flops comp and softmax scale in mla 

Edenzzzz:
Fixes some issues.
1. The benchmark cannot run due to missing arg
image

2. The FLOPs computation should use the full head dim `d = dpe + dv`. The softmax scale is adjusted accordingly. See  
https://github.com/flashinfer-ai/flashinfer/blob/0f8ecab1121ad52dd22b47317662cbbd7f0d343c/benchmarks/bench_deepseek_mla.py#L73

I ran the bench against Flashinfer main branch on H100, and looks like it has packed in more optimizations than when the last comparision was made. Looking forward to learning more Tilelang tricks to speed it up!
image

## Summary by CodeRabbit

- New Features
  - softmax_scale is now optional and auto-computed at runtime when omitted; callers remain backward compatible.

- Refactor
  - Renamed Flash Inference backend from "flash_infer" to "flashinfer" across targets and dispatch.

- Improvements
  - FLOPs reporting updated to use d only (was d + dv), aligning metrics with the new attention scaling.

### DIFF
--- a/examples/deepseek_mla/benchmark_mla.py
+++ b/examples/deepseek_mla/benchmark_mla.py
@@ -87,7 +87,7 @@ def run_flash_mla(q, block_table, blocked_k, max_seqlen_pad, block_size, b, s_q,
 
 
 @torch.inference_mode()
-def run_flash_infer(q, block_table, blocked_k, max_seqlen_pad, block_size, b, s_q, cache_seqlens,
+def run_flashinfer(q, block_table, blocked_k, max_seqlen_pad, block_size, b, s_q, cache_seqlens,
                     h_q, h_kv, d, dv, causal, dtype):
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@Edenzzzz Thanks for your contribution! And PR ID 900 is a pretty cool number.

We know the trick behind FlashMLA, but for now, we need to implement it using an identically warp‑specialized kernel — for example, as shown here: https://github.com/tile-ai/tilelang/blob/main/examples/warp_specialize/example_warp_specialize_flashmla.py with a carefully designed layout.

We’ll work on improving performance in the future. For now, our focus will be on programmability and robustness.

LeiWang1999:
LGTM, Merge :)

PR #936, opened 2025-10-03
[Profiler]Adds CUPTI profiler support

Cunxiao2002:
## Summary by CodeRabbit

- New Features
  - Added built-in benchmarking to the GEMM example, printing measured latency.
  - Expanded profiling capabilities with event- and Kineto-based benchmarking, optional per-kernel timings, and adaptive warmup/repeat for more stable results.
  - Added support for quantile statistics, configurable return modes (min/max/mean/median), multi-kernel profiling, and optional output suppression.
  - Enabled trace export for deeper analysis.
  - Extended the public profiling API with new optional parameters to configure the above behaviors.


LeiWang1999:
would be better if we can combine do_bench_kineto and do_bench_event into a same api `do_bench`.

tilelang/profiler/__init__.py
```diff
@@ -223,6 +223,10 @@ def do_bench(
         n_warmup: int = 1,
         n_repeat: int = 1,
         input_tensors: List[torch.Tensor] = None,
+        backend: Literal["event", "cupti"] = "event",
+        grad_to_none: Optional[List[torch.Tensor]] = None,
```
LeiWang1999: do we really need parameter `grad_to_none`?

tilelang/profiler/bench.py
```diff
@@ -67,34 +117,74 @@ def do_bench(
...
+        )
+        if quantiles is not None:
+            ret = torch.quantile(times, torch.tensor(quantiles, dtype=torch.float)).tolist()
+            if len(ret) == 1:
+                ret = ret[0]
+            return ret
+        return getattr(torch, return_mode)(times).item()
+
+    elif backend == "cupti":
+        using_nsys = int(os.environ.get('DG_NSYS_PROFILING', 0))
```
LeiWang1999: DG means `DeepGEMM`, would be better to remove this env and always with using_nsys = False.

tilelang/profiler/bench.py
```diff
@@ -2,6 +2,45 @@
 
 import torch
 from typing import Callable, List, Literal, Optional, Union
+import os
+import sys
+
+
+# from https://github.com/deepseek-ai/DeepGEMM/blob/main/deep_gemm/testing/bench.py
+class suppress_stdout_stderr:
```
LeiWang1999, line 10: would be better to removed

Cunxiao2002:
some benchmarks for two profiler
1. example/gemm/example_gemm.py
- event Latency: 0.027775999158620834ms
- cupti Latency: 0.023777000000000045ms
2. example/flash_attention/example_mha_fwd_bhsd.py
- event Latency: 0.01020800042897462845ms
- cupti Latency: 0.00630400000000008731ms
3. example/grouped_gemm/example_grouped_gemm_fwd.py
- event Latency: 0.4819520115852356ms
- cupti Latency: 0.4754229999999989ms

LeiWang1999 (approved):

PR #939, opened 2025-10-05
[Build] Migrate to scikit-build-core

### LINKED ISSUE #756: ARM wheels

### LINKED ISSUE #833: [Feature Request] Migrate to scikit-build-core
scikit-build-core provides better cmake integration and I hope it can simplify current setup.py and build/wheel packaging workflow.

I can work on this when I have time.

oraluben:
Resolves #833
Closes #756

```
pip install . -v  # CUDA / Metal build on Linux / Mac
USE_ROCM=ON pip install . -v  # ROCm build
USE_CUDA=OFF pip install . -v  # CPU build
```

Detailed information could be found at the [doc](https://github.com/oraluben/tilelang/blob/sk-build-core/docs/get_started/Installation.md)

This PR basically migrates to a fully cmake-based build system and eliminates python-based building process, and therefore simplifies the building and installing stage.

User and developers of tilelang could just install from source (without setting `PYTHONPATH` and so on).

- [x] Use one wheel for different python version via stable abi
   `cp38-abi3` wheels for >= python 3.8 (needs https://github.com/tile-ai/tvm/pull/12)
- [x] CUDA wheel
   - [x] ROCm wheel
   - [x] ~~Unify CUDA and ROCm wheel? Is that possible?~~
- [x] Metal wheel
- [x] Build cython ext
   - [x] ~~Fix~~ Remove in-tree jit compile
- [x] git hash and cuda/rocm/metal extension in version. Support build-time toolchain info for CUDA. (e.g. in the [artifacts](https://github.com/tile-ai/tilelang/actions/runs/18362923640?pr=939))
  Now tilelang reads version (for cache, etc.) from the package version.
- [x] Validation
   - [x] Wheel works
   - [x] linux aarch64 https://github.com/tile-ai/tilelang/issues/756
   - [x] editable install
- [x] Cleanup
   - [x] Remove tox-based build scripts (currently we may have three host platform (linux+{x86,aarch64}, metal), and for each platform we only need 1 wheel. That makes tox unnecessary for building different python versions and insufficient for building against different platform) and refractor the scripts to build locally.
- [x] CI related fixes
   - [x] auditwheel and [delocate](https://github.com/matthew-brett/delocate) for linux / darwin wheels
- [x] Update [doc](https://github.com/oraluben/tilelang/blob/sk-build-core/docs/get_started/Installation.md)
   - [x] Build wheel
   - [x] Editable install
   - [x] Use custom tvm
   - [ ] ...

## Summary by CodeRabbit

* **New Features**
  * Added a scheduled cross-platform CI workflow for daily and release wheel builds; introduced dynamic build-time version metadata.

* **Documentation**
  * Installation guide updated: raised Python minimum, tightened CUDA requirements, clarified bundled vs external backend and nightly guidance.

* **Refactor**
  * Modernized CMake and packaging flow; unified backend selection and simplified native/Python integration and library discovery.

* **Chores**
  * Removed legacy install/build scripts, multi-version tox/docker workflows, and replaced monolithic packaging script with modern toolchain.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -5,7 +5,6 @@ on:
     - cron: '0 22 * * *'
   release:
     types: [ published ]
-  pull_request:
 
 env:
   PYTHON_VERSION: '3.12'


.github/workflows/dist.yml
XuehaiPan, line 1: FYI, you may read this: https://cibuildwheel.pypa.io

- https://github.com/metaopt/optree/blob/09d65cb330dd0dd52d8dc05be07996cdb407ad3b/pyproject.toml#L92-L136
- https://github.com/metaopt/optree/blob/09d65cb330dd0dd52d8dc05be07996cdb407ad3b/.github/workflows/build.yml#L268-L319

oraluben: `cibuildwheel` actually do not works well here. As proper docker image (manylinux_2_28 + cuda) is too large for GA runner. This is just a POC and test to build on GA. [Building locally](https://github.com/tile-ai/tilelang/pull/939/files/5592fd06c9904d3008d336f5a2bca81ada8dcc2d#diff-61d1811c43ee3a7a99754d36cbd32699d7f7a7f27e9f4da98fa023cd2c3e05b4) could be more feasible for tilelang.

XuehaiPan: > docker image (manylinux_2_28 + cuda) is too large for GA runner

You do not need a full CUDA container but a clean `nvcc` installation (maybe without cuDNN). That can be achieved by the official manylinux container + a manual `nvcc` installation.

XuehaiPan: Here is a very old reference: https://github.com/metaopt/torchopt/blob/960fb0aa53f27bad1af0bdf018b5ad4be077949a/pyproject.toml#L133-L135 (the image was based on CentOS at that moment, while now Ubuntu)

.github/workflows/dist.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Build wheels
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+        echo "whl_name=$(ls wheelhouse | head -n1)" >> $GITHUB_OUTPUT
+
+    - uses: actions/upload-artifact@v4
+      with:
+        name: ${{ steps.repair.outputs.whl_name }}.zip
+        path: wheelhouse/${{ steps.repair.outputs.whl_name }}
+        compression-level: 0
```
XuehaiPan:
```suggestion
      - name: Set up QEMU
        if: runner.os == 'Linux'
        uses: docker/setup-qemu-action@v3
        with:
          platforms: all

      - name: Build wheels
        uses: pypa/cibuildwheel@v3.2
        env:
          CIBW_BUILD: cp312-*
          CIBW_ARCHS: auto64
          CIBW_ARCHS_MACOS: auto64 universal2
        with:
          package-dir: .
          output-dir: wheelhouse
          config-file: "{package}/pyproject.toml"

      - uses: actions/upload-artifact@v4
        with:
          name: wheels-${{ env.PYTHON_VERSION }}-${{ runner.os }}-${{ runner.arch }}
          path: wheelhouse/*.whl
          if-no-files-found: error
```

You could use the `pypa/cibuildwheel@v3.2` action directly.

oraluben: Sure. But I wonder why would you prefer qemu over linux aarch64 runner?

XuehaiPan: You will not need QEMU if you are building AArch64 on `ubuntu-*-arm`. You will need QEMU if you are building other arches (e.g., `s390x`, `riscv64`). Passing on this if they are not planned.

oraluben:
This should be ready for review.

Main parts that this PR touched:
1. Build system
   1. Now the project builds fully by cmake driven by sk-build-core. `pip install .` is supposed to cover all use case.
   2. Build dir is set to `./build` intentionally, to help IDE indexing, but it's not suppose to use libs in build dir.
   3. All libs will be installed into python's `site-packages`, no matter if it's a editable install.
   4. Add a patch in `tile-ai/tvm` https://github.com/tile-ai/tvm/pull/12 so that tvm's cython ext could be used by py>=3.8. But upstream tvm seems to have something new? cc @Hzfengsy 
   5. Thanks to @XuehaiPan , we're using a pure manylinux2014 builder for x86_64, which means we can support torch<2.6 easier. (But manylinux2014 only have a up-to-12.4 cuda, does that mean we need another wheel for CUDA-13?)
   6. Use ccache if possilbe.
2. Runtime compile and cache
   1. `cython_wrapper.so` will be compiled at build time, and installed directly under `site-packages`, to support `import cython_wrapper`. This is not the best practice, but I think we have plan to migrate to tvm-ffi soon?
   2. Then I removed the runtime compilation logic of `cython_wrapper.so`. Developers who want to modify that file needs to use an editable install and re-install with pip to make sure the file is re-generated. This takes ~30 seconds each time, because pip needs to pack and unpack each time it installs. I wonder if this looks good to you @LeiWang1999 , or shell we restore the old logic?
   3. Now tilelang use the package version (e.g. `0.1.6.post1+cu121.gitxxxxxxxx`) as the version used in its cache folder.
7. CI and releases
   1. Added a CI job to generate wheels for linux+{x86,aarch64} and mac. I plan to set it to a daily cron job and for new releases, for nightly users to use and for maintainers to be uploaded to pypi. I'm not sure if this aligned with your workflow. @LeiWang1999 
   2. And do we need different release for CUDA 12/13?

.github/workflows/dist.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  build-wheels:
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, ubuntu-22.04-arm, macos-14]
+        include:
+        - os: ubuntu-22.04
+          cuda_version: 12.1
+        - os: ubuntu-22.04-arm
+          cuda_version: 12.8
```
XuehaiPan:
```suggestion
        - os: ubuntu-22.04
          cuda_version: "12.1"
        - os: ubuntu-22.04-arm
          cuda_version: "12.8"
```

This should be interpreted as a string rather than number.

LeiWang1999:
1. I think it’s currently difficult to have a single wheel that supports both CUDA 12 and CUDA 13, and it seems that there aren’t many users on CUDA 13 yet. We will hold off on providing a CUDA 13 release until there is clear demand from the community.  
2. We do have a plan to migrate to `tvm_ffi` — see issue #970.  
3. Phasing out the JIT compilation of cython sounds good to me, as re‑compiling Cython on the fly is a very rare requirement during the development, not to mention users instead of developers. 
4. The daily build CI looks very helpful.

PR #941, opened 2025-10-05
[Example] Disable TMA and enable FastMath for NSA Examples (#941)

LeiWang1999:
In this case, `cp.async` will be about 4× faster than TMA because the block size is too small and register allocation/deallocation instructions are not used efficiently.

```bash
Benchmark Results (TileLang):
Configuration: batch=32, seq_len=1024, heads=1, head_query=16, dim=128, blocks=16, block_size=32
Average time: 0.66 ms
Performance: 207.89 TFLOPs
Benchmarking Triton implementation:

Benchmark Results (Triton):
Configuration: batch=32, seq_len=1024, heads=1, head_query=16, dim=128, blocks=16, block_size=32
Average time: 0.96 ms
Performance: 143.60 TFLOPs
```

## Summary by CodeRabbit

- Bug Fixes
  - Standardized block selection in the NSA forward path for consistent accumulation per token.
  - Improved buffer index handling to promote small integer indices to 64-bit when needed, preventing potential overflow during flattening.

- Chores
  - Updated JIT/pass configurations across DeepSeek NSA examples to enable fast math and disable certain compiler specializations; no public API changes.

### DIFF
--- a/src/transform/flatten_buffer.cc
+++ b/src/transform/flatten_buffer.cc
@@ -62,6 +62,43 @@ private:
   using IRMutatorWithAnalyzer::VisitStmt;
   using IRMutatorWithAnalyzer::VisitStmt_;
 
+  class Int64Promoter : public tir::IndexDataTypeRewriter {
+  public:
+    using Parent = IndexDataTypeRewriter;
+
+    PrimExpr VisitExpr_(const VarNode *op) final {
+      if (op->dtype.is_int() && op->dtype.bits() < 64) {
+        return cast(DataType::Int(64), GetRef<Var>(op));
+      }
+      return GetRef<PrimExpr>(op);
+    }
+
+    PrimExpr VisitExpr_(const IntImmNode *op) final {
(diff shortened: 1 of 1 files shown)


LeiWang1999:
Some takeaway of this pr:

1. **Automatic TMA usage in TileLang**  
   TileLang currently utilizes Tensor Memory Accelerator (TMA) automatically. For small tile sizes—especially in this case—this can be inefficient.  
   - Issuing TMA commands incurs several hundred cycles of overhead compared to `cp.async`.
   - For small memory copies, the overhead outweighs any potential benefit.

2. **TMA with Sparse Load increases register usage**  
   Using TMA together with Sparse Load can result in more registers being allocated to producer warps.  
   - By default, producers have limited registers.
   - This extra register pressure may reduce performance efficiency.

3. **FastMath setting**  
   In previous PRs, we disabled FastMath by default to ensure numerical accuracy, whereas Triton enables FastMath by default.  
   - This difference also impacts performance measurements.

Issue #814, opened 2025-09-15
[Bug] Incorrect Inlining of Local Variables in TileLang


LJC00118:
TileLang produces different code for semantically equivalent programs when inlining local variables.

Example:
```python
# Code snippet 1
tmp = T.max(A[i], 1)
B[i] = tmp / 2
A[i] = tmp * 2

# Code snippet 2  
tmp = T.max(A[i], 1)
A[i] = tmp * 2
B[i] = tmp / 2
```

These two snippets are semantically identical, but TileLang inlines them differently:

```python
# Inlined result 1
B[i] = T.max(A[i], 1) / 2
A[i] = T.max(A[i], 1) * 2

# Inlined result 2
A[i] = T.max(A[i], 1) * 2  
B[i] = T.max(A[i], 1) / 2
```

The repeated evaluation of T.max(A[i], 1) combined with different inlining order causes the modification of A[i] to affect the subsequent computation of B[i], breaking the original semantics.

LeiWang1999:
PR #947 attempts to resolve this issue. Previously, TileLang forced all expressions to be inlined, but now we rely on the simplify pass to retain `let` statements that may have side effects (as in this case). However, this introduces some complications.

One major concern is the design choice in TVM’s `LetStmt`, which follows the form `(var, value, body)`. This structure can make software pipeline analysis more complex. For example, in a loop like:

```python
for i:
    let:
        copy
        copy
        gemm
```

the nested structure means we must also reason about how the `let` behaves across different pipeline stages.

I wonder if a flatter structure like this:

```python
for i:
    let
    copy
    copy
    gemm
```

—where they are at the same level—might be a cleaner design and simplify stage analysis.

would you mind take a look @Hzfengsy ?

LeiWang1999:
@LJC00118 , now resolved.

The current TileLang code:

```python
@tilelang.jit
def _tmp_var_kernel(N, block_N, dtype="float"):

    @T.prim_func
    def kernel(
            A: T.Tensor((N,), dtype),
            B: T.Tensor((N,), dtype),
    ):
        with T.Kernel(T.ceildiv(N, block_N), threads=128) as bx:
            for i in T.Parallel(block_N):
                idx = bx * block_N + i
                tmp = T.max(A[idx], 1)
                B[idx] = tmp / 2
                A[idx] = tmp * 2

    return kernel
```

will be generated as:

```cpp
[...]

PR #947, opened 2025-10-07
[Bugfix] Do not force inline let stmt

LeiWang1999:
as title.

- [x] Add a doc about let stmt inline.

## Summary by CodeRabbit

- New Features
  - FusedMoE example main() is now configurable via function parameters (model size, experts, batch, seq_len).

- Bug Fixes
  - More robust software-pipeline injection/planning that preserves wrappers and attributes.
  - Relaxed shared-memory boundary checks and improved buffer remapping/versioning for correctness.
  - Safer let-binding handling and IR simplification to avoid incorrect aliasing/inlining.
  - New pass option to control eager let inlining.

- Documentation
  - Added “Compiler Internals: LetStmt Inlining” page and updated docs index.

- Tests
  - Added regression tests for temporary-variable handling and multi-version buffers; updated example tests to pass explicit parameters.

### DIFF
--- a/docs/compiler_internals/letstmt_inline.md
+++ b/docs/compiler_internals/letstmt_inline.md
@@ -1,10 +1,10 @@
 # LetStmt Inlining in TileLang
 
-This document explains how `LetStmt` inlining works in TileLang's simplification pass, which is an important optimization that affects code generation and performance.
+This document explains how `LetStmt` inlining works in TileLang's simplification pipeline, which is an important optimization that affects code generation and performance.
 
 ## Overview
 
-A `LetStmt` (Let Statement) is a temporary variable binding in the IR (Intermediate Representation). During compilation, TileLang's simplifier may choose to inline these temporary variables to simplify the code. However, not all `LetStmt` nodes can be safely inlined.
+A `LetStmt` (Let Statement) is a temporary variable binding in the IR (Intermediate Representation). During compilation, TileLang's simplifier may choose to inline these temporary variables to simplify the code. TileLang also provides a standalone `LetInline` pass that performs eager substitution before the main legalization pipeline. However, not all `LetStmt` nodes can be safely inlined.
 
 ## When Does LetStmt Get Inlined?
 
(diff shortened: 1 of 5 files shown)


LeiWang1999:
a fix for issue #814

LeiWang1999:
# PR Summary via GPT-5 Pro – LetStmt-Aware Pipeline Fixes (Issue #814)

## Overview
- Addressed correctness regressions that appear when `LetInline` is disabled for CUDA pipelines.
- Hardened multiple lowering passes so they can work with `LetStmt`, `AttrStmt`, and single-branch `IfThenElse` wrappers that now remain in the IR.
- Added documentation and regression tests to capture the behaviour.

## Compiler Pass Updates

- **Remove unconditional `LetInline`** (`tilelang/engine/phase.py:88`): the pass was masking issues in downstream transforms; we now rely on updated passes that explicitly tolerate `LetStmt` wrappers.

- **Pipeline planning** (`src/transform/pipeline_planning.cc:410`):
  - Walks through `LetStmt`/single-branch `IfThenElse` wrappers to recover the underlying `SeqStmt`.
  - Records shared buffers defined in ancestor blocks and handles `address_of` calls whose operand is a `Var`, ensuring dependency analysis still sees shared memory accesses.

- **Software-pipeline injection** (`src/transform/inject_pipeline.cc:846`):
  - Uses the same unwrapping logic and re-applies the stripped wrappers after the pipeline is rewritten so attributes, lets, and guards survive the transform intact.

- **Multi-version buffer rewriter** (`src/transform/multi_version_buffer_rewriter.cc:141`):
  - Tracks ancestor statements and block-local allocations to find shared buffers even when their LCA is outside the loop body.
  - Treats `Role::kBoth` stages as both producers and consumers and falls back to “first write / last read” indices to decide when double-buffering is required.
  - Restricts multi-versioning to `shared` / `shared.dyn` scopes so locals are not duplicated accidentally and handles let/if wrappers around the pipeline body.

- **Shared-memory merger** (`src/transform/merge_shared_memory_allocations.cc:194`):
  - Allows reads in the same scope level where the buffer is allocated, preventing false negatives in the presence of let-wrapped expressions.

- **Tile op lowering** (`src/transform/lower_tile_op.cc:330`):
  - Resolves let-bound `BufferLoad` expressions before remapping, so `address_of` / `tvm_access_ptr` seen after `LetInline` removal still convert to the permuted layouts required by Hopper intrinsics.
  - Caches let bindings during visitation to avoid repeated re-resolution.

- **Docs** (`docs/compiler_internals/letstmt_inline.md`, `docs/index.md:35`):
  - Document how and why `LetStmt` inlining works, providing context for the removal in the lowering pipeline.

## Examples & Tests

- **FusedMoE example** (`examples/fusedmoe/example_fusedmoe_tilelang.py:521`):
  - Parameterized the demo so CI can run a reduced problem size, keeping coverage while respecting resource limits.
  - Updated test harness (`examples/fusedmoe/test_example_fusedmoe.py:3`) to pass the smaller configuration.

- **Issue regression test** (`testing/python/issue/test_tilelang_issue_814.py`):
  - Captures the scenario where a let-bound temporary feeds both load and store paths, ensuring future changes do not reintroduce silent inlining bugs.

## Verification

- `python examples/deepseek_deepgemm/example_deepgemm_fp8_2xAcc.py`
- `python -m pytest testing/python/issue/test_tilelang_issue_814.py`
- `python -m pytest examples/fusedmoe/test_example_fusedmoe.py`

LeiWang1999:
have to modify some codes for aws related pass, @chengyupku should take a look.

LeiWang1999:
ci has failed due to limited storage, but tests and examples have been passed in my local environment.

PR #799, opened 2025-09-10
[Backend] Add metal backend

oraluben:
Add metal backend via tvm's metal codegen and torch's `torch.mps.compile_shader`, and create a new `torch` backend, who only supports MPS for now. Some tasks are left for future works.

Performance numbers:

PyTorch's MPS backend uses [matrixMultiplicationWithPrimaryTensor](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/matrixmultiplication(primary:secondary:name:)?language=objc) by default, which is faster than default implementation.
For pytorch's native MPS mm impl (enable with `PYTORCH_MPS_PREFER_METAL=1`), we can observe significant speedup.

In general, tilelang is 2-3x faster than torch native impl and 2-3x slower than MPSGraph impl. tilelang impl could be optimized via implementating `T.gemm` by matrixMultiplicationWithPrimaryTensor.

Outliners in the table below seem to be error and disappear after rerun.

```
config                     tl_run1    tl_run2        torch    torch_native
----------------------  ----------  ---------  -----------  --------------
1024_1024_1024,float32    0.203224   0.119603    0.0732187        0.454423
1024_1024_1024,float16    0.118059   0.171631    0.0509571        0.329795
1024_1024_2048,float32    0.21623    0.277052    0.0956315        1.01536
1024_1024_2048,float16    0.163326   0.245721    0.0952914        0.738723
1024_1024_4096,float32    0.364069   0.58848     0.18605          2.02241
1024_1024_4096,float16    0.321528   0.519458    0.164556         1.51384
1024_2048_1024,float32    0.17366    0.236656    0.0963473        0.991771
1024_2048_1024,float16    0.174649   0.251009    0.0902327        0.663332
1024_2048_2048,float32    0.350103   0.55103     0.177765         1.99991
1024_2048_2048,float16    0.322436   0.529719    0.160975         1.3848
1024_2048_4096,float32    0.711705   1.1608      0.3412           3.81918
1024_2048_4096,float16    0.649519   1.15448     0.308582         2.75274
1024_4096_1024,float32    0.348693   0.502921    0.17556          1.92968
1024_4096_1024,float16    0.320613   0.553743    0.162928         1.35127
1024_4096_2048,float32    0.705229   1.10684     0.335839         3.74767
1024_4096_2048,float16    0.634583   1.01857     0.305798         2.6813
1024_4096_4096,float32    1.42189    2.32234     0.657488         7.39324
[...]

PR #973, opened 2025-10-10
[CI][Refactor] Merge test CI workflow files into one

XuehaiPan:
This PR is a follow-up PR to:

- #971

Please review and merge that first if you are interested in this one.

------

This PR merges the three workflow files for supported platforms (currently CUDA, ROCm, and Metal) into a single file with matrix strategies.

This enables several advantages:
- Easier maintenance: Only one workflow file to maintain instead of three separate files.
- Scalability: Easier to add support for new platforms and Python versions in the future by simply extending the matrix.
- Consistent testing: All platforms are tested with the same steps and configurations, ensuring consistency across different environments.
- Reduce duplication: Common steps are defined once and reused for all platforms, reducing redundancy.
- Better caching logic: Improved caching logic to avoid redundant installations and speed up the workflow.
- Improved readability: A single workflow file provides a clearer overview of the CI process for the project.

## Summary by CodeRabbit

* **Chores**
  * Major CI/workflow overhaul: unified lint & cross-toolkit test pipelines, multi-runner matrices (CUDA/ROCm/Metal/macOS), conditional triggers, caching/ccache, per-target wheel builds, artifact staging, and repository-owner gating; legacy ROCm CI removed.

* **Refactor**
  * Examples updated to accept optional argv for programmatic invocation; tests updated to call mains with explicit args.

* **Tests**
  * Deterministic test seeding added; separate test requirement manifests for CUDA/ROCm/Metal introduced; one numeric tolerance relaxed.

* **Style**
  * Added a multiprocessing-safe formatting pre-commit hook.

* **Config**
  * Packaging, build tooling, and dev/test requirements reorganized; project metadata and build configs updated.


.github/workflows/ci.yml
```diff
@@ -1,125 +1,248 @@
...
+        if: contains(matrix.runner.toolkit, 'Metal')
+        run: |
+          export CMAKE_CONFIGURE_OPTIONS="${CMAKE_CONFIGURE_OPTIONS} -DUSE_METAL=ON"
+
+          echo "USE_METAL=ON" | tee -a "${GITHUB_ENV}"
+          echo "CMAKE_CONFIGURE_OPTIONS=${CMAKE_CONFIGURE_OPTIONS}" | tee -a "${GITHUB_ENV}"
+
+      - name: Setup venv
+        id: setup-venv
+        if: steps.venv-cache.outputs.cache-hit != 'true'
```
XuehaiPan: The `venv` cache key contains the Python version and the accelerator toolkit version. We no longer need to manage the cache markers manually.

.github/workflows/ci.yml
```diff
@@ -1,125 +1,248 @@
...
+      - labeled
+      - unlabeled
+      - opened
+      - synchronize
+      - reopened
+  # Allow to trigger the workflow manually
+  workflow_dispatch:
+
+permissions:
+  contents: read
```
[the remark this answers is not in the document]
XuehaiPan: False positive!!!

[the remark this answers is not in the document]
XuehaiPan: You shall read this: <https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents>

`actions: write` means the current action can create or cancel runs. Not storing caches.
`contents: read` is for listing commits/branches, while `contents: write` can create and push commits/releases.

oraluben:
I think this increase the maintain efforts in some ways: the self-hosted runners seems to be setup in peking university and have a much slower connection with github and pypi/uv. That makes restoring GA cache takes more time than it saves.
On the other hand, those runners have more than enough cpus to compile tilelang, which the metal runner don't, so ccache is necessary for metal. And it seems better to just use a pip mirror for cuda and rocm runner

Considering the different situation for those runners, it might be better to keep them separated? (or use multiple jobs with dependencies, do not merge them in one single job)

Some of my previous test could be found around: https://github.com/tile-ai/tilelang/pull/939/commits/76e9df739824b0ec3f8be3b4fcf39b6a8769ccbf. I used uv on self-hosted runner locally to speedup package installation, but do not upload them as cache: https://github.com/tile-ai/tilelang/pull/939/commits/df9ae253e9a1f43540e56088855c9a7e0dc5eea4, and uv cache are preserved on host.
You can find full experiment in  https://github.com/tile-ai/tilelang/pull/939

XuehaiPan:
> That makes restoring GA cache takes more time than it saves.

Hi @oraluben, if I understand correctly, you mean downloading/uploading the `venv` caches from GitHub? The `actions/cache` action does support self-hosted runners for Windows. But the support status for other OSs remains unknown.

I have implemented a manual caching mechanism in commit https://github.com/tile-ai/tilelang/pull/973/commits/de88d356f38972ac95455bc40c74e9053539d3f8 with some code comments. I believe it will resolve your concerns.

The new `venv` cache on the self-hosted runners can be shared between multiple CI runs if they are run concurrently. Previously, we used a hardcoded path for the shared `venv` on self-hosted runners (maybe with a concurrency limit of 1 for self-hosted runners).

.github/workflows/ci.yml
```diff
@@ -1,125 +1,319 @@
...
+      # The GitHub-hosted runners have fast internet access to cache storage,
+      # so we can use the actions/cache action to cache the venv directory.
+      - name: Setup venv cache (GitHub-hosted runners)
+        id: venv-cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            venv
+            .cache
+          key: venv-${{ runner.os }}-${{ runner.arch }}-${{ matrix.python-version }}-${{ matrix.runner.name }}-${{ matrix.runner.toolkit }}-${{ hashFiles('**/requirements*.txt') }}
```
oraluben: Given that `uv pip` has a significant speedup against `pip` by pre-unpack wheel file and use link as much as possible, with cache available, the time of creating venv and install dependencies could be safely ignored.

This also keeps one single cache entity, instead of keep multiple cache archive and significantly reduce complexity.

I propose to replace current setup with uv and use a single directory on each runner as uv cache.
This also works for concurrent run, if there's plan.

This is just part of the change, we might need something like cleaning-up when cache reaches size limit.

```suggestion
      - name: Setup Python via uv
        uses: astral-sh/setup-uv@v6
        with:
          enable-cache: false
          cache-local-path: ${{ runner.tool_cache }}/uv
          activate-environment: true
          python-version: ${{ matrix.python-version }}
```

XuehaiPan: SGTM. Updated.

XuehaiPan:
Seems there is a segmentation fault in the example: https://github.com/tile-ai/tilelang/actions/runs/18415877369/job/52479419949?pr=973#step:14:43

```diff
- Error: Process completed with exit code 139.
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,315 @@
...
+        run: |
+          uv pip install --no-build-isolation -v .
+
+      - name: Run examples with Python ${{ matrix.python-version }} (${{ matrix.runner.toolkit }})
+        if: contains(matrix.runner.toolkit, 'CUDA')
+        run: |
+          cd testing
+          uv run --no-project -m -- \
+            pytest --verbose --durations=0 --cache-clear \
+              -k 'not (test_example_mha_sink_bwd_bhsd or test_example_gqa_sink_bwd_bhsd)' \
```
XuehaiPan: I'm temporarily disabling these examples, which will cause segmentation faults. They need to be fixed in a separate PR because that will be out of the scope of this PR.

XuehaiPan:
cc @LeiWang1999 @oraluben for code review

.github/workflows/ci.yml
LeiWang1999: cc @Rachmanino

.github/workflows/ci.yml
```diff
@@ -0,0 +1,315 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  CMAKE_CONFIGURE_OPTIONS: "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON" # to be updated
```
oraluben: This seems already in cmake file

.github/workflows/ci.yml
Rachmanino: It seems that the problem is caused by incorrect behaviour when lowering  `T.atomic_add`. Note that we introduced regional atomic add for the first time in these 2 examples in a recent pr. @yyttt6 would you mind taking a look?

.github/workflows/ci.yml
XuehaiPan: I renamed `CMAKE_CONFIGURE_OPTIONS` -> `CLANG_TIDY_CMAKE_OPTIONS` for clarity.

.github/workflows/ci.yml
Rachmanino: suppose to be fixed via 883. @XuehaiPan could you try adding these 2 examples to ci workflow? thanks

.github/workflows/dist.yml
```diff
@@ -2,60 +2,78 @@ name: Dist
...
 
 jobs:
   build-wheels:
     strategy:
       matrix:
-        os: [ubuntu-22.04, ubuntu-22.04-arm, macos-16]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, macos-26]
+        python-version:
+          - "3.8"
+          # Wait for tvm fix
```
oraluben: tvm is fine after https://github.com/tile-ai/tvm/pull/12. Now we only enable python 3.8 because we requested a cp38-api3 build. All higher version python should give same file name.

.github/workflows/dist.yml
```diff
@@ -2,60 +2,78 @@ name: Dist
...
+            cuda-version: "12.1"
+          - os: ubuntu-24.04-arm
+            cuda-version: "12.8"
       fail-fast: true
     runs-on: ${{ matrix.os }}
     env:
-      CUDA_VERSION: ${{ matrix.cuda_version }}
-      NO_VERSION_LABEL: ${{ github.event_name != 'release' }}
+      CUDA_VERSION: ${{ matrix.cuda-version || '' }}
+      NO_VERSION_LABEL: ${{ github.event_name == 'release' && 'OFF' || 'ON' }}
```
oraluben, line 56: thanks for the fix!

.github/workflows/dist.yml
```diff
@@ -2,60 +2,78 @@ name: Dist
...
-        compression-level: 0
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          submodules: recursive
+
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1
+        with:
```
oraluben, line 69: Does cibuildwheel work well with ccache? I didn't test it and just enable cache for metal.

XuehaiPan: You're right. It has no effect on Linux. Reverted.

pyproject.toml
```diff
@@ -133,49 +133,43 @@ ignore = [
...
+
+[tool.cibuildwheel.linux]
 repair-wheel-command = [
-  "auditwheel repair --exclude libcuda.so.1 --exclude /usr/local/cuda\\* -w {dest_dir} {wheel}",
-  "pipx run abi3audit --strict --report {wheel}",
+    "auditwheel repair --exclude libcuda.so.1 --exclude '/usr/local/cuda*' -w {dest_dir} {wheel}",
+    "pipx run abi3audit --strict --report {wheel}",
 ]
-
+# Equlivant to `source /opt/rh/gcc-toolset-12/enable`, safe when gcc-toolset-12 is not installed
```
oraluben: This is an outdated comment I forgot to remove

.github/workflows/dist.yml
```diff
@@ -2,60 +2,118 @@ name: Dist
 on:
   schedule:
     # gemini said this is 6:00 china time
-    - cron: '0 22 * * *'
+    - cron: "0 22 * * *"
```
oraluben, line 5: GA have a 2000 minutes limit per month. I think building release on each PR (even just for changes on specific files) would be too often.

LeiWang1999 (dismissed):

.github/workflows/dist.yml
XuehaiPan: GitHub only charges for private repositories. See: https://docs.github.com/en/billing/concepts/product-billing/github-actions

> GitHub Actions usage is **free** for **self-hosted runners** and for **public repositories** that use standard GitHub-hosted runners.

oraluben: It's free but not unlimited: https://docs.github.com/en/actions/reference/limits#storage-limits-for-all-github-hosted-runners

See: https://github.com/tile-ai/tilelang/pull/799#issuecomment-3370226915

oraluben (changes requested):
One concern about GA quota commented above.

.github/workflows/dist.yml
XuehaiPan: > It's free but not unlimited: [docs.github.com/en/actions/reference/limits#storage-limits-for-all-github-hosted-runners](https://docs.github.com/en/actions/reference/limits#storage-limits-for-all-github-hosted-runners)

This means the runtime hard disk in the runner, such as temporary built objects. It is a per-run limit, but not a limit for how many runs we can have.

Additionally, the generated artifacts and logs will be rotated if the total storage reaches or exceeds the 90-day limit. See also: https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization

XuehaiPan: I disabled wheel artifact upload in PR.

oraluben: Sure. I remember hitting some runner limit before but I didn't find other docs about it. This LGTM.

oraluben (approved):

LeiWang1999:
cc @Alex4210987 for the rocm ci issues.

Alex4210987:
@XuehaiPan hi,you should set different env var for each version.
For example,USE_ROCM=True, for rocm ci.
otherwise, the tests would fail.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,388 @@
...
+          fi
+          export UV_INDEX="${PIP_EXTRA_INDEX_URL}"
+          export CLANG_TIDY_CMAKE_OPTIONS="${CLANG_TIDY_CMAKE_OPTIONS} -DUSE_ROCM=ON"
+
+          echo "USE_ROCM=ON" | tee -a "${GITHUB_ENV}"
+          echo "ROCM_VERSION=${ROCM_VERSION}" | tee -a "${GITHUB_ENV}"
+          echo "ROCM_VERSION_MAJMIN=${ROCM_VERSION_MAJMIN}" | tee -a "${GITHUB_ENV}"
+          echo "ROCM_VERSION_MAJMIN_NODOT=${ROCM_VERSION_MAJMIN_NODOT}" | tee -a "${GITHUB_ENV}"
+          echo "PIP_EXTRA_INDEX_URL=${PIP_EXTRA_INDEX_URL}" | tee -a "${GITHUB_ENV}"
+          echo "UV_INDEX=${UV_INDEX}" | tee -a "${GITHUB_ENV}"
```
XuehaiPan, line 204: > For example,USE_ROCM=True, for rocm ci.

@Alex4210987 It is defined here.

LeiWang1999:
LGTM, thanks for your great job @XuehaiPan ! and thanks for @oraluben 's review.

PR #1019, opened 2025-10-13
[Feature]: Add test for atomicadd auto vectorize and remove useless code

yyttt6:
## Summary by CodeRabbit

- **New Features**
  - Added an element-wise tile-level atomic-add intrinsic.

- **Refactor**
  - Switched atomic-add vectorization to intrinsic-based detection, enabling x2/x4 vectorized variants with safe fallbacks to scalar.
  - Simplified layout handling for atomic-add paths (layout inference no longer delegated).

- **Chores**
  - Updated transformation flow and messages to recognize the new intrinsic and emit clearer warnings for unsupported vectorization patterns.

### DIFF
--- a/src/transform/atomicadd_vectorize.cc
+++ b/src/transform/atomicadd_vectorize.cc
@@ -225,10 +225,8 @@ private:
         legal_vectorize = false;
     }
     if (legal_vectorize) {
-      const BufferLoad dst_node =
-          Downcast<BufferLoad>(node->args[0].as<BufferLoadNode>());
-      const BufferLoad value_node =
-          Downcast<BufferLoad>(node->args[1].as<BufferLoadNode>());
+      const BufferLoad dst_node = Downcast<BufferLoad>(node->args[0]);
+      const BufferLoad value_node = Downcast<BufferLoad>(node->args[1]);
 
       Call address_of_dst =
           Call(DataType::Handle(), builtin::address_of(), {dst_node});


src/op/atomic_add.cc
```diff
@@ -288,7 +287,7 @@ For AtomicAddNode::MakeSIMTLoop(arith::Analyzer *analyzer) const {
   new_args.push_back(src_value);
 
   Call atomicadd_call =
-      tvm::tir::Call(dst->dtype, builtin::call_extern(), new_args);
+      tvm::tir::Call(dst->dtype, tvm::tl::tl_atomicadd_elem_op(), new_args);
```
LeiWang1999: replace with `tl_atomicadd_elem_op()`

src/op/builtin.h
```diff
@@ -501,6 +501,13 @@ TVM_DLL const Op &initialize_descriptor();
  * tilelang.
  */
 TVM_DLL const Op &increase_descriptor_offset();
+/*!
+ * \brief tilelang intrinsic for element-wise atomic addition.
+ *
+ *  This op is used to represent an element-wise atomic add operation in
+ * tilelang.
+ */
+TVM_DLL const Op &tl_atomicadd_elem_op();
```
LeiWang1999: I think we can remove the prefix `tl` as we already in tl namespace

LeiWang1999 (changes requested):
Overall LGTM, but some comments :)

LeiWang1999 (approved):

PR #1020, opened 2025-10-13
[CI] Removes redundant environment variable 

Cunxiao2002:
## Summary by CodeRabbit

* **Chores**
  * Updated continuous integration configuration by removing the UV_INDEX_URL environment variable from formatting and NVIDIA build/test jobs. This aligns builds with default package index settings and simplifies the workflow. No changes to runtime behavior, features, or public interfaces. Build and test steps otherwise remain the same. Expect identical results for consumers, with reduced maintenance overhead in our CI.


LeiWang1999:
LGTM.

PR #1021, opened 2025-10-13
[BugFix] fix torch MPS  check

Cunxiao2002:
I discovered that in PyTorch versions below 2.4, there's no torch.mps.is_available() interface, only torch.backends.mps.is_available(). To support older PyTorch versions, we need to modify this interface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of Apple Metal (MPS) support on macOS, resulting in more reliable automatic device selection. This reduces incorrect CPU/GPU fallbacks and applies MPS-specific behavior only when supported, improving stability and reducing errors for users with compatible hardware. CUDA and CPU behavior remains unchanged.

### DIFF
--- a/tilelang/utils/device.py
+++ b/tilelang/utils/device.py
@@ -1,7 +1,7 @@
 import torch
 
 IS_CUDA = torch.cuda.is_available()
-IS_MPS = torch.mps.is_available()
+IS_MPS = torch.backends.mps.is_available()
 
 
 def get_current_device():


oraluben:
Currently torch's mps backend rely on api [since 2.7](https://github.com/pytorch/pytorch/pull/148972). Refactor to support lower version is planned but I don't have specific time yet. One solution is to add something like `torch>=2.7; platform='Darwin'` in requirements.

Cunxiao2002:
@oraluben  Thank you for your response. I understand now. I will close this PR and look forward to the refactoring!

PR #1032, opened 2025-10-14
[Bugfix] Recover code for flexible parallel

LeiWang1999:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of loop-to-thread mappings to detect invalid inner-variable mappings earlier and raise clearer errors.
  * Added checks for cross-thread access and direct store-from-load patterns to ensure correct predicate insertion and prevent incorrect parallel updates.

* **Refactor**
  * Adjusted inference flow to fail fast on conflicting mappings and apply additional bounding predicates when needed, increasing robustness of parallel layout computation.


LeiWang1999:
Local Test can pass, but it's weird for the failed ci test.

PR #1040, opened 2025-10-15
[Enhancement] Remove constraint requiring last dimension stride to be 1

LJC00118:
After #798, we can remove this constraint.

## Summary by CodeRabbit

* **New Features**
  * Accept tensors with a non‑unit stride on the last dimension, reducing errors for non‑contiguous layouts and enabling more flexible memory layouts.

* **Tests**
  * Added CUDA vectorization tests that validate kernel vector-width selection and numerical correctness across varied stride configurations.

### DIFF
--- a/testing/python/language/test_tilelang_language_vectorize.py
+++ b/testing/python/language/test_tilelang_language_vectorize.py
@@ -3,8 +3,7 @@ import tilelang.testing
 import tilelang.language as T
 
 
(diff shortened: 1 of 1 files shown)


LeiWang1999:
Would be better if we can add a simple test for this pr.

LeiWang1999:
LGTM.

PR #1043, opened 2025-10-15
[CI] Fix ROCm CI

XuehaiPan:
## Summary by CodeRabbit

* **Chores**
  * Updated self-hosted GPU test environment from Nightly ROCm 7.0 to ROCm 6.3.

* **Tests**
  * ROCm test failures now fail the pipeline instead of being ignored.
  * Removed a recurring erroneous error log to reduce CI noise.
  * Added a termination-time check that causes CI to fail if no tests are collected.


XuehaiPan:
cc @LeiWang1999 @Alex4210987

PR #1044, opened 2025-10-15
[CI][Lint] Retire `format.sh` and add `clang-tidy` to GHA workflow

XuehaiPan:
## Summary by CodeRabbit

- Chores
  - CI now runs clang-tidy on Linux with fresh CMake builds, automatic source discovery, artifact cleanup, improved error reporting, dynamic runner selection, and optional auto-fixes.
  - Reworked formatting/orchestration into a multi-mode formatter with pre-commit integration and merge-base-based changed-file handling.
  - Added CMake build outputs to .gitignore and bumped clang-format/clang-tidy/ruff versions; clang-tidy config layout restructured.

- Bug Fixes
  - Adjusted atomic operation fast-path conditions to ensure correct handling for reduced-type cases.

- Style
  - Widespread non-functional formatting and whitespace changes plus local const-reference refinements.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -84,7 +84,7 @@ elif [[ "${#FILES[@]}" -gt 0 ]]; then
 fi
 
 # If pre-commit is not installed, install it.
-if python3 -m pre_commit --version &>/dev/null; then
+if ! python3 -m pre_commit --version &>/dev/null; then
     python3 -m pip install pre-commit
 fi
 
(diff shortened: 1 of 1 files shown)


format.sh
LeiWang1999, line 1: would be better to maintain format.p referable, as it is more convenient though provides similar functionality to pre-commit. Developers should not be forced to install `pre-commit` when they're developing; instead, we could modify the content of `format.sh` to install and run pre-commit directly.

XuehaiPan: How about moving `format.sh` as a `Makefile` with the following target (Unix-like systems always have `make` executable):

```bash
make format
```

We can add more targets like:

```bash
make install
make docs
make test
make clang-tidy
make clean
...
```

XuehaiPan: I opened a new PR to resolve this.

- #1083

LeiWang1999: LGTM.

PR #1052, opened 2025-10-16
Automatically initialize submodule if missing

LeiWang1999:
This pull request adds logic to the project's `CMakeLists.txt` to ensure that git submodules are initialized automatically when configuring the build. This helps prevent build errors due to missing submodule dependencies and improves the developer experience by automating a common setup step.

Build system improvements:

* Added a check for the existence of `.gitmodules` and `.git`, and if present, attempts to initialize and update git submodules using CMake's `execute_process` with `git submodule update --init --recursive`. If git is not found or the submodule update fails, a clear fatal error message is displayed to guide the user.

## Summary by CodeRabbit

* **Chores**
  * Enhanced build configuration to automatically initialize Git submodules during setup with improved error handling and clearer guidance when Git is unavailable or submodule initialization fails.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,30 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitmodules" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+  find_package(Git QUIET)
+  if(Git_FOUND)
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      RESULT_VARIABLE TILELANG_GIT_SUBMODULE_RESULT
+    )
+    if(NOT TILELANG_GIT_SUBMODULE_RESULT EQUAL 0)
+      message(
+        FATAL_ERROR
+          "Failed to initialize git submodules. Please run "
+          "`git submodule update --init --recursive` and re-run CMake."
+      )
+    endif()
+  else()
+    message(
+      FATAL_ERROR
+        "Git is required to initialize TileLang submodules. "
+        "Please install git or fetch the submodules manually."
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -10,6 +10,30 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitmodules" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+  find_package(Git QUIET)
+  if(Git_FOUND)
```
oraluben, line 15: Building tvm will fail if git is not found, we can safely assume git is installed

CMakeLists.txt
```diff
@@ -10,6 +10,30 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitmodules" AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git")
```
XuehaiPan, line 13: FYI, you could use [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html) to enable building without the `.git` directory. This will allow us to completely remove `.gitmodules` and the `3rdparty` directory.

An example: https://github.com/metaopt/optree/blob/4b2aad3276ec391cc68b9445cf18d163dea8aa9c/CMakeLists.txt#L258-L283

CMakeLists.txt
LeiWang1999: I think this might have happened before the TVM build?

CMakeLists.txt
LeiWang1999: Thanks, I'll take alook

LeiWang1999: I think at least the `3rdparty/tvm` directory isn’t very stable — we occasionally need to update it. Managing it with Git would make things much more convenient. Once it becomes more stable, we can consider switching to that approach.

CMakeLists.txt
oraluben: Ah I see, you're right

PR #1054, opened 2025-10-17
[Enhancement] Improve CUDA compiler detection in CMake

LJC00118:
## Summary by CodeRabbit

* **Chores**
  * Improved CUDA build configuration: the build now explicitly assigns and configures the CUDA compiler path when CUDA is detected, reducing manual setup and increasing build reliability and stability on systems with CUDA support.


LeiWang1999:
likely helpful when `cuda/bin` is not in `PATH`.

PR #1073, opened 2025-10-20
[Layout] Utilizing IsEqual instead of StructuralEqual

LeiWang1999:
This pull request updates the way buffer layout equality is checked in the `BufferUseDefCollector` class within `src/transform/layout_inference.cc`. The change replaces the use of the general `StructuralEqual` function with the more specific `IsEqual` method for layout comparison.

Improvement to layout comparison logic:

* Changed buffer layout equality check from using `StructuralEqual()` to the more precise `layout->IsEqual()` method in the `BufferUseDefCollector` class, ensuring that layout objects are compared using their specialized equality logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved buffer layout validation accuracy during runtime checks.

### DIFF
--- a/src/transform/layout_inference.cc
+++ b/src/transform/layout_inference.cc
@@ -159,7 +159,7 @@ public:
           }
         }
         // If already in map, ensure they are structurally equal
-        ICHECK(StructuralEqual()(layout, layout_map[buffer]))
+        ICHECK(layout->IsEqual(layout_map[buffer].get()))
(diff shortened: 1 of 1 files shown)


LeiWang1999:
StructuralEqual is too strict for this case: 
6e1d91efef7ab47903dcbbc68b8b715a

Those two layouts is fully replicated and they can have no conflicts.

PR #1090, opened 2025-10-21
[BugFix] Correct direct copy from bf16 to fp8

Cunxiao2002:
## Summary by CodeRabbit

* **Improvements**
  * Added two FP8 reduced-precision float formats with bfloat16→FP8 conversion support for CUDA paths.
  * Unified type-mapping so GEMM kernels resolve runtime numeric types via the new mapping layer, reducing type-mismatch risk and improving precision handling.

* **Chores**
  * Minor import and code hygiene cleanup.


LeiWang1999:
We're good to go if we can resolve this conflict

PR #1116, opened 2025-10-23
[Feature]:Add device assert

yyttt6:
## Summary by CodeRabbit

* **New Features**
  * Device-side assertions for CUDA kernels with optional messages; available via the language API and active when CUDA debug support is present.
  * Additional language utilities (atomic ops, reshape/view helpers, loop break, etc.) exported for user code.

* **Tests**
  * New tests validating assertion behavior in triggering and non-triggering scenarios.

* **Documentation**
  * Updated docstrings describing assertion behavior and CUDA availability notes.

### DIFF
--- a/tilelang/language/print.py
+++ b/tilelang/language/print.py
@@ -143,7 +143,8 @@ _IS_CUDA_AVAILABLE = check_cuda_availability()
 def device_assert(condition: tir.PrimExpr, msg: str = ""):
     """
     Device-side assert emulation.
-    In debug mode (TILELANG_DEBUG != 0), emits a device-side assert call on CUDA targets only.
+    Emits a device-side assert call on CUDA targets when CUDA is available.
+    The assert is always enabled and cannot be disabled at runtime.
     """
     if _IS_CUDA_AVAILABLE:
         if msg == "":


src/tl_templates/cuda/debug.h
```diff
@@ -257,3 +257,12 @@ __device__ void debug_print_buffer_value<int16_t>(const char *msg,
          msg, blockIdx.x, blockIdx.y, blockIdx.z, threadIdx.x, threadIdx.y,
          threadIdx.z, buf_name, index, (int32_t)var);
 }
+
+__device__ void device_assert(bool cond) { assert(cond); }
```
LeiWang1999: Why don't we use TL_DEVICE?

testing/python/debug/test_device_assert.py
```diff
@@ -0,0 +1,47 @@
...
+                T.device_assert(tid == tid)
+
+        jit_kernel = tilelang.compile(program, target="cuda")
+        profiler = jit_kernel.get_profiler()
+        profiler.run_once()
+    finally:
+        if prev is None:
+            os.environ.pop("TILELANG_DEBUG", None)
+        else:
+            os.environ["TILELANG_DEBUG"] = prev
```
LeiWang1999: why we need `TILELANG_DEBUG` and likely I didn't find any configs to exclude this file in ci?

PR #1171, opened 2025-11-01
[Release] Unify local build scripts to use `cibuildwheel` and reduce size of sdist

oraluben:
1. sdist: Update sdist includes, reduce size to meet the 100m limit of pypi;
2. GA: List artifacts for pull requests with title `"[Release]"`, then we can download built wheel/sdist from artifacts, instead of use qemu on local machine. This also produces macos wheels, example: https://github.com/tile-ai/tilelang/actions/runs/19007774631?pr=1171 @LeiWang1999 ;
3. sdist: Generates a `.git_commit.txt` in sdist. `pip` do not allow a sdist to have different version with the installed package (we cannot have a sdist named `tilelang-0.1.6.post2+gitxxxx.tar.gz` and install a package named `tilelang==0.1.6.post2+cu128.gitxxxx`). Since sdist should not contain toolchain info, we need to store its git commit separately. When installed from sdist, the git info will be read from that file.
4. scripts: delete some file, use `cibuildwheel` to generate wheels for linux. However this seems not necessary, I hope Github Actions can eliminated most of local builds.

cc @XuehaiPan 

## Summary by CodeRabbit

* **Chores**
  * Removed legacy manifest file and moved to explicit vendored subcomponent packaging in pyproject.
  * Simplified build/distribution: removed multi-arch Docker orchestration, using direct build-tool invocations for local/CI builds.
  * Updated container builder bases and per-architecture CUDA defaults for manylinux builds.
  * Persisted version stamping to a git-pin file (now ignored) and fallback to "gitunknown" when commit info is unavailable.
  * CI: sdist/wheels/artifacts also produced for PRs titled “[Release]”.


oraluben:
is `3rdparty/tvm/src` necessary for wheels?

LeiWang1999:
> is `3rdparty/tvm/src` necessary for wheels?

yes, this should be included

pyproject.toml
```diff
@@ -60,19 +60,30 @@ metadata.version.provider = "version_provider"
...
+# build.verbose = true
+# logging.level = "DEBUG"
+
 [tool.scikit-build.sdist]
-# See MANIFEST.in for details
 include = [
-    "VERSION",
-    "LICENSE",
+    "./VERSION",
+    "./_git_commit.txt",
```
XuehaiPan: nit: You will not need the `./` folder prefix here. Also, in Unix-like systems, the misc file is better to be prefixed with `.`, not `_`.

oraluben: > nit: You will not need the `./` folder prefix here.

`scikit-build-core`'s include/exclude mechanism is not that straightforward. The `LICENSE` rule will include `LICENSE` file in other excluded folders, e.g. `.venv`. So I have to explicitly include the files under project root.

> Also, in Unix-like systems, the misc file is better to be prefixed with `.`, not `_`.

Since files started with `.` will be hidden by default, that might cause confusion to developers. For example, `setuptools-scm` uses `pkg/_version.py` as the example for its auto-generated file: https://github.com/pypa/setuptools-scm/

XuehaiPan: > For example, `setuptools-scm` uses `pkg/_version.py`

`.` is not a valid character in Python module names, so they have to use `_`, while the new file is a `txt` file.

XuehaiPan: > Since files started with `.` will be hidden by default, that might cause confusion to developers.

The file is auto-generated, and they are not intended to modify the file, right?

oraluben: > `.` is not a valid character in Python module names, so they have to use `_`, while the new file is a `txt` file.

That makes sense.

pyproject.toml
XuehaiPan, line 216: Out of curiosity, is there any reason not to use `manylinux_2_28` for x86_64? GLIBC 2.28 was released 7 years ago. The EOL-ed Ubuntu 20.04 has GLIBC 2.31.

oraluben: The only reason is that torch migrated from manylinux 2.17 to manylinux 2.28 since 2.6. I guess torch 2.4/2.5 is not that old to be dropped.

XuehaiPan: The shared libraries in our wheel are not linked against `torch`'s, right?

oraluben: That's correct. We only align with torch's glibc version here.

XuehaiPan: > That's correct. We only align with torch's glibc version here.

I think there will not be any runtime effects as they are not linked against each other. And there will be only one libc library used in the process, which is very likely 2.28+.

If you persist, could you add a TODO comment in `pyproject.toml` for this choice?

oraluben: I'm fine with that, since that change is quite independent and small, maybe we can open an issue to see if that's ok for other users cc @LeiWang1999

LeiWang1999: both of them lgtm.

LeiWang1999 (approved):

PR #1184, opened 2025-11-04
[CI] Enable `ccache` for CIBW on Linux

### LINKED ISSUE #1155: [Feature Request][CI] Enable `ccache` or `sccache` for `cibuildwheel` for `manylinux`
### Required prerequisites

### Motivation

CIBW runs natively on macOS, allowing us to use `ccache` directly in our dist CI. However, the Linux wheels are being built inside Docker containers without build caches, which is very slow.

### Solution

As per the issue title, enable `ccache` or `sccache` and store/restore the cache for wheel building for Linux. cc @oraluben

### Alternatives

_No response_

### Additional context

Xref:

- pypa/cibuildwheel#1030
- NVIDIA/cuda-python#1156

oraluben:
Closes #1155

Reduce build time of linux wheels from ~30m to ~3m (all cache hit, env setup included)

Mostly copied from https://github.com/pypa/cibuildwheel/issues/1030#issuecomment-2443820382

cc @XuehaiPan 

## Summary by CodeRabbit

* **Chores**
  * Improved cross-platform build caching for faster, more reliable builds; unified cache behavior, added Linux host-cache handling, disabled timestamped caches, and record project base directory on Apple hosts. Updated ignore rules to use a dot-prefixed commit file name.

* **Tests**
  * CI adjusted to use built sdist/wheel artifacts for test runs, simplifying runner setup and improving reproducibility.

### DIFF
--- a/examples/flash_attention/example_gqa_bwd_tma_reduce.py
+++ b/examples/flash_attention/example_gqa_bwd_tma_reduce.py
@@ -5,6 +5,8 @@ import tilelang.language as T
 from tilelang.contrib import nvcc
 import argparse
 
+tilelang.disable_cache()
+
 
 @tilelang.jit(
     out_idx=[3, 4], pass_configs={
@@ -44,7 +46,9 @@ def flashattn_fwd(batch, heads, seq_len, dim_qk, dim_v, is_causal, block_M, bloc
             T.copy(Q[bz, bx * block_M:(bx + 1) * block_M, by, :], Q_shared)
             T.fill(acc_o, 0)
             T.fill(logsum, 0)
-            T.fill(scores_max, -T.infinity(accum_dtype))
+            # Warning: in causal/varlen/unaligned seqlen scenarios, the -inf will cause undefined behavior in exp ops
+            # We should set it to negative large number instead
+            T.fill(scores_max, T.Cast(accum_dtype, -1e30))
             loop_range = (
                 T.ceildiv(
                     (bx + 1) * block_M, block_N) if is_causal else T.ceildiv(seq_len, block_N))
@@ -53,7 +57,7 @@ def flashattn_fwd(batch, heads, seq_len, dim_qk, dim_v, is_causal, block_M, bloc
                 if is_causal:
                     for i, j in T.Parallel(block_M, block_N):
                         acc_s[i, j] = T.if_then_else(bx * block_M + i >= k * block_N + j, 0,
-                                                     -T.infinity(acc_s.dtype))
+                                                     T.Cast(accum_dtype, -1e30))
                 else:
                     T.clear(acc_s)
                 T.gemm(Q_shared, K_shared, acc_s, transpose_B=True, policy=T.GemmWarpPolicy.FullRow)
@@ -265,17 +269,17 @@ def flashattn_bwd_atomic_add(batch,
 @tilelang.jit(pass_configs={
     tilelang.PassConfigKey.TL_ENABLE_FAST_MATH: True,
 })
-def flashattn_bwd_split(batch,
-                        heads,
-                        seq_len,
-                        dim_qk,
-                        dim_v,
-                        is_causal,
-                        block_M,
-                        block_N,
-                        threads=256,
-                        num_stages=2,
-                        groups=1):
+def flashattn_bwd_split_novarlen(batch,
+                                 heads,
+                                 seq_len,
+                                 dim_qk,
+                                 dim_v,
+                                 is_causal,
+                                 block_M,
+                                 block_N,
+                                 threads=256,
+                                 num_stages=2,
+                                 groups=1):
     sm_scale = (1.0 / dim_qk)**0.5
     scale = (1.0 / dim_qk)**0.5 * 1.44269504  # log2(e)
     head_kv = heads // groups
@@ -424,7 +428,7 @@ class _attention(torch.autograd.Function):
             kernel(q, k, v, do, lse, delta, dq, dk, dv)
             dq, dk, dv = mod_post(dq, dk, dv)
         else:
-            kernel = flashattn_bwd_split(
+            kernel = flashattn_bwd_split_novarlen(
                 BATCH,
                 H,
                 N_CTX,
--- a/examples/flash_attention/example_gqa_bwd_tma_reduce_varlen.py
+++ b/examples/flash_attention/example_gqa_bwd_tma_reduce_varlen.py
@@ -7,6 +7,8 @@ import argparse
 from einops import rearrange, repeat
 from bert_padding import pad_input, unpad_input
 
+# tilelang.disable_cache()
+
 
 def generate_random_padding_mask(max_seqlen, batch_size, device, mode="random"):
     assert mode in ["full", "random", "third"]
@@ -29,6 +31,7 @@ def generate_random_padding_mask(max_seqlen, batch_size, device, mode="random"):
 def flashattn_fwd(batch,
                   total_q,
                   total_kv,
+                  N_CTX,
                   heads,
                   max_seq_len,
                   dim_qk,
@@ -54,7 +57,7 @@ def flashattn_fwd(batch,
             cu_seqlens_q: T.Tensor([batch + 1], "int32"),  # type: ignore
             cu_seqlens_k: T.Tensor([batch + 1], "int32"),  # type: ignore
             Output: T.Tensor(o_shape, dtype),  # type: ignore
-            lse: T.Tensor([total_q, heads], accum_dtype),  # type: ignore
+            lse: T.Tensor([batch, heads, N_CTX], accum_dtype),  # type: ignore
     ):
         with T.Kernel(T.ceildiv(max_seq_len, block_M), heads, batch, threads=256) as (bx, by, bz):
             Q_shared = T.alloc_shared([block_M, dim_qk], dtype)
@@ -86,7 +89,9 @@ def flashattn_fwd(batch,
 
             T.fill(acc_o, 0.0)
             T.fill(logsum, 0.0)
-            T.fill(scores_max, -T.infinity(accum_dtype))
+            # Warning: in causal/varlen/unaligned seqlen scenarios, the -inf will cause undefined behavior in exp ops
+            # We should set it to negative large number instead
+            T.fill(scores_max, T.Cast(accum_dtype, -1e30))
             loop_range = T.ceildiv(k_current_seqlen, block_N)
             for k in T.Pipelined(loop_range, num_stages=1):
                 for i, d in T.Parallel(block_N, dim_qk):
@@ -100,12 +105,12 @@ def flashattn_fwd(batch,
                         acc_s[i, j] = T.if_then_else((bx * block_M + i >= k * block_N + j) and
                                                      (bx * block_M + i < q_current_seqlen and
                                                       k * block_N + j < k_current_seqlen), 0,
-                                                     -T.infinity(acc_s.dtype))
+                                                     T.Cast(accum_dtype, -1e30))
                 else:
                     for i, j in T.Parallel(block_M, block_N):
                         acc_s[i, j] = T.if_then_else(
                             bx * block_M + i < q_current_seqlen and
-                            k * block_N + j < k_current_seqlen, 0, -T.infinity(acc_s.dtype))
+                            k * block_N + j < k_current_seqlen, 0, T.Cast(accum_dtype, -1e30))
                 T.gemm(Q_shared, K_shared, acc_s, transpose_B=True, policy=T.GemmWarpPolicy.FullRow)
                 for i, d in T.Parallel(block_N, dim_v):
                     if k * block_N + i < k_current_seqlen:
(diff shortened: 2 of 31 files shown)


.github/workflows/dist.yml
```diff
@@ -73,18 +73,24 @@ jobs:
         uses: hendrikmuhs/ccache-action@v1
         with:
           create-symlink: true
-          key: ccache-${{ runner.os }}-${{ runner.arch }}
+          key: ccache-sdist-${{ runner.os }}-${{ runner.arch }}
           evict-old-files: "7d"
 
       - name: Test SDist buildable
+        # We extract sdist to same folder (./build) and install package from there
+        # to make sure ccache don't get a lot of miss because of different working dir.
```
XuehaiPan: This might be caused by the GHA cache-key collision. You should add relevant information to the cache key to avoid collection.

.github/workflows/dist.yml
```diff
@@ -126,13 +132,11 @@ jobs:
...
 
-      # NB: CIBW builds wheels in containers on Linux
-      - name: Setup ccache (macOS only)
-        if: runner.os == 'macOS'
+      - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1
         with:
           create-symlink: true
-          key: ccache-${{ runner.os }}-${{ runner.arch }}-${{ matrix.python-version }}-${{ matrix.target.toolkit }}
+          key: ccache-${{ runner.os }}-${{ runner.arch }}
```
XuehaiPan: Please revert this. See https://github.com/tile-ai/tilelang/pull/1184/files#r2488892874

.github/workflows/dist.yml
oraluben: https://github.com/tile-ai/tilelang/pull/1184/files/f69e458d3da7d1e4e87b978f8934d85b09d4e6e3#diff-3135dbaba3b8a311ffab711e7791abf763ffeae8c00aa3cf58ee43e4c930203dR76

.github/workflows/dist.yml
oraluben: This should be fine. `${{ runner.os }}-${{ runner.arch }}` resolves to `Linux-X86` or `Linux-ARM64`. While most part of tvm is independent to python/CUDA, most part of the cache should be reusable.

.github/workflows/dist.yml
XuehaiPan: Since the ccache problem is due to the malformed GHA cache. You will not need to manually untar the SDist and do a manual build below. Just pip install from the SDist.

oraluben: `pip install sdist.tar.gz` will extract source in different tmp dir and will make ccache (sccache as well) not able to cache compilation.

.github/workflows/dist.yml
XuehaiPan: The problem is not about downloading and reusing the cache. It is a potential problem that multiple workflows will upload/push the cache with the same cache key. Since we do not change `matrix.python-version` and `matrix.target.toolkit` very often, it will not be a performance problem with a detailed cache key.

oraluben: Currently we're running exactly one job for each `${{ runner.os }}-${{ runner.arch }}` so that should be fine?

XuehaiPan: Since there is only one job for each plat-arch. So the argument for this modification in this PR for cache reuse does not hold, right?

.github/workflows/dist.yml
XuehaiPan: Solution given by Claude, I haven't tested it yet:

image

.github/workflows/dist.yml
oraluben: different python/cuda version should be orthogonal with ccache, i.e. even when we build for different python versions at the same time, those builds can benefits from previous cache, no matter the python version used to generate that cache. There's no reason to store cache for different python version/cuda version.

About the collision issue you mentioned, the action already add timestamp on each uploaded artifacts so I don't see any issue there.
```
Post job cleanup.
ccache stats
/usr/bin/ccache --evict-older-than 7d
Save cache using key "ccache-ccache-Linux-X64-2025-11-04T06:57:23.439Z".
```

.github/workflows/dist.yml
oraluben: `CCACHE_NOHASHDIR` looks helpful, trying

.github/workflows/dist.yml
XuehaiPan: Let me clarify, for example, when we support more archs, such as `s390x` / `riscv64` with QEMU, the build cache is not usable between archs. While the host Linux runner is still X64, that will give the same cache key prefix. An invalid cache will surpass the valid cache due to a newer timestamp.

------

> those builds can benefits from previous cache

As I mentioned earlier in https://github.com/tile-ai/tilelang/pull/1184#discussion_r2488938500:

> Since we do not change `matrix.python-version` and `matrix.target.toolkit` very often, it will not be a performance problem with a detailed cache key.

.github/workflows/dist.yml
XuehaiPan: I guess we can set the following in `CMakeLists.txt`:

```cmake
set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "C compiler launcher")
set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "CXX compiler launcher")
set(CMAKE_CUDA_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "CUDA compiler launcher")
```

oraluben: > I guess we can set the following in `CMakeLists.txt`:
> 
> ```cmake
> set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "C compiler launcher")
> set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "CXX compiler launcher")
> set(CMAKE_CUDA_COMPILER_LAUNCHER "${CCACHE_PROGRAM};base_dir=${CMAKE_SOURCE_DIR}" CACHE STRING "CUDA compiler launcher")
> ```

This seems a little bit complicated.

Simply setting `CCACHE_NOHASHDIR` and `CCACHE_BASEDIR` for sdist doesn't work for some reason, I'll try to dig into it locally.
It's confusing that the cache generating from building sdist works for next wheel building, but not for next sdist job.

.github/workflows/dist.yml
oraluben: > for example, when we support more archs, such as `s390x` / `riscv64` with QEMU, the build cache is not usable between archs

While the current `cibuildwheel` config supports cross compiling with qemu, I don't think our GA workflows will involve that.

Even if we want to cover that case, `matrix.python-version` and `matrix.target.toolkit` won't help.

oraluben:
Even if [using a hard-coded path](https://github.com/tile-ai/tilelang/pull/1184/files#diff-3135dbaba3b8a311ffab711e7791abf763ffeae8c00aa3cf58ee43e4c930203dR96-R104), ccache [doesn't work](https://github.com/tile-ai/tilelang/actions/runs/19070915624/job/54473213915?pr=1184) when building from sdist.

Given that (and the complexity trying to make it work), I think we may just generate a sdist on GA runner, and maybe build from sdist on self-hosted runners? @XuehaiPan do you have any suggestion here?

.github/workflows/dist.yml
oraluben: not working :( See comments above

XuehaiPan:
> Even if [using a hard-coded path](https://github.com/tile-ai/tilelang/pull/1184/files#diff-3135dbaba3b8a311ffab711e7791abf763ffeae8c00aa3cf58ee43e4c930203dR96-R104), ccache [doesn't work](https://github.com/tile-ai/tilelang/actions/runs/19070915624/job/54473213915?pr=1184) when building from sdist.
> 
> Given that (and the complexity trying to make it work)

The solution I posted in https://github.com/tile-ai/tilelang/pull/1184#discussion_r2489246467 works perfectly for me.

```console
$ rm -rf "$(ccache --get-config cache_dir)"
$ ccache --show-stats
Local storage:
  Cache size (GiB): 0.0 / 5.0 ( 0.00%)
$ uv build --sdist
$ uv cache clean
$ uv pip uninstall tilelang
$ uv pip install -v dist/tilelang-*.tar.gz
$ ccache --show-stats
Cacheable calls:    631 / 631 (100.0%)
  Hits:               0 / 631 ( 0.00%)
    Direct:           0
    Preprocessed:     0
  Misses:           631 / 631 (100.0%)
Local storage:
  Cache size (GiB): 0.1 / 5.0 ( 1.61%)
  Hits:               0 / 631 ( 0.00%)
  Misses:           631 / 631 (100.0%)

$ uv cache clean
$ uv pip uninstall tilelang
$ uv pip install -v dist/tilelang-*.tar.gz
$ ccache --show-stats
Cacheable calls:    1262 / 1262 (100.0%)
  Hits:              631 / 1262 (50.00%)
    Direct:          631 /  631 (100.0%)
    Preprocessed:      0 /  631 ( 0.00%)
  Misses:            631 / 1262 (50.00%)
Local storage:
  Cache size (GiB):  0.1 /  5.0 ( 1.61%)
  Hits:              631 / 1262 (50.00%)
  Misses:            631 / 1262 (50.00%)
```

CMakeLists.txt
```diff
@@ -41,7 +41,13 @@ endif()
 
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
-  message(STATUS "Using ccache: ${CCACHE_PROGRAM}")
+  message(STATUS "Using ccache: ${CCACHE_PROGRAM} with base_dir=${CMAKE_SOURCE_DIR}")
+  if(APPLE)
+    # passing configs like `ccache base_dir=/xxx gcc ...` is supported
+    # (likely) since ccache 4.x, which has been provided by homebrew.
+    # For linux, setting ccache without base_dir already works well.
```
XuehaiPan: Seems this is due to the build directory being a fixed location inside the container: https://github.com/tile-ai/tilelang/actions/runs/19092933810/job/54546981073?pr=1184#step:5:845

```txt
  -- Using ccache: /usr/bin/ccache with base_dir=/project
  -- Enable CUDA support by default.
  -- Found CUDAToolkit: /usr/local/cuda/targets/x86_64-linux/include (found version "12.1.105")
  -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
  -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
  -- Found Threads: TRUE
  -- Forbidding undefined symbols in shared library, using -Wl,--no-undefined on platform Linux
  -- Build with RPC support...
  -- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY) 
  -- Build Alloc alignment set to 64
  -- CMAKE_C_COMPILER_LAUNCHER already defined.  Not using ccache.
```

However, without setting the `base_dir` for `ccache` on Linux, we cannot reuse the cache between different workflows.

oraluben: Currently I already set different key for caches from different workflow. You're right about this:

> It is a potential problem that multiple workflows will upload/push the cache with the same cache key.

In the case where some workflow is aborted during cache upload, the key will be "locked" for ~24h and will prevent new cache being uploaded: https://github.com/actions/cache/issues/485#issuecomment-744145040

Both `manylinux2014` and `manylinux_2_28` provides ccache 3.7 which does not support that form. So for our current GA with `cibuildwheel` needs to set their `base_dir` during setup. I can do that, but we're only using linux runners when building wheel, so it should be fine?

XuehaiPan: > Both `manylinux2014` and `manylinux_2_28` provides ccache 3.7 which does not support that form.

Could you add a comment for this? And also, attaching that `cibuildwheel` builds the wheel with a fixed location.

Maybe we can find another way to set the `base_dir` via custom command `ccache -o ...` with `if("$ENV{CIBUILDWHEEL}")` or [`VERSION_GREATER_EQUAL`](https://cmake.org/cmake/help/latest/command/if.html#version-greater-equal). But I think that would be too complicated, and the current state looks fine to me.

XuehaiPan:
I have cleared the GHA cache:

```console
$ gh cache delete --all
✓ Deleted 102 caches from tile-ai/tilelang
```

I will trigger the Dist workflow twice; let's see if it works.

oraluben:
run time after cache's been cleared just for record: https://github.com/tile-ai/tilelang/actions/runs/19094657395/usage

XuehaiPan (dismissed):

XuehaiPan (approved):

LeiWang1999 (approved):

PR #1189, opened 2025-11-04
[Feature] Enhance fill operation to support various buffer types

LeiWang1999:
- Added support for `BufferLoad` in the `fill` function to handle different buffer types.
- Updated `Fill` class to process region descriptors and buffer regions, improving flexibility in buffer handling.
- Introduced checks for static bounds in region definitions to ensure safety during operations.
- Refactored loop induction variable handling in `FillNode` to accommodate sliced regions.

## Summary by CodeRabbit

* **New Features**
  * Fill operation now accepts additional input forms and normalizes region-based inputs for broader use.
  * Region-based handling unified so fill always targets an explicit region descriptor.

* **Bug Fixes**
  * Improved region bound checks with safer static validation and conditioned upper-bound checks.
  * Corrected SIMT loop indexing to respect sliced regions.
  * Hardened lowering behavior for unsupported destination scopes.

* **Tests**
  * Added tests covering static and dynamic region fill behavior on CUDA.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_1008.py
+++ b/testing/python/issue/test_tilelang_issue_1008.py
@@ -8,44 +8,46 @@ from tilelang import language as T
     pass_configs={
         tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
         tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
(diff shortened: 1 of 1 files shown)


LeiWang1999:
for issue #1008

PR #1215, opened 2025-11-10
[Build] Explicitly add `libtvm` as a dep of `libtilelang`

### LINKED ISSUE #1028: [Bug][Build][TVM] Build issue in Darwin aarch64
Currently building tilelang on Darwin requires extra workaround for TVM:

1. `libtvm` needs to be linked to `libtilelang`, which seems wrong, as `libtilelang` is supposed to work with `libtvm_runtime` only.:
  https://github.com/tile-ai/tilelang/blob/7a5077e4aa8e30533b6fe1f0716b2c28cf6f661b/CMakeLists.txt#L126-L129
2. When building with `cibuildwheel`, libbacktrace will be built for x86_64, and `python -mbuild -w` seems fine.
  https://github.com/tile-ai/tilelang/blob/0f515b86fc752cf5a9fd65ae9a46e8f229a84226/CMakeLists.txt#L78-L79

Maybe cc @Hzfengsy ?

oraluben:
Closes #1028

After digging into #1028, it seems that `libtilelang` depends on `libtvm` anyway. Linux works because it enables lazy loading by default and the old behavior works fine. Darwin requires all symbols to be found when linking, causes the previous error.

## Summary by CodeRabbit

* **Chores**
  * Updated library linking configuration to apply consistently across all platforms.
  * Enhanced runtime dependency linking for the build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,12 +169,8 @@ target_include_directories(tilelang_objs PRIVATE ${TILE_LANG_INCLUDES})
 
 add_library(tilelang SHARED $<TARGET_OBJECTS:tilelang_objs>)
 add_library(tilelang_module SHARED $<TARGET_OBJECTS:tilelang_objs>)
-target_link_libraries(tilelang PUBLIC tvm_runtime)
+target_link_libraries(tilelang PUBLIC tvm_runtime tvm)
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@codex review

PR #1218, opened 2025-11-10
Fix various issues under `int64_t` static and dynamic shape.

Elevator14B:
This PR tries to resolve various issues under `int64_t` static and dynamic shape.

TODOs:

* Wrapper/adapter should be further improved to use corresponding data type instead of enforcing `c_int64`;
* Add corresponding cases as unit tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Zero-value checks are now dtype-aware to avoid incorrect comparisons.
  * Dynamic kernel argument types now follow each symbol's declared dtype instead of assuming integer, improving correctness across backends.

* **Tests**
  * Added integration tests that exercise very large int64-shaped workloads to validate dynamic-dimension and kernel-argument behavior.

### DIFF
--- a/src/transform/inject_assumes.cc
+++ b/src/transform/inject_assumes.cc
@@ -63,7 +63,8 @@ private:
     Stmt build(Stmt body) {
       auto analyzer = arith::Analyzer{};
       for (const auto &e : items) {
-        auto simplified = analyzer.Simplify(GT(e.expr, make_zero(e.expr->dtype)));
+        auto simplified =
+            analyzer.Simplify(GT(e.expr, make_zero(e.expr->dtype)));
         std::stringstream ss;
         ss << "Buffer shape should be greater than 0: shape `" << e.expr
            << "` from buffer ";


LeiWang1999:
@codex review

src/transform/inject_assumes.cc
```diff
@@ -62,7 +62,7 @@ class AssumeInjector : public tvm::tir::StmtExprMutator {
     Stmt build(Stmt body) {
       auto analyzer = arith::Analyzer{};
       for (const auto &e : items) {
-        auto simplified = analyzer.Simplify(GT(e.expr, 0));
+        auto simplified = analyzer.Simplify(GT(e.expr, IntImm(e.expr->dtype, 0)));
```
LeiWang1999: I  think we can use `make_zero(e.expr->dtype)` instead.

src/transform/loop_vectorize_dynamic.cc
```diff
@@ -189,7 +189,7 @@ class VectorizePlannerDynamic : public arith::IRVisitorWithAnalyzer {
       }
       PrimExpr offset = buffer.OffsetOf(indices).back();
       // condition for alignment, maybe useless
-      condition_ = (FloorMod(offset, vector_size_) == 0);
+      condition_ = (FloorMod(offset, IntImm(offset.dtype(), vector_size_)) == IntImm(offset.dtype(), 0));
```
LeiWang1999: `make_zero(e.expr->dtype)`

LeiWang1999 (approved):

PR #1260, opened 2025-11-15
[BugFix] Remove memory_order in atomic constexpr and fix NSA bwd

KevinZeng08:
What this PR do?

- Remove memory_order in atomic constexpr temporally to avoid compile error, which is introduced in #1081 
- Fix NSA bwd tilelang example

TODO in future

- atomic_ref does not support 1 or 2 byte types, [link](https://godbolt.org/z/TEaoasbc9), maybe add inline PTX for non-vectorized atomic when dtype is fp16/bf16 for different memory orders
- refactor NSA example and benchmark, since it contains some redundant code

## Summary by CodeRabbit

* **Refactor**
  * Clarified internal loop variable naming to reduce shadowing in parallel computations.

* **Performance**
  * Standardized CAS-based handling for half/bfloat16 atomics and refined memory‑order paths to streamline specialized atomic operations.

* **Tests**
  * Converted several JIT-wrapped checks to plain tests, expanded dtype coverage (float/float16/bfloat16), and adjusted memory‑order scenarios to improve test coverage.

### DIFF
--- a/src/tl_templates/cuda/atomic.h
+++ b/src/tl_templates/cuda/atomic.h
@@ -52,8 +52,7 @@ TL_DEVICE void AtomicMax(T1 &ref, T2 val,
     // We simulate this process by atomicCAS loop.
     unsigned short *address_as_ushort =
         reinterpret_cast<unsigned short *>(address);
-    unsigned short val_as_ushort =
-        *reinterpret_cast<unsigned short *>(&val);
+    unsigned short val_as_ushort = *reinterpret_cast<unsigned short *>(&val);
     unsigned short old_val_ushort = *address_as_ushort;
     while (val > *reinterpret_cast<T1 *>(&old_val_ushort)) {
       unsigned short assumed_val_ushort = old_val_ushort;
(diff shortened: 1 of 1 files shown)


LeiWang1999:
cc @tzj-fxz @Rachmanino

tzj-fxz:
@KevinZeng08 Thank you for your valuable suggestion. It will be more reliable to directly inline PTX for atomic operations with different memory orders, like what has already been done in vectorized dtype. I will check it and fix it later.

KevinZeng08:
Thanks for your kind reply. Looking forward to your implementation.

tzj-fxz:
@KevinZeng08 This new PR (https://github.com/tile-ai/tilelang/pull/1264) was created based on your modification and suggestion. For the AtomicAdd operator, we inline PTX to handle different memory orders. For the AtomicMin/Max operators, we implement the atomicCAS loop to simulate the process because the original CUDA does not support a single 16-byte AtomicMin/Max. Could you check it if you are interested? :)

KevinZeng08:
Sure, let me have a check

Issue #1266, opened 2025-11-16
[BUG]  example_tilelang_nsa_fwd.py is inconsistent with the reference implementation and specific parameters .


sea-with-sakura:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.6.post2+cuda.giteb415744

### System information

- Device: CUDA 12.4
- Dependencies: torch 2.6.0+cu124
- tilelang: 0.1.6.post2+cuda.giteb415744
- sys: 3.11.13 | packaged by conda-forge | (main, Jun  4 2025, 14:48:23) [GCC 13.3.0] linux

### Problem description

The output of the example in examples/deepseek_nsa/example_tilelang_nsa_fwd.py is inconsistent with the reference implementation when specific parameters are set (example_tilelang_nsa_fwd.py L136-L137)

Problem Description
In examples/deepseek_nsa/example_tilelang_nsa_fwd.py, after setting the parameters to the following combination, running the script will trigger a significant numerical inconsistency between 'ref' and 'out', resulting in an assertion failure.

### Reproducible example code

Reproduction Steps
- Change SEQ_LEN from 64 to 128
- Change S to 2
- Keep other parameters unchanged
- Run directly: python examples/deepseek_nsa/example_tilelang_nsa_fwd.py

The Python snippets:

```python

B=2, SEQ_LEN=128, H=1, HQ=16, D=32, S=2, block_size=32, dtype=torch.float16, scale=0.1

```

### Traceback

```pytb
Traceback (most recent call last):
  File "/root/sakura/work/deepseek_nsa/tilelang-dllm/main_fwd.py", line 111, in <module>
    main()
  File "/root/sakura/work/deepseek_nsa/tilelang-dllm/main_fwd.py", line 106, in main
    torch.testing.assert_close(ref, out, atol=1e-2, rtol=1e-2)
  File "/root/miniconda3/envs/env_torch/lib/python3.11/site-packages/torch/testing/_comparison.py", line 1530, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 35891 / 131072 (27.4%)
Greatest absolute difference: 0.634765625 at index (0, 88, 3, 6) (up to 0.01 allowed)
[...]

PR #1276, opened 2025-11-18
fix the issue of #1266 in examples/deepseek_nsa/example_tilelang_nsa_…

sea-with-sakura:
Fix the bug in issue #1266 and modify duplicate loop variables to avoid warning messages.

## Summary by CodeRabbit

* **Chores**
  * Updated example configuration parameters (sequence length, block size, and block counts).
  * Refined block computation logic to derive block counts from actual indices instead of random generation, improving alignment between configuration values.

### DIFF
--- a/src/target/rt_mod_cuda.cc
+++ b/src/target/rt_mod_cuda.cc
@@ -2,6 +2,7 @@
 #include "runtime/cuda/cuda_module.h"
 #include "runtime/pack_args.h"
 #include <tvm/ffi/reflection/registry.h>
+#include <tvm/ir/transform.h>
 
 namespace tvm {
 namespace codegen {
@@ -66,7 +67,10 @@ ffi::Module BuildTileLangCUDA(IRModule mod, Target target) {
   std::string ptx;
   if (const auto f =
           ffi::Function::GetGlobal("tilelang_callback_cuda_compile")) {
-    ptx = (*f)(code, target).cast<std::string>();
+    // Fetch current pass context config and pass into the compile callback
+    tvm::transform::PassContext pass_ctx =
+        tvm::transform::PassContext::Current();
+    ptx = (*f)(code, target, pass_ctx->config).cast<std::string>();
     if (ptx[0] != '/')
       fmt = "cubin";
   } else {
--- a/tilelang/contrib/nvcc.py
+++ b/tilelang/contrib/nvcc.py
@@ -78,7 +78,7 @@ def compile_cuda(code,
         out_file.write(code)
 
     file_target = path_target if path_target else temp_target
-    cmd = ["nvcc"]
+    cmd = [get_nvcc_compiler()]
     cmd += [f"--{target_format}", "-O3"]
     if kernels_output_dir is not None:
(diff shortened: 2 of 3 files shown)


examples/deepseek_nsa/example_tilelang_nsa_fwd.py
```diff
@@ -90,9 +90,9 @@ def native_sparse_attention(
                     T.copy(K[i_b, i_s:i_s + BS, i_h, :], K_shared)
 
                     if is_causal:
-                        for i, j in T.Parallel(G, BS):
-                            acc_s[i, j] = T.if_then_else(i_t >= (i_s + j), 0,
-                                                         -T.infinity(acc_s.dtype))
+                        for _i, j in T.Parallel(G, BS):
```
LeiWang1999: for what reason we need to change `i` into `_i`?

sea-with-sakura: Because both the outer loop and the inner loop use `i`, a warning is triggered:

2025-11-19 03:11:27  [TileLang:tilelang.language.v2.builder:WARNING]: Variable `i` shadows another declared value. Are you forgetting to allocate it as a var?

Stack (most recent call last):
  File "/root/sakura/work/tilelang/examples/deepseek_nsa/example_tilelang_nsa_fwd.py", line 189, in <module>
    main()
  File "/root/sakura/work/tilelang/examples/deepseek_nsa/example_tilelang_nsa_fwd.py", line 139, in main
    kernel = native_sparse_attention(
  File "/4090data2/sakura/work/tilelang/tilelang/jit/__init__.py", line 287, in __call__
    self._kernel_cache[key] = self.compile(*args, **kwargs, **tune_params)
  File "/4090data2/sakura/work/tilelang/tilelang/jit/__init__.py", line 237, in compile
    func = self.get_tir(*args, **kwargs)
  File "/4090data2/sakura/work/tilelang/tilelang/jit/__init__.py", line 206, in get_tir
    program_result = program_result_source(*args, **kwargs)
  File "/root/sakura/work/tilelang/examples/deepseek_nsa/example_tilelang_nsa_fwd.py", line 53, in native_sparse_attention
    @T.prim_func
  File "/4090data2/sakura/work/tilelang/tilelang/language/v2/builder.py", line 735, in prim_func
    return impl(func) if func is not None else impl
  File "/4090data2/sakura/work/tilelang/tilelang/language/v2/builder.py", line 722, in impl
    return prim_func_generator(**annot)
  File "/4090data2/sakura/work/tilelang/tilelang/language/v2/builder.py", line 707, in prim_func_generator
    ir_gen.gen(builder)(*args, **kwargs)
  File "/root/sakura/work/tilelang/examples/deepseek_nsa/example_tilelang_nsa_fwd.py", line 93, in native_sparse_attention
    for i, j in T.Parallel(G, BS):

LeiWang1999: I see, I think we can remove this warning or enhance it in future, for now I think we can recover those modifications and we're good to go to merge this pr :)

PR #1281, opened 2025-11-19
[Fix] Fix memory leak bug

kurisu6912:
This pr fix a memory leak bug and add a test in DSLMutator, it use a expression of {**func.__globals__}, which make the globals unable to free

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for tilelang capture functionality.

* **Refactor**
  * Improved internal handling of closure variable capture and type hint resolution.
  * Marked internal utility function as deprecated; users relying on it should prepare for future removal.

### DIFF
--- a/tilelang/language/v2/utils.py
+++ b/tilelang/language/v2/utils.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 import ast
 import inspect
 from typing import Any, Callable, Literal
-from deprecated import deprecated
 from tilelang import env
 from hashlib import sha256
 import linecache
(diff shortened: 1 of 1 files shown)


SiriusNEO:
LGTM, let's rebase with new tvm-ffi changes and fix lint

tilelang/language/v2/builder.py
```diff
@@ -593,22 +594,8 @@ def get_type_hints(func):
...
-        if closure_bindings:
-            localns.update(closure_bindings)
-            # Also update globals so ForwardRef eval sees them uniformly
-            globalns.update(closure_bindings)
-    except Exception:
-        # Be permissive: absence or access issues with closure shouldn't crash
-        pass
-
+    globalns = func.__globals__
+    localns = utils.get_func_nonlocals(func)
```
SiriusNEO: Let's add a comment here to briefly explain the reason (why "nonlocals" of func should be passed into "localns" of eval_type)

PR #1331, opened 2025-11-25
[Enhancement] Add support for vectorized loading in gemm_v1 on ROCm.

Gongen-Ali:
Using vectorization to transfer data from shared memory to registers in gemm_v1 can achieve significant performance gains. Moreover, when computing fp8 GEMM with k_pack=2, gemm_v1 performs better than gemm_v2.
Mi308X：
M=N=K=16384, dtype=float8_e4m3fnuz
v1:
`Best latency (s): 33.353660583496094
Best TFlops: 263.722
Best config: {'block_M': 128, 'block_N': 128, 'block_K': 64, 'num_stages': 1, 'thread_num': 256, 'policy': <GemmWarpPolicy.Square: 0>, 'enable_rasteration': False}`
v2:
`Best latency (s): 36.72024154663086
Best TFlops: 239.543
Best config: {'block_M': 128, 'block_N': 128, 'block_K': 128, 'num_stages': 1, 'thread_num': 256, 'policy': <GemmWarpPolicy.Square: 0>, 'enable_rasteration': True}`

### DIFF
--- a/src/tl_templates/hip/common.h
+++ b/src/tl_templates/hip/common.h
@@ -86,6 +86,7 @@ typedef
     __attribute__((__vector_size__(4 * sizeof(short)))) short bfloat16x4_vec;
 
 using int32x4 = __attribute__((__vector_size__(4 * sizeof(int)))) int;
+using float32x2 = __attribute__((__vector_size__(2 * sizeof(float)))) float;
 using float32x4 = __attribute__((__vector_size__(4 * sizeof(float)))) float;
 using float32x16 = __attribute__((__vector_size__(16 * sizeof(float)))) float;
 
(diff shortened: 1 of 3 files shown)


LeiWang1999:
Thanks @Gongen-Ali , we're good to go if we can fix the lint with `./format.sh`.

It's also surprised that v2 is slower than v1, would you mind dig further and help us fix the performance issue?

Gongen-Ali:
> Thanks @Gongen-Ali , we're good to go if we can fix the lint with `./format.sh`.
> 
> It's also surprised that v2 is slower than v1, would you mind dig further and help us fix the performance issue?
>
>👌 We have already analyzed this issue and preliminarily concluded that, when k_pack=2, the vectorized transfer from ‘buf_dyn_shmem‘ to ’local‘ uses only fp8_e4_8_t, whereas fp8_e4_16_t could be used instead to improve transfer efficiency. We are currently testing this approach and plan to modify TVM's automatic vectorization handling accordingly.

Gongen-Ali:
We have fix the v2 performance. And temporarily close the current PR.
https://github.com/tile-ai/tilelang/pull/1344

PR #1358, opened 2025-12-01
[Analysis] Enhance NestedLoopChecker with tile op cases

SiriusNEO:
This PR explicitly adds check for cases like:

```python
for i in T.Parallel():
    T.tile_op(...)
```

For now we only let atomic operations like `T.atomic_add` and TIR operators present in T.Parallel.

## Summary by CodeRabbit

* **Tests**
  * Added tests validating that tiled operations inside parallel loops are rejected, including matmul/atomic patterns and reference checks.

* **Bug Fixes**
  * Enforced validation to prevent TiledOp usage within parallel constructs with clearer semantic errors.

* **Chores**
  * Improved debug logging for AST printing.
  * Updated internal operator routing to use the tile-op registration/pathway.

### DIFF
--- a/tilelang/analysis/nested_loop_checker.py
+++ b/tilelang/analysis/nested_loop_checker.py
@@ -5,7 +5,6 @@ from tvm.tir import (
     PrimFunc,
     PyStmtExprVisitor,
 )
-from tvm.ir import Op
 from tvm.tir.transform import prim_func_pass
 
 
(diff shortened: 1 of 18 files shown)


tilelang/analysis/nested_loop_checker.py
```diff
@@ -17,6 +19,12 @@ def is_pipelined_for(op: For) -> bool:
     return any(key in op.annotations for key in anno_keys)
 
 
+def is_atomic_op(op: Call) -> bool:
+    """Check if a call is an atomic operation."""
+
+    return op.op == Op.get("tir.call_extern") and op.args[0].value.startswith("Atomic")
```
LeiWang1999: I think this is a bit tricky, we can collect the definition of tile intrins, let me take a look

PR #1375, opened 2025-12-04
[Layout] Enhance Free Layout Inference

LeiWang1999:
- skip invalid layout candidate (when loop layout is not bijective)
- optimize bfs path.
- support layout inference for T.view

## Summary by CodeRabbit

* **New Features**
  * Public injectivity detection API and a new loop-layout exception for injectivity failures.
  * Reshape APIs now accept element-size rescaling parameters (rescale_num/rescale_den).

* **Improvements**
  * Layout inference queue now prioritizes ready buffers for faster propagation.
  * 1D bulk copy/load path handled alongside existing bulk cases.
  * Aggressive shared-memory merge enabled in a backward kernel example.

* **Bug Fixes**
  * Stricter injectivity validation with clearer error reporting.

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -567,7 +567,10 @@ LayoutMap ParallelOpNode::InferLayout(const LayoutInferArgs &T,
   auto injective_res = loop_layout_->DetectInjective();
   if (!injective_res->errors.empty()) {
     std::ostringstream oss;
-    oss << "Loop layout is not injective: " << loop_layout_->DebugOutput();
+    oss << "Loop layout is not injective: " << loop_layout_->DebugOutput()
+        << '\n'
+        << "  errors: " << injective_res->errors << '\n'
+        << "  loop AST: " << root_;
     throw LoopLayoutInjectiveException(oss.str());
   }
 
--- a/src/op/parallel.h
+++ b/src/op/parallel.h
@@ -24,24 +24,6 @@ namespace tl {
 
 using namespace tir;
 
-class LayoutConflictException : public std::exception {
-public:
-  const char *what() const noexcept override { return msg_.c_str(); }
-  LayoutConflictException(const std::string &msg) : msg_(msg) {}
-
-private:
-  std::string msg_;
-};
-
-class LoopLayoutInjectiveException : public std::exception {
-public:
-  const char *what() const noexcept override { return msg_.c_str(); }
-  LoopLayoutInjectiveException(const std::string &msg) : msg_(msg) {}
-
-private:
-  std::string msg_;
-};
-
 bool ProveFragmentContains(Fragment small_frag, Fragment large_frag,
                            Array<PrimExpr> small_frag_indices,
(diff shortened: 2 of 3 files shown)


src/transform/layout_inference.cc
```diff
@@ -72,7 +73,7 @@ class BufferUseDefCollector : public IRVisitorWithAnalyzer {
 
   void RunInferStep(int cur_infer_id, InferLevel level, bool update_queue,
                     LayoutMap &layout_map, const LayoutMap &strict_layout_map,
-                    std::queue<int> &q, std::vector<bool> &in_queue) {
+                    std::deque<int> &q, std::vector<bool> &in_queue) {
```
SiriusNEO, line 76: Is there any particular reason to replace the BFS queue with a dequeue?

src/op/parallel.h
```diff
@@ -33,6 +33,15 @@ class LayoutConflictException : public std::exception {
   std::string msg_;
 };
 
+class LoopLayoutInjectiveException : public std::exception {
```
SiriusNEO: Do you think it's a good idea to move `LayoutConflictException` and `LoopLayoutInjectiveException` into `layout/layout.h`? Just a suggestion since I think they are more related to layout issues.

src/layout/layout.cc
```diff
@@ -536,6 +536,21 @@ bool FragmentNode::IsCompletedReplicated() const {
                          ReplicationPlaceholder());
 }
 
+arith::IterMapResult FragmentNode::DetectInjective() const {
```
SiriusNEO, line 556: Given that current implementation is to do bijective check (which is a stricter condition), let's just name it `DetectBijective` to avoid confusion. We have comments here so it's OK if we want to improve this in the futher.

src/op/parallel.cc
```diff
@@ -562,6 +563,13 @@ LayoutMap ParallelOpNode::InferLayout(const LayoutInferArgs &T,
   } else {
     return {};
   }
+  // check loop_layout_ is injective
+  auto injective_res = loop_layout_->DetectInjective();
+  if (!injective_res->errors.empty()) {
+    std::ostringstream oss;
+    oss << "Loop layout is not injective: " << loop_layout_->DebugOutput();
```
SiriusNEO: I suggest also print the loop which is not injective here. Otherwise we need to locate the loop when we see this error next time (with only a layout->DebugOutput() info)

src/transform/layout_inference.cc
LeiWang1999: For example:

```
op a: uses buffer A
op b: uses buffer A, buffer B
op c: uses buffer B, buffer C
```

Previously, when we inferred the layout for op a, the queue was initialized arbitrarily. Inferring op c before op b is problematic because op c’s layout in the next stage is free, which would assign a free layout to buffer B. That result could later conflict with the layout already inferred for op b.

Therefore, we need a deque that places hot ops (common‑path) on the left and cold ops (free‑path) on the right.

PR #1240, opened 2025-11-12
[Feature]: Add benchmark scripts for examples

yyttt6:
## Summary by CodeRabbit

* **New Features**
  * Adds a lightweight benchmarking framework with warmup/repeat timing, failure tracking, aggregated results table, and PNG visualization.
  * Registers many new example benchmark entry points across the examples suite for unified discovery and execution.

* **Chores**
  * CI/perf workflow updated to produce machine-readable benchmark tables and upload/embed generated plots in PR feedback.

PR #1398, opened 2025-12-10
[Enhancement] Improve vectorization invariant check

LJC00118:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loop vectorization with an earlier invariant-boundary check, allowing more loops to be vectorized and yielding better runtime performance for eligible operations.
* **Tests**
  * Added new invariant-index vectorization tests and expanded coverage across multiple sizes to validate correctness and emitted SIMD hints.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -292,6 +292,15 @@ bool IndiceCanVectorize(const PrimExpr &expr, Var var,
     return false;
 
   // Check if expr is invariant within vector boundaries
+  // We're trying to prove the access expression A[f(var)] depends only on
+  // floor(var/vecsize), not on var%vecsize
+  // Mathematically:
+  // \forall var, f(floor(var/vecsize)*vecsize + var%vecsize) ==
+  // f(floor(var/vecsize)*vecsize + 0)
+  // Example: for i in T.vectorized(8):
+  //     A[i] = B[i] * C[i//4]
+  // if vecsize=4, f(i)=i//4 depends only on i//4
+  // Therefore A[i] = B[i] * C[i//4] can be vectorized with vecsize=4
   PrimExpr var_aligned =
       floordiv(var, target_vectorized_size) * target_vectorized_size;
   PrimExpr expr_aligned = Substitute(expr, {{var, var_aligned}});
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@LJC00118 we're good to go to add some test cases

src/transform/loop_vectorize.cc
```diff
@@ -290,6 +290,15 @@ bool IndiceCanVectorize(const PrimExpr &expr, Var var,
   if (!analyzer->CanProveEqual(FloorMod(iter_var_size, target_size_for_iter),
                                0))
     return false;
+
+  // Check if expr is invariant within vector boundaries
```
kurisu6912, line 294: @LJC00118 we're good to go to add some comments here, an example helps a lot

We're trying to prove the access expression `A[f(i)]`, can be written by `A[g(i//len)]`, or more formally:

$$
\exists g(x),\ \forall a,\ g\left(\left\lfloor\frac{a}{\text{vecsize}}\right\rfloor\right) = f(a)
$$

The above equation is equivalent to:

$$
\forall a,\ f\left(\left\lfloor\frac{a}{\text{vecsize}}\right\rfloor * \text{vecsize} + 0\right) = f\left(\left\lfloor\frac{a}{\text{vecsize}}\right\rfloor * \text{vecsize} + a\ \\%\ \text{vecsize}\right)
$$

yyttt6:
This PR brings significant performance improvements for sparse_mla_bwd, as well as other cases such as fusedmoe and sparse_mla_fwd_pipelined.
In our benchmarks, sparse_mla_bwd achieves up to 1.2× speedup.
We observed that the generated CUDA code applies vectorization, which were not present in the original commit, and this appears to account for the performance gain.
speedup_bar2

LeiWang1999:
surprise to find speed up for mla_bwd!

88ec454e44dc3c539a78ae0ee101bab4

kurisu6912:
The plot is fantastic!

Can we add it to the CI to obtain the performance each time?

LeiWang1999:
> The plot is fantastic!
> 
> Can we add it to the CI to obtain the performance each time?

yes we should and pr #1240 is working for this part.

LeiWang1999 (approved):

PR #1405, opened 2025-12-11
[Enhancement] Improve InjectAssumes logic and make assumes work after SplitHostDevice

SiriusNEO:
Previous InjectAssumes doesn't properly handle the root block case, which causes repeatedly assumes injections (specifically, for shape boundary check in function arguments) in two positions. This PR fixes this logic.

Another problem is that previous host-side assumes are missed after SplitHostDevice, which brings some problems when we use analyzer after splitting. This PR maintains the assumes context after splitting by copying the host-side assumes to the start of device-side in SplitHostDevice pass.

## Summary by CodeRabbit

* **New Features**
  * Added detection helpers for assume-style Evaluate statements and automatic conversion to assume attributes.

* **Refactor**
  * Consolidated assume creation and grouping; simplified root-block and buffer handling; host-side assumes are now collected and reapplied.

* **Bug Fixes / Stability**
  * Enforced assume processing before host/device splitting and prevented assumes from remaining in device regions.

* **Documentation**
  * Expanded docstring explaining assume-injection behavior.

* **Chores**
  * Updated third-party submodule references.


src/transform/inject_assumes.cc
```diff
@@ -79,32 +89,37 @@ class AssumeInjector : public tvm::tir::StmtExprMutator {
...
+      //    T.assume(cond2)
+      // This SeqStmt will be converted to:
+      //    With(attr::tilelang_assume, cond1) {
+      //      Stmt1
+      //      Stmt2
+      //    }
+      //    With(attr::tilelang_assume, cond2) {
+      //      ...
+      //    }
+      if (auto e = GetAssumeExprInEvaluateForm(stmt)) {
```
kurisu6912, line 122: This is good!

src/transform/split_host_device.cc
kurisu6912, line 126: A little strange, we want to destroy use_def?

Another hint, we can use the statement expression in C/C++ (But I'm not very sure it is supported by the gcc used in CI)

```suggestion
    auto [params, buffers_to_declare] = ({
        tir::VarUseDefAnalyzer use_def(/*defined_vars=*/{},
        /*visit_thread_extent=*/true);
        use_def(body);
        
        // Sort first by variable type, then by variable name
        std::vector<tir::Var> params{use_def.undefined_.begin(),
        use_def.undefined_.end()};
        std::sort(params.begin(), params.end(),
            [](const tir::Var &a, const tir::Var &b) {
            auto sort_key = [](const tir::Var &var) {
                return std::tuple{
                    !var->dtype.is_handle(),
                    var->name_hint,
                };
            };
            return sort_key(a) < sort_key(b);
        });
        {params, use_def.undefined_buffers_};
    });
```

src/transform/split_host_device.cc
```diff
@@ -51,10 +55,29 @@ class HostDeviceSplitter : public tir::StmtMutator {
...
       return SplitDeviceFunc(op->body, device_target);
+    } else if (op->attr_key == tir::attr::tilelang_assume) {
+      // NOTE(chaofan): the assumes collected here must be in host-side.
+      //    This is because when the collector reaches the split region,
+      //    it will start to split and return. For safety, we add a check here.
+      ICHECK(!found_device_region_)
+          << "Assumes collection should not be in device region.";
+      // We first push back the outside assume, then visit the child.
+      // So when moving assumes to device side, we need to do the building
+      // process in a reverse order.
```
kurisu6912, line 73: I'm a little confused.

The pass is in pre-order traversal, for example:

```py
T.attr('tl.assume', expr_1, 'assume_1')
T.attr('tl.assume', expr_2, 'assume_2')
```

We first push `assume_1`, and go to `VisitStmt_` to visit `assume_2`, and then push `assume_2`. The host_assumes_ is:

````
[assume_1, assume_2]
````

Why do we need to reverse it?

**Review by kurisu6912 — COMMENTED** [review]
This pass traverses the AST and copies all assume attribute statements to the device side.

1. Traverse AST and collect all assume statements into `host_assumes_`.
2. Until the first AttrStmtNode with `tvm::attr::kTarget`.
3. Call `SplitDeviceFunc` to convert it to device function

File: src/transform/common/assume.h [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+namespace tl {
+
+using namespace tir;
+
+// Get the expression inside an assume statement, if any. Returns nullopt if
+// the statement is not an assume statement.
+std::optional<PrimExpr> GetAssumeExprInEvaluateForm(Stmt stmt);
+
+// Check if a statement is an assume statement.
+bool IsAssumeInEvaluateForm(const Stmt &stmt);
```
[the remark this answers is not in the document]
kurisu6912: This is a problem

src/transform/split_host_device.cc
SiriusNEO: Because we need to "wrap" the body repeatedly using AttrNode(body=...). This process is reverse, first do assume_2 then assume_1, and we can get:

```
With assume_1:
    With assume_2:
```

src/transform/split_host_device.cc
kurisu6912: This logic is from tvm, confusing

PR #1434, opened 2025-12-15
[Bugfix] Convey  `compile_flags` to ffi compilation path with pass_configs

LeiWang1999:
`compile_flags` is helpful for users to customize compile options for nvcc or other compilers. However, previously it only works for those lib generator that directly invoke nvcc to compile a .so files. After we change the default build pipeline into tvm ffi, the source compilation has been shifted into cpp for cubin compilation and which will ignore the compile_flags. Now we fix it via transform compile_flags into pass_configs

## Summary by CodeRabbit

* **New Features**
  * Add a pass configuration option to supply extra device compiler flags; these flags are parsed and merged with existing compiler options from JIT calls.
* **Chores**
  * JIT entrypoints no longer coerce string compile_flags into single-element lists; compile_flags may be provided or propagated as raw strings or lists.

### DIFF
--- a/tilelang/jit/kernel.py
+++ b/tilelang/jit/kernel.py
@@ -223,10 +223,11 @@ class JITKernel(Generic[_P, _T]):
 
         compile_flags = self.compile_flags
 
-        compile_flags_cfg = pass_configs.get(PassConfigKey.TL_DEVICE_COMPILE_FLAGS)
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@codex review

LeiWang1999:
Local Test can Pass, Merge :)

PR #1443, opened 2025-12-16
[Bugfix] Open Rocm ci test and fix some bugs.

Gongen-Ali:
## Summary by CodeRabbit

* **New Features**
  * Broader HIP support: warp reductions, device-side atomic helpers, 1D/2D cumulative-sum utilities, unified device debug printing, and expanded FP8 datatype support.

* **Tests**
  * Many GPU tests now gated to CUDA; several tests updated to use higher-precision accumulation; new FP8 test variants added; tests rely more on default backend selection.

* **Refactor**
  * Consolidated HIP/CUDA utilities and moved include/template path resolution to environment-driven configuration.

* **Chores**
  * CI adjusted to skip specific ROCm test paths.

### DIFF
--- a/testing/python/language/test_tilelang_language_pdl.py
+++ b/testing/python/language/test_tilelang_language_pdl.py
@@ -34,6 +34,7 @@ def kernels_with_pdl_sync(N, block_size=256, dtype=T.float32):
     return main
 
 
+@tilelang.testing.requires_cuda
 def test_pdl_trigger():
     N = 64
     program = kernels_with_pdl_trigger(N)
@@ -43,6 +44,7 @@ def test_pdl_trigger():
     assert "cudaTriggerProgrammaticLaunchCompletion" in code
 
 
+@tilelang.testing.requires_cuda
 def test_pdl_sync():
     N = 64
     program = kernels_with_pdl_sync(N)


src/op/gemm.cc
```diff
@@ -154,6 +154,9 @@ std::pair<int, int> GemmWarpPolicyNode::computeWarpPartition(
   if (TargetIsVolta(target)) {
     kNPerWarp = 16;
   }
+  if (TargetIsCDNA(target)) {
```
LeiWang1999: would be better to use `else if` ?

testing/python/debug/test_device_assert.py
```diff
@@ -13,7 +13,7 @@ def program():
             tid = T.get_thread_binding()
             T.device_assert(tid > 0, "Assertion Trigger !")
 
-    jit_kernel = tilelang.compile(program, target="cuda")
+    jit_kernel = tilelang.compile(program, target="auto")
```
LeiWang1999: The default value of target is `None` and then can be analyzed into `auto`, so this argument would be better to be removed.

testing/python/debug/test_device_assert.py
```diff
@@ -25,7 +25,7 @@ def program():
             tid = T.get_thread_binding()
             T.device_assert(tid == tid)
 
-    jit_kernel = tilelang.compile(program, target="cuda")
+    jit_kernel = tilelang.compile(program, target="auto")
```
LeiWang1999: same :)

testing/python/debug/test_tilelang_debug_print.py
```diff
@@ -74,7 +66,7 @@ def program(Q: T.Tensor((M, N), dtype)):
             if tid == 0:
                 T.print(bx + by + bz)
 
-    jit_kernel = tilelang.compile(program, target="cuda")
+    jit_kernel = tilelang.compile(program, target="auto")
```
LeiWang1999: same target issues for this file.

testing/python/language/test_tilelang_language_annot.py
```diff
@@ -4,6 +4,8 @@
 import torch
 
 
+# TODO(Gong): Here is a bug on build with cython backend, disable for now
```
LeiWang1999: would be better to add more details, for example, disable for rocm or `enable only for cuda backend` :)

LeiWang1999 (commented):
Overall LGTM, left some comments.

testing/python/language/test_tilelang_language_annot.py
Gongen-Ali: I have opened an issue and changed the description.

PR #1445, opened 2025-12-16
[Bugfix] Improve autotune from elementwise_add function in examples

senlyu163:
This pull request updates the `example_elementwise_add.py` example to improve the compilation and execution of the element-wise addition kernel. The main change is the removal of the `@tilelang.jit` decorator, which caused `autotuner.run` to fail when autotune was enabled in this example.

Kernel compilation and execution improvements:

* Removed the `@tilelang.jit(out_idx=[-1])` decorator from the `elementwise_add` function, so it now returns a program instead of a compiled kernel.
* Updated the main function to first generate the program using `elementwise_add`, then compile it with `tilelang.compile`, specifying `out_idx` and `target`. This makes the compilation step explicit and configurable.

## Summary by CodeRabbit

* **Refactor**
  * Updated the elementwise example to expose size and autotune flags, add an autotuning configuration generator, and switch the build flow to support runtime autotune or a default config; removed legacy autotune helper utilities.
* **Tests**
  * Added a test exercising the autotune execution path for the elementwise example.

### DIFF
--- a/examples/elementwise/example_elementwise_add.py
+++ b/examples/elementwise/example_elementwise_add.py
@@ -3,13 +3,22 @@ import itertools
 import torch
 import tilelang
 import tilelang.language as T
-from tilelang.autotuner import AutoTuner
 
 
 def ref_program(x, y):
     return x + y
 
 
+def get_configs():
+    block_M = [64, 128, 256]
+    block_N = [64, 128, 256]
+    threads = [64, 128, 256]
+    configs = list(itertools.product(block_M, block_N, threads))
+    return [{"block_M": bm, "block_N": bn, "threads": th} for bm, bn, th in configs]
+
+
+@tilelang.autotune(configs=get_configs())
+@tilelang.jit(out_idx=[-1])
 def elementwise_add(M, N, block_M, block_N, in_dtype, out_dtype, threads):
     @T.prim_func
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@senlyu163 Thanks, but for what reason we should remove this decorator？

senlyu163:
> @senlyu163 Thanks, but for what reason we should remove this decorator？

Hi, this decorator will cause `autotuner.run` run to fail when `use_autotune` was enabled in this example. I'll refine the test scripts later in this pr.

Detailed explanation:
The `jit` decorator constructs a JITKernel for a function. When tuning without using the `autotune` decorator format, `self.jit_compile` is initialized to `None`, which will cause an error during program compilation due to the JITKernel being passed in. See here: https://github.com/tile-ai/tilelang/blob/main/tilelang/autotuner/tuner.py#L357-L362

LeiWang1999:
@senlyu163 I see, thanks, I think the `get_best_config` is a bit legacy, would be better for us to use: 

```python
@tilelang.autotune
@tilelang.jit
def func():
    ...
```

senlyu163:
@LeiWang1999 Morden style is very concise, I have refactor it :)

LeiWang1999 (approved):

PR #1464, opened 2025-12-18
[Refactor] Rename test for curand & add triton baseline in `test_tilelang_language_rand.py`

silentCoder-dev:
## Summary by CodeRabbit

* **Tests**
  * Removed an outdated RNG test and added a new RNG parity test suite that compares two GPU backends for reproducible random outputs.
  * The suite runs on CUDA, is parameterized over sizes and seeds, asserts cross-backend equality, and includes a script entry point for convenient execution.

### DIFF
--- a/testing/python/language/test_tilelang_language_rand.py
+++ b/testing/python/language/test_tilelang_language_rand.py
@@ -20,7 +20,9 @@ def tilelang_rand_1d(M=1024, seed=42):
             T.rng_init(seed, 0, bx * blk_M + tx * num_per_thread)
             for i, j in T.Parallel(threads, num_per_thread):
                 offsets = (bx * threads + i) * num_per_thread
-                A[offsets + j] = T.rng_rand()
+                idx = offsets + j
+                if idx < M:
+                    A[idx] = T.rng_rand()
 
     return rand_kernel
 


testing/python/language/test_tilelang_language_rand.py
```diff
@@ -0,0 +1,53 @@
...
+import triton.language as tl
+import pytest
+import tilelang.testing
+
+tilelang.disable_cache()
+
+
+@tilelang.jit
+def tilelang_rand_1d(M=1024, seed=42):
+    blk_M = M
```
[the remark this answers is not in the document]
silentCoder-dev: Setting `blk_M=M` and calling `rng_rand()` four times is to align the results between tilelang and triton.

kurisu6912 (approved):
LGTM!

testing/python/language/test_tilelang_language_rand.py
```diff
@@ -0,0 +1,54 @@
+import tilelang
+import tilelang.language as T
+import torch
+import triton
+import triton.language as tl
+import pytest
+import tilelang.testing
+
+tilelang.disable_cache()
```
LeiWang1999: should remove

testing/python/language/test_tilelang_language_rand.py
```diff
@@ -0,0 +1,54 @@
...
+def test_rand_1d(M, seed):
+    kernel = tilelang_rand_1d(M, seed)
+    tilelang_result = torch.empty(M, dtype=torch.uint32, device="cuda")
+    kernel(tilelang_result)
+
+    triton_result = torch.empty(M, dtype=torch.uint32, device="cuda")
+    grid = (M // 128,)
+    triton_rand_1d[grid](triton_result, tl.constexpr(M), tl.constexpr(128), seed)
+
+    torch.testing.assert_close(tilelang_result, triton_result)
```
LeiWang1999, line 55: surprised to find that those two rand outputs are the same. Do they rely on the same implementation?

testing/python/language/test_tilelang_language_rand.py
silentCoder-dev: This line is removed in 38d3b34

testing/python/language/test_tilelang_language_rand.py
silentCoder-dev: Triton implements its own Philox algorithm and generates 4 random numbers each time via `randint4x`. Now it has been switched to the `randint4x` implementation, which does not discard the last three random numbers.

```python
def randint(seed, offset, n_rounds: tl.constexpr = N_ROUNDS_DEFAULT):  
    ret, _, _, _ = randint4x(seed, offset, n_rounds)  
    return ret
```

PR #1466, opened 2025-12-18
[Language] Make TL scripts friendly to Python syntax highlights

SiriusNEO:
## Summary by CodeRabbit

* **New Features**
  * Expanded reduce operations: new reducers (sum, abs-sum/abs-max, bitwise reducers), cumsum, finalize and warp-reducer variants.

* **Refactor**
  * Reorganized public operation imports for clearer module organization.
  * Serial and Unroll loop constructs converted to callable wrappers.

* **Documentation**
  * Updated module docstrings to clarify operation categories.

* **Chores**
  * Removed redundant header lines from example files.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1476, opened 2025-12-19
[Language] Fix gemm syntax highlight

SiriusNEO:
Following https://github.com/tile-ai/tilelang/pull/1466, this PR fixes the highlight problem of `T.gemm`

## Summary by CodeRabbit

* **Refactor**
  * GEMM entry now uses a runtime-dispatching callable wrapper instead of a static alias, preserving behavior while enabling runtime selection.
  * Loop helpers Serial and Unroll now expose explicit, typed function signatures for clearer argument handling while forwarding to the same underlying implementations.

### DIFF
--- a/tilelang/language/gemm_op.py
+++ b/tilelang/language/gemm_op.py
@@ -1,6 +1,7 @@
 """GEMM (General Matrix Multiplication) operators exposed on the TileLang language surface."""
 
 from __future__ import annotations
+
 from tilelang.tileop.base import GemmWarpPolicy
 import tilelang.language as T
 from tvm import tir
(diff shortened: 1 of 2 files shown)


tilelang/language/gemm_op.py
```diff
@@ -184,4 +184,8 @@ def gemm_v2(
 
 
 # Default to v2; allow forcing v1 via environment variable
-gemm = gemm_v1 if _env.use_gemm_v1() else gemm_v2
+# gemm = gemm_v1 if _env.use_gemm_v1() else gemm_v2
+
+
+def gemm(*args, **kwargs):
```
LeiWang1999: one small question, if we replace the alias with function proxy, can lint still tell us the arguments of gemm is?

SiriusNEO: Maybe we should explicit add arguments in the sigature

LeiWang1999:
@codex review

tilelang/language/gemm_op.py
```diff
@@ -184,4 +184,45 @@ def gemm_v2(
...
+    C: tir.Buffer | tir.Var,
+    transpose_A: bool = False,
+    transpose_B: bool = False,
+    policy: GemmWarpPolicy = GemmWarpPolicy.Square,
+    clear_accum: bool = False,
+    k_pack: int = 1,
+    wg_wait: int = 0,
+    mbar: tir.Buffer | None = None,
+):
+    return (
```
LeiWang1999: Would be better to be refactored into:

```python
impl = gemv_v1 if ... else gemm_v2
return impl(
...
)
```
?

PR #1481, opened 2025-12-19
[Enhancement] Allow `import tilelang` on CPU-only machines without CUDA libraries

### LINKED ISSUE #1478: [Feature Request] Cannot `import tilelang` on pure CPU machine with CUDA build
### Required prerequisites

### Motivation

For a training framework, most of the cases run with GPUs, but in some cases, e.g. debugging CPU dataloader, we don't want to waste GPU hours. So `import tilelang` **with the same CUDA build** should work on a machine without CUDA driver/GPU.

Now, an error may occur:
```
OSError: libcuda.so.1: cannot open shared object file: No such file or directory
```

XuehaiPan:
Resolves #1478

## Summary by CodeRabbit

* **New Features**
  * Added set_log_level() for runtime logging control.
  * Added a CUDA driver stub that provides a safe CPU-only fallback with runtime-loaded CUDA wrappers.

* **Performance / UX**
  * Heavy native dependencies now lazy-load to reduce import-time overhead.
  * Header inclusion tightened to enforce stub usage.

* **Chores**
  * Unified library output/install layout and removed direct CUDA runtime links at install time.

* **CI Improvements**
  * Added CI env vars, caching and setup steps, replaced yum with dnf, adjusted CUDA build matrix, and added Python 3.12 wheel validation/testing.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -194,15 +194,18 @@ jobs:
 
       - name: Test built wheels
         run: |
-          # Clean uv cache to avoid using compatible but undesired torch versions in cache
-          uv cache clean torch
-
           for WHEEL in wheelhouse/*.whl; do
             echo "Testing wheel: ${WHEEL}"
             (
               set -e
               uv venv --python=3.12 test-venv
               source test-venv/bin/activate
+
+              uv pip install --upgrade pip setuptools wheel
+              if [[ "${UV_INDEX}" == *"/nightly/"* ]]; then
+                uv pip install --prerelease=allow -v torch
+              fi
+
               uv pip install -v "${WHEEL}"
               (
                 set -e


pyproject.toml
```diff
@@ -260,10 +260,16 @@ esac
 
 cudaver="$(echo "${CUDA_VERSION:-$DEFAULT_CUDA_VERSION}" | cut -d '.' -f-2)"
 v="${cudaver//./-}"
-yum install -y "cuda-minimal-build-${v}" "cuda-driver-devel-${v}" "cuda-nvrtc-devel-${v}" nvidia-driver-cuda-libs
-yum clean all
+dnf install -y "cuda-minimal-build-${v}" "cuda-driver-devel-${v}" "cuda-nvrtc-devel-${v}" nvidia-driver-cuda-libs
```
oraluben: We don't need `libcuda.so.1` anymore

```suggestion
dnf install -y "cuda-minimal-build-${v}" "cuda-driver-devel-${v}" "cuda-nvrtc-devel-${v}"
```

pyproject.toml
```diff
@@ -260,10 +260,16 @@ esac
...
+dnf clean all
 """
 repair-wheel-command = [
+    # Exclude libcuda.so.1 to allow importing on a CPU-only machine.
+    "pipx run wheel unpack --dest {dest_dir}/tmp_dir {wheel}",
+    "find {dest_dir}/tmp_dir -name '*.so*' -exec patchelf --remove-needed libcuda.so.1 '{}' '+'",
```
oraluben: Can we move this to cmake script? Current we already introduced `patchelf` as a build-time dependency to resolve some z3 issue: 

https://github.com/tile-ai/tvm/blob/03ad7ccd20058ad9576571375771129a98e72756/CMakeLists.txt#L568-L583

LeiWang1999: agree.

.github/workflows/dist.yml
```diff
@@ -149,12 +154,15 @@ jobs:
           if [[ "${{ matrix.target.toolkit }}" == *"CUDA"* ]]; then
             CUDA_VERSION="${{ matrix.target.toolkit }}"
             CUDA_VERSION="${CUDA_VERSION#CUDA-}"
+            CUDA_VERSION_MAJMIN="$(echo ${CUDA_VERSION} | cut -d '.' -f-2)"
+            CUDA_VERSION_MAJMIN_NODOT="${CUDA_VERSION_MAJMIN//./}"
             echo "CUDA_VERSION=${CUDA_VERSION}" | tee -a "${GITHUB_ENV}"
+            echo "UV_INDEX=https://download.pytorch.org/whl/cu${CUDA_VERSION_MAJMIN_NODOT}" | tee -a "${GITHUB_ENV}"
```
oraluben:
```suggestion
            echo "UV_TORCH_BACKEND=cu${CUDA_VERSION_MAJMIN_NODOT}" | tee -a "${GITHUB_ENV}"
```

Ref: https://docs.astral.sh/uv/guides/integration/pytorch/#automatic-backend-selection

XuehaiPan: The backend enums are compiled into the `uv` binary. If the backend does not exist, the command will error out.

CMakeLists.txt
```diff
@@ -303,33 +335,88 @@ install(TARGETS tilelang_cython_wrapper
 # Copy libz3.so to build folder to workaround isolated build env issue
 if(USE_Z3 AND USE_PYPI_Z3)
   get_target_property(Z3_LIBRARY_PATH z3::libz3 IMPORTED_LOCATION)
-  install(FILES "${Z3_LIBRARY_PATH}" DESTINATION "${CMAKE_BINARY_DIR}/tvm")
+  install(FILES "${Z3_LIBRARY_PATH}" DESTINATION "${CMAKE_BINARY_DIR}/lib")
```
oraluben, line 332: This seems not right as `libtvm.so` will look into the same folder for `libz3.so`

CMakeLists.txt
```diff
@@ -237,13 +263,19 @@ endif()
 # Include tvm after configs have been populated
 add_subdirectory(${TVM_SOURCE} tvm EXCLUDE_FROM_ALL)
 
+# Link TVM libraries against CUDA stub for lazy loading
+if(USE_CUDA)
```
oraluben: duplicated linking

.github/workflows/dist.yml
oraluben: Since we're using cuda 13.0 now I think we can use `UV_TORCH_BACKEND` for these reasons:
1. `UV_INDEX` is treated as additional index and the behavior of multiple indexes providing same package is undocumented;
2. The minor version of cuda toolchain used to compile tilelang is not important, as long as all kernels are jit. We could use wheels from 13.0 for all 13.x, so it's ok if uv don't have the latest backend.

.github/workflows/dist.yml
```diff
@@ -35,6 +35,11 @@ env:
   COLUMNS: "100"
   FORCE_COLOR: "1"
   CLICOLOR_FORCE: "1"
+  UV_INDEX_STRATEGY: "unsafe-best-match"
```
oraluben, line 38: See comment about `UV_INDEX` below.

oraluben (commented):
Overall LGTM, amazing work!

CMakeLists.txt
XuehaiPan: Now everything is under `build/lib`.

.github/workflows/dist.yml
XuehaiPan: `UV_INDEX` combined with `UV_INDEX_STRATEGY: "unsafe-best-match"` is more consistent with the default behavior with `pip`.

If you are using `UV_INDEX` with default `UV_INDEX_STRATEGY: "first-index"`, that can cause problems. See also:

- pytorch/pytorch#156694

oraluben: It's not clear if `uv pip` would use `torch==2.9.1+cu130` from pytorch pip or `torch==2.9.1+cu128` from pypi, I think it's ok to leave it to `uv` with `UV_TORCH_BACKEND`, which is more explicit.

.github/workflows/dist.yml
oraluben: Do we still need this since we have `UV_TORCH_BACKEND`?

XuehaiPan:
cc @LeiWang1999 to stamp this

oraluben (dismissed):

oraluben (approved):

LeiWang1999 (approved):

Issue #1465, opened 2025-12-18
[BUG] Compilation time increases greatly after Z3 introduced


lucifer1004:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7

### System information

3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] linux
0.1.7+cu130.gitf996239f
2.10.0a0+b558c986e8.nv25.11

### Problem description

Current:

<img width="743" height="445" alt="Image" src="https://github.com/user-attachments/assets/d315462e-b162-4352-8fe1-b045e637904b" />

Before Z3:

<img width="743" height="445" alt="Image" src="https://github.com/user-attachments/assets/247dcc68-9d50-4b6f-8941-aca71f09912d" />

### Reproducible example code

See https://github.com/lucifer1004/tilelang-cutedsl-benchmarks.

### Traceback

```pytb

```

kurisu6912:
z3 has a large launch overhead. Even if proving 1!=2, it cost 200us to start up. I will restrict the use of z3 to prove strength only symbolic bound. This should save a lot of time.

LeiWang1999:
Compare these two commits: 899f7bd5 and 1a3a64fb using the simple MHA compilation script [mha_compile.py](https://github.com/user-attachments/files/24250057/mha_compile.py). 

I was unable to reproduce the issue on my Intel CPUs with H800. This suggests the problem may be specific to AMD processors. Would be great if sb. have access to an AMD CPUs can help test this.

LeiWang1999:
This has been addressed :) closed.

PR #1482, opened 2025-12-19
[Enhancement] Use static Z3 context 

LeiWang1999:
We use a static Z3 context so all analyzers can share a common context, because Z3 initialization is slow on some CPUs (e.g., AMD EPYC 7502 32-Core). which's super helpful for issue #1465 

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependency reference


LeiWang1999:
LTCP, Merged:)

Issue #1463, opened 2025-12-18
[Feature Request] PDL Support


LeiWang1999:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

[Programmatic Dependent Launch and Synchronization](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#programmatic-dependent-launch-and-synchronization)

helpful for us to overlap two kernels. 

### Solution

Introduce `T.trigger_pdl` and `T.pdl_sync` for `cudaTriggerProgrammaticLaunchCompletion` and `cudaGridDependencySynchronize`.

Introduce pdl initialization which we can borrow from `T.annotate_l2_hit_ratio`

w169q169:
I can come help with this

LeiWang1999:
Thanks @w169q169 feel free to connect if you come across any issues :)

w169q169:
@LeiWang1999 Hey! I’ve got this PR #1494 ready for review. This feature was kinda tricky for me as a beginner, so any tips or feedback would be super helpful. Thanks!

PR #1494, opened 2025-12-22
[Feat] PDL Support

w169q169:
Add PDL support as per Issue #1463

## Summary by CodeRabbit

* **New Features**
  * Added CUDA PDL primitives pdl_trigger() and pdl_sync() and end-to-end support for programmatic dependent launches on CUDA CC ≥ 9.

* **Documentation**
  * Instruction Reference gains a new "Synchronization helpers" subsection documenting the new primitives.

* **Tests**
  * Added tests exercising NVRTC, JIT (ctypes/FFI), and TileLang PDL integration.

* **Behavior**
  * Host launch and codegen paths now recognize and honor per-kernel PDL synchronization hints.

### DIFF
--- a/tilelang/jit/adapter/cython/adapter.py
+++ b/tilelang/jit/adapter/cython/adapter.py
@@ -385,3 +385,7 @@ class CythonKernelAdapter(BaseKernelAdapter):
             # Wrapper only has host kernel source
             assert self.host_kernel_source is not None, "Wrapped source is not available"
             return self.host_kernel_source
+
+    def get_host_source(self):
+        """Returns the source code of the host function."""
+        return self.host_kernel_source


src/transform/lower_pdl.cc
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+private:
+  bool has_trigger_launch_ = false;
+  bool has_grid_sync_ = false;
+
+  MarkCudaSyncCalls() = default;
+};
+
+class ElininateCudaSyncCalls : public StmtExprMutator {
```
silentCoder-dev: There's a typo in the word "Eliminate".

testing/python/jit/test_tilelang_jit_cython.py
silentCoder-dev, line 1: Test failure with `kernel_cache` enabled.
```plaintext
2025-12-22 17:12:03  [TileLang:tilelang.cache.kernel_cache:ERROR]: Error saving host kernel source code to disk
  File "/home/silentCoder-dev/tilelang/tilelang/cache/kernel_cache.py", line 316, in _save_kernel_to_disk
    KernelCache._safe_write_file(host_kernel_path, "w", lambda file: file.write(kernel.adapter.get_kernel_source()))
  File "/home/silentCoder-dev/tilelang/tilelang/cache/kernel_cache.py", line 266, in _safe_write_file
    operation(temp_file)
  File "/home/silentCoder-dev/tilelang/tilelang/cache/kernel_cache.py", line 316, in <lambda>
    KernelCache._safe_write_file(host_kernel_path, "w", lambda file: file.write(kernel.adapter.get_kernel_source()))
                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: write() argument must be str, not None
```

w169q169: This looks like an existing issue on main rather than something introduced by this PR.
The adapter can return None for the kernel source, which breaks the cache write path.

Happy to fix this in a follow-up PR if that makes sense, but I’d prefer to keep this PR focused on its original scope.

silentCoder-dev: I looked into it, and the issue seems to stem from the existing `kernel_cache` implementation. Since this PR doesn't modify that component, we might need to address it separately.

silentCoder-dev:
The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.

src/transform/lower_pdl.cc
```diff
@@ -0,0 +1,154 @@
+/*!
+ * \file lower_pdl.cc
+ * \brief Mark Device PrimFunc with attributes if CUDA sync functions are called
```
silentCoder-dev: "CUDA PDL functions are called"

testing/python/jit/test_tilelang_jit_cython.py
silentCoder-dev, line 1: move this into the test specifically for the "cython" backend.

w169q169: This issue is already tracked in #1508

w169q169: Just to clarify, when you say “move this into the test specifically for the cython backend”, do you prefer:

1. Renaming testing/python/jit/test_tilelang_jit_ctypes.py to testing/python/jit/test_tilelang_jit_cython.py, or
2. Keeping the existing ctypes test and adding a new dedicated test file testing/python/jit/test_tilelang_jit_cython.py, or
3. Some other restructuring you have in mind (e.g. splitting backend-specific cases into separate files)?

I’m happy to follow whichever structure you think is more appropriate.

w169q169:
> The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.

Thanks for the clarification. From my understanding, the Cython backend itself is fine.
The difficulty we’re running into is that PDL requires modifying the launchConfig for the secondary kernel, and I’m not sure how this can be done cleanly when going through tvm_ffi.

In particular, with the tvm_ffi path we eventually call cudaLaunchKernelEx, but it’s not obvious to me where (or whether) we can intercept and adjust the cudaLaunchConfig_t there.
If there is a supported or recommended way to modify the cudaLaunchKernelEx config in the tvm_ffi flow, that would be ideal. Otherwise, this seems to be the main blocker for enabling PDL in this backend.

silentCoder-dev:
> The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.



> > The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.
> 
> Thanks for the clarification. From my understanding, the Cython backend itself is fine. The difficulty we’re running into is that PDL requires modifying the launchConfig for the secondary kernel, and I’m not sure how this can be done cleanly when going through tvm_ffi.
> 
> In particular, with the tvm_ffi path we eventually call cudaLaunchKernelEx, but it’s not obvious to me where (or whether) we can intercept and adjust the cudaLaunchConfig_t there. If there is a supported or recommended way to modify the cudaLaunchKernelEx config in the tvm_ffi flow, that would be ideal. Otherwise, this seems to be the main blocker for enabling PDL in this backend.

You can manipulate this in `src/runtime/runtime.cc`. For example, see the test `test_tvm_ffi_l2_persistent_map` in `testing/python/jit/test_tilelang_jit_tvm_ffi.py`, which invokes `tvm_cuda_stream_set_access_policy_window`.

testing/python/jit/test_tilelang_jit_cython.py
silentCoder-dev: I think Choice 1 is sufficient.

w169q169:
> > The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.
> 
> > > The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.
> > 
> > 
> > Thanks for the clarification. From my understanding, the Cython backend itself is fine. The difficulty we’re running into is that PDL requires modifying the launchConfig for the secondary kernel, and I’m not sure how this can be done cleanly when going through tvm_ffi.
> > In particular, with the tvm_ffi path we eventually call cudaLaunchKernelEx, but it’s not obvious to me where (or whether) we can intercept and adjust the cudaLaunchConfig_t there. If there is a supported or recommended way to modify the cudaLaunchKernelEx config in the tvm_ffi flow, that would be ideal. Otherwise, this seems to be the main blocker for enabling PDL in this backend.
> 
> You can manipulate this in `src/runtime/runtime.cc`. For example, see the test `test_tvm_ffi_l2_persistent_map` in `testing/python/jit/test_tilelang_jit_tvm_ffi.py`, which invokes `tvm_cuda_stream_set_access_policy_window`.

Thanks for the suggestion. Unfortunately, this approach cannot achieve what we need.

`tvm_cuda_stream_set_access_policy_window` only modifies the behavior of the CUDA stream, which is fine for L2 persistence, but PDL specifically requires modifying the launchConfig of the kernel. In the tvm_ffi path, cudaLaunchKernel is implemented inside TVM itself, at:

https://github.com/apache/tvm/blob/main/src/runtime/cuda/cuda_module.cc#L184-L207


There is an existing draft PR here: https://github.com/apache/tvm/pull/18604
Once this is merged in TVM, we can continue implementing PDL support in the tvm_ffi backend.

silentCoder-dev:
> > > The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.
> > 
> > 
> > > > The Cython backend implementation is correct. The issue lies in the missing support for tvm_ffi. I'm currently working on adding it.
> > > 
> > > 
> > > Thanks for the clarification. From my understanding, the Cython backend itself is fine. The difficulty we’re running into is that PDL requires modifying the launchConfig for the secondary kernel, and I’m not sure how this can be done cleanly when going through tvm_ffi.
> > > In particular, with the tvm_ffi path we eventually call cudaLaunchKernelEx, but it’s not obvious to me where (or whether) we can intercept and adjust the cudaLaunchConfig_t there. If there is a supported or recommended way to modify the cudaLaunchKernelEx config in the tvm_ffi flow, that would be ideal. Otherwise, this seems to be the main blocker for enabling PDL in this backend.
> > 
> > 
> > You can manipulate this in `src/runtime/runtime.cc`. For example, see the test `test_tvm_ffi_l2_persistent_map` in `testing/python/jit/test_tilelang_jit_tvm_ffi.py`, which invokes `tvm_cuda_stream_set_access_policy_window`.
> 
> Thanks for the suggestion. Unfortunately, this approach cannot achieve what we need.
> 
> `tvm_cuda_stream_set_access_policy_window` only modifies the behavior of the CUDA stream, which is fine for L2 persistence, but PDL specifically requires modifying the launchConfig of the kernel. In the tvm_ffi path, cudaLaunchKernel is implemented inside TVM itself, at:
> 
> https://github.com/apache/tvm/blob/main/src/runtime/cuda/cuda_module.cc#L184-L207
> 
> There is an existing draft PR here: [apache/tvm#18604](https://github.com/apache/tvm/pull/18604) Once this is merged in TVM, we can continue implementing PDL support in the tvm_ffi backend.

I locally implement similar functionality with that draft PR (: I believe this implementation is sufficient, and we can proceed by adding `has_programmatic_dependent_launch` soon after that pr is merged.

LeiWang1999:
Thanks. Since the ctypes execution backend was removed in #1510, we can also remove the related caching mechanisms in this PR :)

src/transform/lower_pdl.cc
```diff
@@ -0,0 +1,154 @@
...
+#include <tvm/tir/stmt_functor.h>
+#include <tvm/tir/transform.h>
+
+namespace tvm {
+namespace tl {
+
+namespace attr {
+// Attributes to mark CUDA sync calls
+constexpr const char *kHasTriggerLaunch = "has_cuda_pdl_trigger";
+constexpr const char *kHasGridSync = "has_cuda_pdl_sync";
```
silentCoder-dev: These attributes should be declared in the header file to ensure proper visibility across other files, such as `src/target/codegen_cuda.cc`.

src/transform/lower_pdl.cc
```diff
@@ -0,0 +1,154 @@
...
+    if (mutator.has_trigger_launch_) {
+      new_f = WithAttr(std::move(new_f), attr::kHasTriggerLaunch, 1);
+    }
+    if (mutator.has_grid_sync_) {
+      new_f = WithAttr(std::move(new_f), attr::kHasGridSync, 1);
+    }
+    return new_f;
+  }
+
+  Stmt VisitStmt_(const EvaluateNode *op) final {
```
silentCoder-dev: The `VisitStmt_` definition appears unnecessary here. `VisitExpr_` can already handle "pdl_sync" cases.

w169q169:
image

CI is failing with ruff-check (F821 Undefined name 'autotune') in an example file that is untouched by this PR.
The same code passes locally.
Not sure if this is caused by CI ruff config or environment differences.
Any help would be appreciated.

silentCoder-dev:
> image
> CI is failing with ruff-check (F821 Undefined name 'autotune') in an example file that is untouched by this PR. The same code passes locally. Not sure if this is caused by CI ruff config or environment differences. Any help would be appreciated.

You can update your local branch by pulling from `tilelang/main`.

silentCoder-dev (changes requested):
When `have_pdl` is not supported, we should skip the `lower_pdl` pass. Running it to try and eliminate PDL calls would be incorrect or unnecessary.

w169q169:
> When `have_pdl` is not supported, we should skip the `lower_pdl` pass. Running it to try and eliminate PDL calls would be incorrect or unnecessary.



I have thought about this concern, and my reasoning is as follows.

`cudaGridDependencySynchronize()` and
`cudaTriggerProgrammaticLaunchCompletion()`
are not just semantic markers; they lower to concrete device instructions whose **placement in the instruction stream matters**.

For example, in the following kernel:

```cuda
__global__ void square(int* array, int n) {
    int tid = blockDim.x * blockIdx.x + threadIdx.x;
    if (tid < n)
        array[tid] = array[tid] * array[tid];
        cudaGridDependencySynchronize();
        array[tid] = array[tid] * array[tid];
}
```

`cudaGridDependencySynchronize()` is lowered to specific SASS instructions (`BSYNC B0` and `ACQBULK`), which are emitted at precise locations in the control flow:

```sass
# nvcc -gencode arch=compute_90,code=sm_90
# compiled via godbolt.org

...
 STG.E desc[UR4][R2.64], R5 
.L_x_1:
 BSYNC B0 
.L_x_0:
 ACQBULK 
 LDG.E R0, desc[UR4][R2.64] 
 IMAD R5, R0, R0, RZ 
 STG.E desc[UR4][R2.64], R5 
 EXIT 
...
```

This positional dependency is the key reason I am cautious about aggressively running `lower_pdl` when `have_pdl` is not supported.

If we model PDL via `tir.call_extern`, then I agree this introduces extra passes whose sole purpose is to eliminate these calls on non-CUDA targets. That is redundant, but at least it preserves **exact placement semantics** until codegen.

I also considered an alternative using `block.attr`, similar to `T.annotate_l2_hit_ratio`. This avoids enabling a CUDA-specific lowering pass on non-CUDA targets. However, the drawback is that we lose precise location information. For example:

```python
with T.Kernel(T.ceildiv(N, block_size), threads=block_size) as (bx2,):
    for i in T.Parallel(block_size):
        idx = bx2 * block_size + i
        if idx < N:
            C[idx] = B[idx] * 2.0
    T.pdl_sync()
    for i in T.Parallel(block_size):
        idx = bx2 * block_size + i
        T.pdl_sync()
        if idx < N:
            C[idx] = B[idx] * 2.0
```

In this case, `pdl_sync` becomes attached to the `T.Kernel` block, but its **exact position relative to surrounding instructions is no longer explicit**, which is problematic for something that lowers to synchronization instructions.

One possible direction is to avoid `call_extern` and instead use `attr`, then explicitly materialize the corresponding device-side calls during codegen for each backend. However, the trade-off is that this approach still requires backend-specific handling and is conceptually similar in cost to eliminating function calls via passes.


These are my current thoughts. If there are mistakes in this reasoning, or if you see a cleaner or more idiomatic approach, I would greatly appreciate your guidance.

silentCoder-dev:
> > When `have_pdl` is not supported, we should skip the `lower_pdl` pass. Running it to try and eliminate PDL calls would be incorrect or unnecessary.
> 
> I have thought about this concern, and my reasoning is as follows.
> 
> `cudaGridDependencySynchronize()` and `cudaTriggerProgrammaticLaunchCompletion()` are not just semantic markers; they lower to concrete device instructions whose **placement in the instruction stream matters**.
> 
> For example, in the following kernel:
> 
> ```cuda-c++
> __global__ void square(int* array, int n) {
>     int tid = blockDim.x * blockIdx.x + threadIdx.x;
>     if (tid < n)
>         array[tid] = array[tid] * array[tid];
>         cudaGridDependencySynchronize();
>         array[tid] = array[tid] * array[tid];
> }
> ```
> 
> `cudaGridDependencySynchronize()` is lowered to specific SASS instructions (`BSYNC B0` and `ACQBULK`), which are emitted at precise locations in the control flow:
> 
> ```sass
> # nvcc -gencode arch=compute_90,code=sm_90
> # compiled via godbolt.org
> 
> ...
>  STG.E desc[UR4][R2.64], R5 
> .L_x_1:
>  BSYNC B0 
> .L_x_0:
>  ACQBULK 
>  LDG.E R0, desc[UR4][R2.64] 
>  IMAD R5, R0, R0, RZ 
>  STG.E desc[UR4][R2.64], R5 
>  EXIT 
> ...
> ```
> 
> This positional dependency is the key reason I am cautious about aggressively running `lower_pdl` when `have_pdl` is not supported.
> 
> If we model PDL via `tir.call_extern`, then I agree this introduces extra passes whose sole purpose is to eliminate these calls on non-CUDA targets. That is redundant, but at least it preserves **exact placement semantics** until codegen.
> 
> I also considered an alternative using `block.attr`, similar to `T.annotate_l2_hit_ratio`. This avoids enabling a CUDA-specific lowering pass on non-CUDA targets. However, the drawback is that we lose precise location information. For example:
> 
> ```python
> with T.Kernel(T.ceildiv(N, block_size), threads=block_size) as (bx2,):
>     for i in T.Parallel(block_size):
>         idx = bx2 * block_size + i
>         if idx < N:
>             C[idx] = B[idx] * 2.0
>     T.pdl_sync()
>     for i in T.Parallel(block_size):
>         idx = bx2 * block_size + i
>         T.pdl_sync()
>         if idx < N:
>             C[idx] = B[idx] * 2.0
> ```
> 
> In this case, `pdl_sync` becomes attached to the `T.Kernel` block, but its **exact position relative to surrounding instructions is no longer explicit**, which is problematic for something that lowers to synchronization instructions.
> 
> One possible direction is to avoid `call_extern` and instead use `attr`, then explicitly materialize the corresponding device-side calls during codegen for each backend. However, the trade-off is that this approach still requires backend-specific handling and is conceptually similar in cost to eliminating function calls via passes.
> 
> These are my current thoughts. If there are mistakes in this reasoning, or if you see a cleaner or more idiomatic approach, I would greatly appreciate your guidance.

Testing on SM_80 reveals that NVCC fails due to the `griddepcontrol` instruction requiring SM_90+. To ensure consistent behavior, the lower_pdl pass has been updated to throw the same error. In my view, addressing this hardware limitation (by targeting a compatible architecture) is a user responsibility; the pass should not automatically eliminate PDL calls as a workaround.

A similar pattern can be observed with the `__ldg` intrinsic. For example, in `testing/python/language/test_tilelang_language_intrinsics_codegen.py`, targeting a non-CUDA backend like "metal" results in the error: "tvm.error.InternalError: Unresolved call Op(tl.__ldg)".

tilelang/jit/adapter/nvrtc/libgen.py
```diff
@@ -197,6 +197,14 @@ def compile_lib(self, timeout: float | None = None):
...
                 f"-D__CUDACC_VER_MAJOR__={__CUDACC_VER_MAJOR__}",
             ]
+
+            # Add CUDA C++ standard library include path.
+            # CUDA 13+ uses the CCCL-based cuda::std layout, while older versions use the legacy path.
+            if __CUDACC_VER_MAJOR__ >= 13:
+                options += [f"-I{cuda_home}/targets/{target_arch}/include/cccl/cuda/std"]
+            else:
+                options += [f"-I{cuda_home}/targets/{target_arch}/include/cuda/std"]
+
```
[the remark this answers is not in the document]
silentCoder-dev: @w169q169 Could you review it again to confirm it's correct?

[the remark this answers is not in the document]
w169q169: Thanks for the suggestion. I believe this change would be incorrect in this context.

The headers referenced here are not CCCL headers themselves, but the CUDA std / libc++ compatibility headers provided by CUDA. These include paths are added solely to ensure consistent behavior between NVRTC and NVCC, rather than to directly consume CCCL’s public headers.

Historically, the NVRTC implementation has used {cuda_home}/targets/{target_arch}/include/..., and this patch intentionally follows that existing convention to avoid introducing behavioral differences. The targets/{target_arch}/include directory is already part of CUDA’s toolchain layout for device compilation, and NVRTC relies on it to mirror NVCC’s include resolution.

In other words, this change is about alignment with prior NVRTC behavior, not about relocating CCCL headers. Switching to {cuda_home}/include/cccl/... here would diverge from the established NVRTC setup and could introduce inconsistencies rather than fix them.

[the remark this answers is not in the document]
w169q169: I agree that NVRTC and NVCC have many behavioral differences, and include-path discrepancies have already caused multiple issues . Debugging NVRTC-specific problems is indeed painful.

That said, I think this concern is out of scope for this PR. The goal here is to keep behavior consistent with the existing NVRTC implementation, not to reconcile or redesign the broader NVRTC vs. NVCC include model.

I suggest we track the include-path divergence and NVRTC-related issues in a separate issue, where we can discuss a more systematic solution without blocking or overloading this change.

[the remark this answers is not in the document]
silentCoder-dev: @coderabbitai Read the comment above.

silentCoder-dev (approved):
LGTM!

LeiWang1999:
@codex review

tilelang/jit/adapter/wrapper.py
```diff
@@ -138,6 +138,22 @@
...
+\t\tcudaLaunchAttribute attribute[1];
+\t\tattribute[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
+\t\tattribute[0].val.programmaticStreamSerializationAllowed = 1;
+\t\tconfig.attrs = attribute;
+\t\tconfig.numAttrs = 1;
+\t\tconfig.stream = stream;
+\t\tconfig.gridDim = {0};
+\t\tconfig.blockDim = {1};
+\t\tconfig.dynamicSmemBytes = {2};
+\t\tcudaLaunchKernelEx(&config, {4}, {3});
```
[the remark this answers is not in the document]
silentCoder-dev: `cudaLaunchKernelExC` uses a `void**` for its arguments, while `cudaLaunchKernelEx` employs a template to accept `ArgsTypes&&... args`.

LeiWang1999:
Thanks @w169q169 @silentCoder-dev. I tested the example, and the kernels overlap in both the `cython` and `tvm_ffi` execution backends. Note that for `tvm_ffi`, overlapping occurs after the first launch, as the initial run includes JIT overhead."
image

LeiWang1999:
Merged.

PR #1512, opened 2025-12-23
[Misc] add env for default target/backend/verbose

lucifer1004:
With these env vars, we can batch run examples for a specific target/backend.

## Summary by CodeRabbit

* **New Features**
  * Environment-driven defaults: new TILELANG_* vars to set default target, execution backend, and verbosity.
  * Public APIs now accept optional parameters (target, execution backend, verbose, etc.) and will read defaults from the environment when omitted.
  * Example updated to use the direct JIT decorator instead of the previous autotune wrapper.

* **Documentation**
  * Docstrings and usage docs updated to describe environment-driven defaults and the new environment variables.

### DIFF
--- a/examples/flash_decoding/example_gqa_decode_varlen_logits.py
+++ b/examples/flash_decoding/example_gqa_decode_varlen_logits.py
@@ -7,7 +7,6 @@ import tilelang
 import tilelang.language as T
 
 torch.manual_seed(0)
-tilelang.disable_cache()
 
 
 def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
@@ -197,7 +196,6 @@ def get_configs():
     return configs
 
 
-@tilelang.autotune(configs=get_configs(), warmup=10, rep=10)
 @tilelang.jit(out_idx=[-2, -1])
 def flashattn(
     batch, heads, k_heads, max_seqlen_kv, total_seqlen_k, dim, has_sink, block_N=128, block_H=64, num_split=1, num_stages=1, threads=128


tilelang/jit/__init__.py
```diff
@@ -470,26 +512,45 @@ def jit(  # This is the new public interface
...
+
     Returns
     -------
     Callable
         Either a JIT-compiled wrapper around the input function, or a configured decorator
         instance that can then be applied to a function.
     """
 
+    # Apply environment variable defaults if parameters are not explicitly set
+    if target is None:
```
LeiWang1999: It would be better to move this part after kernel = self._kernel_cache.get(key, None). Otherwise, the outermost kernel cache will have to handle the target every time it's invoked.

tilelang/jit/__init__.py
```diff
@@ -118,36 +139,57 @@ def compile(
...
+    TILELANG_TARGET : str
+        Default compilation target (e.g., "cuda", "llvm"). Defaults to "auto".
+    TILELANG_EXECUTION_BACKEND : str
+        Default execution backend. Defaults to "auto".
+    TILELANG_VERBOSE : str
+        Set to "1", "true", "yes", or "on" to enable verbose compilation by default.
     """
+
+    # Apply environment variable defaults if parameters are not explicitly set
+    if target is None:
```
LeiWang1999: same as `def jit` below

LeiWang1999 (changes requested):
Overall LGTM, but left some performance concerns.

tilelang/jit/__init__.py
lucifer1004: But kernel cache key depends on target and backend anyway?

lucifer1004: Moved all env resolve logic to the inner-most kernel cache.

tilelang/jit/__init__.py
lucifer1004: Moved all env resolve logic to the inner-most kernel cache.

tilelang/jit/__init__.py
```diff
@@ -118,36 +111,49 @@ def compile(
...
-    execution_backend: Literal["auto", "dlpack", "tvm_ffi", "cython", "nvrtc", "torch", "cutedsl"] = "auto",
-    target: str | Target = "auto",
+    execution_backend: Literal["auto", "dlpack", "tvm_ffi", "cython", "nvrtc", "torch", "cutedsl"] | None = None,
+    target: str | Target | None = None,
     target_host: str | Target | None = None,
-    verbose: bool = False,
+    verbose: bool | None = None,
     pass_configs: dict[str, Any] | None = None,
     compile_flags: list[str] | str | None = None,
     num_workers: int = None,
```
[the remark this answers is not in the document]
lucifer1004: Fixed

tilelang/cache/__init__.py
```diff
@@ -17,10 +17,10 @@ def cached(
     func: PrimFunc = None,
     out_idx: list[int] = None,
     *args,
-    target: str | Target = "auto",
+    target: str | Target | None = None,
     target_host: str | Target = None,
-    execution_backend: Literal["auto", "tvm_ffi", "cython", "nvrtc", "torch"] | None = "auto",
-    verbose: bool | None = False,
+    execution_backend: Literal["auto", "tvm_ffi", "cython", "nvrtc", "torch"] | None = None,
+    verbose: bool | None = None,
```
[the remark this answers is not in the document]
lucifer1004: Fixed

LeiWang1999 (approved):

Issue #1138, opened 2025-10-28
[BUG] example_mha_fwd_varlen.py failed


lemyx:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.6.post1+cu126.git7a5077e4

### System information

```python
>>> print(sys.version, sys.platform)
3.10.16 (main, Dec 11 2024, 16:24:50) [GCC 11.2.0] linux
>>> print(tilelang.__version__)
0.1.6.post1+cu126.git7a5077e4
>>> print(torch.__version__)
2.7.1+cu126
```

### Problem description

The provided example codes of example_mha_fwd_varlen.py cannot be executed **on the causal mode**.

Note: the `causal` is set to `False` in the provided snippet by default, the snippet can pass the check as this moment. However, the snippet fails when the `causal` is modified to `True`.

### Reproducible example code

The Python snippets:

```python
python3 example_mha_fwd_varlen.py
```

### Traceback

```pytb
    torch.testing.assert_close(out, fla_out, rtol=1e-2, atol=1e-2)
  File "/usr/local/conda/lib/python3.10/site-packages/torch/testing/_comparison.py", line 1519, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 94582131 / 134217728 (70.5%)
Greatest absolute difference: 3.916015625 at index (5, 4, 4, 27) (up to 0.01 allowed)
Greatest relative difference: inf at index (0, 0, 0, 0) (up to 0.01 allowed)

```

Rachmanino:
The same reason as mentioned in #1139

DoubleRedX:
Any progress on this? I met the same problem.

Rachmanino:
Closed as resolved via #1530, and testing of causal varlen forward has been intergrated into TileLang CI

PR #1530, opened 2025-12-25
[BugFix] Fix bugs of varlen attention forward examples caused by `S_q != S_kv`

hukongyi:
In example_gqa_fwd_varlen.py, the bx index is already relative to each sequence's start. Adding q_current_seqlen to the loop range calculation was redundant and potentially incorrect for causal masking logic. This PR cleans up the calculation to correctly reflect the block range.

## Summary by CodeRabbit

* **New Features**
  * Added optional causal mode with a CLI flag to run causal attention variants.

* **Refactor**
  * Validation now uses Flash Attention as the reference and removes the legacy reference comparison.
  * Unified key/value handling for variable-length workloads; improved causal masking and iteration logic.
  * Numerical stability improvements in causal paths (replaced extreme infinities with stable values).

* **Tests**
  * Expanded coverage to include GQA example and both causal/non-causal MHA variants.

### DIFF
--- a/examples/flash_attention/example_gqa_fwd_varlen.py
+++ b/examples/flash_attention/example_gqa_fwd_varlen.py
@@ -117,7 +117,7 @@ def flashattn(batch_size, groups, UQ, UKV, heads, dim, is_causal, block_M=64, bl
             for i, j in T.Parallel(block_M, dim):
                 # When sq > skv, some tokens can see nothing
                 acc_o[i, j] = 0 if is_causal and bx * block_M + i + offset < 0 else acc_o[i, j] / logsum[i]
-           
+
             T.copy(acc_o, O_shared)
             for i, d in T.Parallel(block_M, dim):
                 if bx * block_M + i < q_current_seqlen:


LeiWang1999:
cc @Rachmanino

Rachmanino:
hi @hukongyi, thanks for your feedback. Actually instead of removing `q_current_seqlen` here, i think we should replace it with `kv_current_seqlen - q_current_seqlen` to correctly handle cases where `seqlen_q != seqlen_kv`. This is because that we usually use **right alignment** (checkout fa2 https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/mask.py#L197 as an example). I'll double check this later

LeiWang1999:
and I think we should also add a test to test the causal implementation :)

hukongyi:
> hi @hukongyi, thanks for your feedback. Actually instead of removing `q_current_seqlen` here, i think we should replace it with `kv_current_seqlen - q_current_seqlen` to correctly handle cases where `seqlen_q != seqlen_kv`. This is because that we usually use **right alignment** (checkout fa2 https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/mask.py#L197 as an example). I'll double check this later

Hi @Rachmanino ,
I have converted this PR to a Draft and pushed my latest changes so you can review the full implementation directly.
I attempted to implement the right-alignment logic as discussed (using offset = kv_len - q_len) and updated both the TileLang kernel and the PyTorch reference accordingly. However, the correctness tests are currently failing.
As I am new to TileLang, I suspect I might be missing some nuances regarding the loop bounds or masking syntax. I would really appreciate it if you could take a look at the code when you have a moment and point me in the right direction.
Thanks for your patience!

Rachmanino:
> > hi @hukongyi, thanks for your feedback. Actually instead of removing `q_current_seqlen` here, i think we should replace it with `kv_current_seqlen - q_current_seqlen` to correctly handle cases where `seqlen_q != seqlen_kv`. This is because that we usually use **right alignment** (checkout fa2 https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/mask.py#L197 as an example). I'll double check this later
> 
> Hi @Rachmanino , I have converted this PR to a Draft and pushed my latest changes so you can review the full implementation directly. I attempted to implement the right-alignment logic as discussed (using offset = kv_len - q_len) and updated both the TileLang kernel and the PyTorch reference accordingly. However, the correctness tests are currently failing. As I am new to TileLang, I suspect I might be missing some nuances regarding the loop bounds or masking syntax. I would really appreciate it if you could take a look at the code when you have a moment and point me in the right direction. Thanks for your patience!

Sure, really appreciate your contribution to tilelang! I'll check this and add test for you later.

Rachmanino:
This is supposed to also fix #1138

examples/flash_attention/example_gqa_fwd_varlen.py
```diff
@@ -158,7 +115,10 @@ def main(
                 T.gemm(acc_s_cast, V_shared, acc_o, policy=T.GemmWarpPolicy.FullRow)
 
             for i, j in T.Parallel(block_M, dim):
-                acc_o[i, j] /= logsum[i]
+                if is_causal and bx * block_M + i + offset < 0:
```
LeiWang1999: How about: 

```python
acc_o[i, j] = 0 if is_causal and bx * block_M + i + offset < 0 else ( acc_o[i, j]  / logsum[i])
```
It’s simpler and may improve performance by generating predicate-based instructions.

Rachmanino: copy that

LeiWang1999 (approved):
Awesome contributions, LGTM now.

PR #1545, opened 2025-12-26
[Typing] Update Kernel signature and add type hints for buffer operations

clouds56:
I know it's not perfect, but better than nothing
```python
def __getitem__(self, idx) -> Buffer
```

This is not breaking, I think `T.Kernel` is intended to call with `T.Kernel(64, 128)` but not `T.Kernel([64], [128])`

## Summary by CodeRabbit

* **New Features**
  * Added new utility operations for expression handling.
  * Typing support added for indexing on buffer-like objects.

* **API Changes**
  * Simplified kernel-like call signature to accept block extents directly (no per-block list wrapping required).


clouds56:
@coderabbitai generate docstrings

kurisu6912 (approved):
LGTM!

PR #1547, opened 2025-12-26
improve benchmark on mps

clouds56:
## Summary by CodeRabbit

* **New Features**
  * Unified, cross-backend profiling: consistent timing and synchronization across CUDA, MPS, and CPU with automatic device selection and CPU fallback.
  * Metal backend: added ability to retrieve kernel source for inspection or diagnostics.

* **Bug Fixes**
  * More robust kernel caching: gracefully skips backends that do not produce on-disk cacheable artifacts and avoids erroneous write attempts.

### DIFF
--- a/tilelang/profiler/bench.py
+++ b/tilelang/profiler/bench.py
@@ -57,51 +57,37 @@ class suppress_stdout_stderr:
         self.errnull_file.close()
 
 
-IS_CUDA = torch.cuda.is_available()
-IS_MPS = hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
-if IS_CUDA:
-    device = "cuda:0"
-elif IS_MPS:
-    device = "mps:0"
-else:
-    device = "cpu"
-
-
-class Event:
+class CpuEvent:
     """Dummy event class for CPU timing compatibility."""
 
     def __init__(self, *args, **kwargs):
-        self.inner = None
-        if IS_CUDA:
-            self.inner = torch.cuda.Event(*args, **kwargs)
-        elif IS_MPS:
-            self.inner = torch.mps.Event(*args, **kwargs)
         self.enable_timing = kwargs.get("enable_timing", False)
         self.record_time = None
 
     def record(self):
-        if self.inner is not None:
-            self.inner.record()
-        elif self.enable_timing:
+        if self.enable_timing:
             self.record_time = time.perf_counter_ns()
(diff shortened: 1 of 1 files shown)


oraluben:
@clouds56 Did you run profile on MPS? I think it's blocked by upstream issue https://github.com/pytorch/pytorch/issues/162872. Also extra abstraction on top of event might introduce noice.

clouds56:
I just want to enable profiling on mps.
We could make Event alias of torch.cuda.Event/torch.mps.Event. but this might make type annotation a mess.

tilelang/profiler/bench.py
```diff
@@ -57,8 +58,57 @@ def __exit__(self, *_):
 
 
 IS_CUDA = torch.cuda.is_available()
-device = "cuda:0" if IS_CUDA else "mps:0"
```
oraluben, line 60: I think we can keep these and do same thing for the `synchronize` function, and use `Event`/`synchronize` below, instead of creating a new `Event` class.

clouds56: Ok, and what's the story for CPU? Shall we create a CpuEvent that uses `time`? Or leave it unsupported?

clouds56: Ok, and what's the story for CPU? Shall we create a CpuEvent that uses `time`? Or leave it unsupported?

oraluben: > Ok, and what's the story for CPU? Shall we create a CpuEvent that uses `time`? Or leave it unsupported?

I didn't test how the cpu backend works, I think we can leave it uncahnged.

tilelang/profiler/bench.py
```diff
@@ -56,9 +57,44 @@ def __exit__(self, *_):
...
+    Event = torch.cuda.Event
+elif IS_MPS:
+    device = "mps:0"
+    Event = torch.mps.Event
+else:
+    device = "cpu"
+    Event = CpuEvent
+
+
+def _synchronize() -> None:
```
oraluben, line 93: We don't want extra function call here, so might be better assign different impl to `synchronize`, as `Event` above.

clouds56:
I don't like global variables, it's hard to migrate if one platform could do benchmark with multiple targets.
So I'm closing this.

I think someone could do better after the blocking torch issue is resolved.

PR #1550, opened 2025-12-28
[Bugfix] Fix regression test to use installed package instead of source directory

xwhzz:
## Summary by CodeRabbit

* **Chores**
  * Reorganized CI regression workflow to add a dedicated permission-check step and streamline subsequent job flow.
  * Renamed report messaging to "Performance Regression Test Report" for clarity.
  * Improved CI cleanup to remove build/artifact directories during baseline setup.
  * Simplified and consolidated local performance/regression tooling, replacing older scripts with a unified regression runner and tightening test invocation.

### DIFF
--- a/.github/workflows/pr-regression-test-bot.yml
+++ b/.github/workflows/pr-regression-test-bot.yml
@@ -191,6 +191,10 @@ jobs:
           uv pip install -v .
           rm -rf tilelang build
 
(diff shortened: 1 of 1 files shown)


LeiWang1999:
Would be better for us to enable ccache as we should build from source twice?

LeiWang1999:
likely we will automatically enable ccache, lgtm.

PR #1559, opened 2025-12-29
[Parallel][Infer] Free-mode chooses minimal replication between buffer-based and PlanLoopPartition

LeiWang1999:
Title
- Minimize replication in free layout inference by comparing compute-from-buffer vs PlanLoopPartition and choosing the smallest valid replication.

Summary
- In `ParallelOpNode::InferLayout` (free mode), generate two loop layout candidates:
  - compute_loop_layout_from_buffer: always correct; may over-replicate
  - PlanLoopPartition: often yields smaller replication
- Choose the layout that minimizes replication while ensuring correctness:
  - Use `ProveFragmentContains` on loop vars to check mutual containment
  - If buffer-based contains plan-based → pick PlanLoopPartition (smaller rep)
  - If plan-based contains buffer-based → pick buffer-based
  - If neither contains the other, compare `ReplicateExtent`; pick smaller provably; otherwise fall back to buffer-based (safest)
- Preserve all existing guards/checks:
  - Injectivity check on chosen loop layout
  - `ProveFragmentContains` against all fragments in the loop
  - Predicates for replicated cross-thread stores to shared/global

Why
- `compute_loop_layout_from_buffer` is guaranteed correct but can inflate `replicate` and hurt performance.
- Some cases can use `PlanLoopPartition` with fewer replicas without sacrificing correctness. This change automatically prefers that smaller replication when containment holds.

Changes
- src/op/parallel.cc:
  - Free-mode: try both candidates, then select by containment and replicate extent
  - Keep vectorization + coalesced width handling and predicate guard logic unchanged
- src/transform/layout_inference.cc:
  - Prioritize queueing ops that touch any buffer with known layout (previously required all) to improve propagation order
  - Simplify touched-buffer dedup logic
- tilelang/language/annotations.py:
  - Enforce that fragment buffers annotated via `annotate_layout` must use `Fragment` objects (type guard)
- testing/python/issue/test_tilelang_issue_layout.py:
  - Add test ensuring free inference picks the smallest replication when feasible

Impact
- Only affects free layout inference path for parallel loops
- Expected to reduce redundant replication and improve performance in qualifying cases
- Keeps strict/common paths and all safety checks intact

Notes
- Logged selection decisions to help diagnose layout choices
- Happy to add more targeted tests if desired

## Summary by CodeRabbit

* **New Features**
  * Layout inference now evaluates two candidate layouts and deterministically selects the best (preferring contained candidates or those with smaller replication), improving memory efficiency.

* **Bug Fixes**
  * Prioritization logic refined to trigger sooner when any known layout anchor exists; improved candidate validation and replication-guard insertion for safer parallel access.

* **Tests**
  * New test covering the free-layout inference selection behavior.

* **Chores**
  * Added runtime validation to ensure fragment buffers carry fragment layouts.


LeiWang1999:
@regression-perf

Issue #1257, opened 2025-11-14
[BUG] Missing `__syncthreads()` after `AtomicAdd` in generated CUDA kernel


LJC00118:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.6.post2+cuda.git5eb30a4f

### System information

3.12.9 | packaged by Anaconda, Inc. | (main, Feb  6 2025, 18:56:27) [GCC 11.2.0] linux
0.1.6.post2+cuda.git5eb30a4f
2.7.0+cu128

### Problem description

In the CUDA kernel code generated by TileLang, there is a missing __syncthreads() synchronization call after the AtomicAdd operation.

#### The Python snippets:

```python
import torch
import tilelang
from tilelang import language as T


tilelang.disable_cache()
@tilelang.jit
def get_kernel(m: int):
    @T.prim_func
    def test_kernel(
        a: T.Tensor[(m, ), "int32"],
    ):
        with T.Kernel(1, threads=64) as (bx):
            shared = T.alloc_shared((64, ), "int32")
            tx = T.get_thread_binding(0)
            shared[tx ^ 1] = 0
            T.atomic_add(shared[tx], 1)
            a[tx] = shared[tx ^ 1]
    return test_kernel


m = 64
kernel = get_kernel(m)
print(kernel.get_kernel_source())
a = torch.zeros((m, ), device="cuda", dtype=torch.float32).to(torch.int32)
kernel(a)
```
#### Generated Code:
```c
extern "C" __global__ void __launch_bounds__(64, 1) test_kernel_kernel(int* __restrict__ a) {
  extern __shared__ __align__(1024) int shared[];
  shared[(((int)threadIdx.x) ^ 1)] = 0;
  __syncthreads();
  AtomicAdd(shared[((int)threadIdx.x)], 1);
  // missing '__syncthreads()' here
  a[((int)threadIdx.x)] = shared[(((int)threadIdx.x) ^ 1)];
}
```

### Traceback

```pytb

```

LeiWang1999:
hopefully this can be addressed from issue #1423

Issue #1423, opened 2025-12-13
[Feature Request] Replace all `address_of` with `access_ptr`


LeiWang1999:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

access_ptr carries richer semantic information—such as a pointer’s accessible range and its read/write permissions—which makes analysis easier and more precise. Many passes currently rely on the less strict address_of; we should improve this by replacing it with more precise constructs like access_ptr so the analysis becomes more rigorous.

Dayuxiaoshui:
@LeiWang1999 I can come help with this

PR #1581, opened 2025-12-31
[BugFix] : Add missing __syncthreads() after AtomicAdd and replace address_ of with access_ptr

Dayuxiaoshui:
This commit fixes two issues:

1. Issue #1257: Add missing __syncthreads() after AtomicAdd operations
   - Added ThreadSyncAfterAtomicInserter class in thread_storage_sync.cc
   - Automatically inserts __syncthreads() after AtomicAdd on shared memory
   - Integrated into TileLangThreadSync() pass
   - Fixes synchronization issue in generated CUDA kernels

2. Issue #1423: Replace address_of with access_ptr (tvm_access_ptr)
   - Updated atomic_add.cc to use MakeAccessPtrFromRegion
   - Updated atomicadd_vectorize.cc to use MakeAccessPtrFromRegion
   - Provides richer semantic information for analysis
   - Maintains backward compatibility

Changes:
- src/op/atomic_add.cc: Replace address_of with access_ptr in AtomicAdd operations
- src/transform/atomicadd_vectorize.cc: Replace address_of with access_ptr in vectorization
- src/transform/thread_storage_sync.cc: Add ThreadSyncAfterAtomicInserter for syncthreads insertion

Verification:
- Compilation: Success
- Runtime: Verified with test case
- Generated CUDA code: Confirmed __syncthreads() is present after AtomicAdd

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically insert thread synchronization barriers after atomic operations targeting shared memory to ensure correct visibility and ordering.

* **Refactor**
  * Standardized buffer argument handling to use region-based access pointers across atomic, vectorized, and transfer paths, improving consistency while preserving public interfaces and remaining compatible with legacy pointer forms.

### DIFF
--- a/src/op/atomic_add.cc
+++ b/src/op/atomic_add.cc
@@ -273,7 +273,6 @@ For AtomicAddNode::MakeSIMTLoop(arith::Analyzer *analyzer) const {
     src_value = Cast(dst->dtype, src_value);
 
   // Build a pointer to destination element using tvm_access_ptr
-  BufferLoad dst_load = BufferLoad(dst, dst_indices);
   Array<Range> dst_ranges;
   for (const PrimExpr &index : dst_indices) {
     dst_ranges.push_back(Range::FromMinExtent(index, 1));
--- a/src/transform/atomicadd_vectorize.cc
+++ b/src/transform/atomicadd_vectorize.cc
@@ -285,14 +285,8 @@ private:
               MakeAccessPtrFromRegion(dst_region, 2); // 2 = write access
           new_args.push_back(dst_ptr);
         } else if (const auto *call = node->args[0].as<CallNode>()) {
-          // If it's already an address_of or access_ptr, forward it; otherwise,
-          // keep original.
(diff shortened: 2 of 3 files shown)


Dayuxiaoshui:
@LeiWang1999 We have successfully fixed issue #1257 by inserting __syncthreads() synchronization after AtomicAdd operations on shared memory. The fix has been verified: it correctly inserts sync for shared memory AtomicAdd operations and does not insert sync for global memory AtomicAdd operations. There are 5 failing tests in CI related to backward pass, but all AtomicAdd operations in these tests target global memory (dQ, dV, dK), which should not be affected by our fix. The basic fix has been verified as correct, and the failing tests are likely caused by other factors that require further investigation.

LeiWang1999:
Thanks for your valuable contribution! But after further consideration, we introduced a new op to fundamentally solve the issues to combine tvm_access_ptr and address_of (as they both have their advs and disadvs) in pr #1827. closed for now :)

PR #1582, opened 2025-12-31
[Feature] Add more curand operations & support vectorization

silentCoder-dev:
## Summary by CodeRabbit

* **New Features**
  * Added floating-point RNG output (single/double precision) and a selectable RNG generator option for initialization.
* **Performance**
  * Improved vectorization behavior so RNG calls can be vectorized more effectively across lanes.
* **Tests**
  * Expanded tests to exercise multiple RNG generators and multiple RNG-generated outputs per run.


LeiWang1999 (commented):
Can we merge those `rng_rand*` functions into a same `rng_rand`?

Issue #1549, opened 2025-12-27
[BUG] Scalar variable incorrectly vectorized in T.Parallel loop (regression in 0.1.7.post1)


WXY277:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7.post1

### System information

torch      2.5.0+cu121

### Problem description

Pre-allocated scalar variables using `T.alloc_var()` are incorrectly treated as arrays when assigned inside `T.Parallel` loops, causing compilation errors.  This is a regression from 0.1.6.post2 where the same code works correctly.

### Reproducible example code

The Python snippets:

```import torch
import tilelang
import tilelang.language as T

@tilelang.jit(
    pass_configs={
        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
    },
)
def minimal_repro_kernel(
    batch:  int,
    seq_len:  int,
    feature_dim: int,
    BLOCK_SIZE: int = 16,
    num_threads: int = 64,
):
    
    INT32 = "int32"
    data_shape = [batch, seq_len, feature_dim]

    @T.prim_func
    def kernel_func(
        DataIn: T. Tensor(data_shape, INT32),
        DataOut: T.Tensor(data_shape, INT32),
    ):
        with T. Kernel(
            tilelang.cdiv(seq_len, BLOCK_SIZE), 
            batch, 
            threads=num_threads
        ) as (bx, bz):
            i_b = bz
            base_l = bx * BLOCK_SIZE

            shared_mem = T.alloc_shared([BLOCK_SIZE, feature_dim], dtype=INT32)

            # Load data to shared memory
            for l_idx, k in T.Parallel(BLOCK_SIZE, feature_dim):
                lq = base_l + l_idx
                if lq < seq_len:
                    shared_mem[l_idx, k] = DataIn[i_b, lq, k]
                else:
                    shared_mem[l_idx, k] = 0
            T.sync_threads()

            # Pre-allocated scalar variables (causes issue in 0.1.7.post1)
            idx = T.alloc_var(INT32)
            val_a = T.alloc_var(INT32)
            val_b = T.alloc_var(INT32)
            temp = T.alloc_var(INT32)

[...]

PR #1607, opened 2026-01-05
[BugFix] Fix some bugs in lowering ParallelOp and VectorizeLoop

SiriusNEO:
https://github.com/tile-ai/tilelang/issues/1549 exposes some bugs in current lowering ParallelOp and VectorizeLoop implementation, mainly two problems:
1. Some `IsLocal` conditions don't take `local.var` into account.
2. If indices of a BufferStore are invarient, we shouldn't vectorize this loop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved vectorization issue when allocating scalar variables within parallel operations
  * Improved handling of store operations during loop vectorization to prevent incorrect vector size calculations
  * Enhanced local buffer classification for more accurate tile-lowering decisions

* **Tests**
  * Added regression test for variable vectorization edge cases

### DIFF
--- a/tilelang/engine/phase.py
+++ b/tilelang/engine/phase.py
@@ -168,8 +168,6 @@ def LowerAndLegalize(mod: IRModule, target: Target) -> IRModule:
     mod = tilelang.transform.LayoutInference()(mod)
     # Visualize the layout
     LayoutVisual(mod)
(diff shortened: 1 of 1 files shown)


SiriusNEO:
@regression-perf

PR #1629, opened 2026-01-07
[Fix] Add z3 dependency

douyixuan:
## Summary by CodeRabbit

* **Documentation**
  * Updated installation instructions to include an additional system dependency requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Or install locally:
 ```bash
 # install required system dependencies
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@douyixuan likely z3 already in https://github.com/tile-ai/tilelang/blob/main/requirements.txt#L13

kurisu6912:
Tilelang uses pypi z3 package, no need to install it as a system dependency (and the system z3 is usually older and slower than the pypi)

Issue #1633, opened 2026-01-07
[Feature Request] Add CUDA graph backend support to autotuner


cscyuge:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

## Problem
The [AutoTuner](https://github.com/tile-ai/tilelang/blob/main/tilelang/autotuner/tuner.py#L451) currently uses the default `"event"` backend for benchmarking, which includes kernel launch overhead. For short-duration kernels, this overhead can significantly affect configuration selection. When kernels are executed with CUDA graphs (minimizing launch overhead), the selected configuration may not be optimal.

**Request:** Add support for CUDA graph backend in the profiler, allowing users to specify the profiling backend during auto-tuning.

## Details

We compared different profiling methods and found significant discrepancies. But we don't known which measurement method is most reliable.

Benchmarking a small GEMM kernel (M=1, N=1024, K=1024) result:
| Method | Latency |
|--------|---------|
| tilelang profiler `event` backend | 8.77us |
| tilelang profiler `cupti` backend | 4.64us |
| `triton.testing.do_bench` | 8.40us |
| `triton.testing.do_bench_cudagraph` | 3.65us |
| `ncu` (direct kernel) | 5.38us |
| `ncu` (CUDA graph) | 5.18us |

test script:
```python
import tilelang
import tilelang.language as T
import triton


@tilelang.jit()
def matmul(M, N, K, block_M, block_N, block_K, dtype=T.float16, accum_dtype=T.float32):
    @T.prim_func
    def gemm(
        A: T.Tensor((M, K), dtype),
        B: T.Tensor((K, N), dtype),
        C: T.Tensor((M, N), dtype),
    ):
        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128) as (bx, by):
            A_shared = T.alloc_shared((block_M, block_K), dtype)
            B_shared = T.alloc_shared((block_K, block_N), dtype)
            C_local = T.alloc_fragment((block_M, block_N), accum_dtype)

            T.clear(C_local)
            for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=4):
                T.copy(A[by * block_M, k * block_K], A_shared)
                T.copy(B[k * block_K, bx * block_N], B_shared)
                T.gemm(A_shared, B_shared, C_local)

[...]

PR #1658, opened 2026-01-12
[Feat] profiler support cudagraph backend

cscyuge:
this pr is related to https://github.com/tile-ai/tilelang/issues/1633, add cudagraph backend to profiler and allows AutoTuner to specify the profiling backend.

### 1. Profiler CUDA Graph Backend

- **`tilelang/profiler/bench.py`**:
  - Added `_bench_with_cudagraph()` function implementing CUDA Graph-based benchmarking
  - Updated `do_bench()` to support `"cudagraph"` backend option
  - Implementation follows `triton.testing.do_bench_cudagraph` pattern

- **`tilelang/profiler/__init__.py`**:
  - Extended `backend` parameter to include `"cudagraph"` option
  - Fixed default values for `n_warmup` and `n_repeat` (changed from 1 to 0)

### 2. AutoTuner Backend Selection

- **`tilelang/autotuner/param.py`**:
  - Added `backend` parameter to `ProfileArgs` class with type `Literal["event", "cupti", "cudagraph"]`

- **`tilelang/autotuner/tuner.py`**:
  - Added `backend` parameter to `set_profile_args()` method
  - Updated `_profile()` to use specified backend for both kernel and reference program benchmarking

### 3. Example Updates

- **`examples/gemm/example_gemm_autotune.py`**:
  - Added `profile_backend` parameter to `get_best_config()` and `main()`
  - Added `--profile_backend` CLI argument
  - Updated benchmark calls to use specified backend

## Usage

```python
# AutoTuner with CUDA Graph backend
autotuner.set_profile_args(backend="cudagraph", warmup=3, rep=20)

# Profiler with CUDA Graph backend
profiler.do_bench(backend="cudagraph")
```

## Bench Results

```
# python examples/gemm/example_gemm_autotune.py --m 1 --n 1024 --k 1024 --use_autotune --profile_backend cudagraph
{'block_M': 16, 'block_N': 32, 'block_K': 128, 'num_stages': 3, 'thread_num': 128, 'enable_rasteration': False}
TileLang latency: 0.0037138615734875202
Ref latency: 0.005500588566064835
TileLang TFlops: 0.5646823282189969
Ref TFlops: 0.38125956428337626

# python examples/gemm/example_gemm_autotune.py --m 1 --n 1024 --k 1024 --use_autotune --profile_backend cupti
{'block_M': 16, 'block_N': 32, 'block_K': 128, 'num_stages': 3, 'thread_num': 128, 'enable_rasteration': False}
TileLang latency: 0.0053819006309147
Ref latency: 0.006137294351630924
TileLang TFlops: 0.38966754383266483
Ref TFlops: 0.3417062763891556

# python examples/gemm/example_gemm_autotune.py --m 1 --n 1024 --k 1024 --use_autotune --profile_backend event
{'block_M': 32, 'block_N': 16, 'block_K': 128, 'num_stages': 3, 'thread_num': 128, 'enable_rasteration': False}
TileLang latency: 0.008737378753721714
Ref latency: 0.009419833309948444
TileLang TFlops: 0.24002072693789445
Ref TFlops: 0.22263154038884772
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable profiling backends: "event", "CUPTI", and "CUDA graph" for benchmarking
  * New --profile_backend command-line option to choose the profiler backend
  * CUDA graph-based benchmarking implemented for lower-overhead, repeated measurements

* **Improvements**
  * Autotuning now honors the selected profiling backend for more consistent tuning results

### DIFF
--- a/examples/gemm/example_gemm_autotune.py
+++ b/examples/gemm/example_gemm_autotune.py
@@ -76,9 +76,9 @@ def get_configs(M, N, K, with_roller=False, topk=20):
             config["enable_rasteration"] = hint.rasterization_plan is not NoRasterization
             configs.append(config)
     else:
-        block_M = [64, 128, 256] if M > 32 else [16, 32]
-        block_N = [64, 128, 256] if M > 32 else [16, 32]
-        block_K = [32, 64, 128]
+        block_M = [64, 128, 256]
+        block_N = [64, 128, 256]
+        block_K = [32, 64]
         num_stages = [0, 1, 2, 3]
         thread_num = [128, 256]
         enable_rasterization = [True, False]


tilelang/profiler/__init__.py
```diff
@@ -211,10 +211,10 @@ def do_bench(
         func: Callable | None = None,
         warmup: int = 25,
         rep: int = 100,
-        n_warmup: int = 1,
-        n_repeat: int = 1,
+        n_warmup: int = 0,
```
LeiWang1999, line 226: Why we need to change the default value from 1 to 0?

tilelang/profiler/bench.py
```diff
@@ -202,3 +204,56 @@ def _bench_with_cupti(
 
     kernel_time_us = (total_cuda_time - excluded_time) / n_repeat
     return kernel_time_us * 1e-3  # Convert microseconds to milliseconds
+
+
+def _bench_with_cudagraph(
```
LeiWang1999, line 209: just want better understand the advantages of using CUDA Graphs for benchmarking. While I understand it likely reduces CPU overhead and kernel launch times, could we not simply use the CUPTI profiler if we only need to measure the kernel execution time?

tilelang/profiler/__init__.py
cscyuge: For the "cudagraph" backend, n_repeat = 0 avoids measuring a single-iteration graph where launch overhead is still present.
We also align n_repeat / n_warmup with bench.py::do_bench so the default behavior is consistent.

tilelang/profiler/bench.py
cscyuge: You’re right that, functionally, CUPTI could be used to measure kernel execution time here. The main reason we prefer CUDA Graphs is that they better reflect inference-style execution, where the same kernels are executed repeatedly in a steady-state loop rather than as isolated launches.

This distinction can also matter for autotuning, since different measurement contexts may bias the tuning process differently. While we haven’t observed divergent optimal configurations in our current tests, using CUDA Graphs helps ensure that the benchmarking setup is aligned with the intended inference execution pattern.

LeiWang1999: I see. I think we can keep the cudagraph backend implementation if for this reason.

However, I haven't seen other projects use this method for benchmarking, so I do not recommend making it the default behavior in `examples/gemm/example_gemm_autotune.py`. cudagraph construction and replay introduce additional overhead and potential instability.

If we can revert the changes to the example file (to restore the default behavior), I can approve this PR.

LeiWang1999:
While valuable, this approach should be treated as an alternative metric rather than a default replacement for standard event-based benchmarking.

The benefits of this feature from my understanding: It accurately measures very short kernels where launch/scheduling overheads or Python/driver jitter might distort the GPU timeline. t measures performance under 'graph replay,' which typically yields more optimistic results than eager execution (due to reduced launch overhead and optimized execution flows)."

Issue #1678, opened 2026-01-16
[BUG] Multiple `T.alloc_var()` Declarations Cause Unexpected Behavior in TileLang


LJC00118:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7.post2+cuda.gitb27fb928

### System information

>>> print(sys.version, sys.platform)
3.12.9 | packaged by Anaconda, Inc. | (main, Feb  6 2025, 18:56:27) [GCC 11.2.0] linux
>>> print(tilelang.__version__)
0.1.7.post2+cuda.gitb27fb928
>>> print(torch.__version__)
2.7.0+cu128

### Problem description

When using multiple T.alloc_var() declarations in a TileLang kernel, the generated CUDA code incorrectly overwrites the first variable's declaration.

### Reproducible example code

The Python snippets:

```python
import torch
import tilelang
from tilelang import language as T


@tilelang.jit
def qwq():
    @T.prim_func
    def qwq_kernel():
        with T.Kernel(4096, 1, threads=1) as (pid_y, pid_x):
            i = T.alloc_var('int32')
            i = 1
            tmp_row = T.alloc_local((4, ), 'float32')
            amax_local = T.alloc_var('float32')
            j = 0
            amax_local = T.max(amax_local, tmp_row[j])

    return qwq_kernel


kernel = qwq()
print(kernel.get_kernel_source())
```

### Traceback

```pytb
Traceback (most recent call last):
  File "~/workspace/qwq/test7.py", line 21, in <module>
    kernel = qwq()
             ^^^^^
  File "~/tilelang/tilelang/jit/__init__.py", line 435, in __call__
    kernel = self.compile(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/tilelang/tilelang/jit/__init__.py", line 374, in compile
    kernel_result = compile(
                    ^^^^^^^^
  File "~/tilelang/tilelang/jit/__init__.py", line 98, in compile
    return cached(
           ^^^^^^^
  File "~/tilelang/tilelang/cache/__init__.py", line 74, in cached
    return _dispatch_map[execution_backend].cached(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1683, opened 2026-01-16
[Bugfix] Avoid reusing local.var allocations (#1678)

LudovicoYIN:
Summary

This PR fixes incorrect local variable reuse in StorageRewrite that caused
multiple T.alloc_var() declarations to be merged into a single local variable,
leading to unexpected overwrites in generated CUDA code.

Root Cause

StorageRewrite previously allowed reuse of local.var allocations.
However, T.alloc_var() is semantically closer to a distinct scalar declaration
rather than a reusable buffer.

When multiple alloc_var are present in the same scope, reuse breaks
declaration and initialization ordering, producing invalid code such as:

```cpp
int i = 0;
float tmp_row[4];
i = 1;
i = max(i, tmp_row[0]);  // overwrites previous alloc_var
```

Fix

* Disable reuse for local.var in StorageRewrite (storage_rewrite.cc)
* Ensure each T.alloc_var() results in a unique local variable
* Preserve correct declaration and initialization semantics

This change only affects local.var and does not impact buffer or shared memory reuse.

Tests

* test_tilelang_issue_1678.py

Related Issue

Fixes [#1678](https://github.com/tile-ai/tilelang/issues/1678) 

## Release Notes

* **Bug Fixes**
  * Improved allocation handling for specific scope scenarios.

* **Tests**
  * Added test coverage for allocation with mixed data type scenarios in CUDA environments.

### DIFF
--- a/src/transform/storage_rewrite.cc
+++ b/src/transform/storage_rewrite.cc
@@ -1094,6 +1094,10 @@ private:
       return NewAlloc(op, attach_scope, scope, const_nbits);
     }
 
+    if (scope.tag == ".var") {
+      return NewAlloc(op, attach_scope, scope, const_nbits);
+    }
+
     if (is_known_size) {
       // constant allocation.
       auto begin = const_free_map_.lower_bound(const_nbits / match_range);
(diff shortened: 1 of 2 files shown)


LeiWang1999:
Thanks for your contribution! We dug a bit deeper and decided to disable storage reuse to address the root cause of this issue. Therefore, we are closing this PR. We look forward to your future contributions!

LudovicoYIN:
Thanks for the explanation and review! Looking forward to contributing again.

PR #1693, opened 2026-01-19
[Enhancement] Log warnings for OOB acceses to non-global buffers

SiriusNEO:
For some thread programs (non-tile programs), OOB accesses to non-global buffers may also happen. For better developing these kernels, we add additional warnings if the compiler detects such cases.

## Summary by CodeRabbit

* **Chores**
  * Improved internal memory access safety validation mechanisms to enhance coverage and robustness.


SiriusNEO:
@regression-perf

PR #1701, opened 2026-01-20
[AMD] Fix bugs about AMD FA kernel

danielhua23:
As tilelang updates, I found some issues when I running AMD FA kernel, this pr is to propose the workaround for AMD FA example
- issue1: when saving tuned kernel, AMD FA will go through kernel_global_source of CythonKernelAdapter, but that attr is lost
- issue2: original tuned config like block_M = 32 and block_N = 32 does not work now, so remove it, specifically, the issues are 
```
Layout infer conflict between acc_s and acc_s_cast in T.Parallel loop:
    loop Fragment([32, 32] -> [8], replicate: 2, thread: 256, ...)
    fragment Fragment([32, 32] -> [4], replicate: 1, thread: 256, ...)
```
and 
```
Check failed: pb->value != 0 (0 vs. 0) : Divide by zero
```

are the two issues expected? thanks

## Summary by CodeRabbit

* **Chores**
  * Optimized AMD Flash Attention example: narrowed tuning options, improved kernel execution for correct softmax-related computation and reduced problematic configurations.

* **Refactor**
  * Added a compatibility alias on the kernel adapter to mirror existing kernel source access, improving integration with callers.

### DIFF
--- a/examples/amd/example_amd_flash_attn_fwd.py
+++ b/examples/amd/example_amd_flash_attn_fwd.py
@@ -16,12 +16,7 @@ def supply_tensors_gpu(params):
             # Force creation on GPU device
             shape = [int(s) for s in param.shape]
             # Convert TileLang dtype to PyTorch dtype
-            if hasattr(param.dtype, "as_torch"):
-                torch_dtype = param.dtype.as_torch()
-            else:
-                torch_dtype = param.dtype
-            if torch_dtype is None:
-                raise ValueError(f"Cannot convert dtype '{param.dtype}' to torch.dtype")
+            torch_dtype = param.dtype.as_torch()
             tensor = torch.randn(shape, dtype=torch_dtype, device="cuda")
             tensors.append(tensor)
         else:


examples/amd/example_amd_flash_attn_fwd.py
```diff
@@ -8,15 +8,24 @@
...
+    }
+
     tensors = []
     for param in params:
         if hasattr(param, "shape") and hasattr(param, "dtype"):
             # Force creation on GPU device
             shape = [int(s) for s in param.shape]
-            tensor = torch.randn(shape, dtype=param.dtype, device="cuda")
+            # Convert TileLang dtype to PyTorch dtype
+            torch_dtype = dtype_map.get(param.dtype, torch.float16)
```
LeiWang1999: `parm.dtype.as_torch()`

examples/amd/example_amd_flash_attn_fwd.py
```diff
@@ -124,7 +133,7 @@ def main(
             bx = T.alloc_var(T.int32)
             bx = b_split
 
-            while bx < num_q_blocks:
+            with T.While(bx < num_q_blocks):
```
LeiWang1999: why we need to change `while` into `T.While`?

PR #1708, opened 2026-01-21
[BugFix] Corrected when proving loop layout contains a fragment buffer layout

LeiWang1999:
Only prove loop contain loop we need index checking.

## Summary by CodeRabbit

* **Refactor**
  * Updated validation logic to use default forward-index checking behavior during candidate validation and selection, resulting in improved control flow handling.


LeiWang1999:
@regression-perf

PR #1713, opened 2026-01-22
[Layout] Fix Layout Bugs in Parallel and Reduce

kurisu6912:
1. Add read write checking in `T.Parallel`: When a loop read a buffer, the buffer layout contains the loop layout (every read can hit the buffer); and when the loop write a buffer, loop layout contains the buffer layout (every place in the buffer is written)
2. Refactor layout inference of `T.reduce_xx`: First infer a reducer layout, then use the reducer layout to do reduce, and finally copy the reducer to target buffer.

## Summary by CodeRabbit

* **Refactor**
  * Access tracking centralized to record per-buffer indices and read/write intent, simplifying parallel access iteration and planning.

* **Behavior Change**
  * Layout inference and conflict reporting now use reducer-aware checks and per-buffer access metadata; guarded/conditional writes added to avoid unsafe writes during reductions.

* **Tests**
  * Added regression tests exercising multiple layout and reduction scenarios to prevent regressions.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_1719.py
+++ b/testing/python/issue/test_tilelang_issue_1719.py
@@ -43,10 +43,10 @@ def test_issue_1719_layout_2():
 @tilelang.testing.requires_cuda
 def test_issue_1719_layout_3():
     @tilelang.jit
-    def _buggy_kernel(A, B, dtype=T.float32):
+    def _buggy_kernel(A, dtype=T.float32):
         M, N = T.const("M, N")
         A: T.Tensor[(M, N), dtype]
-        B: T.Tensor[(M,), dtype]
+        B = T.empty((M,), dtype)
         with T.Kernel(1, threads=32) as _:
             A_local = T.alloc_fragment((M, N), dtype)
             B_local = T.alloc_fragment((M,), dtype)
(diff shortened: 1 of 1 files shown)


kurisu6912:
@regression-perf

kurisu6912:
@regression-perf

src/op/reduce.cc
```diff
@@ -204,6 +240,10 @@ Stmt ReduceOpNode::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
     size_t src_dim = src_layout->InputDim();
     size_t dst_dim = dst_layout->InputDim();
 
+    auto red_layout = ComputeReducerLayout(src_layout, dim);
+    auto dst_rep = *as_const_int(dst_layout->ReplicateExtent());
```
LeiWang1999: this implementation looks better in my opinion
```
auto dst_rep = dst_layout->ReplicateExtent();
auto red_rep = red_layout->ReplicateExtent();

and analyzer->CanProve(red_rep > dst_rep);
```

testing/python/issue/test_tilelang_issue_1719.py
```diff
@@ -0,0 +1,155 @@
+import tilelang
+import torch
+import tilelang.testing
+import tilelang.language as T
+
+
+@tilelang.testing.requires_cuda
```
LeiWang1999, line 7: eager style jit looks better.

src/op/reduce.cc
kurisu6912: We should check `!analyzer->CanProve(red_rep >= dst_rep)`, and since red contains dst, we can only check `!analyzer->CanProve(red_rep == dst_rep)`

PR #1715, opened 2026-01-22
[Refactor] Refactor cross vars handling logic to a copy-remap strategy in SplitHostDevice

SiriusNEO:
Prior to this PR, after SplitHostDevice, TIR vars which are "defined in host-side and used in device-side" (we call them "cross vars") are directly appended to the parameter list of new generated device function (e.g. main_kernel). This causes the device function body will not uses vars in parameter list, but uses a new renamed var. The root cause is complex. But I believe an important reason is that the old logic will let two "identical TIR vars" present in two different PrimFuncs, which is a dangerous case and may cause strange use-def analysis bug.

This PR refactors the logic of cross vars part, first copying them and doing remap. This makes the function form more reasonable and solve the problem.

## Release Notes

* **Refactor**
  * Enhanced host-device code splitting logic with improved cross-variable identification and buffer management on the device side.

* **Tests**
  * Added comprehensive tests for host-device transformation with user-supplied constraints.

### DIFF
--- a/src/transform/split_host_device.cc
+++ b/src/transform/split_host_device.cc
@@ -110,35 +110,61 @@ private:
     return body;
   }
 
+  std::tuple<Array<tir::Var>, Array<tir::Var>, Array<tir::Buffer>,
+             Map<tir::Var, tir::Var>>
+  collectCrossVars(tir::Stmt body) {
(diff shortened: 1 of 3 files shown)


LeiWang1999:
closed as pr #1723  introduced another solution.

Issue #1712, opened 2026-01-22
[Question] Add language specs


Da1sypetals:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Questions

As a python DSL, it would be benificial if some specs can be added.

Without some language specs, I can only mimic syntax found in examples. I don't even know I can slice on buffers before I checked the test suite. 

I don't know if it is for me or generally someone have the same question. I am just too afraid if I used a feature supported in python but not in tilelang, the compiler spits some 1000 lines of internal error.

Da1sypetals:
<img width="2176" height="1354" alt="Image" src="https://github.com/user-attachments/assets/9f1f309a-536e-4955-8bb0-9ec50a63abe6" />

Some picture to illustrate my mindset

LeiWang1999:
Thanks for your report @Da1sypetals , think we can have a spec list based on our parser.

LeiWang1999:
@Da1sypetals Thanks and I made a related pr #1745 , would you mind take a look and help review?

SiriusNEO:
I believe this is a challenging yet critical issue for any Python-embedded DSL that requires long-term maintenance and community collaboration. This is also a key goal in TileLang's development roadmap.

I believe we need to strengthen our documentation in the following three areas:

- D1 (DSL Spec): A reference for available **TileLang APIs** (T.xx) and their constraints.
- D2 (Syntactic Support): A guide on valid **Python syntax** within TileLang. As a Python-embedded DSL, TileLang supports many native Python constructs, but not all.
- D3 (Best Practices): A guide on how to write **efficient** TileLang kernels (i.e., the programming philosophy behind TileLang).

As TileLang is still an active project, we are constantly handling corner cases and applying fixes. Consequently, it is currently hard to completely avoid situations where the compiler "spits out 1000 lines of internal errors." We plan to address this in three ways:
- User-friendly warnings / errors report. We have some Python-side checks which can report many friendly illegal cases and avoid entering TVM stacks and "spits out 1000 lines of internal errors". (But this is also not very trivial. We still do this case by case, according to community feedback).
- Enhance documentation: Including detailed language specifications and tutorials.
[...]

PR #1745, opened 2026-01-28
[Docs] Add Python Compatibility document of TileLang

LeiWang1999:
## Summary

- Add a comprehensive TileLang Semantics guide to help users understand the DSL syntax
- Documents Python compatibility: what works and what doesn't in `@T.prim_func` kernels
- Covers loops, conditionals, buffer operations, variables, functions, and operators
- Includes detailed section on integer division/modulo semantics (floordiv vs truncdiv)
- Highlights that TileLang follows Python semantics for `//` and `%`, unlike Triton which uses C-style truncation

## Test plan

- [x] Documentation renders correctly in markdown
- [x] All code examples are syntactically correct
- [x] Links to other documentation pages are valid

## Summary by CodeRabbit

* **Documentation**
  * Added Python compatibility guide for TileLang, documenting supported Python features in kernel scripts with status indicators (✅, ❌, ⚠️).
  * Updated documentation index to include the new Python compatibility section.

* **Chores**
  * Updated submodule reference to latest version.


docs/programming_guides/tilelang_semantics.md
```diff
@@ -0,0 +1,428 @@
...
+# Negative indexing
+last = A[-1]
+A[-1] = 0
+
+# Shape access
+M = A.shape[0]
+N = A.shape[1]
+
+# Type casting
+x = value.astype("float32")
```
Elevator14B: So what's the canonical way to cast types? `Buffer.astype` mentioned here, or `T.cast` from TVM, or `T.Cast` defined in TileLang Python package?

docs/programming_guides/tilelang_semantics.md
```diff
@@ -0,0 +1,428 @@
...
+### Early Exit
+
+```python
+# break/continue work as expected:
+for i in range(N):          # or T.serial(N)
+    if found(i):
+        break
+```
+
+### Max/Min Finding
```
Elevator14B: Maybe together with sum? Also, I think it's good to mention difference between var-based serial reduction and fragment-based parallel reduction, and `alloc_reducer` as another option for the latter.

Elevator14B:
For the "What (not) works" section, I think we should include more explanation for what's happening behind the interfaces and what's the constraints. E.g., fragments should be accessed with expressions consisting of `T.Parallel` indices (but with a few exceptions as syntax sugar) so that it can be mapped to some thread during layout inference, `T.reduce_max/min/sum` do parallel reduction inside a block, etc.

docs/programming_guides/tilelang_semantics.md
SiriusNEO: @Elevator14B They are the same, just syntax sugar. Both `T.cast` and `T.Cast` are from TVM. And the difference between `T.cast` and `T.Cast` is just the order of arguments. I think the canonical way is to use `T.cast` or `astype`.

Da1sypetals: What is the behavior of `Buffer.astype`? Return a new buffer, or casting the elements in-place? If it is the latter one, should the buffer size change if the bitwidth of src/dst types are different?

Instead of immutable tiles model in triton/cuTile, we allocate before writing to a buffer in Tilelang, which (for me) in many cases causes this kind of confusion.

SiriusNEO: @Da1sypetals It returns a new pointer to the original data

SiriusNEO: @Elevator14B I submit a PR which unifies the cast-related ops: https://github.com/tile-ai/tilelang/pull/1757. Any suggestions?

Da1sypetals: > @Da1sypetals It returns a new pointer to the original data

So is it correct that it modifies the content of original buffer? Also, if the bitwidth are different (and thus space required to hold the same amount of element change), how can it point to the original buffer? What is the internal behavior?

SiriusNEO: @Da1sypetals 1. This instruction itself don't modifies the content of original buffer since it just returns the pointer. But we can use this pointer to read/write the data, which may modify the content. 2. The internal behavior is just it do a static cast, like:

```
buf1 = T.cast(buf, T.float16)
# Codegen result
float16* buf1 = static_cast<float16*>(buf);
```

You can write a simple TileLang program and use `get_kernel_source()` to get the `.cu` code to have a better understanding.

Da1sypetals:
> For the "What (not) works" section, I think we should include more explanation for what's happening behind the interfaces and what's the constraints. E.g., fragments should be accessed with expressions consisting of `T.Parallel` indices (but with a few exceptions as syntax sugar) so that it can be mapped to some thread during layout inference, `T.reduce_max/min/sum` do parallel reduction inside a block, etc.

@Elevator14B I have different opinions with you, IMO documenting every corner cases is way more important than explaining compiler/hardware internals and restrictions to ***users*** who most likely don't care about these. Maybe appendix is a good place for these info.

docs/programming_guides/tilelang_semantics.md
Da1sypetals: > @Da1sypetals 1. This instruction itself don't modifies the content of original buffer since it just returns the pointer. But we can use this pointer to read/write the data, which may modify the content. 2. The internal behavior is just it do a static cast, like:
> 
> ```
> buf1 = T.cast(buf, T.float16)
> # Codegen result
> float16* buf1 = static_cast<float16*>(buf);
> ```
> 
> You can write a simple TileLang program and use `get_kernel_source()` to get the `.cu` code to have a better understanding.

So it is not casting by value like `tensor.to(dtype)` in PyTorch do, it is just casting the pointer? 

In this case, if a buffer is casted to a dtype with different bitwidth(e.g. float32 -> float16, which in the same space holds 2x elements), how will the buffer's shape change? I think this should be documented.

Elevator14B:
Another missing piece: what's allowed in the `@T.prim_func`ed function signature? This includes `T.Tensor` and equivalently `T.Buffer`, their two flavours of marking shape and dtype (`[shape, dtype]` and `(shape, dtype)`); `T.ptr` (together with its usage); scalars like `T.float32`, etc.

Elevator14B:
> > For the "What (not) works" section, I think we should include more explanation for what's happening behind the interfaces and what's the constraints. E.g., fragments should be accessed with expressions consisting of `T.Parallel` indices (but with a few exceptions as syntax sugar) so that it can be mapped to some thread during layout inference, `T.reduce_max/min/sum` do parallel reduction inside a block, etc.
> 
> @Elevator14B I have different opinions with you, IMO documenting every corner cases is way more important than explaining compiler/hardware internals and restrictions to _**users**_ who most likely don't care about these. Maybe appendix is a good place for these info.

I get your point. Maybe we need some compiler/hardware agnostic way to explain the semantics (not just the allowed lexical subset), so that users don't get confused with the internal details while still know what their written code is expected to behave.

Da1sypetals:
Is it required (in some cases) that the thrid argument of `T.gemm`, the accumulator, must be on fragment? If so, it should be documented. Sometimes I encounter this error:

```
  File "/opt/conda/lib/python3.11/site-packages/tilelang/tileop/gemm/__init__.py", line 23, in gemm_py_infer_layout
    return gemm_py.infer_layout(target, thread_nums)

  File "/opt/conda/lib/python3.11/site-packages/tilelang/tileop/gemm/__init__.py", line 142, in infer_layout
    return impl_class(self).infer_layout(target, thread_nums)

  File "/opt/conda/lib/python3.11/site-packages/tilelang/tileop/gemm/gemm_wgmma.py", line 42, in infer_layout
    self.C: mma_emitter.make_mma_store_layout(self.C),

  File "/opt/conda/lib/python3.11/site-packages/tilelang/intrinsics/wgmma_macro_generator.py", line 581, in make_mma_store_layout
    assert is_fragment(local_buf), "local_buf must be a fragment"

AssertionError: local_buf must be a fragment
```

SiriusNEO:
@LeiWang1999 I think the goal (A comprehensive "language spec" of TileLang) is not easy to achieve in this PR and should be a long-term goal. So for now I preserve the Python compatibility part in this PR and merge it as a temporary document enhancement. In the future we need a deep refactor of the "Programming Guides" section of our document to address all feedbacks in this thread as well as https://github.com/tile-ai/tilelang/issues/1712.

PR #1756, opened 2026-01-29
[BugFix] Always build guard in loop partitioning to prevent out-of-bounds access

LeiWang1999:
## Summary
- Fixed a bug where loop partitioning could generate out-of-bounds memory access when loop extent differs from Fragment layout input shape
- Simplified guard construction by always building guard conditions and letting the analyzer decide if they can be proved true
- This handles both non-bijective layouts and cases like loop extent=4 with Fragment([8]->[1]) where inverse index `tx % 8` ranges 0-7 but buffer only has 4 elements

## Test plan
- [x] Tested with issue_1744.py reproduction case - no more illegal memory access error
- [x] Generated CUDA code now correctly includes boundary check `if ((threadIdx.x & 7) < 4)`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced guard logic for transformation loops to ensure correctness in non-bijective and shape-different cases.

* **Tests**
  * Added a CUDA-enabled test validating the kernel behavior and preventing regressions for the addressed issue.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_1744.py
+++ b/testing/python/issue/test_tilelang_issue_1744.py
@@ -1,0 +1,37 @@
+import torch
+import tilelang
+import tilelang.testing
+import tilelang.language as T
+
+
+@tilelang.jit(
+    pass_configs={
+        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@perf-regression

LeiWang1999:
@regression-perf

PR #1757, opened 2026-01-29
[Refactor] Unify the usage of cast-related operators

SiriusNEO:
TVM IR has the following type cast related operators:

```python
buffer.astype(dtype)
T.cast(value, dtype)
T.Cast(dtype, value)

# And the reinterpret cast operator
T.reinterpret(dtype, value)
```

We can see that APIs are not very consistent and may cause some confusions. This PR unifies them to:

```python
buffer.astype(dtype)
T.cast(value, dtype)
# We don't recommend use T.Cast now. And all examples are refactored to use T.cast
T.reinterpret(value, dtype) # Reorder to consistent with T.cast. Always value first, dtype second
```

## Summary by CodeRabbit

* **Refactor**
  * Standardized cast/reinterpret APIs: unified argument order and introduced new cast/reinterpret call forms; compatibility handling added for the old form.
* **New Features**
  * Exposed several new public operation aliases in the language bindings for GPU/low-level ops.
* **Tests / Examples**
  * Updated tests and examples to adopt the new cast/reinterpret APIs and validate compatibility.

### DIFF
--- a/tilelang/language/tir/ir.py
+++ b/tilelang/language/tir/ir.py
@@ -276,7 +276,9 @@ anylist_setitem_call_packed = _op_wrapper(_tir_op.anylist_setitem_call_packed)
 anylist_setitem_call_cpacked = _op_wrapper(_tir_op.anylist_setitem_call_cpacked)
 vscale = _op_wrapper(_tir_op.vscale)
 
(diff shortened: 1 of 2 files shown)


Elevator14B:
Maybe detect which one is dtype in `T.reinterpret`? The current modification would be a breaking change.

PR #1758, opened 2026-01-29
Add an example: mHC residual projection backward

Da1sypetals:
I’d like to add an example for the mHC backward pass, but I’m running into some bugs that I haven’t been able to resolve on my own. I would really appreciate it if someone could take a look.

The algorithm is described [here in my blog](https://da1sypetals.netlify.app/posts/sinkhorn-bwd/). More importantly, there is a Triton implementation available [here](https://gist.github.com/Da1sypetals/e5cc3a9ec27f1d569e0d47adc609f023).

I attempted to translate the Triton code into TileLang more or less word-for-word, but I couldn’t get correct results. The current implementation always produces `NaN`s. Interestingly, if I add a `print` inside the `T.serial` loop, the results become correct (or at least very close).

I’d really appreciate it if someone could review the code and point out whether there is a bug in my implementation or if something else is going wrong 🌹

## Summary by CodeRabbit

* **Documentation**
  * Added new example demonstrating backward pass computation for Sinkhorn operations with autotuning and performance optimization.

* **Chores**
  * Updated spelling wordlist.

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -1,4 +1,5 @@
 cancelled
+dout
 hsa
 ist
 LOD
--- a/examples/deepseek_mhc/example_mhc_bwd.py
+++ b/examples/deepseek_mhc/example_mhc_bwd.py
@@ -1,6 +1,5 @@
 # NOTE: This bwd script is not an official upstream script; it is community-written and provided for reference only.
 # checkout pr: https://github.com/tile-ai/tilelang/pull/1758
-
 import torch
 
 import tilelang
@@ -175,9 +174,7 @@ def sinkhorn_bwd_implicit_cg(n_stream: int, tilesize: int = 32, threads: int = 1
             # Conjugate gradient: iteration ends
 
             for i_tile, i_nx, i_ny in T.Parallel(tilesize, n_stream, n_stream):
-                res_tile[i_tile, i_nx, i_ny] = (
-                    dR[i_tile, i_nx, i_ny] - x1[i_tile, i_nx] - x2[i_tile, i_ny]
-                ) * R[i_tile, i_nx, i_ny]
+                res_tile[i_tile, i_nx, i_ny] = (dR[i_tile, i_nx, i_ny] - x1[i_tile, i_nx] - x2[i_tile, i_ny]) * R[i_tile, i_nx, i_ny]
 
             T.copy(res_tile, res[i_seq * tilesize : (i_seq + 1) * tilesize, :, :])
 


LeiWang1999:
I’ll help and take look, one helpful trick for debugging is replace `T.alloc_shared([tilesize, n_stream, n_stream], dtype=dtype)` with `T.alloc_shared([tilesize, n_stream, n_stream], dtype=dtype, scope="shared")` to use static shared memory for more readable codegen

Da1sypetals:
> Use `T.sync_threads()` after reductions to prevent read-after-write races.

It seems like LLM's suggestion is correct, I got correct results after inserting `T.sync_threads`. 

Probably triton inserted implicit synchonization in its reduce operations `triton.language.sum`, and I was not aware of the need to insert synchronizations in tile-level programming. 

Is this a design choice to leave this to user? Or did I misunderstood something? @LeiWang1999

LeiWang1999:
@Da1sypetals Yes and I also found the `sync_threads` issue when retrieving the generated cuda code via `print(kernel.get_kernel_source())`.

```
// 1. Perform reduction sum to calculate pAp_frag[0]
pAp_frag[0] = tl::AllReduce<tl::SumOp, 4, 1, 0, 128>::run_hopper(pAp_frag[0]);

// 2. Write reduction results to Shared Memory
// Note: Only a subset of threads (tid % 4 == 0) are responsible for writing
if ((((int)threadIdx.x) % 4) == 0) { 
  ((float*)buf_dyn_shmem)[((((int)threadIdx.x) >> 2) + 20640)] = pAp_frag[0]; 
} 

// === CRITICAL FLAW: Missing __syncthreads(); here ===
// There is no guarantee that the write operation is complete 
// or visible to all Warps at this point.

// 3. All threads immediately read from Shared Memory to calculate alpha
// This read operation is highly likely to retrieve:
// a) Stale data from the previous iteration
// b) Garbage data (incomplete writes), leading to division by zero or NaNs
for (int i_28 = 0; i_28 < 4; ++i_28) { 
  float alpha = (((float*)buf_dyn_shmem)[...] / (((float*)buf_dyn_shmem)[... + 20640] + epsilon)); 
  // ... Update x1, r1, etc.
}
```


This is indeed a bug. The previous algorithm invoked `__syncthreads()` inside the if block, which caused program hangs. We refactored the code in PR #1631 to remove that synchronization to fix the deadlock. However, removing it entirely was incorrect for this case. We need to improve the pass to inject `__syncthreads()` outside the if block instead."

```
if ((((int)threadIdx.x) % 4) == 0) { 
  ((float*)buf_dyn_shmem)[((((int)threadIdx.x) >> 2) + 20640)] = pAp_frag[0]; 
  // previously, syncthreads will lead to a hang.
} 
```


Two solution:

1. use fragment as reduce's output instead of shared
```python
r_new_normsq = T.alloc_shared([tilesize], dtype=dtype)
->
r_new_normsq = T.alloc_fragment([tilesize], dtype=dtype)


pAp = T.alloc_shared([tilesize], dtype=dtype)
->
pAp = T.alloc_fragment([tilesize], dtype=dtype)
```

2. direct invoke `T.sync_thread()` after `dot` .


We will have a fix asap, and sorry for the trouble.

LeiWang1999:
I made a fix at #1760 , thanks for pointing out the issue!

Da1sypetals:
@LeiWang1999 Thanks! Is it correct that after the fix, it will no longer be required to manually call `T.sync_threads()` in the kernel? Shall I remove those then?

LeiWang1999:
@Da1sypetals Yes, I double-checked, and we no longer need to inject `T.sync_threads` manually.

Da1sypetals:
@LeiWang1999  Great! I have removed the synchronizations in mHC res-proj backward kernel. So long as the bug is fixed, can these changes be merged now, or do we need to wait until the next release?

LeiWang1999:
By caching some buffers into fragments instead of shared memory, we achieved a 2x speedup than the original implementation. However, I can't reproduce the triton results, possibly because I used the wrong triton version. Thanks for your contribution I think we can let this pr in.

PR #1760, opened 2026-01-30
[ThreadSync] Use Z3 for constraint equivalence checking

LeiWang1999:
This PR replaces the `const_int_bound` based range comparison with Z3 SMT solver for checking thread constraint equivalence in `FindConflict`.

The previous implementation used `const_int_bound` to compare thread variable ranges, which only checks the min/max bounds without considering additional constraints (e.g., `threadIdx_x % 4 == 0`). This led to incorrect equivalence judgments when constraints narrowed the actual executing thread set.

## Approach

We use Z3 to formally verify constraint set equivalence through bidirectional implication:

Let $P(t)$ denote the predicate for the previous access's constraint set and $C(t)$ denote the predicate for the current access's constraint set, where $t = (\texttt{threadIdx.x}, \texttt{threadIdx.y}, \texttt{threadIdx.z})$.

We check:
1. $P(t) \Rightarrow C(t)$: Every thread executing `prev` also executes `curr`
2. $C(t) \Rightarrow P(t)$: Every thread executing `curr` also executes `prev`

If both hold, then $P(t) \Leftrightarrow C(t)$, meaning the exact same set of threads execute both accesses. Combined with `has_same_index` (identical buffer index expressions), this guarantees each thread only accesses memory locations it wrote itself, eliminating cross-thread RAW/WAR conflicts.

### Implementation

The implication $P \Rightarrow C$ is checked by proving $\neg P \vee C$ using the Z3 prover:

```cpp
bool prev_implies_curr = analyzer.z3_prover.CanProve(tir::Or(tir::Not(prev_constr), curr_constr));
bool curr_implies_prev = analyzer.z3_prover.CanProve(tir::Or(tir::Not(curr_constr), prev_constr));
```

## Changes

- Added `ToConjunction()` method to `ConstrSet` for converting constraint sets to a single conjunction expression
- Replaced `const_int_bound` range comparison with Z3-based equivalence checking in `FindConflict`

## Summary by CodeRabbit

* **Refactor**
  * Improved constraint comparison and storage-synchronization logic with a stronger equivalence check to reduce false conflicts.

* **API**
  * Added a method to obtain a combined constraint expression from a constraint set.

* **Tests**
  * Added a CUDA-targeted test for a modulo-index shared-memory pattern and tightened validation for synchronization placement in generated kernels.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_1106.py
+++ b/testing/python/issue/test_tilelang_issue_1106.py
@@ -32,13 +32,12 @@ def test_issue_1106():
     m = 200
     kernel = get_kernel(m)
     source = kernel.get_kernel_source()
-    print(source)
     # Ensure __syncthreads is not inside the for loop
     for_start = source.find("for (int i = 0;")
     for_end = source.find("__syncthreads")
     assert for_end > for_start, "__syncthreads should be after the for loop, not inside it"
     # Check that __syncthreads appears after the closing brace of the outer for loop
-    assert source[for_end - 4:for_end - 2] == "}\n", "__syncthreads should not be inside any for loop"
+    assert source[for_end - 4 : for_end - 2] == "}\n", "__syncthreads should not be inside any for loop"
 
 
 if __name__ == "__main__":


LeiWang1999:
This is also a fix for #1758

Issue #1761, opened 2026-01-30
[Feature Request] Better performance for T.reduce


bucket-xv:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

Currently, the implementation for T.reduce is far from satisfactory. Compared to CCCL implementation, there are the following limitations. Let me explain further in the `solution` section.

### Solution

Let's see how CCCL solves this:
1. For threads ==32, use `redux.sync` instruction. A specialized instruction for sm>=80 arch.
2. For threads > 32:
    a. Do per warp reduction first. using  `redux.sync`.
    b. Each leader thread per warp writes to temporary shared mem. Then do sync.
    c. Read all (at most 31, do unroll to jump itself) from shared mem

The current solution suffers from:
1.  Low speed, as too much block level sync, the above solution only sync once at most. Also fail to use the latest inst.
2.  More space. Should first do warp level reduction to save space.
3.  High constraint. The above solution can handle any number of threads as a multiple of 32. But the current impl can only handle the power of 2.

### Alternatives

Use cub/block/block_reduce.cuh reduction directly, but need to construct a Temp storage. You may also learn from PTX of this.
```C++
#include <cub/block/block_reduce.cuh>
#include <cuda/atomic>
#include <cuda/cmath>
#include <cstdio>

template <int block_size>
__global__ void reduce(int* data, int* global_sum) {
  using BlockReduce = cub::BlockReduce<int, block_size>;
  __shared__ typename BlockReduce::TempStorage temp_storage;

  int const index = threadIdx.x + blockIdx.x * blockDim.x;
  int sum = data[index];
  sum = BlockReduce(temp_storage).Sum(sum);
  global_sum[blockIdx.x] = sum;
}


int main() {

    int N = 1024;
    int* d_data;
    int* d_result;
    cudaMalloc(&d_data, N * sizeof(int));
    cudaMalloc(&d_result, N * sizeof(int));
    reduce<1024><<<1, 1024>>>(d_data, d_result);
    return 0;
}
```

tzj-fxz:
Thank you for the suggestions! PR https://github.com/tile-ai/tilelang/pull/1762 add these features. Welcome to review and test.

PR #1762, opened 2026-01-31
[Feature] Hierarchical reduction and warp reduction intrinsics support

tzj-fxz:
For https://github.com/tile-ai/tilelang/issues/1761
* Add hierarchical reduction from warp to block to reduce workspace size
* Add `redux.sync` PTX templates to support faster reduction on (u)int32 with `__CUDA_ARCH__>=800`

## Summary by CodeRabbit

* **Performance**
  * Faster and more memory-efficient reductions on recent GPUs via hierarchical and architecture-aware reduction paths.

* **Bug Fixes**
  * Improved handling and validation of integer reduction results to ensure correct casting and accurate outcomes.

* **Tests**
  * Expanded test coverage exercising many reduction ops, dtypes and thread configurations with stronger reference comparisons.

### DIFF
--- a/src/op/reduce.cc
+++ b/src/op/reduce.cc
@@ -356,11 +356,9 @@ Stmt ReduceOpNode::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
         // space.  When the thread-to-data mapping for the reduce dimension is
         // normalized as  threadIdx = source * scale + ...,
         //   * scale == 1  means threads are contiguous (0, 1, 2, ...),
-        //     which enables the optimized hierarchical warp-reduce path
-        //     (workspace = reducing_threads / 32).
         //   * scale  > 1  means threads are interleaved (0, scale, 2*scale,
-        //     ...), requiring the general recursive XOR-butterfly reduce
-        //     (workspace = total thread count).
+        //     ...).
+        // Both cases use the recursive XOR-butterfly reduce.
         // `extent` is the number of distinct thread positions along the reduce
         // dimension, so reducing_threads = extent * scale covers the full
         // thread index range that participates in the reduction.
@@ -386,23 +384,11 @@ Stmt ReduceOpNode::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
         }
         Array<PrimExpr> thread_reduce_args = {
             StringImm(ss.str()), BufferLoad(clear_buffer, red_indices)};
-        // The hierarchical path (contiguous threads, warp-aligned) only
-        // needs one shared-memory slot per warp; the butterfly path needs
-        // one slot per thread in the block.
-        bool is_hierarchical = [&]() {
-          if (reducing_threads <= 32)
-            return false;
-          if (reducing_threads % 32 != 0)
-            return false;
-          if (*scale != 1)
-            return false;
-          return true;
-        }();
+        // The butterfly reduce path needs one shared-memory slot per
+        // thread in the block.
         if (reducing_threads > 32) {
           int workspace_size =
-              is_hierarchical
-                  ? reducing_threads / 32
-                  : static_cast<int>(*as_const_int(T.thread_bounds->extent));
+              static_cast<int>(*as_const_int(T.thread_bounds->extent));
           PrimExpr workspace =
               T.AddWorkspace(workspace_size, clear_buffer->dtype);
           thread_reduce_args.push_back(workspace);
--- a/src/tl_templates/cuda/reduce.h
+++ b/src/tl_templates/cuda/reduce.h
@@ -77,28 +77,6 @@ template <int all_threads> struct NamedBarrier {
   }
 };
 
-// Performs the cross-warp reduction within the first warp of each group.
-// After all warps have written their partial results into red_buf, this
-// function loads them, reduces via warp shuffles, and writes the final
-// result back.
-template <class Reducer, int num_warps, typename T>
-TL_DEVICE void warp_inter_reduce(T *red_buf, int group_base, int lane_id) {
-  T val = (lane_id < num_warps) ? red_buf[group_base + lane_id]
-                                : red_buf[group_base];
-  for (int offset = 16; offset > 0; offset >>= 1) {
-    T y = tl::shfl_xor_sync(uint32_t(-1), val, offset);
-    if ((lane_id ^ offset) < num_warps) {
-      if (lane_id < num_warps) {
-        val = Reducer()(val, y);
-      } else {
-        val = y;
-      }
-    }
-  }
-  if (lane_id == 0)
-    red_buf[group_base] = val;
-}
-
 // AllReduce performs a cross-thread reduction over a group of `threads`
 // threads.
 //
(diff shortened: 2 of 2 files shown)


tzj-fxz:
Maybe we also need the performance regression tests. Where can I trigger them? @LeiWang1999

Rachmanino:
@regression-perf

bucket-xv:
Thanks for your contribution! I've further investigated this problem this weekend and have some suggestions for the code:
1. Use intrinsic functions instead of raw `redux` PTX for maintainability. Refer to https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-reduce-functions. Note that this is also for sm80+.
2. May leverage `redux` instruction for f32 types? This is supported since sm100a. Refer to https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-redux-sync.
3. At warp level reduction, maybe any floating types can be cast to f32 and integral types cast to int32. This cast helps leverage the  `redux` inst. This is almost always better in performance since there is also an implicit cast for `shfl.sync`, which requires b32 types.(https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-shfl-sync)

LeiWang1999:
@tzj-fxz Would you mind take a look.

tzj-fxz:
> Thanks for your contribution! I've further investigated this problem this weekend and have some suggestions for the code:
> 
> 1. Use intrinsic functions instead of raw `redux` PTX for maintainability. Refer to https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-reduce-functions. Note that this is also for sm80+.
> 2. May leverage `redux` instruction for f32 types? This is supported since sm100a. Refer to https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-redux-sync.
> 3. At warp level reduction, maybe any floating types can be cast to f32 and integral types cast to int32. This cast helps leverage the  `redux` inst. This is almost always better in performance since there is also an implicit cast for `shfl.sync`, which requires b32 types.(https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-shfl-sync)

Thank you for the advice. I will further implement these features.

src/op/reduce.cc
```diff
@@ -376,8 +376,12 @@ Stmt ReduceOpNode::Lower(const LowerArgs &T, arith::Analyzer *analyzer) const {
         Array<PrimExpr> thread_reduce_args = {
             StringImm(ss.str()), BufferLoad(clear_buffer, red_indices)};
         if (reducing_threads > 32) {
-          PrimExpr workspace = T.AddWorkspace(
-              *as_const_int(T.thread_bounds->extent), clear_buffer->dtype);
+          int workspace_size = (reducing_threads > 32 &&
```
LeiWang1999: It would be better to refactor this into a lambda function; that way, we can have cleaner code and more useful comments

src/tl_templates/cuda/reduce.h
```diff
@@ -60,45 +63,126 @@ struct BitXorOp {
 template <class Reducer, int threads, int scale, int thread_offset = 0,
           int all_threads = threads>
 struct AllReduce {
-  static_assert(threads == 1024 or threads == 512 or threads == 256 or
```
LeiWang1999, line 63: why do we need to remove this piece of code?

tzj-fxz: It is more flexible to support reduction across any warps (not constrained to powers of 2) by doing warp-level reduction after inner-warp reduction.

bucket-xv:
As far as I know, `__reduce_max_sync` does not accept float types. May you use assembly to use `redux.sync` for floating types since sm_100a?

It's indeed complex. It seems that  `__reduce_max_sync` can be used to substitute ASM, but no function warps `redux.sync` for floating types since sm_100a.

tzj-fxz:
> As far as I know, `__reduce_max_sync` does not accept float types. May you use assembly to use `redux.sync` for floating types since sm_100a?
> 
> It's indeed complex. It seems that `__reduce_max_sync` can be used to substitute ASM, but no function warps `redux.sync` for floating types since sm_100a.

Fixed in the latest commit. :)

src/tl_templates/cuda/reduce.h
```diff
@@ -250,14 +336,63 @@ template <int threads, int Axis = 0, bool reverse = false> struct CumSum2D {
...
+    (defined(__CUDA_ARCH_FEAT_SM100_ALL) || defined(__CUDA_ARCH_FEAT_SM100_F))
+  float value_cast = 0.0f;
+  if constexpr (std::is_same_v<T, half_t>) {
+    value_cast = __half2float(value);
+  } else if constexpr (std::is_same_v<T, bfloat16_t>) {
+    value_cast = __bfloat162float(value);
+  } else {
+    value_cast = static_cast<float>(value);
+  }
+  if constexpr (std::is_same_v<ReduceOp, MaxOp>) {
```
bucket-xv: As the AI bot points out, the integer value may also go through this branch, causing correctness problems

src/tl_templates/cuda/reduce.h
```diff
@@ -250,14 +336,63 @@ template <int threads, int Axis = 0, bool reverse = false> struct CumSum2D {
...
+  } else if constexpr (std::is_same_v<ReduceOp, MinOp>) {
+    float res;
+    asm("redux.sync.min.f32 %0, %1, %2;"
+        : "=f"(res)
+        : "f"(value_cast), "r"(mask));
+    return static_cast<T>(res);
+  }
+#endif
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 800)
+  auto run_reduce_sync = [&]<typename T_cast>(T_cast val) {
```
bucket-xv, line 311: These __reduce_sync functions only accept integers.

tzj-fxz:
@regression-perf

tzj-fxz:
Threads=128. One block.

| Op | M | N | Original (ms) | Redux (ms) | Original (tpt) | Redux (tpt) | Speedup |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **MAX** | 256 | 32 | 0.0077 | **0.0076** | 4.37 GB/s | 4.46 GB/s | **+1.32%** |
| **MIN** | 256 | 32 | 0.0067 | **0.0062** | 5.05 GB/s | 5.41 GB/s | **+7.13%** |
| **MAX** | 256 | 64 | 0.0068 | **0.0066** | 9.83 GB/s | 10.09 GB/s | **+3.03%** |
| **MIN** | 256 | 64 | 0.0067 | **0.0063** | 9.90 GB/s | 10.49 GB/s | **+6.35%** |
| **MAX** | 256 | 128 | 0.0067 | **0.0062** | 19.70 GB/s | 21.31 GB/s | **+8.06%** |
| **MIN** | 256 | 128 | 0.0070 | **0.0065** | 18.88 GB/s | 20.48 GB/s | **+7.69%** |

The regression performance test also shows that `example_topk.py`, which heavily uses `reduce_max`, has a 10%+ gain.

src/tl_templates/cuda/reduce.h
tzj-fxz: Yes. This lambda expression will only be triggered when the input is an integer.

bucket-xv:
May you add some more tests? i.e:
1. All datatypes and all types of reduce op. from fp4, fp8 to fp32 and int8 to int64. Also, max, min, add, absmax, absmin etc.
2. All shapes and block sizes. The current ones do not involve num threads < 32.
If that is all prepared, I have no further suggestions.

tzj-fxz:
1. Add more tests for various datatypes and operations. Since we rarely do reductions on low-precision datatypes (warp shuffle instructions also do not support them), we skip the tests for them.
2. Add tests for threads < 32.

bucket-xv:
I have no further concerns. Shall we test and merge it? @LeiWang1999

tzj-fxz:
I don't have a ROCm machine on hand to reproduce the error. The Hopper passes this test. It seems weird because the test function `test_reduce_max_clear` and the source code are not modified in the latest commit. Could anyone help?

LeiWang1999:
@regression-perf

tzj-fxz:
@regression-perf

PR #1764, opened 2026-02-01
[Feature] Support tcgen5mma lowering for `.kind::i8`

Rachmanino:
As titled.
Thanks @Hamerlate for providing the blackwell machine for testing.

## Summary by CodeRabbit

* **New Features**
  * Adds end-to-end support for int8 and uint8 tensor-core GEMM paths, including encoding for 8-bit integral types and int32 accumulators.

* **Tests**
  * Adds an int8/uint8 -> int32 GEMM correctness test targeting CUDA.

* **Refactor**
  * Standardizes 8-bit integer handling in code generation and template emission for tensor-core instructions.

### DIFF
--- a/maint/gemm_v2/correctness_evaluation_tcgen05.py
+++ b/maint/gemm_v2/correctness_evaluation_tcgen05.py
@@ -131,25 +131,38 @@ M_VALUES = [32, 64, 128, 256]
 N_VALUES = [64, 128, 256, 512]
 K_VALUES = [16, 32, 64, 128]
 K_VALUES_8Bit = [32, 64, 128]
-FALSE_TRUE_CASES = [
-    pytest.param(
-        k,
-        T.float16,
-        T.float32,
-        T.float32,
-        id=f"K{k}-float16-float-float",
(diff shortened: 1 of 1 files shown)


src/target/codegen_cuda.cc
```diff
@@ -2345,7 +2345,7 @@ void CodeGenTileLangCUDA::VisitExpr_(const CallNode *op, std::ostream &os) {
   } else if (op->op.same_as(tl::ptx_tcgen05_mma_ss())) {
     ICHECK_EQ(op->args.size(), 14U)
         << "ptx_tcgen05_mma_ss args is " << op->args;
-    std::string C_dtype = Downcast<StringImm>(op->args[0])->value;
+    std::string kind_dtype = Downcast<StringImm>(op->args[0])->value;
```
Rachmanino: I think this is a bugfix, because according to both tcgen5mma macro emitter side and codegen for ptx_tcgen05_mma_ts, `args[0]` is the dtype of AB for kind selection.

PR #1781, opened 2026-02-03
[Bugfix] Fix thread storage sync conflict detection for loop carry write-after-read

LeiWang1999:
## Summary
- Fix thread storage synchronization logic in `thread_storage_sync.cc` to correctly identify conflicts between read and write operations based on loop carry conditions
- The previous logic incorrectly checked for double buffer write with read operations without loop carry, but the correct behavior should detect write-after-read conflicts when loop carry is present

## Test plan
- [ ] Verify existing thread sync tests pass
- [ ] Confirm the fix addresses the synchronization issue in loop carry scenarios

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate placement and hoisting of thread-storage synchronizations, including loop-aware and runtime-dependent condition handling to avoid incorrect barriers.

* **New Features**
  * Loop-aware conflict analysis and target-specific warp-size support; ability to detect runtime-dependent conditions for hoisting.
  * Added a pure-Python reference attention implementation for decoding comparisons.

* **Tests**
  * New CUDA-focused tests for loop-carried sync patterns, hoisting behavior, and buffering.

* **Chores**
  * Reduced example matrix defaults, removed a paged decoding example and its test, updated third-party submodule, and exposed a constraint-accessor helper.

### DIFF
--- a/src/transform/thread_storage_sync.cc
+++ b/src/transform/thread_storage_sync.cc
@@ -1465,7 +1465,8 @@ private:
         // For loop-carry analysis, we compare iteration i with iteration i+1.
         // Since i+1 must be a valid iteration, i can only range from min to
         // min+extent-2 (i.e., extent-1 valid pairs instead of extent).
-        PrimExpr adjusted_extent = loop->extent - make_const(loop->extent.dtype(), 1);
+        PrimExpr adjusted_extent =
+            loop->extent - make_const(loop->extent.dtype(), 1);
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #1785, opened 2026-02-04
[Enhancement] Enhance register vectorize inference

LeiWang1999:
This pull request introduces an enhancement to the vectorization planning logic in `loop_vectorize.cc`, specifically improving how buffer indices are validated for vectorization at different vector sizes. The changes make the vectorization planner more robust by allowing it to recognize additional cases where indices can be safely vectorized, not just when they are invariant but also when they meet certain vectorization criteria.

Improvements to vectorization boundary checks:

* The loop that checks buffer indices for invariance at the new vector size now also considers whether indices are vectorizable using the `IndiceCanVectorize` method, expanding the set of cases where vectorization is allowed.
* Added a clarifying comment to indicate that the validation checks for both invariance and vectorizability at the new vector size boundary.

Code cleanup:

* Removed unnecessary blank lines in the buffer stride and offset calculation logic for better readability.

## Summary by CodeRabbit

* **Refactor**
  * Improved loop-vectorization revalidation so buffer vector-size feasibility is assessed with a broader feasibility check, enabling more loops to be safely vectorized.
* **Bug Fixes**
  * Fixed edge cases where vector-size adjustments were overly strict; the planner now safely reduces vector sizes when needed and logs adjustments to preserve correctness.
* **Chores**
  * Renamed an internal helper used by the vectorization planner (internal-only).

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -279,9 +279,8 @@ public:
                   << ")" << "\n";
       }
       // vector_size may be greater than local/fragment buffers' vector_size.
-      // In such case, we need to re-validate if the indices are invariant
-      // or is vectorizable at the new vector_size boundary
-      // at the new vector_size boundary. If not invariant, take GCD.
+      // In such case, we need to re-validate if the indices are vectorizable
+      // at the new vector_size boundary. If not, take GCD.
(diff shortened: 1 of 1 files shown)


SiriusNEO:
@regression-perf

PR #1795, opened 2026-02-05
[ThreadSync] Skip (tx1 != tx2) checking for loop carry analysis

LeiWang1999:
unnecessary and which may lead to performance regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of thread conflict detection by refining how different access patterns are handled.
  * Removed unnecessary warning messages from thread synchronization operations.

* **Performance**
  * Optimized thread variable management for more efficient computation in synchronization scenarios.


LeiWang1999:
@regression-perf

PR #1796, opened 2026-02-05
[Feature] Support passing PrimExpr value in tile-level atomic operation

SiriusNEO:
This PR supports operations like `T.atomic_add(A_buf, 1.0)`, which is primarily only supported in thread-level atomic add operations. And it also enhances the whole structure of atomic operators as well as the argument checking part of copy operators.

## Release Notes

* **New Features**
  * Atomic operations now support both buffer regions and scalar values as sources
  * Added Tensor Memory Accelerator (TMA) path for atomic add operations
  * Enhanced vectorization support for atomic operations in parallel contexts

* **Tests**
  * Expanded atomic operation test coverage with scalar sources, expressions, and memory ordering scenarios


SiriusNEO:
@regression-perf

Issue #1576, opened 2025-12-31
[BUG] Cannot dump tilelang autotune cache due to TVMFFIKernelAdapter API


retonym:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7post1

### System information

```
>>> print(sys.version, sys.platform)
3.13.9 (main, Nov 19 2025, 22:47:49) [Clang 21.1.4 ] linux
>>> print(tilelang.__version__)
0.1.7.post1
>>> print(torch.__version__)
2.9.1
```

### Problem description

When using tilelang autotuner, I encounter the below issue
```
2025-12-31 11:47:30  [TileLang:tilelang:ERROR]: Error saving kernel source code to disk: 'TVMFFIKernelAdapter' object has no attribute 'kernel_global_source'
2025-12-31 11:47:30  [TileLang:tilelang:ERROR]: Error saving kernel library to disk: Module[library] does not support SaveToBytes
```

### Reproducible example code

The Python snippets:

```python

```

### Traceback

```
Bench configurations: 100%|██████████| 6/6 [00:03<00:00,  1.73it/s, best_latency=16.3]
Tuned Latency 16.287647247314453 with config {'block_I': 32, 'num_stages': 0, 'num_threads': 128} at index 0
Tuned Latency 30.409759521484375 with config {'block_I': 64, 'num_stages': 0, 'num_threads': 256} at index 1
Tuned Latency 19.79088020324707 with config {'block_I': 64, 'num_stages': 0, 'num_threads': 128} at index 2
Tuned Latency 52.33155059814453 with config {'block_I': 128, 'num_stages': 0, 'num_threads': 256} at index 3
Tuned Latency 61.295616149902344 with config {'block_I': 128, 'num_stages': 0, 'num_threads': 512} at index 4
Tuned Latency 51.058719635009766 with config {'block_I': 128, 'num_stages': 0, 'num_threads': 128} at index 5
2025-12-31 11:47:30  [TileLang:tilelang:ERROR]: Error saving kernel source code to disk: 'TVMFFIKernelAdapter' object has no attribute 'kernel_global_source'
2025-12-31 11:47:30  [TileLang:tilelang:ERROR]: Error saving kernel library to disk: Module[library] does not support SaveToBytes
```

[...]

PR #1812, opened 2026-02-08
Handle staled autotune state with tvm-ffi adapter.

### LINKED ISSUE #1813: [BUG] Autotuner fails to save kernel library after crash recovery due to stale JIT cache state
### Required prerequisites

0.1.7.post3+cuda.git47869154

The nightly build version on the main branch of the repository.

### Problem description

Issue #1576 mentions that the autotune result cannot be saved because `kernel_global_source` was missing from the `TVMFFIKernelAdapter` object, which was fixed by #1589. However, it has recently come to my attention that the following error can happen again, if events occur in a particularly bad sequence, which leads to a stale state in the autotune cache.

```
2026-02-07 20:46:54  [TileLang:tilelang:ERROR]: Error saving kernel source code to disk: 'TVMFFIKernelAdapter' object has no attribute 'kernel_global_source'
2026-02-07 20:46:54  [TileLang:tilelang:ERROR]: Error saving kernel library to disk: Module[library] does not support SaveToBytes
```

**In the normal case, where we run a fresh instance of autotune:**

1. Autotuner cache miss.
2. For each config, tilelang compiles -> KernelCache.cached() -> JIT cache miss -> compiles fresh -> saves .so to JIT cache -> returns fresh kernel.
3. Autotuner profiles all kernels, picks best.
4. Autotuner saves to its own cache -> `_safe_write_executable` -> works because the executable is still the freshly-compiled `Executable`.
5. Next run -> autotuner cache hits.

**However, if a crash happens in-between 2 and 3, the autotune cache would run into issues:**

1. First run: autotuner cache miss -> starts compiling configs.
2. KernelCache.cached() compiles each kernel fresh and saves them to the JIT cache immediately.
3. Ctrl+C / crash -> kills the process before the autotuner reaches `_save_result_to_disk`.
4. State on disk: JIT cache has .so files for some/all configs. Autotuner cache has nothing.

**Next run after crash:**

1. Autotuner cache miss (was never written).
2. For each config, tilelang compiles -> KernelCache.cached() -> JIT cache **HIT** -> `from_database()`  -> `runtime.load_module(...)` -> produces a `Module[library]`.
3. Autotuner profiles, picks best.
4. Autotuner tries to save -> `_safe_write_executable(Module[library]` -> `Module[library] does not support SaveToBytes`, which raises an error and aborts without saving anything to disk. We're back at where we started right after the crash, and thus stuck in a stale state....

So the root cause isn't really a "bug in normal operation". It's a crash recovery gap. In particular, for the `tvm_ffi` execution backend, before calling `_safe_write_executable` in the following code snippet, we should check if the kernel was loaded from a previous cache, or freshly compiled.

https://github.com/tile-ai/tilelang/blob/47869154b63d1a2995f8206992de39ca7c682d50/tilelang/autotuner/param.py#L239-L244

---

The fresh compilation path (in `_convert_torch_func`) creates a serializable `Executable`:

https://github.com/tile-ai/tilelang/blob/47869154b63d1a2995f8206992de39ca7c682d50/tilelang/jit/adapter/tvm_ffi.py#L170-L171

But `from_database` loads a raw `.so`, which can execute but not re-export:

https://github.com/tile-ai/tilelang/blob/47869154b63d1a2995f8206992de39ca7c682d50/tilelang/jit/adapter/tvm_ffi.py#L284

### Reproducible example code

This issue can happen with autotuning in any kernel usign tvm-ffi execution backend, whose JIT cache is present on disk, but autotune cache has nothing.

### Expected behavior

The autotuner result should be properly saved back to disk and uncover from the stale state, regardless of crash in-between.

haok1402:
Fix #1813 

Hopefully, the root cause analysis from #1813 provides a more resillient understanding into why #1576 occurs in the first place and how to fix the issue in a grounded manner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved kernel library handling for the TVM FFI backend by optimizing how kernel libraries are cached and loaded, preferring available library copies when present.

### DIFF
--- a/tilelang/autotuner/param.py
+++ b/tilelang/autotuner/param.py
@@ -237,7 +237,7 @@ class AutotuneResult:
                     logger.debug(f"Saving kernel library to file: {kernel_lib_path}")
                 self._safe_write_file(kernel_lib_path, "wb", lambda f: f.write(self._load_binary(src_lib_path)))
             elif kernel.execution_backend == "tvm_ffi":
-                if hasattr(kernel.adapter, 'libpath') and kernel.adapter.libpath:
+                if hasattr(kernel.adapter, "libpath") and kernel.adapter.libpath:
                     src_lib_path = kernel.adapter.libpath
                     if verbose:
                         logger.debug(f"Copying kernel library to file: {kernel_lib_path}")


LeiWang1999:
LGTM!

PR #1814, opened 2026-02-08
[Release][Build] Merge libtilelang and libtilelang_modules

oraluben:
Also skip `z3-solver==4.15.5.0`: https://github.com/Z3Prover/z3/issues/8536

cc @Blealtan

## Summary by CodeRabbit

* **Chores**
  * Exclude a problematic z3-solver version from package metadata and build config.
  * Consolidate native library outputs into a single public artifact, simplify target list, and adjust CUDA stub handling when CUDA is enabled.
  * Remove explicit runtime linkage from the native target.

* **Refactor**
  * Loader now consistently resolves the consolidated native library name.

### DIFF
--- a/examples/gemm/example_gemm_autotune.py
+++ b/examples/gemm/example_gemm_autotune.py
@@ -107,7 +107,13 @@ def get_configs(M, N, K, with_roller=False, topk=20):
     return configs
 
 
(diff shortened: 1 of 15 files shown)


kurisu6912 (approved):
LGTM!

This will reduce the binary size from ~89M to ~74M.

PR #1827, opened 2026-02-09
[Refactor] Introduce `T.access_of` to combine `T.address_of` and `access_ptr`

LeiWang1999:
## Motivation

TileLang previously lowered `T.access_ptr(...)` to one of two existing TIR patterns:

- `tir.address_of(BufferLoad(...))`
  - Pros: keeps the original `BufferLoad`.
  - Cons: does not carry explicit access intent / scope metadata (read vs write), which forces downstream passes to be conservative (e.g. inserting sync even when a narrower access footprint would be legal).
- `tir.builtin.tvm_access_ptr(dtype, data, offset, extent, rw_mask)`
  - Pros: carries `rw_mask` and `extent`.
  - Cons: drops the original `BufferLoad` and therefore loses indexing/stride context. This breaks expectations in passes that rely on the buffer access structure (e.g. “safe memory access” legalization), and `tvm_access_ptr` is not desired to appear in the frontend IR.

To resolve this mismatch, TileLang introduces a dedicated frontend-only op that preserves both:

- the **structured buffer access** (`BufferLoad`) and
- the **access intent** (`rw_mask` + `extent`).

### 1) New TileLang Op: `tl.access_ptr(base_load, extent, rw_mask)`

A new TileLang-only intrinsic is registered and used as the frontend representation of `T.access_ptr`:

- **Op name:** `tl.access_ptr`
- **Signature:**
  - `arg0`: `tir.BufferLoad` (`base_load`)
  - `arg1`: 1D extent in **elements** (`extent`)
  - `arg2`: access mask (`rw_mask`): `1=read`, `2=write`, `3=read-write`
- **Effect kind:** pure (metadata-only)

Files:

- `src/op/builtin.h`
- `src/op/builtin.cc`

### 2) Frontend `T.access_ptr` now emits `tl.access_ptr`

`tilelang/language/builtin.py` was updated so `T.access_ptr(...)` no longer dispatches to
`tir.address_of(...)` or `tir.builtin.tvm_access_ptr(...)`. Instead it creates:

```python
tir.call_intrin(
    "handle",
    tir.op.Op.get("tl.access_ptr"),
    base_load,     # BufferLoad
    extent_1d,     # PrimExpr
    rw_mask,       # IntImm(int32)
)
```

Key frontend semantics:

- The first argument is always a `BufferLoad`, so downstream passes can recover:
  - the referenced `tir.Buffer`
  - strides (explicit `buffer->strides` if present, otherwise row-major)
  - storage scope / layout metadata associated with the buffer
- Extent defaults follow typical pointer intent:
  - `BufferLoad` base: `extent = 1` (element pointer)
  - `BufferRegion` base: `extent = product(region.extents)`
  - `Buffer` base: `extent = product(buffer.shape)`
- `ignore_last_ndim` is implemented by zeroing the trailing indices in the emitted `BufferLoad`.
- `offset != 0` is only supported for **1D** bases when emitting `tl.access_ptr` (for N-D, users should express the offset via indexing, e.g. `A[i + off, j]`).

File:

- `tilelang/language/builtin.py`

### 3) New lowering pass: `tl.transform.LowerAccessPtr`

To keep interoperability with downstream infrastructure that expects
`tir.builtin.tvm_access_ptr`, TileLang adds a dedicated lowering pass:

- Rewrites:
  - `tl.access_ptr(BufferLoad(buf, indices), extent, rw_mask)`
  - into:
  - `tir.builtin.tvm_access_ptr(dtype, buf.data, linear_offset, extent, rw_mask)`
- `linear_offset` computation:
  - uses `buffer->strides` when defined (and matches ndim)
  - otherwise falls back to row-major strides from `buffer->shape`
  - for vector indices (e.g. `Ramp`, `Broadcast`), uses the base/value to derive a conservative base offset

Files:

- `src/transform/lower_access_ptr.cc`
- `tilelang/transform/__init__.py` (Python binding)

### 4) Compilation pipeline integration

`LowerAccessPtr` is inserted into the standard TileLang lowering pipeline so that:

- early passes can still see structured buffer access (`BufferLoad`) for analysis and safety checks
- later stages can operate on standardized `tvm_access_ptr` when needed for codegen patterns

Current placement:

- immediately after `LegalizeSafeMemoryAccess()`

File:

- `tilelang/engine/phase.py`

### 5) Compatibility updates in existing passes

Some existing passes/utilities previously assumed all pointer metadata would appear as
`tvm_access_ptr` (or `address_of`). They were extended to also accept `tl.access_ptr`:

- `src/op/operator.cc`
  - `GetVarFromAccessPtr(...)` now supports extracting the underlying `Var` from `tl.access_ptr`
- `src/transform/lower_tile_op.cc`
  - permuted-layout and related pointer rewrite logic now recognizes `tl.access_ptr`
  - avoids hard failures for `T.access_ptr(...)` under layout transformations
- `src/transform/loop_vectorize.cc`
  - atomic-add dtype inference accepts `tl.access_ptr` as the destination pointer argument

## Tests

Tests were updated/added to reflect the new frontend shape and the explicit lowering step:

- `testing/python/language/test_tilelang_language_access_ptr.py`
  - verifies `T.access_ptr(...)` builds `tl.access_ptr(BufferLoad, extent, rw_mask)`
  - verifies `tilelang.transform.LowerAccessPtr()` rewrites to `tir.tvm_access_ptr` with expected `offset/extent`
- `testing/python/language/test_tilelang_language_access_ptr_codegen.py`
  - smoke-test: `T.access_ptr -> tl.access_ptr -> tvm_access_ptr -> cp.async` remains codegen-able on CUDA

### DIFF
--- a/examples/bitnet-1.58b/kernel_benchmark/tilelang_bitnet_158_int8xint2_decode.py
+++ b/examples/bitnet-1.58b/kernel_benchmark/tilelang_bitnet_158_int8xint2_decode.py
@@ -139,8 +139,8 @@ def bitnet_158_int8xint2_decode(
                 T.call_extern(
                     "handle",
                     "decode_i2u_to_i8s",
-                    T.address_of(B_quant_local[0]),
-                    T.address_of(B_dequantize_local[0]),
+                    T.access_ptr(B_quant_local, "r"),
+                    T.access_ptr(B_dequantize_local, "w"),
                 )
 
                 if use_dp4a:
--- a/examples/bitnet-1.58b/kernel_benchmark/tilelang_bitnet_158_int8xint2_prefill.py
+++ b/examples/bitnet-1.58b/kernel_benchmark/tilelang_bitnet_158_int8xint2_prefill.py
@@ -253,8 +253,8 @@ def bitnet_158_int8xint2_prefill(
                     T.call_extern(
                         "handle",
                         "decode_i2u_to_i8s",
-                        T.address_of(B_local[0]),
-                        T.address_of(B_dequantize_local[0]),
+                        T.access_ptr(B_local, "r"),
+                        T.access_ptr(B_dequantize_local, "w"),
                     )
 
                     for v in T.vectorized(0, local_size):
--- a/examples/dequantize_gemm/example_dequant_gemm_bf16_fp4_hopper.py
+++ b/examples/dequantize_gemm/example_dequant_gemm_bf16_fp4_hopper.py
@@ -180,8 +180,8 @@ def matmul(
                 # Then, dequant.
                 T.call_extern(
                     func_name,
-                    T.address_of(B_local_thread[0]),
-                    T.address_of(B_dequantize_local_thread[0]),
+                    T.access_ptr(B_local_thread, "r"),
+                    T.access_ptr(B_dequantize_local_thread, "w"),
                     1,
                     dtype=out_dtype,
                 )
--- a/examples/dequantize_gemm/example_dequant_gemm_bf16_mxfp4_hopper.py
+++ b/examples/dequantize_gemm/example_dequant_gemm_bf16_mxfp4_hopper.py
@@ -246,8 +246,8 @@ def matmul(
                 # Then, dequant.
                 T.call_extern(
                     func_name,
-                    T.address_of(B_local_thread[0]),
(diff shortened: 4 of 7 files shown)


LeiWang1999:
@regression-perf

PR #1834, opened 2026-02-11
[LoopVectorize] Loop Independent Var Optimization in IfThenElse Expr

kurisu6912:
The following loop fails to vectorized for a expression `tx * 4 + i < N` related to `i`.

```py
T.assume(N % 4 == 0)
for i in T.vectorized(4):
    a_shared_1[i] = T.if_then_else(tx * 4 + i < N, A[tx * 4 + i], T.float32(0.0))
```

However, we can prove that `tx * 4 + i < N` is equvilant to `tx * 4 < N`, because we have assumed `N % 4 == 0`. Then the loop can be vectorized to:

```py
a_shared_1[0:4] = T.if_then_else(T.Add(tx * 4, 0) < N, A[tx * 4:tx * 4 + 4], T.Broadcast(T.float32(0.0), 4))
```

This pr add the check in loop vectorize pass and allows better vectorize performance in some predicated load senarios.

## Summary by CodeRabbit

* **Chores**
  * Updated third-party dependency reference


src/transform/vectorize_loop.cc
```diff
@@ -473,7 +476,12 @@ class TLVectorizer : public StmtMutator,
 
   // IfThenElse expr
   PrimExpr MutateIfThenElseExpr_(const CallNode *op) {
-    PrimExpr cond = this->VisitExpr(op->args[0]);
+    PrimExpr cond = op->args[0];
+    PrimExpr cond_zeroed = Substitute(cond, {{var_, IntImm(var_->dtype, 0)}});
+    if (analyzer_->CanProve(cond == cond_zeroed)) {
+      cond = cond_zeroed;
+    }
+    cond = this->VisitExpr(cond);
```
kurisu6912: Here is the main change!

LeiWang1999:
@regression-perf

PR #1835, opened 2026-02-11
[Refactor] Treat `local.var` as `local` buffers when deciding vectorization for stable actions

LeiWang1999:
## Summary by CodeRabbit

* **Refactor**
  * Improved buffer detection logic to better handle local memory optimization.

* **Tests**
  * Added test coverage for vectorization behavior with local variable handling.


LeiWang1999:
@regression-perf

PR #1838, opened 2026-02-11
[MISC] Fixed the image display issue under different themes.

1195343015:
The original images show different problems under different themes.
Like:
In a light theme, images should not have a black background.
image
In dark mode, the black text in the code is completely unreadable.
image

After modification
image

image

## Summary by CodeRabbit

* **Documentation**
  * Updated image format in README from PNG to SVG.


LeiWang1999:
@1195343015 Thanks, can we still keep the pink style?

1195343015:
> @1195343015 Thanks, can we still keep the pink style?

Okay, let me try it in a little while. The current color really isn't very appealing.

1195343015:
The original pink style doesn't look very friendly on a white background (Plan A)

So I came up with the following three options. Which one do you think is the best? @LeiWang1999 

Plan A 
image
Plan B 
image
Plan C 
image

Rachmanino:
Thanks for the contribution! We don't think this change is necessary at the moment, so closing this for now. Appreciate your interest! 🙏

Issue #1648, opened 2026-01-09
[BUG] TMA fails


bucket-xv:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7.post2+cuda.gita56212de

### System information

3.12.9 | (main, Feb  6 2025, 18:56:27) [GCC 11.2.0] linux

0.1.7.post2+cuda.gita56212de
2.9.0+cu130

### Problem description

Failure once remove disable tma lower

### Reproducible example code

The Python snippets:

```python
import tilelang
from tilelang import language as T
import torch

@tilelang.jit(
    pass_configs={
        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
        tilelang.PassConfigKey.TL_ENABLE_FAST_MATH: False,
    },
)
def get_sample_kernel(k: int):
    m = T.symbolic('m')
    block_m = 4
    block_k = 128
    num_threads = 32
    in_dtype = T.float8_e4m3fn
    @T.prim_func
    def sample_kernel(x: T.Tensor[(m, k), in_dtype]):
        with T.Kernel(T.ceildiv(m, block_m), T.ceildiv(k, block_k), threads=num_threads) as (
            pid_m,
            pid_k,
        ):
            # Local buffers
            x_in_shared = T.alloc_shared((block_m, block_k), dtype=in_dtype)

            # Load x to fragment
            T.fill(x_in_shared, 0)
            T.copy(
                x[pid_m * block_m: (pid_m + 1) * block_m, pid_k * block_k: (pid_k + 1) * block_k],
                x_in_shared
            )

    return sample_kernel

kernel = get_sample_kernel(128)
print(kernel.get_kernel_source())

```

### Traceback

```pytb
/tmp/tmpmgb2e0ij/tvm_kernels.cu(23): error: identifier "mbarrier" is undefined
      tl::tma_load(x_desc, mbarrier[0], (&(x_in_shared[0])), 0, (((int)blockIdx.x) * 4));
```

Rachmanino:
Currently auto warp-specialized pass is coupled with TMA pass, which means we could not use TMA with warp-specialization disabled for now. The reason in detail is that the detection of mbarrier lies in the wasp pass. We may decouple them in the future.

[...]

PR #1840, opened 2026-02-11
[BugFix] Fix Hopper TMA lowering without warp specialization

### LINKED ISSUE #737: Lower to TMA correctly without automatic warp specialization
Currently the warp specialization pass takes responsibility of inserting barriers waiting for TMA operations to complete, so if automatic warp specialization is not expected, `T.copy` will not be able to lower to TMA safely.

In specific, this combination of `pass_configs` should work correctly, but now it isn't:

```python
@tilelang.jit(pass_configs={
    tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: False,
    tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
})
def f():
    ...
```

Henry-Jessie:
## Summary

When `tl.disable_warp_specialized=True` but `tl.disable_tma_lower=False`, the optimization pipeline skipped the `WarpSpecialized` + `InjectTmaBarrier` passes, leaving TMA load instructions referencing undeclared `mbarrier` symbols. This caused NVCC compilation failure: `identifier "mbarrier" is undefined`.

**Root cause:** `OptimizeForTarget` in `phase.py` gated the entire Hopper pipeline on `allow_tma_and_warp_specialized()`, which required both TMA *and* WS to be enabled. When WS was disabled, the pipeline fell through to the generic branch that lacks barrier allocation/injection.

**Fix:**
- Add `allow_tma_lower()` that only checks TMA capability, decoupled from warp specialization
- Change the Hopper pipeline entry condition to `allow_tma_lower()` so the `WarpSpecialized` pass (which supports a mbarrier-only mode when WS is disabled) and `InjectTmaBarrier` still run
- Fix `inject_tma_barrier.cc` to promote raw `IntImm` barrier ids to `get_mbarrier()` calls for multi-dimensional (descriptor-based) TMA loads in the degenerate producer-only path

Fixes #737, #1648

## Test plan

- [x] `pre-commit run --all-files` passes
- [x] New regression test `test_tilelang_issue_tma_no_ws.py` covers both 1D and 2D-descriptor TMA paths
- [x] Full test suite on H20 (SM90): 784 passed, 107 skipped — all 19 failures are pre-existing NVRTC/cutedsl issues (verified identical on clean `origin/main`)

## Summary by CodeRabbit

* **New Features**
  * TMA lowering can be enabled independently of warp specialization for more flexible compilation paths.

* **Bug Fixes**
  * Improved barrier insertion and promotion so non-1D and 1D TMA load patterns are handled reliably.

* **Tests**
  * Added CUDA tests that validate TMA lowering without warp specialization, barrier emission, descriptor usage, and end-to-end kernel execution.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_tma_no_ws.py
+++ b/testing/python/issue/test_tilelang_issue_tma_no_ws.py
@@ -105,8 +105,7 @@ def test_tma_lower_no_warp_specialized_2d_descriptor_uses_args1_barrier():
     flat_src = " ".join(src.split())
     pattern = r"tl::tma_load\([^,]+,\s*mbarrier\[[0-9]+\]"
     assert re.search(pattern, flat_src), (
-        f"Expected regex {pattern!r} to match flattened CUDA source. "
-        f"Generated source (truncated):\n{src[:1000]}"
+        f"Expected regex {pattern!r} to match flattened CUDA source. Generated source (truncated):\n{src[:1000]}"
     )
 
     x = torch.randn((M, K), device="cuda", dtype=torch.float16)


Henry-Jessie:
Thanks for the thorough review!

`is_1d_tma_load` inconsistency: I unified the 1D-TMA detection into a shared `Is1DTmaLoad()` helper used by both `TmaExpectTxRewriter` and `TmaBarrierRewriter`. The helper treats only `tma_load` as potentially 1D and excludes both `create_tma_descriptor` and `create_tma_im2col_descriptor`, so both rewriters now agree on the barrier argument index. 

Duplicate `LowerOpaqueBlock` / redundant `InjectFenceProxy` gating: These redundancies are pre-existing and not introduced by this PR. To keep the change focused on the TMA lowering bugfix, I'm leaving them as-is here.

Regex assertion failure message: Added an explicit failure message including the expected pattern and a truncated source snippet.

Henry-Jessie:
It seem that the ROCm CI failure is pre-existing and unrelated to this PR, the same HIPBLAS_STATUS_INVALID_VALUE error appears in all recent PRs (e.g. #1870,  #1863 ). CUDA and Metal CI pass cleanly.

LeiWang1999:
@regression-perf

LeiWang1999 (approved):

LeiWang1999:
LGTM!

PR #1845, opened 2026-02-13
[Enhancement] Optimize templates for half/bfloat16

LJC00118:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected min/max behavior for bfloat16 and float16 scalars and added explicit bfloat16/float16 comparisons in reductions.
  * Ensured proper initialization, duplication, and update semantics for max/min/absmax reduction paths when using temporary buffers.

* **Refactor**
  * Optimized 16-bit float shuffle and math helpers to use bitwise-safe pathways for improved correctness and performance.


LJC00118:
@regression-perf

Issue #1849, opened 2026-02-14
[Enhancement] Enhance the conditions for async proxy fence injection


Rachmanino:
Currently TileLang's `inject_fence_proxy` pass relies on the detection of transition from generic to async proxy to  inject proxy fence, which produces uncessary fence e.g. between `mbarrier_expect_tx` and `tma_load`, because they don't contain any w-r dependency indeed. Furthermore, `LDGSTS`-kind instructions belong to generic proxy but are mistaken as async ones in this pass.

PR #1850, opened 2026-02-15
[Enhancement] Enhance the conditions for async proxy in `InjectFenceProxy`

Rachmanino:
#1849
- Remove CpAsync intrins as they belong to generic proxy, according to NV doc.
- DOES NOT inject fence before TmaLoad, as it has no dependency on previous data from generic proxy

## Summary by CodeRabbit

* **Refactor**
  * Refined internal handling of asynchronous intrinsics to a smaller, focused set for more predictable compilation.
  * Removed explicit special-casing of certain async copy operations and clarified proxy ownership in documentation.
  * Minor doc updates and comments to explain behavior.
  * No changes to public APIs or user-facing functionality.

### DIFF
--- a/src/transform/inject_fence_proxy.cc
+++ b/src/transform/inject_fence_proxy.cc
@@ -80,7 +80,7 @@ bool IsAsyncIntrinsic(const CallNode *call) {
 
   // TileLang async intrinsics
   // Although tma_load is also an async intrinsic, it has no dependency on data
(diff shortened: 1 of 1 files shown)


LeiWang1999:
It would be better to provide a test similar to `test_tilelang_transform_thread_sync.py`. Ensuring this modification is safe is critical, as fence synchronization bugs are notoriously hard to debug

Issue #1846, opened 2026-02-13
[BUG] Calling eager mode function with no T.Tensor input directly failed


ColmaLiu:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.7.post3+cuda.git1f16d9a5

### System information

3.12.12 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 20:16:04) [GCC 11.2.0] linux
0.1.7.post3+cuda.git1f16d9a5
2.10.0+cu128

### Problem description

A function written in eager mode whose inputs do not include T.Tensor objects cannot be called directly.

### Reproducible example code

The Python snippets:

```python
import tilelang
import tilelang.language as T

@tilelang.jit
def test(fill, M):
    X = T.empty((M), dtype="float32")
    with T.Kernel(1, threads=M) as _:
        for i in T.Parallel(M):
            X[i] = fill
    return X

# This is OK.
# kernel = test.compile(fill=1.0, M=32)
# print(kernel.get_kernel_source())
# X = kernel()
# print(X)

# Failed.
X = test(fill=1.0, M=32)
print(X)
```

### Traceback

```pytb
Traceback (most recent call last):
  File "/home/lyn/workspace/test.py", line 19, in <module>
    X = test(fill=1.0, M=32)
        ^^^^^^^^^^^^^^^^^^^^
  File "/home/lyn/workspace/tilelang/tilelang/jit/__init__.py", line 446, in __call__
    return kernel(*kernel_args.values())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lyn/workspace/tilelang/tilelang/jit/kernel.py", line 207, in __call__
    return self.torch_function(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lyn/workspace/tilelang/tilelang/jit/adapter/tvm_ffi.py", line 199, in func
    raise ValueError(f"Kernel expected {expected_inputs} inputs, but {len(inputs)} are provided.")
ValueError: Kernel expected 0 inputs, but 2 are provided.
```

LeiWang1999:
closed as it's completed.

PR #1851, opened 2026-02-15
[BugFix] Fix eager mode where there is no tensor args

Rachmanino:
#1846
When there is no tensor args, we should let the kernel accept an empty runtime kwargs, rather than the original full kwargs

## Release Notes

* **Bug Fixes**
  * Ensured argument parsing returns a consistent empty mapping when no tensor arguments are present, improving reliability.

* **Tests**
  * Added a CUDA-restricted regression test verifying an eager/JIT kernel correctly allocates and fills a 1D tensor, asserting shape and values to prevent regressions.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_1846.py
+++ b/testing/python/issue/test_tilelang_issue_1846.py
@@ -1,0 +1,35 @@
+import torch
+
+import tilelang
+import tilelang.testing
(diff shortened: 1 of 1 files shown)


LeiWang1999:
would you mind add a simple test for this issue?

PR #1877, opened 2026-02-25
[Build] Disable gtest in tvm

oraluben:
By default, tvm tries to detect if gtest is installed. When the library is installed but header is not, configuration may fail.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to disable Google Test detection during CMake configuration.


LeiWang1999:
LGTM as we do not have cpp tests.

Issue #1868, opened 2026-02-22
[BUG] tvm.error.InternalError: Check failed: (analyzer_->CanProveEqual(abs(source->scale), 1)) is false


a858438680:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.8

### System information

3.12.3 (main, Jan 22 2026, 20:57:42) [GCC 13.3.0] linux
0.1.8
2.8.0+cu128

<frozen runpy>:128: RuntimeWarning: 'torch.utils.collect_env' found in sys.modules after import of package 'torch.utils', but prior to execution of 'torch.utils.collect_env'; this may result in unpredictable behaviour
Collecting environment information...
PyTorch version: 2.8.0+cu128
Is debug build: False
CUDA used to build PyTorch: 12.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.2 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version: Could not collect
CMake version: version 4.0.2
Libc version: glibc-2.39

Python version: 3.12.3 (main, Jan 22 2026, 20:57:42) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-6.8.0-85-generic-x86_64-with-glibc2.39
Is CUDA available: True
CUDA runtime version: Could not collect
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: 
GPU 0: NVIDIA GeForce RTX 4090
GPU 1: NVIDIA GeForce RTX 4090

Nvidia driver version: 570.172.08
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        52 bits physical, 57 bits virtual
Byte Order:                           Little Endian
CPU(s):                               256
On-line CPU(s) list:                  0-255
Vendor ID:                            AuthenticAMD
Model name:                           AMD EPYC 9554 64-Core Processor
CPU family:                           25
Model:                                17
Thread(s) per core:                   2
Core(s) per socket:                   64
Socket(s):                            2
Stepping:                             1
[...]

PR #1884, opened 2026-02-27
[Analysis] Refactor FragmentLoopChecker visiting style

SiriusNEO:
Check the cases: https://github.com/tile-ai/tilelang/issues/1868

## Summary by CodeRabbit

* **Refactor**
  * Improved nested-loop validation for fragment buffer accesses using a scoped loop stack, ensuring checks consider innermost loop context and symbolic-range loops.
  * Error messages and rules updated to refer to fragment buffers consistently.
* **Bug Fix**
  * Fixed missed detections in complex loop nesting by collecting fragment accesses at the correct scope and validating against all relevant symbolic-range parallel loops.


SiriusNEO:
Found some examples which violate the check rules. Need to verify

LeiWang1999:
somehow correct, guess we should allow this pattern

PR #1904, opened 2026-03-06
[Feature] Add T.gemm support for CPU target

xyyy1420:
## Summary

Enable `T.gemm` to work with `target="c"` by adding a CPU scalar fallback implementation. Previously, `T.gemm` on CPU would crash with `ICHECK(0) << "Unsupported target for gemm"`.

Depends on #1899 and #1901.

## Changes

**C++ (gemm_v2 path)**
- Add `GemmInst::kCPU` enum value in `gemm.h`
- Add `TargetIsCPU()` helper in `target/utils.{h,cc}`
- Add CPU branch in `GemmPyNode::getGemmInst()` — returns `kCPU` when target is CPU

**Python**
- Add `GemmInst.CPU = 4` and `is_cpu()` in `inst.py`
- Add `GemmScalar` class in `gemm_scalar.py` — triple-loop scalar gemm implementation
- Add CPU dispatch in `GemmPy._get_implementation_class()` — `is_cpu() → GemmScalar`
- Update docstring for `_select_gemm_instruction()` to reflect all supported targets

**Tests**
- Add `test_tilelang_cpu_tgemm.py` with 11 tests:
  - 4 codegen tests (basic, rectangular, trans_A, trans_B)
  - 4 correctness tests for NN layout (multiple sizes: 128³, 256³, 256×512×128, 512³)
  - 3 correctness tests for transpose combinations (TN, NT, TT)

## Design

The CPU path only uses the gemm_v2 (Python) pipeline:

T.gemm() → GemmPyNode::getGemmInst() → kCPU
        → GemmPy._get_implementation_class() → GemmScalar
        → GemmScalar.lower() → triple-loop IR

The gemm_v1 (C++) path is unchanged — it will still `ICHECK` on CPU targets since we don't plan to support it for CPU.

## Future work

`GemmScalar` is the universal CPU fallback. Architecture-specific implementations (e.g., RVV, AVX, AMX, SME) can be added in two ways:

1. **Sub-dispatch under `is_cpu()`**: Extend `_get_implementation_class()` to further dispatch based on CPU architecture (e.g., `is_riscv() → GemmRVV`, default → `GemmScalar`). No C++ changes needed.

2. **Peer-level with `kCPU`**: Add new `GemmInst` values (e.g., `kRVV`, `kAMX`) and corresponding `TargetIs*()` checks in `getGemmInst()`, making them siblings of `kCPU`/`kMMA`/`kMFMA`. This requires renaming `kCPU` to `kScalar` or similar to reflect it's the scalar-only fallback.

Both approaches are straightforward — option 1 requires only Python changes, option 2 requires minor naming adjustments in both C++ and Python.

### DIFF
--- a/src/op/gemm.h
+++ b/src/op/gemm.h
@@ -39,7 +39,7 @@ inline const char *GemmWarpPolicyTypeToString(GemmWarpPolicyType type) {
 }
 
 // Target GEMM instruction
-enum class GemmInst : uint8_t { kMMA, kWGMMA, kTCGEN5MMA, kMFMA, kCPU };
+enum class GemmInst : uint8_t { kMMA, kWGMMA, kTCGEN5MMA, kMFMA, kScalar };
 
 /// Convert GemmInst enum to string for debugging
 inline const char *GemmInstToString(GemmInst inst) {
@@ -52,8 +52,8 @@ inline const char *GemmInstToString(GemmInst inst) {
     return "TCGEN5MMA";
   case GemmInst::kMFMA:
     return "MFMA";
-  case GemmInst::kCPU:
-    return "CPU";
+  case GemmInst::kScalar:
+    return "Scalar";
   default:
     return "Unknown";
   }
--- a/src/op/gemm_py.cc
+++ b/src/op/gemm_py.cc
@@ -137,7 +137,7 @@ GemmInst GemmPyNode::getGemmInst(int block_size, Target target) const {
   } else if (TargetIsCuda(target)) {
     return GemmInst::kMMA;
   } else if (TargetIsCPU(target)) {
-    return GemmInst::kCPU;
+    return GemmInst::kScalar;
   } else {
     ICHECK(0) << "Unsupported target for gemm: " << target->str();
     return GemmInst::kMMA; // This line will never be reached due to ICHECK, but
--- a/tilelang/tileop/gemm/__init__.py
+++ b/tilelang/tileop/gemm/__init__.py
@@ -131,7 +131,7 @@ class GemmPy(Node, Scriptable):
         2. WGMMA for Hopper architecture with sufficient matrix size and warp count
         3. MFMA for CDNA (AMD) architecture
         4. MMA for CUDA architecture
-        5. CPU for CPU target (scalar fallback)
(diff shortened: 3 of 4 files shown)


LeiWang1999:
Thanks for your contribution! think can have a better name instead of gemm.cpu, I will also take a look then :)

xyyy1420:
Thanks for the feedback! 

Looking at the existing naming convention (`kMMA`, `kWGMMA`, `kMFMA`, `kTCGEN5MMA`), they all describe "how to compute" rather than "where". So something like `kScalar` might be more consistent? since this is a scalar triple-loop fallback.

One open question: CPU targets have diverse ISA extensions (RVV, AMX, AVX, SVE...). Should we add them as peer-level enum values (`kRVV`, `kAMX`, ...) or keep a single entry point and sub-dispatch by ISA extensions in Python only? Happy to hear your thoughts.

LeiWang1999:
> Thanks for the feedback!
> 
> Looking at the existing naming convention (`kMMA`, `kWGMMA`, `kMFMA`, `kTCGEN5MMA`), they all describe "how to compute" rather than "where". So something like `kScalar` might be more consistent? since this is a scalar triple-loop fallback.
> 
> One open question: CPU targets have diverse ISA extensions (RVV, AMX, AVX, SVE...). Should we add them as peer-level enum values (`kRVV`, `kAMX`, ...) or keep a single entry point and sub-dispatch by ISA extensions in Python only? Happy to hear your thoughts.

Sure, kScalar lgtm.

xyyy1420:
> > Thanks for the feedback!
> > Looking at the existing naming convention (`kMMA`, `kWGMMA`, `kMFMA`, `kTCGEN5MMA`), they all describe "how to compute" rather than "where". So something like `kScalar` might be more consistent? since this is a scalar triple-loop fallback.
> > One open question: CPU targets have diverse ISA extensions (RVV, AMX, AVX, SVE...). Should we add them as peer-level enum values (`kRVV`, `kAMX`, ...) or keep a single entry point and sub-dispatch by ISA extensions in Python only? Happy to hear your thoughts.
> 
> Sure, kScalar lgtm.

already renamed

PR #683, opened 2025-07-30
[Enhancement] Output cache-file-related messages with verbose=True

chenyang78:
This is a minor enhancement to output verbose messages indicating where cache files are saved and loaded. These messages are useful for examining the relevant intermediate files.

PR #1906, opened 2026-03-06
[Enhancement] Add eager-mode support for tilelang.autotune

ColmaLiu:
Currently `@tilelang.autotune` only works with lazy mode kernels. This PR extends `@tilelang.autotune` to better support eager mode.

## Summary by CodeRabbit

* **New Features**
  * Added compile() for direct kernel compilation.
  * Introduced lazy/eager execution modes for autotuning.
  * Enhanced kernel cache to store both kernel and chosen configuration.
  * Autotuner now saves/loads an out_idx.json to preserve selected output index.

* **Tests**
  * Added end-to-end autotuning tests for a tiled matrix-multiply (including compile path and symbolic/dynamic-dimension cases).

### DIFF
--- a/testing/python/autotune/test_tilelang_autotune_eager_mode.py
+++ b/testing/python/autotune/test_tilelang_autotune_eager_mode.py
@@ -16,14 +16,14 @@ def ref_program(A, B):
 
     Parameters
     ----------
-    A : numpy.ndarray
+    A : torch.Tensor
         The matrix with shape (M, K).
-    B : numpy.ndarray
+    B : torch.Tensor
         The matrix with shape (N, K).
 
     Returns
     -------
-    np.ndarray
+    torch.Tensor
         The result of A @ B.T, shape (M, N).
     """
     return A @ B.T
@@ -124,7 +124,7 @@ def test_autotune_matmul():
     Run the autotuning validation for the matmul kernel on a 1024x1024x1024 problem.
 
     This test constructs random CUDA tensors, autotunes the JIT-compiled block-level matrix-multiplication kernel,
-    executes it, and asserts the result matches a reference CPU implementation within tolerances.
+    executes it, and asserts the result matches a reference PyTorch implementation within tolerances.
     """
     run_autotune(1024, 1024, 1024)
 
--- a/testing/python/autotune/test_tilelang_autotune.py
+++ b/testing/python/autotune/test_tilelang_autotune.py
@@ -11,14 +11,14 @@ def ref_program(A, B):
 
     Parameters
     ----------
-    A : numpy.ndarray
+    A : torch.Tensor
         The matrix with shape (M, K).
-    B : numpy.ndarray
+    B : torch.Tensor
         The matrix with shape (N, K).
 
     Returns
     -------
-    np.ndarray
+    torch.Tensor
         The result of A @ B.T, shape (M, N).
     """
     return A @ B.T
--- a/testing/python/autotune/test_tilelang_autotune_with_inputs.py
+++ b/testing/python/autotune/test_tilelang_autotune_with_inputs.py
@@ -16,14 +16,14 @@ def ref_program(A, B):
 
     Parameters
     ----------
-    A : numpy.ndarray
+    A : torch.Tensor
         The matrix with shape (M, K).
-    B : numpy.ndarray
+    B : torch.Tensor
         The matrix with shape (N, K).
 
     Returns
     -------
-    np.ndarray
+    torch.Tensor
         The result of A @ B.T, shape (M, N).
     """
     return A @ B.T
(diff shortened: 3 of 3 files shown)


tilelang/autotuner/param.py
```diff
@@ -378,6 +379,20 @@ def save_to_disk(self, path: Path, verbose: bool = False):
             logger.debug(f"Saving function to file: {path / FUNCTION_PATH}")
         self._safe_write_file(str(path / FUNCTION_PATH), "wb", lambda f: cloudpickle.dump(self.func, f))
 
+        # save out idx (atomic)
```
SiriusNEO, line 382: Is it necessary that we should save `out_idx` in a seperate file? And in my understanding, the out idx info should not be conditioned with `verbose`. Usually, we use `verbose` when we have some debug information (like `latency.json`) to dump.

tilelang/autotuner/param.py
```diff
@@ -421,6 +436,12 @@ def load_from_disk(cls, path: Path, compile_args: CompileArgs) -> AutotuneResult
         with open(path / FUNCTION_PATH, "rb") as f:
             func = cloudpickle.load(f)
 
+        # load out idx
+        if verbose:
+            logger.debug(f"Loading out idx from file: {path / OUT_IDX_PATH}")
+        with open(path / OUT_IDX_PATH) as f:
```
SiriusNEO, line 442: Maybe we need to handle FileNotFoundError?

tilelang/autotuner/tuner.py
```diff
@@ -680,21 +682,64 @@ def get_tunner(self):
...
-        key_args_tuple = args
-        key_kwargs_tuple = tuple(sorted(kwargs.items()))
-        key = (key_args_tuple, key_kwargs_tuple)
+    def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> JITKernel | _T:
+        return_kernel = kwargs.pop("__return_kernel", False)
+
+        mode = self.jit_impl.initialize_jit_mode(*args, **kwargs)
+
+        def _normalize_for_key(x):
+            import torch
```
SiriusNEO: No need to import torch here. We have imported it in the start of the file.

tilelang/autotuner/tuner.py
```diff
@@ -680,21 +682,64 @@ def get_tunner(self):
...
+
+            if isinstance(x, torch.Tensor):
```
SiriusNEO: Maybe we can abstract a common normalize function together with another closure `_normalize_param` in the outer side?

tilelang/autotuner/param.py
ColmaLiu: The reason I stored it separately is that eager mode relies on `PrimFunc.out_idx_override`, and I verified this runtime-attached attribute is lost after cloudpickle load. So I wanted to persist it explicitly instead of coupling it to function.pkl behavior.
Also, verbose only guards the log message here, not the actual save behavior. I followed the existing cache-file pattern from #683, where `verbose=True` is used to print save/load locations for cache artifacts.

tilelang/autotuner/param.py
ColmaLiu: I followed the current load_from_disk behavior here: the existing code assumes the cache entry is complete once the cache directory exists, and other cache files are also loaded without per-file FileNotFoundError handling.
So this change keeps the same assumption and behavior as the existing logic. If you prefer, I can also add explicit handling for incomplete cache entries in this PR.

tilelang/autotuner/tuner.py
```diff
@@ -342,7 +342,9 @@ def run(self, warmup: int = 25, rep: int = 100, timeout: int = 30):
                 extra_parameters[var_name] = cell.cell_contents
 
         if isinstance(self.configs, Callable):
-            self.configs = self.configs(*self._kernel_parameters)
+            kernel_args, kernel_kwargs = self._kernel_parameters
+            kernel_kwargs = dict(kernel_kwargs)
+            self.configs = self.configs(*kernel_args, **kernel_kwargs)
```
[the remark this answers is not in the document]
ColmaLiu: fixed in new commits

tilelang/autotuner/tuner.py
```diff
@@ -680,21 +682,64 @@ def get_tunner(self):
...
+                return ("tensor", str(x.dtype), tuple(x.shape))
+            if isinstance(x, (list, tuple)):
+                return tuple(_normalize_for_key(v) for v in x)
+            if isinstance(x, dict):
+                return tuple(sorted((k, _normalize_for_key(v)) for k, v in x.items()))
+            return x
+
+        norm_args = _normalize_for_key(args)
+        norm_kwargs = tuple(sorted((k, _normalize_for_key(v)) for k, v in kwargs.items()))
+        key = (norm_args, norm_kwargs)
```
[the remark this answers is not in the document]
ColmaLiu: Tensor strides is included in `_normalize_value` logic.

testing/python/autotune/test_tilelang_autotune_eager_mode.py
```diff
@@ -0,0 +1,141 @@
...
+    torch.testing.assert_close(c, ref_c, rtol=1e-2, atol=1e-2)
+
+
+def test_autotune_matmul():
+    """
+    Run the autotuning validation for the matmul kernel on a 1024x1024x1024 problem.
+
+    This test constructs random CUDA tensors, autotunes the JIT-compiled block-level matrix-multiplication kernel,
+    executes it, and asserts the result matches a reference CPU implementation within tolerances.
+    """
```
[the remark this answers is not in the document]
ColmaLiu: The misleading docstring is fixed.

SiriusNEO (approved):

PR #1917, opened 2026-03-09
 [Enhancement] Only skip parallel loop partitioning when all stores are to local buffers

LJC00118:
## Summary by CodeRabbit

* **Refactor**
  * Streamlined loop-parallelization and lowering decisions by consolidating non-local store detection and vectorization checks. Loops now only require partitioning when non-local stores are present, and vectorization/lowering paths are chosen more consistently. This reduces edge-case behavior, improves predictability for loop-heavy workloads, and simplifies maintenance.


SiriusNEO:
@regression-perf

Issue #533, opened 2025-05-28
[Feature Request] Support inputs to be a list of tensors


smallscientist1:
How do I implement a kernel with the arguments a list of tensors?

When implementing the grouped gemm kernel, I want the weight input to be a list of tensors. 
I tried to implement the kernel in this way: construct a tensor of data_ptr() as input and then use `T.make_tensor` to reconstruct the tensor in the kernel.
```py
    @T.prim_func
    def kernel(
            A: T.Tensor([batch_sum, K], dtype),  # type: ignore
            B_ptr: T.Tensor([batch_count], "int64"), # type: ignore
            # B: T.Tensor([batch_count, K, N], dtype),  # type: ignore
            C: T.Tensor([batch_sum, N], dtype),  # type: ignore
    ):
        # put here also cause an error
        # B_tensor_0 = T.make_tensor(B_ptr[0], [K, N], dtype=dtype)
        with T.Kernel(
                T.ceildiv(batch_sum, block_M) + batch_count, T.ceildiv(N, block_N),
                threads=threads) as (bx, by):
            B_tensor_0 = T.make_tensor(B_ptr[0], [K, N], dtype=dtype)
```

However, it encountered a compilation error.

a1600012888:
"Inputs to be a list of tensors"

I also want to have such feature. 

Any updates on this?

LeiWang1999:
closed as now resolved.

PR #1923, opened 2026-03-11
Support ptr-table grouped GEMM kernels

LeiWang1999:
## Summary
- add frontend support for pointer-table tensor annotations via `T.Tensor([G], T.ptr)` and keep the runtime ABI lowered to `int64`
- add a ptr-table grouped GEMM example that reconstructs tensors with `T.make_tensor(...)` instead of packing per-group tensors first
- improve compiler handling around ptr-backed kernels by fixing block-uniform thread sync hoisting and hardening software-pipeline buffer access rewriting
- add regression tests for pointer-table language features, grouped matmul, and thread-sync behavior

## Details
This PR makes it possible to write kernels like:

```python
@T.prim_func
def main(src_ptrs: T.Tensor([G], T.ptr), out: T.Tensor([G, N], dtype)):
    with T.Kernel(G, threads=1) as bx:
        Src = T.make_tensor(src_ptrs[bx], (N,), dtype)
        for i in T.serial(N):
            out[bx, i] = Src[i]
```

and use the same pointer-table pattern in grouped GEMM without first concatenating or stacking the source tensors into a packed buffer.

The grouped GEMM example now maps block tiles to groups through a tile-prefix table and uses `T.copy` on ptr-backed tensors. For now the ptr path is intentionally kept on a single-stage pipeline because multi-stage software pipelining on ptr-backed tensors is not correct yet.

## Performance
Kernel-only measurements on the updated ptr grouped GEMM path versus the original packed implementation:

- `batch_sizes=[64, 128, 256], K=N=4096`: `0.0843 ms` packed vs `0.0863 ms` ptr (`1.02x`)
- `batch_sizes=[63, 77, 111, 280], K=N=4096`: `0.1209 ms` packed vs `0.1553 ms` ptr (`1.28x`)
- `batch_sizes=[16, 33, 64, 96, 127], K=2048, N=4096`: `0.0375 ms` packed vs `0.0431 ms` ptr (`1.15x`)

Compared with the earlier ptr prototype, this recovers most of the gap while preserving the no-pack API.

## Testing
- `./format.sh`
  - pre-commit formatting/checks passed
  - clang-tidy failed in the current environment with `dlpack.h: 'stddef.h' file not found`
- `python -m pytest testing/python/language/test_tilelang_language_ptr.py -q`
- `python -m pytest testing/python/transform/test_tilelang_transform_thread_sync.py -q`
- `python examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py --batch_sizes 8,13,21 --K 64 --N 96 --backend tvm_ffi`
- `python examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py --batch_sizes 8,13,21 --K 64 --N 96 --backend cython`

## Summary by CodeRabbit

* **New Features**
  * Added a new example demonstrating pointer-based grouped GEMM workflows with configurable batch sizes and tile dimensions.
  * Introduced `make_tensor_from_addr` API for constructing tensors directly from raw memory addresses.
  * Added pointer table test utilities for copy, multi-copy, and grouped matmul operations.

* **Tests**
  * Added validation tests for pointer table annotations and grouped matmul operations.
  * Added test for thread synchronization behavior with runtime-dependent conditions.

### DIFF
--- a/examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py
+++ b/examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py
@@ -1,0 +1,186 @@
+import argparse
+import math
+import time
+
+import torch
+
+import tilelang as tl
+import tilelang.language as T
+
+
+def make_ptr_table(tensors):
+    assert tensors, "pointer table requires at least one tensor"
+    device = tensors[0].device
+    return torch.tensor([tensor.data_ptr() for tensor in tensors], device=device, dtype=torch.int64)
+
+
+def torch_grouped_gemm_ptr(a_list, b_list):
+    assert len(a_list) == len(b_list), "A/B group count mismatch"
+    outputs = []
+    for a, b in zip(a_list, b_list):
+        assert a.shape[1] == b.shape[0], "incompatible GEMM shapes"
+        outputs.append(torch.matmul(a, b))
+    return outputs
+
+
+def grouped_gemm_ptr(batch_sizes_list, K, N, block_M, block_N, block_K, num_stages=2, threads=128, dtype=T.float16):
+    # Keep per-group tensors separate and pass them via pointer tables.
+    # We currently use a common max_M storage shape per group because
+    # ptr-backed tensors with runtime-varying shapes are not stable enough yet.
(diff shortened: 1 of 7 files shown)


LeiWang1999:
This is also a fix for #533

PR #1937, opened 2026-03-16
Fix predicated cp.async pipeline scheduling

LeiWang1999:
## Summary
- treat predicated `cp.async` conservatively during pipeline planning instead of assuming it fully overwrites the destination shared buffer
- skip producer-side first-stage statements when binding `cp.async` consumers and wait stages so fill/copy/commit are not misclassified as downstream consumers
- add a CUDA regression test that exercises `fill + predicated cp.async + wait` inside a pipelined loop

## Testing
- `python -m pytest testing/python/transform/test_tilelang_transform_pipeline_planning.py`
- `python debug/0316_pipeline/bug.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced pipeline scheduling correctness for GPU async copy operations, improving dependency tracking and stage ordering to ensure proper execution of dependent operations across pipeline stages.

* **Tests**
  * Added CUDA-specific test validating predicated async copy behavior and runtime output correctness.

### DIFF
--- a/src/transform/pipeline_planning.cc
+++ b/src/transform/pipeline_planning.cc
@@ -357,6 +357,14 @@ private:
         }
       }
       if (args.size() == 4) {
+        // Predicated cp.async should not be treated as a proven full
+        // overwrite of the destination buffer at pipeline-planning
+        // granularity. Model a conservative dependence on the destination so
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #1943, opened 2026-03-18
[Warp Specialization] Move producer into the first warp group

LeiWang1999:
as title.

## Release Notes

* **Bug Fixes**
  * Improved thread indexing and boundary condition handling for more accurate partition management.
  * Refined guard conditions and thread remapping logic to ensure correct behavior in edge cases.

* **Tests**
  * Updated test suite to validate thread scheduling logic and boundary conditions.


LeiWang1999:
@regression-perf

LeiWang1999:
closed, we should first move the ws pipeline before the layout inference stage.

Issue #1936, opened 2026-03-16
[BUG] can not view fp4 tensor to uint8


dcaox:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.8

### System information

3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] linux
0.1.8
2.10.0a0+b558c986e8.nv25.11

### Problem description

An FP4 fragment or shared memory buffer cannot be reinterpreted as a uint8 tensor with half the number of elements. In other words, it is invalid to treat two FP4 elements as a single uint8 by simply creating a view with shape [rows_per_cta, mask_k // 2].

for example：
```
# qx_2d_frag_fp4 = T.alloc_fragment((rows_per_cta, mask_k), T.float4_e2m1fn)
# qx_2d_frag_fp4_uint8_view = T.view(qx_2d_frag_fp4, shape=[rows_per_cta, mask_k//2], dtype=T.uint8)

qx_2d_shared_fp4 = T.alloc_shared((rows_per_cta, mask_k), T.float4_e2m1fn)
qx_2d_shared_uint8_view = T.view(qx_2d_shared_fp4, shape=[rows_per_cta, mask_k//2], dtype=T.uint8)
```

### Reproducible example code

The Python snippets:

```python
import tilelang
import tilelang.language as T
import torch
from tilelang.autotuner import set_autotune_inputs

@tilelang.jit(out_idx=[-1,])
def quant_by_mask(
    K, block_size, inp_type, fake_quant_type, s_type
):
    min_k_dim_alignment = block_size * 4
    mask_m = 128
    mask_k = 256
    threads = 256
    elem_per_thread = 16
    threads_per_row = mask_k // elem_per_thread
    rows_per_cta = threads // threads_per_row
    block_per_row = mask_k // block_size

    assert mask_m % rows_per_cta == 0, "mask_m must be multiple of rows_per_cta"
    assert block_size == 32, "block_size must be 32"
    assert K % min_k_dim_alignment == 0, "K must be multiple of MinKDimAlignment"
    assert s_type == T.float8_e8m0fnu, "s_type must be float8_e8m0fnu"
    assert fake_quant_type == T.uint8, "fake_quant_type must be uint8"

    M = T.symbolic("M")

[...]

PR #1947, opened 2026-03-19
Support packed subtype views during layout reshape

LeiWang1999:
## Summary

This fixes dtype-changing alias views on packed sub-byte dtypes such as `float4_e2m1fn -> uint8`.

The main issue was that layout inference still reasoned in bytes, which breaks as soon as two logical elements share the same byte. On top of that, reshape would flatten packed sub-elements and lose the pack-lane information entirely.

This patch:
- switches alias layout propagation from bytes to storage bits
- teaches `Layout::Reshape` and `Fragment::Reshape` to preserve a trailing pack lane for packed subtype views
- adds coverage for `T.view` on `float4_e2m1fn -> uint8`, including a shared-memory compile/run case

## Validation

- `python debug/0319_issue/reproduce.py`
- `python -m pytest testing/python/language/test_tilelang_language_view.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reshape behavior for sub-byte element types, preserving packed-storage semantics when changing element bitwidths (automatic pack-lane handling).
  * More robust dtype-reinterpretation views (e.g., fp4↔uint8) with consistent physical storage preservation.

* **Documentation**
  * Clarified that reshape rescaling uses bit-level (per-element) storage footprint and notes pack-lane implications.

* **Tests**
  * Added tests validating sub-byte view reinterpretation and a CUDA-path compile/run for fp4→uint8 views.

### DIFF
--- a/src/layout/layout.cc
+++ b/src/layout/layout.cc
@@ -101,7 +101,8 @@ Array<PrimExpr> RestoreInputPlaceholders(const Array<PrimExpr> &forward_index,
   return restored;
 }
 
-PrimExpr RestoreInputPlaceholders(const PrimExpr &expr, const Array<Var> &vars) {
+PrimExpr RestoreInputPlaceholders(const PrimExpr &expr,
+                                  const Array<Var> &vars) {
   PrimExpr restored = expr;
   for (size_t i = 0; i < vars.size(); ++i) {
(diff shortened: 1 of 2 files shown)


LeiWang1999:
this is a fix for #1936

LeiWang1999:
@regression-perf

Issue #1844, opened 2026-02-13
[Feature Request] Detailed control on common functions


bucket-xv:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

This is not an emergency issue, but it should be gracefully handled in the long run.
- `T.max/min` may support an optional flag to describe details on comparing values containing `NaN`. You can find functions `__hmax` and `__hmax_nan`(https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____HALF2__COMPARISON.html#group__cuda__math____half2__comparison_1ga59fc7fc7975d8127b202444a05e57e3d). This is also applicable to `T.reduce_max/min`.
- `T.cast` may support an optional flag to describe the rounding behavior (Round to nearest, with even to tie? Round to inf? Round to zero?）and yet another optional flag to describe the safinite behavior (safinite or allow overflow to inf). You may also find relevant details in https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH____BFLOAT16__MISC.html#group__cuda__math____bfloat16__misc_1ga7439bee482ef173dd62f1ef88c8774c9, which provides a list of casting bf16 -> fp32 functions (Just for example).

bucket-xv:
One more possible option for `T.cast`: stochastic cast with supplied random bits. Refer to https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-cvt

haoran35-jpg:
Hey is it possible that I can contribute a PR for this issue? I saw NaN functionality is only for half(float16) in legacy, and I'm thinking to create a new op MaxOpNan, lowering to three cases: cutlass::fast_max(x,y) for general, __hmax(x,y) for bf16 and __hmax_nan(x,y) for half(float16). Then TVM can decide whether to propagate Nan in scalar Max/Min with codegen_cuda.cc and AllReduce in reduce.cc. I'm a rookie for Tilelang project so I'm not sure if this seems like a plan. Feel free to correct me if I'm wrong!

bucket-xv:
Well, nan including comparison applies to bfloat16 and half. You can see the following ref:
[...]

PR #1958, opened 2026-03-21
Nan propagation option for bf16 and half16

haoran35-jpg:
This PR introduces option to propagate NaN specifically for data type bf_16 and half, which modifies the following files:
|file changed | implementation |
|---|---|
 `src/op/builtin.h` | Declares `kReduceMaxMinNanPropagate` and `tl::max_nan` / `tl::min_nan` ops. |
| `src/op/builtin.cc` | Registers TIR intrinsics **`tl.max_nan`** / **`tl.min_nan`**. |
| src/op/reduce.cc | TIR layer: For `AllReduce`, `MakeCodegenReducer() `returns `tl::MaxOpNan`, concatenate into `tl::AllReduce<tl::MaxOpNan, …>::run(...) with tl::MaxOpNan `defined in rumtime template reduce.h. For `MakeReduce`, turn into CallNode then rely on `tl::max_nan/tl::min_nan`.|
| src/target/codegen_cuda.cc| VisitExpr_(MaxNode/MinNode) choose between `__hmax/__hmin` and  `__hmax_nan/__hmin_nan` according to Passcontext. VisitExpr_(CallNode) handles tl::max_nan / tl::min_nan as TVM op defined by buildin.|
| src/op/reduce.h | Define runtime template for MaxOpNan and MinOpNan |
| tilelang/transform/pass_config.py | Python layer: a PassConfigKey button to open NaN propagation functionality. |
| testing/python/language/test_tilelang_language_reduce_maxmin_nan.py | A test file for nan functionality. |

Before this PR, MaxOp/MinOp with parameter type bfloat16 and half will lower to __hmax()/__hmin() forcefully. Now user can optionally lower them to _hmax_nan()/__hmin_nan() by setting kReduceMaxMinNanPropagate to false(default is true). I feel this global config is more elegant and convenient to implement, not sure if this is okay. 
Disclaim: AI is used in code refinement and test generation.

## New Features
- Added `nan_propagate` parameter to `reduce_max()`, `reduce_min()`, and `reduce_absmax()` functions, enabling NaN-aware reduction behavior for fp16/bf16 tensors on CUDA targets.

## Tests
- Added comprehensive test suite validating NaN-propagation semantics and code generation for reduction primitives.


haoran35-jpg:
@https://github.com/tile-ai/tilelang/issues/1844

LeiWang1999:
Thanks @haoran35-jpg , would you mind explain more for introducing `kReduceMaxMinNanPropagate `? sorry I didn't get the point.

haoran35-jpg:
Yes, thank you for comment! It's allowing max(...)/min(...) in frontend to be lowered to Nvidia intrinsic hmax_nan(...)/hmin_nan(...) when setting `kReduceMaxMinNanPropagate` to true, and those two functions return `nan` when either of the parameters are NaN. This is different from hmax(...)/hmin(...) intrinsic where `nan` is returned only when both parameters are NaN. This could make it more convenient and precise to study NaN propagation. I'm still a rookie trying to learn this project, so I'm not sure if this is the best way to do. Thank you for your time!

LeiWang1999:
@haoran35-jpg, Thanks and I think one better solution is to add an annotation for `reduce` tile op (in frontend have sth like `T.reduce_max(nan_propagate=True)`) instead of introducing a pass config.

PR #1962, opened 2026-03-23
[Bugfix] Fix double buffer versioning when TMA is used without warp specialization

LeiWang1999:
## Summary

- When TMA is enabled but warp specialization is disabled, `MultiVersionBuffer` and `InjectSoftwarePipeline` both independently added version dimensions to shared data buffers, resulting in **4x** buffer allocation instead of the intended **2x**.
- Add a `barrier_only` mode to `MultiVersionBuffer`. In the no-WS TMA path, it now only versions barrier buffers (`shared.barrier` scope) for mbarrier parity rewriting, while data buffer versioning is left to `InjectSoftwarePipeline`. The WS path remains unchanged.
- Flatten the `if allow_tma_lower / else` branches in `OptimizeForTarget` into a single linear flow with a three-way branch (WS / TMA-noWS / noTMA), removing the duplicated `LowerOpaqueBlock` call.

## Test plan

- [x] All existing transform tests pass (`test_multi_version_buffer`, `test_pipeline_planning`, `test_Inject_software_pipeline`)
- [x] End-to-end correctness verified on all three paths: WS, TMA+no-WS, no-TMA
- [x] Shared memory usage confirmed at 2x (correct) instead of 4x (broken)

## Summary by CodeRabbit

* **New Features**
  * Exposed an option to enable barrier-only buffer versioning for finer optimization control.

* **Improvements**
  * Compiler now records and detects on-target memory-acceleration to choose different buffer-versioning and pipeline paths.
  * Reordered and unified shared-barrier and software-pipeline stages; refined gating for warp-specialization and synchronization rewrites.
  * Barrier-arrival fusion and target-specific synchronization rewrites are applied more selectively.

* **Tests**
  * Replaced string-based kernel checks with IRModule-based tests and AST-level assertions for more robust validation.

### DIFF
--- a/testing/python/jit/test_tilelang_jit_cutedsl.py
+++ b/testing/python/jit/test_tilelang_jit_cutedsl.py
@@ -361,59 +361,6 @@ def test_cutedsl_dynamic_shape():
     )
 
 
-def run_cutedsl_barrier(
-    M,
-    N,
-    K,
-    block_M,
-    block_N,
-    block_K,
-    mbars,
-    trans_A,
-    trans_B,
-    in_dtype,
-    out_dtype,
-    accum_dtype,
-    num_stages,
-    threads,
-):
-    program = matmul_kernel_with_barrier(
-        M,
-        N,
-        K,
-        block_M,
-        block_N,
-        block_K,
-        mbars,
-        trans_A,
-        trans_B,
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

PR #1970, opened 2026-03-24
[Feature] Introduce T.CUDASourceCodeKernel

SiriusNEO:
This PR introduces `T.CUDASourceCodeKernel`, which is another approach in TileLang to launch a GPU kernel. Instead of expressing the device-side body through TileLang DSL, users can provide CUDA source directly, either as inline code or via a source file path, and launch it from a TileLang PrimFunc.

|  | T.CUDASourceCodeKernel(...) | with T.Kernel(...) |
  |---|---|---|
  | Programming model | Launches an external CUDA kernel directly | Builds the kernel body with the TileLang DSL |
  | Kernel body source | Provided as CUDA source code or a source file path | Written inline inside the with block |
  | Lowering path | Carries external source through IR attrs and injects it during CUDA codegen (**will ignore most TileLang TIR passes**) | Lowers the TileLang DSL body through the normal TileLang lowering pipeline |
  | Best use case | Reusing hand-written CUDA kernels or externally generated CUDA code | Writing new kernels directly in TileLang |
  | Flexibility for TileLang transformations | Limited by the fact that the device body is opaque external code | Higher, since the full kernel body is represented in TileLang IR |

Some valuable points of this new API:
- Users can embed CUDA code directly, and it is compiled **under the same environment** as a regular TileLang kernel. This means the CUDA code can also freely use utilities and headers from `tl_templates`.
- Users can debug the TileLang generated kernel in source level, by first using `kernel..export_sources` and importing the CUDA through `T.CUDASourceCodeKernel`.
- Comparing with exisiting embedding methods (`T.import_source` or `tilelang_callback_cuda_postproc`), this approach is more elegant, more maintainable, and better aligned with TileLang's standard lowering and codegen pipeline.
-This API introduces a more explicit host-driven launch abstraction, which can be a precursor to HostDSL (an early building block for making TileLang a host-side control layer over both TileLang-generated kernels and externally authored device kernels.)

Implementation Sketch:

image

## Summary by CodeRabbit

* **New Features**
  * Add CUDASourceCodeKernel to embed external CUDA source (inline or file) with optional explicit entry selection; external source can be emitted verbatim and mixed with regular kernels.
  * Device ABI extraction can derive parameter order from external CUDA kernels for correct host/device wiring.

* **Bug Fixes**
  * Validate selected CUDA entry names and preserve user-specified device symbol names.
  * Detect and report duplicate CUDA kernel symbols.

* **Tests**
  * New tests for inline/file kernels, explicit entry selection, ABI tracking, mixed kernels, and duplicate-entry rejection.


src/transform/split_host_device.cc
```diff
@@ -104,8 +112,109 @@ class HostDeviceSplitter : public tir::StmtMutator {
...
+    std::vector<tir::Var> params;
+    std::unordered_set<std::string> seen_vars;
+
+    auto push = [&](const tir::Var &var) {
+      if (var.defined() && seen_vars.insert(var->name_hint).second) {
+        params.push_back(var);
+      }
+    };
+
+    // for (const auto &param : host_params_) {
```
LeiWang1999: useless comment?

LeiWang1999 (dismissed):

PR #1973, opened 2026-03-25
Refactor producer-consumer WS access tracking for WGMMA-local state

LeiWang1999:
## Summary
- split producer-consumer WS access summaries into all-buffer accesses vs branch-private accesses
- teach the access collector to recognize raw `buffer.data` operands used by `warpgroup_fence_operand`, `ptx_wgmma_ss`, and `ptx_wgmma_rs`
- make loop-prefix hoisting use branch-private liveness so consumer-private intermediates are not moved into the producer branch
- add a regression test that lowers the real `flashattn_bwd` kernel and checks that WGMMA stays in the consumer branch

## Motivation
The previous logic only tracked explicit buffer loads/stores and a narrow subset of pointer-style accesses. That was too weak for lowered Hopper WGMMA code, where local accumulator buffers are often passed through opaque calls as raw `buffer.data` vars.

As a result, consumer-private compute such as `qkT` and `dsT` could be misclassified as producer-safe prefix work and hoisted across the producer/consumer split, changing kernel semantics.

## Testing
- `python -m py_compile testing/python/transform/test_tilelang_transform_producer_consumer_ws.py`
- `python -m pytest testing/python/transform/test_tilelang_transform_producer_consumer_ws.py -k real_flash_bwd_wgmma -q`
- `PYTHONPATH=/weka-hg/prod/deepseek/permanent/wanglei/tilelang_ref python /weka-hg/prod/deepseek/permanent/wanglei/tilelang_ref/debug/0323_flex/test.py`

## Summary by CodeRabbit

* **Refactor**
  * Enhanced internal buffer tracking and dependency-aware producer-consumer loop hoisting to improve correctness and performance for complex kernel transformations.
* **Tests**
  * Added test helpers for runtime introspection and dynamic module loading to enable more robust validation of producer-consumer optimization scenarios.

### DIFF
--- a/testing/python/transform/test_tilelang_transform_producer_consumer_ws.py
+++ b/testing/python/transform/test_tilelang_transform_producer_consumer_ws.py
@@ -1,9 +1,5 @@
 # ruff: noqa
-import importlib.util
-from pathlib import Path
-
 from tilelang import tvm as tvm
-import tilelang
 import tilelang as tl
 import tilelang.language as T
 import tilelang.testing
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

PR #1975, opened 2026-03-25
Fix wrapped pre-loop TMA prefixes in producer-consumer WS

LeiWang1999:
## Summary
This PR fixes a producer-consumer warp-specialized corner case where a pipelined loop is wrapped by outer `if` guards and has a pre-loop TMA preload.

The change makes the WS rewrite:
- find and rebuild pipeline loops under `IfThenElse` wrappers
- recognize pre-loop TMA producer/wait pairs that have been split or wrapped by `MultiVersionBuffer`
- merge producer/consumer prefixes through non-thread guard wrappers so pre-loop TMA preloads land inside the WS producer branch

The PR intentionally keeps the original pure-TMA barrier grouping behavior to avoid an unnecessary performance regression. The correctness fix is limited to the wrapped pre-loop prefix placement and loop discovery/rebuild logic.

This fixes the softmax varlen repro where the LSE preload stayed outside `kWarpSpecializationScope`, causing the producer election and barrier protocol to run on the wrong thread partition.

## Testing
- `bash ./format.sh --files src/transform/producer_consumer_ws.cc testing/python/transform/test_tilelang_transform_producer_consumer_ws.py`
- `PYTHONPATH=/weka-hg/prod/deepseek/permanent/wanglei/tilelang_ref python -m pytest testing/python/transform/test_tilelang_transform_producer_consumer_ws.py -q`
- manual repro: `PYTHONPATH=/weka-hg/prod/deepseek/permanent/wanglei/tilelang_ref CUDA_LAUNCH_BLOCKING=1 python debug/0325_softmax/softmax_sum.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced support for pipelined loops nested within conditional statements.
  * Improved tensor memory accelerator operation handling in specialized kernel transformations.

* **Tests**
  * Added test coverage for conditional pipeline detection and memory operation relocation patterns.

### DIFF
--- a/src/transform/producer_consumer_ws.cc
+++ b/src/transform/producer_consumer_ws.cc
@@ -1207,13 +1207,12 @@ private:
             wait_insert_pos[i] == wait_insert_pos[i - 1] &&
             arrive_insert_pos[i] == arrive_insert_pos[i - 1] &&
             same_guard(i - 1, i);
-        if (merge_with_prev && current_group_has_tma &&
+        if (merge_with_prev && !remap_pure_tma_barriers_ &&
+            current_group_has_tma &&
             extractor.blocks[i].kind == AsyncProducerKind::kTma) {
-          // The current WS rewrite only supports one TMA producer per merged
-          // group. Reusing a forward barrier for multiple TMA producers would
-          // require group-level handling of:
-          // - mixed TMA/cp.async preserved-protocol arrive semantics; and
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #1976, opened 2026-03-26
[Feature] Batched AllReduce for better T.reduce performance

### LINKED ISSUE #1928: [Feature Request] Better `T.reduce` performance
### Required prerequisites

### Motivation

`T.reduce` behaves poorly when the layout requires each thread to hold multiple reduced values. See the program below. 

The fragment layout is satisfactory, since 8 * sizeof (bfloat16) = 16 bytes is an excellent reading width from shared memory. However, the `T.reduce` code does not leverage this advantage. Instead, it gets the elements from shared memory one by one and calculates the 8 max values sequentially. As we know,  `tl::AllReduce` leverages a butterfly path; there are many thread syncs in the path. This simple piece of code have 8 * log(256 / 32) = 24 thread syncs! I suggest calculating these max values in parallel.

One may argue that computing in parallel may cost more shared memory. However, if you see the issue https://github.com/tile-ai/tilelang/issues/1761, the issue suggests that each warp, instead of each thread, holds one copy of all the values to reduce in shared memory. The memory saved that way is just good for parallel reduction here!

```python
import tilelang

@tilelang.jit(
    pass_configs={
        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
    },
)
def get_sample_kernel():
    @T.prim_func
    def sample_kernel(a: T.Tensor([128, 64], T.bfloat16)):
        with T.Kernel(1, threads=256):
            a_shared = T.alloc_shared([128, 64], T.bfloat16)
            amax_fragment = T.alloc_fragment([64, ], T.bfloat16)
            T.copy(a, a_shared)
            T.reduce_max(a_shared, amax_fragment, dim=0)

    return sample_kernel

kernel = get_sample_kernel()
```

```C++
#include <tl_templates/cuda/gemm.h>
#include <tl_templates/cuda/copy.h>
#include <tl_templates/cuda/reduce.h>
#include <tl_templates/cuda/ldsm.h>
#include <tl_templates/cuda/threadblock_swizzle.h>
#include <tl_templates/cuda/debug.h>
#ifdef ENABLE_BF16
#include <tl_templates/cuda/cuda_bf16_fallbacks.cuh>
#endif

extern "C" __global__ void sample_kernel_kernel(const bfloat16_t* __restrict__ a);
extern "C" __global__ void __launch_bounds__(256, 1) sample_kernel_kernel(const bfloat16_t* __restrict__ a) {
  extern __shared__ __align__(1024) uchar buf_dyn_shmem[];
  bfloat16_t a_shared_frag[32];
  bfloat16_t amax_fragment[8];
  #pragma unroll
  for (int i = 0; i < 4; ++i) {
    *(uint4*)(((bfloat16_t*)buf_dyn_shmem) + ((i * 2048) + (((int)threadIdx.x) * 8))) = *(uint4*)(a + ((i * 2048) + (((int)threadIdx.x) * 8)));
  }
  #pragma unroll
  for (int i_1 = 0; i_1 < 4; ++i_1) {
    *(uint4*)(a_shared_frag + (i_1 * 8)) = *(uint4*)(((bfloat16_t*)buf_dyn_shmem) + ((i_1 * 2048) + (((int)threadIdx.x) * 8)));
  }
  __syncthreads();
  #pragma unroll
  for (int i_2 = 0; i_2 < 8; ++i_2) {
    amax_fragment[i_2] = -std::numeric_limits<bfloat16_t>::infinity();
    #pragma unroll
    for (int rv = 0; rv < 4; ++rv) {
      amax_fragment[i_2] = cutlass::bfloat16_t(__hmax((amax_fragment[i_2]).to_nv_bfloat16(), (a_shared_frag[((rv * 8) + i_2)]).to_nv_bfloat16()));
    }
    amax_fragment[i_2] = tl::AllReduce<tl::MaxOp, 256, 8, 0, tl::NamedBarrier<256>>::run(amax_fragment[i_2], (&(((bfloat16_t*)buf_dyn_shmem)[0])));
  }
}
```

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

kurisu6912:
## Summary

Resolves #1928. When each thread holds multiple values to reduce across warps, the current `AllReduce` serializes butterfly reductions with one call per element, causing excessive thread synchronizations (e.g., 48 syncs instead of 6 for 8 values × 256 threads).

This PR extends the `tl::AllReduce` C++ template with a batched interface that reduces all values in parallel, **sharing synchronization barriers** across all elements. This cuts thread sync count from `N × 2 × log₂(threads/32)` to just `2 × log₂(threads/32)`.

### Changes

- **`src/tl_templates/cuda/reduce.h`** — Add `batch_size` and `workspace_stride` template parameters (defaults preserve backward compatibility). Add batched `run(T *x, T *red_buf)` overload alongside existing scalar `run(T x, T *red_buf)`.
- **`src/tl_templates/hip/reduce.h`** — Same batched extension for the HIP/ROCm AllReduce template.
- **`src/op/reduce.cc`** — Restructure `ReduceOpNode::Lower()` to detect when batching is profitable (`batch_size > 1 && has_cross_warp_reduce`). Splits lowered code into 3 phases: pre-reduce loop, batched AllReduce call, and post-reduce copy-back.
- **`src/op/finalize_reducer.cc`** — Add batched AllReduce path to `FinalizeReducerOpNode::Lower()` with the same detection logic.
- **`testing/python/language/test_tilelang_language_reduce.py`** — Add `test_batched_allreduce_codegen` parametrized test that verifies the batched AllReduce template is emitted in generated CUDA source.

### Backward Compatibility

All template parameters use defaults (`batch_size=1`, `workspace_stride=0`), so existing scalar `AllReduce` call sites are unaffected. The warp-only path (≤32 threads) still uses the scalar interface since there are no shared-memory barriers to amortize.

## Test Plan

- [x] All 28 existing reduce tests pass
- [x] New `test_batched_allreduce_codegen` tests verify batched template instantiation in generated source
- [x] HIP/ROCm codegen path verified (no Barrier type parameter)

Closes #1928

## Summary by CodeRabbit

* **New Features**
  * Batched reduction support: perform multiple independent reductions in parallel.

* **Performance Improvements**
  * More efficient synchronized all-reduce for batched workloads, improving throughput for large/cross-warp reductions and optimizing synchronization per target.

* **Tests**
  * Added parameterized tests to validate batched all-reduce code generation and ensure batch sizes > 1 across configurations.


LeiWang1999 (dismissed):

kurisu6912 (commented):
LLM give a good implementation, but require more tests, I'll add more tests

LeiWang1999:
@regression-perf

PR #2010, opened 2026-04-03
[Example] Migrate eligible examples to eager style

ColmaLiu:
### Summary

This PR converts most examples in the TileLang repo that are suitable for eager style to eager style.

### Motivation

We currently have a mix of styles across examples. Converting eligible examples to eager style makes the example set more consistent, easier to read, and better aligned with the recommended usage pattern.

### Notes

Not all examples are migrated in this PR. Some are intentionally left unchanged because:

* eager style is not a good fit for them, or
* they depend on older TileLang versions that do not yet support eager style (for example, examples with dedicated `requirements.txt`)

## Summary by CodeRabbit

* **Refactor**
  * Simplified TileLang kernel interfaces across all example kernels to use tensor-first calling conventions. Kernels now accept input tensors directly (e.g., `Q, K, V`) instead of explicit shape parameters, with dimensions derived internally. Nested kernel definitions removed and inlined for clearer code structure. Output tensors allocated and returned directly from kernels rather than passed as arguments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2010?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


LeiWang1999:
@regression-perf

PR #2024, opened 2026-04-09
Re-enable deprecated `TL_DISABLE_TMA_LOWER` pass config for TMA store

LJC00118:
## Summary by CodeRabbit

* **Bug Fixes**
  * Temporarily re-enabled the global disable-TMA option so it again prevents auto-lowering of T.copy to TMA stores.

* **Documentation**
  * Clarified deprecation messaging and added an explicit removal target (v0.1.10); advises using per-copy disable_tma=True instead.


LJC00118:
@regression-perf

PR #2031, opened 2026-04-11
[Bugfix] Fix stage-expanded annotated-layout aliases in LayoutInference

TerminusAkivili:
## Summary

Fix `LayoutInference` annotated layout propagation for sibling aliases that only add leading stage dimensions and have matching storage bits.

For `[64, 64] -> [2, 64, 64]`, the layout should be expanded over the leading stage dimension instead of going through the generic `Reshape(...)` fallback.

## Details

`Expand(...)` is used only when:

- storage bits match
- the sibling buffer has additional leading dimensions
- the trailing dimensions are provably equal to the original layout input shape

Exact-shape buffers still reuse the original layout directly, and all other non-exact shape cases continue to use the existing `Reshape(...)` fallback.

## Tests

- Added a regression test for stage-expanded shared-buffer aliases with identity and swizzled layouts.
- Ran `python -m pytest testing/python/transform/test_tilelang_transform_layout_inference.py -q`.
- Ran `python -m pytest testing/python/layout/test_tilelang_layout_inference.py -q`.
- Verified `examples/gemm_fp8/example_tilelang_gemm_fp8_intrinsic.py`.
- Also verified a minimized manual TIR reproducer for `[64, 64] -> [2, 64, 64]`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved layout inference for aliased buffers: better handling when shapes differ (including added leading dimensions) to avoid incorrect reshapes and to prefer shape-matching annotated layouts.

* **Refactor**
  * Centralized and simplified alias-layout compatibility logic for more consistent inference across annotated and propagated layouts.

* **Tests**
  * Added test helpers and parametrized tests covering expanded annotated layout aliases with identity and swizzled layouts.

### DIFF
--- a/src/transform/layout_inference.cc
+++ b/src/transform/layout_inference.cc
@@ -62,8 +62,8 @@ bool ShapesEqual(const Array<PrimExpr> &lhs, const Array<PrimExpr> &rhs,
 }
 
 Optional<Buffer> FindLayoutAnchorBuffer(const Array<Buffer> &buffers,
-                                       const Layout &layout,
-                                       arith::Analyzer *analyzer) {
+                                        const Layout &layout,
+                                        arith::Analyzer *analyzer) {
   for (const auto &buffer : buffers) {
     if (ShapesEqual(layout->InputShape(), buffer->shape, analyzer)) {
       return buffer;
@@ -764,9 +764,10 @@ private:
         ICHECK(!buffers.empty()) << "buffer list for " << var << " is empty";
         Optional<Buffer> anchor_buffer =
             FindLayoutAnchorBuffer(buffers, layout, &analyzer_);
-        int64_t anchor_bits = anchor_buffer.defined()
-                                  ? GetElementStorageBits(anchor_buffer.value()->dtype)
-                                  : GetElementStorageBits(buffers[0]->dtype);
+        int64_t anchor_bits =
+            anchor_buffer.defined()
+                ? GetElementStorageBits(anchor_buffer.value()->dtype)
+                : GetElementStorageBits(buffers[0]->dtype);
         // Apply layout to all buffers associated with this var
         for (const auto &buffer : buffers) {
 
@@ -779,8 +780,7 @@ private:
             annotated_layout_map_.Set(buffer, layout);
           } else {
             auto reshaped_layout =
-                layout->Reshape(buffer->shape, &analyzer_,
-                                Integer(anchor_bits),
+                layout->Reshape(buffer->shape, &analyzer_, Integer(anchor_bits),
                                 Integer(GetElementStorageBits(buffer->dtype)));
             annotated_layout_map_.Set(buffer, reshaped_layout);
           }
--- a/src/transform/layout_inference.cc
+++ b/src/transform/layout_inference.cc
@@ -48,6 +48,30 @@ int64_t GetElementStorageBits(DataType dtype) {
   return static_cast<int64_t>(dtype.bits()) * dtype.lanes();
 }
 
+bool ShapesEqual(const Array<PrimExpr> &lhs, const Array<PrimExpr> &rhs,
+                 arith::Analyzer *analyzer) {
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+  for (size_t i = 0; i < lhs.size(); ++i) {
+    if (!analyzer->CanProveEqual(lhs[i], rhs[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+Optional<Buffer> FindLayoutAnchorBuffer(const Array<Buffer> &buffers,
+                                       const Layout &layout,
+                                       arith::Analyzer *analyzer) {
+  for (const auto &buffer : buffers) {
+    if (ShapesEqual(layout->InputShape(), buffer->shape, analyzer)) {
+      return buffer;
+    }
+  }
+  return Optional<Buffer>();
+}
+
 } // namespace
 
 /*!
@@ -738,31 +762,25 @@ private:
             << "buffer " << var << " is not found in the block";
         const auto &buffers = buffer_data_to_buffers_[var];
         ICHECK(!buffers.empty()) << "buffer list for " << var << " is empty";
+        Optional<Buffer> anchor_buffer =
+            FindLayoutAnchorBuffer(buffers, layout, &analyzer_);
+        int64_t anchor_bits = anchor_buffer.defined()
+                                  ? GetElementStorageBits(anchor_buffer.value()->dtype)
+                                  : GetElementStorageBits(buffers[0]->dtype);
         // Apply layout to all buffers associated with this var
         for (const auto &buffer : buffers) {
 
           // Reshape the layout to match the buffer's shape
           // Check if shapes are structurally equal
           bool shapes_equal =
-              layout->InputShape().size() == buffer->shape.size();
-          if (shapes_equal) {
-            for (size_t i = 0; i < layout->InputShape().size(); ++i) {
-              if (!analyzer_.CanProveEqual(layout->InputShape()[i],
-                                           buffer->shape[i])) {
-                shapes_equal = false;
-                break;
-              }
-            }
-          }
+              ShapesEqual(layout->InputShape(), buffer->shape, &analyzer_);
 
           if (shapes_equal) {
             annotated_layout_map_.Set(buffer, layout);
           } else {
-            // Use the first buffer sharing this var as the base for dtype ratio
-            int64_t base_bits = GetElementStorageBits(buffers[0]->dtype);
-
             auto reshaped_layout =
-                layout->Reshape(buffer->shape, &analyzer_, Integer(base_bits),
+                layout->Reshape(buffer->shape, &analyzer_,
+                                Integer(anchor_bits),
                                 Integer(GetElementStorageBits(buffer->dtype)));
             annotated_layout_map_.Set(buffer, reshaped_layout);
           }
--- a/testing/python/language/test_tilelang_language_view.py
+++ b/testing/python/language/test_tilelang_language_view.py
@@ -124,5 +124,32 @@ def test_view_shared_fp4_to_uint8_compile():
     assert output.dtype == torch.uint8
 
 
+def annotated_layout_on_dtype_changing_view_test():
+    @T.prim_func
+    def main(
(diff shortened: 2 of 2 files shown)


TerminusAkivili:
pre-commit.ci autofix

TerminusAkivili:
A buffer gained a leading stage,
A layout stepped beyond its page.
We taught it when to stretch, not bend,
And made both alias paths amend.

The checks are green, the diff is small—
Thanks, maintainers, for reviewing it all.

TerminusAkivili:
Update:

- Narrowed the fix to the explicit `annotate_layout(...)` path only.
- Kept generic alias propagation and alias completion reshape-only.
- Preserved expand/collapse compatibility only when materializing explicit annotated-layout aliases.

This keeps the original bugfix for stage-expanded annotated layouts while avoiding broader alias semantic changes in unrelated paths.

Additional CI note:

- Metal: passed in the previous run.

- CUDA: the previous remaining failures looked host-side (`export_library -> g++`, returncode `-15`) rather than `LayoutInference` regressions.

- ROCm: the old full log showed two separate signals:
  - `A_shared is 3-dimensional, cannot be indexed with 2-dimensional indices` matched the overly broad generic alias expansion that this narrowing is intended to fix.
  - `tl::rasterization2DRow(10)` was a separate HIP swizzle codegen issue from the old base.

- This branch is now based on the latest `main`, which already includes the HIP swizzle codegen fix, so the old `rasterization2DRow(10)` ROCm failures should no longer apply here.

- The key rerun signal on the latest commit is therefore whether the ROCm `A_shared` failures disappear.

LeiWang1999:
Thanks @TerminusAkivili for your contribution! The layout of a pipelined buffer with annotated layout will be expanded during the inject pipeline pass. For this PR, I believe we're aiming to handle a more complex scenario involving T.view (though a buffer derived from T.view should share the same layout as its raw buffer). My question is: what specific kernel scenario necessitates this fix? Could it be that the frontend is generating something invalid in the first place?

TerminusAkivili:
Thanks @LeiWang1999  for the question!
This was motivated by an actual failure I hit in a pipelined GEMM-style kernel, where a shared buffer gets stage-expanded by the inject-pipeline pass, while a same-storage alias/view is still the object carrying the explicit `annotate_layout(...)`.

A simplified shape is:
`A_stage = T.alloc_shared((num_stages, M, K), ..., scope="shared.dyn")`
`A_view = T.Buffer((M, K), ..., data=A_stage.data, scope="shared.dyn")`

and the annotated layout may be attached to either `A_stage` or `A_view`.

After pipeline injection, `A_stage` is expanded with the stage dimension, while `A_view` still represents the same storage with the pre-expanded logical shape. So my understanding is that this is still valid IR: they should share the same underlying storage layout, but their logical input shapes are not literally identical and need to remain compatible modulo the stage dimension.

Because of that, I treated this as a `LayoutInference` issue rather than assuming the frontend had already produced invalid IR. That was the motivation for narrowing the change to the explicit `annotate_layout(...)` materialization path only, while keeping generic alias propagation / alias completion reshape-only.

I also added regressions for both cases:
annotating the stage-expanded buffer directly
annotating the same-storage alias/view
If there is a stricter intended invariant for this pattern, I’m happy to adjust the patch and tests accordingly.

LeiWang1999:
Thanks @TerminusAkivili , I tried:

```python
import tilelang
import tilelang.language as T


# @tilelang.jit(target="cuda")
# target currently can be "cuda" or "hip" or "cpu".
# if not specified, it will be inferred from the input tensors during compile time
@tilelang.jit
def matmul(M, N, K, block_M, block_N, block_K, dtype=T.float16, accum_dtype=T.float32):
    @T.prim_func
    def matmul_relu_kernel(
        A: T.Tensor((M, K), dtype),
        B: T.Tensor((K, N), dtype),
        C: T.Tensor((M, N), dtype),
        A_plus_one: T.Tensor((M, K), dtype),
    ):
        # Initialize Kernel Context
            A_shared_view = T.alloc_shared((block_M, block_K), dtype)
            B_shared = T.alloc_shared((block_K, block_N), dtype)
            C_local = T.alloc_fragment((block_M, block_N), accum_dtype)

            T.annotate_layout({
                A_shared: tilelang.layout.make_swizzled_layout(A_shared)
            })

            T.clear(C_local)

            T.copy(A[by * block_M, 0], A_shared)
            T.copy(B[0, bx * block_N], B_shared)

            for i, k in T.Parallel(block_M, block_K):
                A_shared_view[i, k] = A_shared[i, k] + T.cast(1, dtype)

            T.copy(A_shared_view, A_plus_one[by * block_M, 0])

                C_local[i, j] = T.max(C_local[i, j], 0)

            # Copy result back to global memory
            T.copy(C_local, C[by * block_M, bx * block_N])

    return matmul_relu_kernel


block_M = 128
block_N = 128
block_K = 32
M = block_M  # M = T.dynamic("m") if you want to use dynamic shape
N = block_N
K = block_K

# Define the kernel (matmul) and compile/lower it into an executable module
matmul_relu_kernel = matmul(M, N, K, block_M, block_N, block_K)
# Test the kernel in Python with PyTorch data
import torch

# Create random input tensors on the GPU
a = torch.randn(M, K, device="cuda", dtype=torch.float16)
b = torch.randn(K, N, device="cuda", dtype=torch.float16)
c = torch.empty(M, N, device="cuda", dtype=torch.float16)
a_plus_one = torch.empty(M, K, device="cuda", dtype=torch.float16)

# Run the kernel through the Profiler
matmul_relu_kernel(a, b, c, a_plus_one)

print(c)
# Reference multiplication using PyTorch
ref_c = torch.relu(a @ b)
ref_a_plus_one = a + 1

# Validate correctness
torch.testing.assert_close(c, ref_c, rtol=1e-2, atol=1e-2)
torch.testing.assert_close(a_plus_one, ref_a_plus_one, rtol=1e-3, atol=1e-3)
print("Kernel output matches PyTorch reference.")

# 4. Retrieve and inspect the generated CUDA source (optional)
cuda_source = matmul_relu_kernel.get_kernel_source()
print("Generated CUDA kernel:\n", cuda_source)

# 5.Profile latency with kernel
profiler = matmul_relu_kernel.get_profiler(tensor_supply_type=tilelang.TensorSupplyType.Normal)

latency = profiler.do_bench()

print(f"Latency: {latency} ms")

```

But still can not reproduce this issue on the main branch, would you mind have a try on the main branch?

TerminusAkivili:
Thanks for checking this!

The issue came from an application-oriented FP8 GEMM / SM120 investigation where I needed to inspect and work with lower-level TVM/TIR-style IR, not only the high-level TileLang frontend API. The key difference from your example is that `T.alloc_shared(...)` creates a separate buffer, while the failing case needs a same-storage alias/view through `data=A_stage.data`.

Here is the minimized reproducer:

```python
from tilelang import tvm as tvm
from tilelang.utils.target import determine_target
import tilelang as tl
import tilelang.language as T

auto_target = tvm.target.Target(determine_target("auto"))

@T.prim_func
def failing_kernel(
    A: T.Tensor((64, 64), "float16"),
    B: T.Tensor((64, 64), "float16"),
):
        # This represents a pipeline-expanded shared buffer.
        A_stage = T.alloc_shared((2, 64, 64), "float16", scope="shared.dyn")

        # Same underlying storage, but a different logical view.
        A_view = T.Buffer((64, 64), "float16", data=A_stage.data, scope="shared.dyn")

        # Explicit annotated layout is attached through the alias/view.
        T.annotate_layout({
            A_view: T.Layout((64, 64), lambda i, j: [i, j])
        })

        for i, j in T.Parallel(64, 64):
            A_view[i, j] = A[i, j]
            B[i, j] = A_stage[0, i, j]

with tvm.target.Target(auto_target):
    mod = tvm.IRModule({"main": failing_kernel})
    mod = tvm.tir.transform.BindTarget(auto_target)(mod)
    mod = tl.transform.LayoutInference()(mod)

print("LAYOUT_INFERENCE_OK")
```
On the old logic, this fails inside LayoutInference with:

```text

InputShape() = [64, 64] shape = [2, 64, 64]
```
So my understanding is that this is a LayoutInference handling issue for a legal same-storage alias/view state after pipeline expansion, rather than the frontend necessarily generating an invalid program.

TerminusAkivili:
I also tried a frontend-only reproducer without the explicit `T.Buffer(..., data=...)` alias construction.

Even if we only use TileLang-provided high-level APIs, the same issue can still appear in kernels that reinterpret shared storage through `T.view(...)`, which is relevant to quantization / dequantization / compression-style workflows.

In this case, `T.view(...)` creates a same-storage reinterpret view of `A_stage`. The explicit layout is attached to that view, while the original staged buffer is still accessed directly. On the old logic, this also fails in `LayoutInference` because the layout attached to the alias/view is reshaped against the staged buffer shape incorrectly.

For example:

```python
import tilelang
import tilelang.language as T


@tilelang.jit(out_idx=[-1])
def repro():
    @T.prim_func
    def kernel(
        A: T.Tensor((64, 64), "float16"),
        B: T.Tensor((64, 128), "int8"),
    ):
            A_stage = T.alloc_shared((2, 64, 64), "float16", scope="shared.dyn")
            A_i8 = T.view(A_stage, (2, 64, 128), "int8")
            T.annotate_layout({A_i8: T.Layout((2, 64, 128), lambda s, i, j: [s, i, j])})

            for i, j in T.Parallel(64, 64):
                A_stage[0, i, j] = A[i, j]

            for i, j in T.Parallel(64, 128):
                B[i, j] = A_i8[0, i, j]

    return kernel


kernel = repro()
print("KERNEL_CREATED")
```
This reproducer passes with the fix, and fails on the old logic with a LayoutInference shape/layout mismatch.

TerminusAkivili:
One additional nuance from my testing: when the stage expansion is introduced implicitly through `T.Pipelined(...)`, I did not reproduce this bug on the old logic in the frontend-only variants I tried.

However, I do not think that removes the need to support the explicit form. An explicitly stage-expanded shared buffer plus a same-storage reinterpret view/alias is still a legitimate expression that users may need for manual staging, quantization / dequantization / compression-style workflows, or low-level debugging.

So the fact that the auto-expanded pipeline path does not trigger the issue in these cases does not mean the explicit case is invalid or unnecessary to support.

LeiWang1999:
I see, I'll take look tomorrow, thanks!

LeiWang1999 (approved):

PR #2034, opened 2026-04-12
Fix TVMDerivedObject slots for apache-tvm-ffi compatibility

ChrisLundquist:
## Summary

The `derived_object` decorator in `3rdparty/tvm/python/tvm/runtime/support.py` creates a `TVMDerivedObject` class that inherits from `CObject` (via `apache-tvm-ffi`). `CObject` is a C extension type with `__slots__ = ()` and **no instance `__dict__`**, so when `__init__` tries to set `self._inst`, it fails with:

```
AttributeError: '_NestedLoopCheckVisitor' object has no attribute '_inst'
```

This breaks any code path using derived TVM objects, including importing the Mamba3 MIMO kernel (`mamba_ssm.ops.tilelang.mamba3.mamba3_mimo`).

## Root cause

tilelang migrated from a custom TVM fork to `apache-tvm-ffi` (October 2025). The old fork's `Object` type allowed arbitrary instance attributes via `__dict__`; the new `CObject` does not — it uses `__slots__` throughout the hierarchy.

## Fix

Add `__slots__ = ("_inst", "__weakref__")` to the dynamically created `TVMDerivedObject` class:

- **`_inst`** — needed for instance storage (set on line 159)
- **`__weakref__`** — needed because the code creates `weakref.ref(self)` on line 170

This is a one-line, backwards-compatible change. Classes that previously worked with `__dict__` will continue to work with explicit slots.

## How to verify

```python
python -c "from mamba_ssm.ops.tilelang.mamba3.mamba3_mimo import mamba3_mimo; print('MIMO import OK')"
```

Should succeed without the `_NestedLoopCheckVisitor` `AttributeError`.

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependencies to maintain compatibility and stability.


ChrisLundquist:
Closing — this fix belongs in https://github.com/TileLang/tvm, the submodule repo, not here.

PR #2043, opened 2026-04-14
[RISCV][CPU] Add tl.cpu_arch pass-config plumbing

xyyy1420:
## Summary
  This PR adds `tl.cpu_arch` pass-config plumbing for CPU lowering.
  It makes `tl.cpu_arch`:
  - a registered pass-config key
  - available through `tilelang.lower(..., pass_configs=...)`
  - available through function-level `tilelang_pass_configs`
  - observable from CPU lowering via a C++ helper
  It also preserves ambient `PassContext` semantics by merging configs into the current active context instead of creating a nested
  pass-context scope.
  ## Notes
  This PR is plumbing only.
  It does not change existing CPU codegen behavior. In particular, scalar CPU codegen remains unchanged with or without `tl.cpu_arch`.
  ## Validation
  Covered by focused CPU tests for:
  - explicit `pass_configs` input
  - function-level pass-config attrs
  - `IRModule` inputs
  - multi-function conflict / override handling
  - `PassContext` instrument preservation
  - CPU-only C++ visibility of `tl.cpu_arch`
  - legacy `tilelang.lower(..., runtime_only)` positional compatibility
  Also verified against the existing CPU baseline:
  - `testing/python/cpu/test_tilelang_cpu_tgemm.py::test_codegen_basic`

## Summary by CodeRabbit

* **New Features**
  * Lowering accepts explicit pass_configs including a new CPU-architecture key to select a target CPU family; configs are normalized, merged with outer context, validated, and can be scoped for temporary overrides.
  * Runtime helpers let lowering detect CPU targets and observe the configured CPU arch.

* **Tests**
  * Added comprehensive tests for CPU-arch configs, merge/precedence rules, observability during lowering, annotated vs. explicit overrides, and related error cases.

### DIFF
--- a/src/op/builtin.cc
+++ b/src/op/builtin.cc
@@ -19,6 +19,7 @@ namespace tl {
 TVM_REGISTER_PASS_CONFIG_OPTION(kDebugMergeSharedMemoryAllocations, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableSafeMemoryLegalize, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableWarpSpecialized, Bool);
+TVM_REGISTER_PASS_CONFIG_OPTION(kCPUArch, String);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableThreadStorageSync, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kConfigIndexBitwidth, Integer);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableTMALower, Bool);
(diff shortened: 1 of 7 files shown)


xyyy1420:
pre-commit.ci autofix

LeiWang1999:
@xyyy1420, thanks for your contributions, how can we benefit from this pass config? I think that information could be automatically extracted when we create target_host.

xyyy1420:
> @xyyy1420, thanks for your contributions, how can we benefit from this pass config? I think that information could be automatically extracted when we create target_host.

Thank you for your reply. As you said, `target_host` can indeed achieve this functionality. It was my oversight that caused you unnecessary work. I will close this pull request.

PR #2046, opened 2026-04-15
[Refactor] Remove obsolete RewriteWgmmaSync pass

LeiWang1999:
## Summary

- Remove the `RewriteWgmmaSync` transform pass, which is no longer needed.
- Delete `src/transform/wgmma_sync_rewriter.cc`, its Python wrapper in `tilelang/transform/__init__.py`, and the call site in `tilelang/engine/phase.py` (along with the now-unused `is_hopper` import).

## Test plan

- [ ] Build tilelang and confirm compilation succeeds without the removed source file.
- [ ] Run existing Hopper (sm90+) GEMM tests to ensure wgmma code generation still works correctly.

## Release Notes

* **Chores**
  * Removed an optimization pass that previously rewrote synchronization operations for GPU memory operations. This may affect performance characteristics on certain hardware targets.


LeiWang1999:
@regression-perf

PR #2047, opened 2026-04-15
[Refactor] Move target gating into InjectFenceProxy pass entry

LeiWang1999:
## Summary
- `InjectFenceProxy` now reads the `PrimFunc`'s target at the pass entry and becomes a no-op when the hardware lacks the TMA / async-proxy programming model (non-CUDA or pre-sm_90). The hardware precondition is an intrinsic property of the pass, so it belongs with the pass rather than being duplicated at every call site.
- `tilelang/engine/phase.py` drops the redundant `allow_warp_specialized` / `allow_fence_proxy` branching around `InjectFenceProxy()`. Both branches previously invoked the same pass, so the whole if/else collapsed into a single unconditional call. The now-unused `allow_fence_proxy` helper is removed.

## Motivation
The former call site looked like this:

```python
if allow_warp_specialized(pass_ctx=pass_ctx, target=target):
    mod = tilelang.transform.InjectFenceProxy()(mod)
else:
    if allow_fence_proxy(target=target):
        mod = tilelang.transform.InjectFenceProxy()(mod)
```

Both predicates ultimately reduced to `have_tma(target)`, so the if/else encoded no real decision — it just made the intent harder to read and tied a hardware check to a warp-specialization flag that has no bearing on whether `fence.proxy.async` is required. Pushing the target check into the pass keeps the scheduling layer target-agnostic and mirrors how other Hopper-era passes (e.g. `LowerPTXAsyncCopy`) gate themselves.

## Test plan
- [ ] CI build + unit tests on CUDA sm_90 (Hopper) to confirm `fence.proxy.async` is still injected where required.
- [ ] CI build on a non-TMA target (e.g. Ampere) to confirm the pass is now a clean no-op and nothing else in the lowering pipeline regresses.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized internal compilation pass logic to improve code maintainability and simplify conditional execution flow.

### DIFF
--- a/src/transform/inject_fence_proxy.cc
+++ b/src/transform/inject_fence_proxy.cc
@@ -19,6 +19,7 @@
 #include "tir/transforms/ir_utils.h"
 
 #include "../op/builtin.h"
+#include "../target/utils.h"
 
 namespace tvm {
 namespace tl {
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #2050, opened 2026-04-16
Replace syntactic loop-var checks with invariance checks

LJC00118:
## Summary by CodeRabbit

* **Refactor**
  * Improved loop/vectorization decision logic so memory access classification is more precise, reducing unnecessary conservative handling. This yields tighter vectorization choices and potential performance gains for workloads with loop-based memory patterns.


LJC00118:
@regression-perf

LeiWang1999 (approved):

Issue #2054, opened 2026-04-16
[BUG] The grouped gemm example failed


Rachmanino:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

latest

### System information

seems due to sth related to recent refactoring of the warp specialize pass

### Problem description

directly run `tilelang/examples/grouped_gemm/example_grouped_gemm_fwd.py`, got:
`❌ Tilelang and Torch mismatch`

### Reproducible example code

The Python snippets:

```python

```

### Traceback

```pytb

```

Rachmanino:
closed as resolved in #2055

PR #2055, opened 2026-04-16
[BugFix] Keep shared-prelude local vars in producer-consumer WS

Rachmanino:
#2054 

- Preserve prelude definitions needed by shared LetStmt values during warp-specialized pipeline rewriting
- Avoid cloning local.var buffers into the producer branch. This keeps grouped GEMM batch index calculations shared and fixes incorrect TMA batch selection.

This fixes the previously failed grouped gemm fwd example.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline prelude liveness so shared prelude variables are preserved when needed and branch-local buffers are less aggressively cloned.

* **Tests**
  * Added CUDA-gated grouped-GEMM tests validating correctness, shared-prelude ordering, and that producer-branch cloning does not occur.
  * Added transform tests that compile/run workloads with workspace optimizations to assert placement and behavior of shared prelude variables.

### DIFF
--- a/examples/grouped_gemm/test_example_grouped_gemm.py
+++ b/examples/grouped_gemm/test_example_grouped_gemm.py
@@ -1,0 +1,60 @@
+import tilelang.testing
+
+import example_grouped_gemm_bwd
+import example_grouped_gemm_fwd
+import example_grouped_gemm_fwd_ptr
+
+
+@tilelang.testing.requires_cuda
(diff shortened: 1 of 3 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2061, opened 2026-04-19
[Feature] Emit named aliases for dynamic shared memory in CUDA codegen for better readability

Rachmanino:
This PR adds an opt-in CUDA codegen path for improving the readability of merged shared.dyn buffers by preserving DSL-levelshared-memory names in generated source.

  What it does:
  - Adds a new pass config: tl.emit_named_smem_pointers
  - Extends MergeSharedMemoryAllocations to emit alias metadata for merged dynamic shared-memory buffers
  - Updates CUDA codegen to emit named aliases such as:
      - `half_t* A_shared = reinterpret_cast<half_t*>(buf_dyn_shmem + ...);`
  - For multiversioned / pipelined shared buffers, emits stage-aware aliases using a small helper:
      - `auto A_shared = tl::PatternVisitor([&](int i) { ... });`
  - Rewrites many downstream shared-memory accesses to use those aliases instead of raw buf_dyn_shmem + offset expressions
  - Covers both non-WS and WS/TMA paths for common GEMM-style kernels, including stage-based descriptor initialization and many cp.async / tma_load sites

  Current limitations / not fully solved yet:

  - The implementation is strongest when the logical buffer identity is still visible in IR (e.g. normal BufferLoad/BufferStore,
    many staged producer accesses).
  - Some paths still lose logical buffer provenance after lowering to tvm_access_ptr(data, offset, extent, rw_mask), especially
    when multiple logical buffers are merged into the same physical buf_dyn_shmem region.
  - Because of that, in some cases codegen can still choose a physically correct but semantically less desirable alias. For
    example, an address logically corresponding to C_shared may still be rendered as A_shared[0] if both map to the same merged
    address range and only physical address information remains.
  - In other words, this PR substantially improves readability, but it does not yet fully preserve logical buffer provenance end-
    to-end through every pass that rewrites tvm_access_ptr.

  Follow-up direction:

  - A more complete solution would preserve a stable logical-buffer identity across lowering (especially for tma_store-style
    paths), instead of relying on address-based alias recovery after buffer merging.

## Summary by CodeRabbit

* **New Features**
  * CUDA codegen can emit named, typed pointer aliases for merged dynamic shared memory; controlled by a new pass config (defaults to True).
  * Multi-version dynamic shared-memory metadata is propagated so device code can emit stage-aware aliases for pipelined kernels.

* **Tests**
  * Added CUDA tests validating default/raw behavior, named alias emission, and pattern-visitor based multiversion indexing.


kurisu6912 (dismissed):
This pr add smem alias by record aliases in `merge_shared_memory_allocations` pass, and then match the alias in codegen passes.

1. At the allocation pass, use prim func attrs to record each buffer's allocation byte range.
2. At the codegen pass, use Analyzer to prove the buffer's index resides in that range, and find out its alias.

Generally this is a very good solution, thanks for contribution!

kurisu6912:
@Rachmanino Tilelang has FP4 buffers, the current pr uses stride to compute the alias, does it handle fp4 buffers correctly?

Sometime fp4 stride is 0.5x, and shape is 2x, which is often counterintuitive.

Rachmanino:
> @Rachmanino Tilelang has FP4 buffers, the current pr uses stride to compute the alias, does it handle fp4 buffers correctly?
> 
> Sometime fp4 stride is 0.5x, and shape is 2x, which is often counterintuitive.

Thanks for your insight! I'll check this when im available. Besides, CI has passed with this feature enabled, and I guess that there's already some fp4 examples in our tests?

Rachmanino:
I guess you are right, thx! I'll fix this soon and rebase to eliminate conflict

LeiWang1999:
pending and wait for #2216 then we can take a look for if there exist better solutions.

LeiWang1999:
Closed. tir has been rebased into tir, where alloc is now a single alloc_buffer statement instead of a declaration and allocation with a body. We also have `handle_add_byte_offset` now, which makes implementing this feature much more convenient. A separate issue will be opened for that.

Issue #2081, opened 2026-04-22
[BUG] Autotuning + kernel value parameters = failure


Triang-jyed-driung:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.8; 0.1.9

### System information

3.14.3 | packaged by Anaconda, Inc. | (main, Feb 24 2026, 22:51:43) [GCC 14.3.0] linux
0.1.8+cuda.git04468a34
2.12.0.dev20260409+cu130

### Problem description

As stated in the title, if any value parameters are present, autotuning fails.

### Reproducible example code

The Python snippets:

```python
import torch
import tilelang
import tilelang.language as T

@tilelang.autotune(configs=[{"threads": 128}, {"threads": 256}], warmup=25, rep=100, timeout=60)
@tilelang.jit
def test_fun(N: int=4096, BLOCK_N: int=512, threads: int=128):
    @T.prim_func
    def kernel(A: T.Tensor((N,), T.float32), s: T.float32): # s: additional parameter
        with T.Kernel(T.ceildiv(N, BLOCK_N), threads=threads) as pid_n:
            A_local = T.alloc_fragment((BLOCK_N,), T.float32)
            T.copy(A[pid_n * BLOCK_N], A_local)
            for i in T.Parallel(BLOCK_N):
                A_local[i] += s
            T.copy(A_local, A[pid_n * BLOCK_N])
    return kernel

A = torch.randn((4096,), device=0)
test_fun()(A, 0.1)
```

### Traceback

```pytb
2026-04-22 16:25:53,516 INFO:Auto-tuning with 0.9 CPU utilizations, 32 CPUs available, 28 CPUs will be used
2026-04-22 16:25:53,928 WARNING:An error occurred while testing config {'threads': 256}, checkout autotuner.log for more details
2026-04-22 16:25:53,929 DEBUG:Error: Traceback (most recent call last):
  File "/home/user/cudaker/tilelang/tilelang/autotuner/tuner.py", line 581, in run
    latency, ref_latency = run_with_timeout(target_fn, timeout, jit_kernel)
                           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2084, opened 2026-04-22
[Enhance] Reject default scalar params and support `do_not_specialize` for autotune 

Rachmanino:
#2081 

## Summary by CodeRabbit

* **New Features**
  * Added a do_not_specialize option to exclude specified inputs from autotune cache-key specialization for more stable cache reuse.
  * New example entry point to run a persistent GEMM regression.

* **Bug Fixes / Improvements**
  * Corrected warp-specialized producer/consumer handling to fix barrier indexing and shared-buffer staging in nested/non-uniform loop nests.

* **Tests**
  * Added CUDA and unit tests validating scalar-autotune input validation, do_not_specialize behavior, and nested WS handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2084)


LeiWang1999 (dismissed):

Triang-jyed-driung:
Like I said in https://github.com/tile-ai/tilelang/issues/2081, I don't want a simple rejection for these kernels. I want a way to manually specify whether this kernel should be retuned if some scalar parameter changes, while for other parameters, I would like to specify their default values during autotuning.

Rachmanino:
> Like I said in #2081, I don't want a simple rejection for these kernels. I want a way to manually specify whether this kernel should be retuned if some scalar parameter changes, while for other parameters, I would like to specify their default values during autotuning.

check my response in your issue, please.

LeiWang1999:
@Triang-jyed-driung Thanks for your suggestion. For this issue, we should introduce a parameter like a `do_not_specify` argument in autotune, but this PR also makes sense to me. However, we should fix the CI first. cc @Rachmanino

LeiWang1999 (dismissed):

PR #2092, opened 2026-04-23
[BugFix] Relax loop wait and adjust trailing drain behavior in async pipeline tests

Rachmanino:
Updated the async pipeline logic to progressively relax loop waits and modify the trailing drain suffix. The test case was also adjusted to reflect changes in the expected behavior, ensuring that the pipeline maintains the correct number of groups in flight and descends through the drain suffix as intended.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced software pipeline optimization for more efficient async operation wait sequencing. The refined approach now generates code with dynamically optimized wait patterns, producing better performance characteristics while maintaining full correctness.

* **Tests**
  * Updated test cases to validate the new pipeline optimization behavior and ensure proper handling of async operations in software pipeline generation.

### DIFF
--- a/src/transform/inject_pipeline.cc
+++ b/src/transform/inject_pipeline.cc
@@ -1792,7 +1792,13 @@ private:
     for (size_t pos = 1; pos < suffix_wait_indices.size(); ++pos) {
       bool changed = false;
       int idx = suffix_wait_indices[pos];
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #2093, opened 2026-04-24
[Feature] Add full Windows support and fix related cross-platform issues

sepcnt:
## Summary
This PR adds end-to-end Windows support for TileLang across build, packaging, runtime loading, and developer setup.
Along the way, it also fixes several cross-platform issues uncovered during the Windows bring-up, especially around CUDA toolkit discovery, MSVC compatibility, runtime DLL loading, logging, and layout/reducer handling.
## Main changes
- add native Windows support in the build and packaging flow
- support pip-installed CUDA toolkits on Windows, including NVCC/NVRTC-based workflows
- add Windows-specific MSVC environment bootstrap helpers
- add a Windows helper script for setting up a local development environment
- improve runtime DLL discovery and PATH setup on Windows
- update installation docs and packaging metadata for Windows
## Bug fixes
- avoid enabling CUDA by default when no CUDA toolkit is available
- fix NVRTC import-library handling on Windows
- fix several MSVC build issues, including:
  - Python debug import library auto-linking
  - `TVM_DLL` import/export behavior
  - designated initializer compilation
  - large CUDA intrinsic dispatch paths in codegen
- normalize Windows file paths for `TVM_LOG_DEBUG`
- make reducer/layout handling more robust when reducer annotations are absent or layout information is still in hint form
- improve runtime library lookup behavior for local/dev builds on Windows
## TODO / Follow-ups
- [x] clean up and optimize the dynamic-loading (`dynload`) polyfill
- [x] investigate the ROCm regression; currently suspected to be a compatibility break introduced by a recent nightly
- [x] optimize and reuse CUDA toolkit lookup results to avoid redundant probing
- [x] fix concurrency/thread-safety issues in the autotuner on Windows
## Known issues
- MSVC is significantly slower than GCC/Clang at compiling CUTE headers; we may need to introduce on-demand / lazy includes to keep build times reasonable on Windows.
- `cutlass-dsl` does not officially support Windows, so any code paths depending on it are currently unavailable in the Windows build.
- CUDA 13 requires 128-byte alignment in some places, but MSVC only supports up to 64-byte alignment for `__declspec(align(...))` if is not newer than `17.10`. As a workaround, we use a torch-style fallback that degrades to the largest alignment MSVC accepts; this matches PyTorch's behavior but means the strict CUDA 13 alignment guarantee is not preserved on MSVC.

## Release Notes
* **New Features**
  * Full Windows platform support with MSVC compiler integration and native CUDA runtime compatibility
  * Windows-specific NVRTC and CUDA toolkit installation via package manager
* **Improvements**
  * Enhanced Windows build infrastructure and CMake configuration
  * Improved Windows installation documentation with Visual Studio Developer Command Prompt guidance
  * Cross-platform dynamic library loading support
  * Better debug logging configuration capabilities
* **Bug Fixes**
  * Fixed CUDA stub compatibility on Windows
  * Corrected platform-specific library path handling


docs/get_started/Installation.md
```diff
@@ -297,15 +343,25 @@ pip install tilelang -f https://tile-ai.github.io/whl/nightly
...
+version, or set `CUDAToolkit_ROOT` to use a specific CUDA toolkit.
 
 `USE_ROCM`: If to enable ROCm support, default: `OFF`. If your ROCm SDK does not located in `/opt/rocm`, set `USE_ROCM=<rocm_sdk>` to enable build ROCm against custom sdk path.
 
 `USE_METAL`: If to enable Metal support, default: `ON` on Darwin.
 
 `TVM_ROOT`: TVM source root to use.
 
-`WITH_PIP_CUDA_TOOLCHAIN`: Path to a pip-installed CUDA toolkit directory (e.g., `/path/to/venv/lib/python3.x/site-packages/nvidia/cu13`). When set, the build system uses this directory instead of a host CUDA installation. If not set and no host CUDA is found, the build system will attempt to auto-detect pip-installed CUDA packages from the current Python environment.
+`WITH_PIP_CUDA_TOOLCHAIN`: Path to a pip-installed CUDA toolkit. This may point
```
SiriusNEO: Same

tilelang/libinfo.py
```diff
@@ -1,7 +1,25 @@
...
+import sys
+
+from .env import TL_LIBS, env
+
 
-from .env import TL_LIBS
+def get_dll_directories():
+    dll_dirs = list(TL_LIBS)
+    if sys.platform.startswith("win32"):
+        cuda_home = getattr(env, "CUDA_HOME", "")
```
SiriusNEO: TileLang has an internal method to auto find CUDA_HOME. Can we use that?

docs/get_started/Installation.md
```diff
@@ -49,6 +49,10 @@ apt-get update
 apt-get install -y python3 python3-dev python3-setuptools gcc zlib1g-dev build-essential cmake libedit-dev
 ```
 
+On Windows, install Python 3, CMake/Ninja, and Visual Studio Build Tools with
```
SiriusNEO: Format issue? The documentation seems to have many line breaks

docs/get_started/Installation.md
```diff
@@ -64,51 +68,93 @@ pip install . -v
 ### With pip-provided CUDA toolchain (no host CUDA required)
 
 If you don't have CUDA installed on the host, you can use pip-provided CUDA packages instead.
+The recommended package is NVIDIA's official `cuda-toolkit` meta-package.
```
SiriusNEO: Same

docs/get_started/Installation.md
```diff
@@ -64,51 +68,93 @@ pip install . -v
...
 git clone --recursive https://github.com/tile-ai/tilelang.git
 cd tilelang
 pip install -r requirements-dev.txt
-pip install "nvidia-cuda-nvcc>=13" "nvidia-cuda-cccl>=13" "nvidia-cuda-nvrtc>=13"
-pip install . -v --no-build-isolation
+pip install "cuda-toolkit[cccl,nvcc,nvrtc]>=13"
+pip install .[nvcc] -v
+```
+
+If the current virtual environment is activated, TileLang will auto-detect the
```
SiriusNEO: Same

src/backend/cuda/codegen/codegen_cuda.cc
```diff
@@ -3554,82 +3554,107 @@ void CodeGenTileLangCUDA::VisitExpr_(const CallNode *op, std::ostream &os) {
...
+  }
+}
+
+bool CodeGenTileLangCUDA::HandleLateIntrinsicCall(const CallNode *op,
+                                                  std::ostream &os) {
+  if (op->op.same_as(tl::__exp())) {
     CUDAFastMath math_func;
     std::string func_name = math_func(op->dtype, "exp");
     os << func_name << "(" << PrintExpr(op->args[0]) << ")";
+    return true;
```
SiriusNEO, line 3680: Perhaps we could tackle this in a follow-up PR, so that this one stays scoped to Windows adaptation

SiriusNEO (commented):
Overall looks good, but we don't have a Windows CI to test it. Maybe we could test it locally? And left some comments.

src/backend/cuda/codegen/codegen_cuda.cc
sepcnt: MSVC has block recursion depth limit C1061.

sepcnt:
At this point, we have enabled support for most operators in both compile and JIT modes on Windows through `uv sync --extra nvcc`, running in isolation mode without requiring a system-wide CUDA toolkit installation. That said, ROCm support remains broken and needs further ablation studies to isolate the root cause.

SiriusNEO:
Yes. Our ROCm CI seems to have some problems, but it's not an issue with this PR.

sepcnt:
> Yes. Our ROCm CI seems to have some problems, but it's not an issue with this PR.

The last available torch version is `2.13.0.dev20260422+rocm7.1` + `triton-rocm 3.7.0+gitb4e20bbe`, unknown change in `2.13.0.dev20260424+rocm7.1` + `triton-rocm 3.7.0+git88b227e2` may break DLTensor layout.

sepcnt:
Update: ccache is now supported on Windows builds, including when PEP 517 build isolation is enabled. This should significantly speed up incremental rebuilds in isolated environments.

sepcnt:
Parallel autotune is now available on Windows as well, though it may be worth further investigation into why `tvm-ffi` works on Unix without the GIL.

sepcnt:
> Overall looks good, but we don't have a Windows CI to test it. Maybe we could test it locally? And left some comments.



> Overall looks good, but we don't have a Windows CI to test it. Maybe we could test it locally? And left some comments.

All tests passed, with the exception of those related to tmcgen05 or cutlass-dsl.

src/backend/cuda/codegen/stubs/nvrtc.cc
```diff
@@ -14,56 +14,70 @@
...
-// Export symbols with default visibility for the shared stub library.
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define TILELANG_NVRTC_STUB_API
+#else
 #define TILELANG_NVRTC_STUB_API __attribute__((visibility("default")))
+#endif
 
 namespace {
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
```
SiriusNEO, line 44: This hardcoded list will inevitably break with the next CUDA release — can we just resolve libnvrtc.so through the normal linker search or probe for the latest version dynamically?

testing/python/language/test_tilelang_language_int64.py
```diff
@@ -1,7 +1,69 @@
...
 
+# 2^32 bf16 elements = 8 GiB. This is the upper-bound stress target; the
+# helper below scales down only when the GPU cannot fit it.
+_DEFAULT_TARGET_N = 2**32
+# 2^28 bf16 elements = 512 MiB. Below this the test is no longer a useful
+# fill-stress — skip instead of running a token-sized fill.
+_FLOOR_N = 2**28
+
+
+def _select_int64_stress_n(
```
SiriusNEO, line 17: I'm worried that this stress test will break our vulnerable CI machine

sepcnt: The workload was previously hardcoded to 2³² (8 GB), which did hang my machine. I've since added an automatic fallback to prevent this.

src/backend/cuda/codegen/stubs/nvrtc.cc
sepcnt: Now we use a best-effort loader.

SiriusNEO:
No more comments, looks good to me. cc @LeiWang1999. Since this PR is quite large, it may take some time to review and merge.

SiriusNEO (dismissed):

src/op/parallel.cc
```diff
@@ -82,11 +82,12 @@ void ParallelLoopNestVisitor::VisitStmt_(const ForNode *op) {
                        IterVar(Range(op->min, op->extent), op->loop_var,
                                IterVarType::kOrdered));
   p->analyzer_.Bind(op->loop_var, Range::FromMinExtent(op->min, op->extent));
-  auto reducer_info_map =
-      op->annotations.Get(attr::kReducerInfo)->as<Map<Var, ReducerInfo>>();
-  if (reducer_info_map) {
-    for (auto &&[buffer, info] : reducer_info_map.value())
-      p->reducer_info_map_.Set(buffer, info);
+  if (auto reducer_info_ref = op->annotations.Get(attr::kReducerInfo)) {
```
LeiWang1999, line 85: Why should we refactor this part?

sepcnt: Otherwise this would cause a null exception of std optional. And this is a simple workaround that just work. Upstream issue needs further investigation.

LeiWang1999:
@SiriusNEO I think this pr is ready, would you mind helping resolve the conflict and run perf regression test, then we can let it in.

SiriusNEO (approved):

PR #2095, opened 2026-04-24
[Fix] Fix tvm_ffi torch tensor handoff

Rachmanino:
## Summary
- convert Torch tensor arguments to tvm.runtime.Tensor before calling runtime.Executable
- add a regression test that checks the adapter does not pass raw torch.Tensor objects into the executable boundary

## Summary by CodeRabbit

* **Tests**
  * Added tests verifying that wrappers bridging PyTorch and the runtime convert tensor inputs as needed and still return PyTorch tensors.

* **Bug Fixes**
  * Improved runtime interoperability by converting PyTorch tensors into the runtime's compatible tensor form before invocation, including correct handling of contiguous and non-contiguous inputs.

### DIFF
--- a/testing/python/jit/test_tilelang_jit_tvm_ffi.py
+++ b/testing/python/jit/test_tilelang_jit_tvm_ffi.py
@@ -4,6 +4,8 @@ import tilelang.testing
 import tilelang
 import torch
 import pytest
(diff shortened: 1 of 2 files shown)


Rachmanino:
@regression-perf

PR #2096, opened 2026-04-24
[AMD] Fix HIP codegen for sync_warp, sync_grid, and local.var initialisation

zhangnju:
when support Tilekernel on AMD GPU, we found TileKerle used some unsupported operation for AMD GPUs, which caused test case to fail. Three fixes to the HIP code generator, plus the cooperative-kernel launch infrastructure needed for T.sync_grid().

1. T.sync_warp() → no-op on HIP (codegen_hip.cc)

NVIDIA Volta introduced Independent Thread Scheduling, making __syncwarp() necessary to re-converge diverged warp threads. AMD wavefronts always execute in lockstep, so the hardware guarantees convergence without any explicit barrier. HIP provides no __syncwarp() equivalent. Previously TileLang hit a LOG(FATAL) "Unresolved call" on any kernel that used T.sync_warp(). Now it is lowered to a no-op (no code emitted), matching the behaviour of the existing "warp" sync path in PrintStorageSync().

2. T.sync_grid() → cooperative_groups::this_grid().sync() on HIP (codegen_hip.cc, codegen_hip.h, stubs/hip.h, stubs/hip.cc, rt_mod_hip.cc)

 T.sync_grid() previously had no HIP handler and caused a fatal codegen error. It is now lowered to cooperative_groups::this_grid().sync(), matching the CUDA path. Supporting infrastructure added:
  - need_cooperative_groups_ flag in CodeGenTileLangHIP that gates inclusion of <hip/hip_cooperative_groups.h>.
  - hipModuleLaunchCooperativeKernel lazy-load stub in the HIP driver shim, so the symbol is available without linking directly against libamdhip64.
  - ExtractFuncInfo in rt_mod_hip.cc now propagates the use_cooperative_groups PrimFunc attr as a kUseCooperativeLaunch launch-param tag, mirroring the CUDA path in rt_mod_cuda.cc.

3. T.alloc_var(init=...) emits initialized scalar on HIP (codegen_hip.cc)

local.var-scoped allocations are single-element variables. The CUDA codegen emits them as `type name = init_value;` and looks up the tl.local_var_init annotation to find the user-supplied initializer. The HIP codegen emitted `type name[1];` unconditionally, silently dropping the init value and leaving the variable undefined. This caused incorrect results on AMD (e.g. unrolled accumulation loops starting from garbage).
Fix: VisitStmt_(AllocateNode) now mirrors the CUDA path for local.var:
  - Emits `type name = 0;` by default.
  - Uses the tl.local_var_init annotation value when present.
  GetBufferRef is also updated to return `vid` directly (no [index] subscript) for local.var, consistent with the scalar declaration.

 New Test Case for this commit:
 testing/python/amd/test_tilelang_hip_codegen.py covers:
  - Absence of __syncwarp in HIP source for T.sync_warp() kernels.
  - Correctness of T.sync_warp() kernels at runtime on AMD.
  - Presence of init value in HIP source for T.alloc_var(init=N).
  - Absence of array subscript (e.g. counter[0]) for local.var scalars.
  - Runtime correctness of T.alloc_var(init=N) on AMD.
  - Codegen correctness of T.sync_grid() (cooperative_groups present i

## Summary by CodeRabbit

* **New Features**
  * Added support for grid-wide synchronization via `sync_grid()` on HIP
  * Added cooperative kernel launch API support for HIP
  * Added device `__habs` overloads for 16-bit floating-point types

* **Bug Fixes**
  * Fixed FP8-E4 packing helpers to correctly handle sign bits

* **Tests**
  * Added comprehensive HIP codegen validation test suite

### DIFF
--- a/src/target/codegen_hip.cc
+++ b/src/target/codegen_hip.cc
@@ -210,6 +210,10 @@ std::string CodeGenTileLangHIP::Finish() {
     decl_stream << "#include <tl_templates/hip/hip_fp8.h>\n";
   }
 
+  if (need_cooperative_groups_) {
+    decl_stream << "#include <hip/hip_cooperative_groups.h>\n";
+  }
+
   decl_stream << "#include <tl_templates/hip/gemm.h>\n";
   decl_stream << "#include <tl_templates/hip/copy.h>\n";
   decl_stream << "#include <tl_templates/hip/reduce.h>\n";
@@ -836,6 +840,15 @@ std::string CodeGenTileLangHIP::GetBufferRef(DataType t,
     buffer_str = temp.str();
   }
 
+  if (scope.empty()) {
+    scope = GetPtrStorageScope(buffer->data);
+  }
+  // local.var is a scalar — no indexing needed.
+  if (scope == "local.var") {
+    os << vid;
+    return os.str();
+  }
+
   std::string index_str = PrintExpr(index);
   if (t.bits() == 4 || (t.bits() == 1 && t.is_int())) {
     // This is a special case, because CodegenCUDA::PrintType()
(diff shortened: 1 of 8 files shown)


zhangnju:
This PR has been included in new PR https://github.com/tile-ai/tilelang/pull/2099

PR #2105, opened 2026-04-27
[AMD][CI issue] add gfx950 guard to fix the CI issues 

zhangnju:
Some CI tests failed because test_gfx950_cp_async_gs_16_in_codegen ran on a non-gfx950 AMD runner and asserted cp_async_gs<16> was in the generated HIP source — which it isn't on gfx942 hardware. 
The fix adds an early pytest.skip when not on gfx950. 

I also fix the issue of async-copy swizzle conflict, re-enable pipelining, fix 32x32 MFMA store layout. 
  - cp_async_gs<16>: drop buffer_load_dwordx4-lds path; swizzled LDS layouts
    from the pipeline planner are incompatible with that instruction's linear
    address requirement. Use synchronous uint4 store unconditionally.
  - pipeline_planning: only skip software pipelining for ROCm targets that
    lack async-copy support (e.g. RDNA); allow gfx942/gfx950 to pipeline.
  - mfma_layout: add thread_id_shared_access_64x16_to_32x32_layout_C_m_n
    with the correct (row, col) mapping for 32x32 MFMA stmatrix stores.
  - test: decouple 16-byte copy codegen check from pipelining via T.async_copy.

This PR also upgraded rocm sdk version to be 7.2, since pytorch nightly release has been on rocm7.2.

## Summary by CodeRabbit

* **New Features**
  * Added a gfx950-specific thread-to-register layout mapper for MFMA stores.

* **Tests**
  * Introduced a gfx950-specific test marker and updated tests to use it; tests skip when the required GPU target isn't present.
  * Strengthened storage-rewrite assertions to explicitly check both inplace and separate-write forms and include generated source on failure.

* **Behavior / Codegen**
  * Refined async-copy and pipeline planning behavior for gfx950 so generated kernels and synchronization are consistent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
             name: self-hosted-amd
             # Format: [Nightly-]ROCm-<major>.<minor>[.<patch>]. E.g., "ROCm-6.4" or "Nightly-ROCm-7.0".
             # Use "Nightly-" prefix to use torch nightly builds.
-            toolkit: Nightly-ROCm-7.1
+            toolkit: Nightly-ROCm-7.2
           - tags: [macos-latest]
             name: macos-latest
             toolkit: Metal # or Nightly-Metal
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,13 +33,10 @@ dependencies = [
     "apache-tvm-ffi~=0.1.0,>=0.1.2",
     # torch-c-dlpack-ext provides prebuilt torch extensions.
     # Without it, TVM FFI may require JIT compilation on first import.
-    "torch-c-dlpack-ext; python_version < '3.14'",
     "cloudpickle",
     "ml-dtypes",
     "numpy>=1.23.5",
     "psutil",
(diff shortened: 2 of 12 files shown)


zhangnju:
now the CI test on MI300 is  
{68F038AF-AD20-417C-AE51-999D738D11E1}

CI test on MI350 is 
{54DBE905-B1AC-4F31-974A-FE54A2E9FF67}

zhangnju:
hi @LeiWang1999 
I run the CI test on local MI300 machine, and no error is reported. but the CI test on remote server still has 24 test cases to fail. it is very weird. 

my local CI test command is "PYTHONDEVMODE=1 PYTHONPATH="" PIP_USER="" \
    uv run --no-project -m -- \
    pytest --verbose --color=yes --durations=0 --showlocals --cache-clear \
    --maxfail=3 --numprocesses=8 \
    --ignore=./python/runtime --ignore=./python/transform \
    ./python " 

Is the above command right?

LeiWang1999 (approved):

PR #2110, opened 2026-04-28
Refactor TileLang backend ownership and registry dispatch

LeiWang1999:
## Summary

This draft PR starts the backend ownership split needed for TileLang's growing backend matrix. The main direction is to make backend-specific behavior explicit instead of spreading NVIDIA/CUDA assumptions across lowering, JIT, cache, wrapper generation, CMake, and runtime code.

The PR separates two layers that are easy to conflate:

- `src/backend/*` is the native build/runtime layer for CMake and C++ sources.
- `tilelang/backend/*` is the Python backend layer for registry metadata, dispatch decisions, pass hooks, FFI builder references, target classification, and execution backend mapping.

That means `tilelang/backend` is intentionally not treated as a C-only location. Backend-specific Python code now has a first-class home next to the native backend layout.

## Backend layout

Native backend layout added under `src/backend`:

- `src/backend/common`
- `src/backend/nvidia`
- `src/backend/amd`
- `src/backend/apple`
- `src/backend/cpu`
- `src/backend/webgpu`

Python backend layout added under `tilelang/backend`:

- `tilelang/backend/common`
- `tilelang/backend/nvidia`
- `tilelang/backend/amd`
- `tilelang/backend/apple`
- `tilelang/backend/cpu`
- `tilelang/backend/webgpu`

Each Python backend package can own its target helpers, execution mapping, pass hooks, and FFI builder references. This gives future backend work a clear place to add Python-side behavior without adding more special cases to central JIT/lowering modules.

## Main changes

- Introduces backend specs and a registry in `tilelang/backend/base.py` and `tilelang/backend/registry.py`.
- Adds explicit concepts for backend metadata, including device backend, host target, execution backend support, FFI builders, adapter settings, cache metadata, and lowering pass hooks.
- Moves target and execution-backend resolution behind backend-owned helpers instead of keeping CUDA/HIP/CPU/Metal/WebGPU logic directly in the JIT entry points.
- Routes JIT wrapper/libgen/cache metadata through backend specs so backend-specific output names, adapters, and cache validation are centralized.
- Keeps optional SDK/runtime imports lazy so unsupported backends do not fail import-time checks on machines without that SDK installed.
- Adds backend pass hook support and moves NVIDIA-specific pass insertion into `tilelang/backend/nvidia/passes.py`.
- Adds native backend CMake subdirectories and moves the CUDA runtime source into `src/backend/nvidia/runtime/runtime.cc`.
- Keeps public FFI registration names stable while changing the source ownership layout.
- Restores the legacy cache `_dispatch_map` compatibility surface as a registry-backed singleton table.
- Adds cache artifact metadata validation for backend and execution backend mismatches.
- Removes trailing blank lines from new backend modules so `git diff --check` passes for the full PR range.

## Commit scope

The PR branch is intentionally limited to backend architecture and cache/JIT dispatch compatibility:

- `Introduce Python backend registry`
- `Complete backend ownership registry staging`
- `Restore cache dispatch compatibility`
- `Remove backend module trailing blank lines`

Language-level correctness fixes found during review are not included in this PR so this draft can stay focused on the backend architecture boundary.

## Design direction

This PR currently introduces a shared lowering pipeline with backend pass hooks. That is a staging step, not necessarily the final architecture.

A stronger long-term direction is to let each backend own its lowering and pass pipeline explicitly:

- Keep `tilelang/engine/lower.py` as the high-level orchestration entry point.
- Move shared pipeline building blocks into `tilelang/backend/common/pipeline.py` or an equivalent common module.
- Add backend-owned pipeline implementations such as `tilelang/backend/nvidia/pipeline.py`, `tilelang/backend/amd/pipeline.py`, `tilelang/backend/apple/pipeline.py`, and so on.
- Let each `DeviceBackend` provide its own pipeline object or factory, instead of only providing hook methods into one global shared pipeline.
- Use the common pipeline as the default implementation for simple backends, while NVIDIA/AMD or other specialized backends can override larger pipeline sections when needed.

TileOP ownership should follow the same direction:

- Keep a common TileOP contract/base layer for frontend IR compatibility and shared behavior.
- Let each backend explicitly register which TileOP implementations it supports.
- Allow a backend to implement an op, inherit the common fallback, or reject the op with a clear unsupported-backend error.
- Move backend-specific TileOP selection logic, such as GEMM instruction selection, toward backend-owned registries instead of keeping all backend choices in one shared TileOP implementation.
- Keep the native C++ `TileOperatorNode` contract as the IR/lowering boundary, while Python-side TileOP implementation choice can become backend-owned.

The current PR creates the backend registry shape needed for this next step, but does not fully migrate pipeline or TileOP ownership yet.

## Testing

Latest checks run on the cleaned PR branch:

- `cmake --build build -j$(nproc)` -> passed
- `pytest -q testing/python/backend/test_backend_registry.py` -> `9 passed`
- `pytest -q testing/python/cache/test_tilelang_kernel_cache.py -rs --tb=short` -> `10 passed, 3 skipped`
- Import-only check for `tilelang`, `tilelang.backend`, `tilelang.cache`, and `tilelang.jit.kernel` -> `import-ok`
- `git diff --check origin/main..HEAD` -> passed

The three cache-suite skips are the existing NVRTC platform gate on this host.

## Known limitations and follow-ups

This PR is intentionally a draft because it defines the first backend ownership boundary rather than completing every backend migration.

Items that still need review or follow-up:

- Decide how much backend-specific C++ code should move under each `src/backend/<name>` directory in later PRs.
- Promote `LowerAndLegalize` / `OptimizeForTarget` from a shared pipeline with hooks into backend-owned pipeline implementations.
- Decide whether backend-specific Python FFI bindings should stay in `tilelang/backend/<name>/ffi.py` long term, or whether they should be split further once each backend grows more runtime-specific code.
- Expand backend-owned pass hooks beyond the current NVIDIA-focused path as AMD, Apple, CPU, and WebGPU passes become more complete.
- Introduce a common TileOP base/contract and backend-owned TileOP implementation registry, so each backend can choose to implement, inherit, or reject each TileOP explicitly.
- Add real execution coverage for ROCm, Metal, CPU, and WebGPU environments. This branch keeps their imports and registry metadata safe, but local runtime coverage was not available in this environment.
- Consider documenting the backend taxonomy in developer docs once the registry shape has been reviewed.

## Review focus

I would especially like review on:

- Whether the split between `src/backend/*` and `tilelang/backend/*` is the right long-term ownership model.
- Whether the registry fields are too broad, too narrow, or missing important backend concepts.
- Whether pass hooks are enough for the first step, and how soon we should move to fully backend-owned lowering/pass pipelines.
- Whether TileOP implementation ownership should become part of `DeviceBackend`, a separate TileOP registry, or a separate backend capability registry.
- Whether cache metadata validation should be stricter before this becomes non-draft.

### DIFF
--- a/src/backend/README.md
+++ b/src/backend/README.md
@@ -1,0 +1,10 @@
+# TileLang Native Backend Ownership
+
+This tree is the native backend ownership surface used by CMake. Backend CMake
+files own source lists, stub libraries, include paths, and compile definitions
+for their backend even when some implementation files still live in transitional
+locations such as `src/target` or `src/transform`.
+
+Public FFI registration names are intentionally preserved during this staged
+move. Physical file relocation should be mechanical once the CMake ownership and
+Python backend metadata are covered by tests.
--- a/tilelang/engine/phase.py
+++ b/tilelang/engine/phase.py
@@ -2,20 +2,9 @@ from __future__ import annotations
 from tvm import tir, IRModule
 from tvm.target import Target
 import tilelang
+from tilelang.backend.base import DeviceBackend
+from tilelang.backend.registry import get_device_backend
 from tilelang.transform import PassContext
-from tilelang.contrib.nvcc import have_tma, have_pdl
-
-
-def allow_warp_specialized(pass_ctx: PassContext | None = None, target: Target | None = None) -> bool:
-    # avoid circular import
-    from tilelang.jit.adapter.utils import is_cuda_target
-
-    if pass_ctx is None:
-        pass_ctx = tilelang.transform.get_pass_context()
-    if (not is_cuda_target(target)) or (not have_tma(target)):
-        return False
-    disable_warp_specialized = pass_ctx.config.get("tl.disable_warp_specialized", False)
-    return not disable_warp_specialized
 
 
 def module_has_tma(mod: IRModule) -> bool:
@@ -42,15 +31,18 @@ def allow_global_thread_synchronization(pass_ctx: PassContext | None = None) -> 
     return enable_global_thread_sync
 
 
-def should_enable_aggressive_merge(pass_ctx: PassContext | None = None, target: Target | None = None) -> bool:
+def should_enable_aggressive_merge(
+    pass_ctx: PassContext | None = None,
+    target: Target | None = None,
+    backend: DeviceBackend | None = None,
+) -> bool:
     if pass_ctx is None:
         pass_ctx = tilelang.transform.get_pass_context()
     enable_aggressive_merge = bool(pass_ctx.config.get(tilelang.PassConfigKey.TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE, False))
-    if allow_warp_specialized(pass_ctx=pass_ctx, target=target):
-        # This is a workaround to avoid the bug in the MergeSharedMemoryAllocations pass
-        # when warp specialization is enabled, as different warp threads may access different
-        # buffers, but the liveness analysis is hard because we need to do pipeline.
-        enable_aggressive_merge = False
+    if backend is None and target is not None:
+        backend = get_device_backend(target)
+    if backend is not None and target is not None:
+        enable_aggressive_merge = backend.pass_hooks.adjust_aggressive_shared_memory_merge(enable_aggressive_merge, target)
     return enable_aggressive_merge
 
 
@@ -142,7 +134,6 @@ def PreLowerSemanticCheck(mod: IRModule) -> None:
 
 
 def LowerAndLegalize(mod: IRModule, target: Target) -> IRModule:
-    # Bind the target device information to the module
     """
     Bind target information and progressively legalize and lower frontend Tile IR into a form suitable for downstream optimization and codegen.
 
@@ -163,6 +154,8 @@ def LowerAndLegalize(mod: IRModule, target: Target) -> IRModule:
     Returns:
         IRModule: The transformed module, ready for target-specific optimization passes.
     """
+    backend = get_device_backend(target)
+    # Bind the target device information to the module
     mod = tir.transform.BindTarget(target)(mod)
 
     if should_force_let_inline():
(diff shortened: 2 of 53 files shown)


tilelang/engine/phase.py
```diff
@@ -225,19 +207,18 @@ def LowerAndLegalize(mod: IRModule, target: Target) -> IRModule:
...
     mod = tilelang.transform.IfStmtBinding()(mod)
     has_tma = module_has_tma(mod)
     # Pipeline barriers are now created at final expanded size by
     # InjectSoftwarePipeline, so no late MVB barrier fixup is needed.
     # Buffer allocation placement is handled uniformly for both paths.
     mod = tilelang.transform.PlanAndUpdateBufferAllocationLocation()(mod)
-    mod = tilelang.transform.LowerSharedBarrier()(mod)
-    if has_tma:
-        mod = tilelang.transform.FuseMBarrierArriveExpectTx()(mod)
+    mod = backend.pass_hooks.lower_shared_barrier(mod, target)
```
SiriusNEO, line 220: This is ugly. Let's directly use different pass Pipeline for different backends.

src/backend/README.md
```diff
@@ -0,0 +1,10 @@
+# TileLang Native Backend Ownership
```
SiriusNEO, line 1: We should have a more concise README to illustrate a backend file structure.

src/op/transpose.cc
```diff
@@ -20,6 +22,44 @@ namespace tl {
 
 using namespace tir;
 
+namespace {
```
SiriusNEO, line 25: Don't include extra changes in this PR.

src/tl_templates/cuda/cuda_fp4.h
```diff
@@ -327,15 +328,16 @@ __tl_cvt_bfloat162_to_fp4x2(const __nv_bfloat162 src) {
 // Load a single fp4 element from packed storage
 // packed: pointer to fp4_e2_2_t array
 // idx: logical index of the fp4 element
-TL_DEVICE fp4_e2_t tl_fp4_packed_load(fp4_e2_2_t *packed, int idx) {
+TL_DEVICE fp4_e2_t tl_fp4_packed_load(fp4_e2_2_t *packed, int64_t idx) {
```
SiriusNEO, line 331: Same. Let's do it in another PR.

SiriusNEO (changes requested):

PR #2122, opened 2026-04-29
[BugFix] Fix redundant runtime bounds checks for BufferLoad indices in LegalizeSafeMemoryAccess

SiriusNEO:
As titled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safe memory bounds checking with a fallback mechanism to reduce unnecessary runtime guards and warnings in cases where symbolic interval analysis is inconclusive but constant bounds can be determined.


SiriusNEO:
@regression-perf

PR #2128, opened 2026-04-30
[FIX] Fix kernel file suffix for cutedsl when only target is set

ur4t:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected debug kernel source file naming logic when using the cutedsl execution backend or target.


kurisu6912 (approved):
LGTM!

PR #2134, opened 2026-04-30
fix: TMA alignment to 1024 bytes on Blackwell

kasper0406:
Previously 1024 byte alignemnt was guarded only by `IsHopper`. This commit changes it to be guarded by `TargetHasBulkCopy` with the same threshold as Hopper is using.

The kernel added in the test was failing on my RTX5090 GPU.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shared-memory alignment for GPU buffer allocation to ensure wider alignment where supported, reducing misalignment issues on modern GPU architectures.

* **Tests**
  * Added regression tests that verify TMA/shared-memory alignment and runtime numerical correctness across multiple GPU architectures and host CUDA testing.

### DIFF
--- a/src/transform/merge_shared_memory_allocations.cc
+++ b/src/transform/merge_shared_memory_allocations.cc
@@ -390,7 +390,7 @@ private:
     if (scope == "shared" || scope == "shared.dyn") {
       auto target = Target::Current();
       ICHECK(target.defined()) << "Target is not defined";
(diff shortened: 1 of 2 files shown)


LeiWang1999 (approved):

LeiWang1999:
LGTM!

PR #2153, opened 2026-05-06
[codex] Split GEMM implementations by backend

LeiWang1999:
## Summary

Split GEMM backend-specific logic out of the shared tile op path. The common `src/op/gemm.cc` now resolves backend implementations through a registry, while CUDA, ROCm, and CPU selection/lowering details live under their own backend directories.

Add a Python GEMM implementation registry under `tilelang.backend`, with per-backend modules registering the implementation classes used by GEMM infer-layout and lowering.

## Why

This keeps hardware instruction selection and backend-specific GEMM implementation choices out of the generic tile op code, matching the backend split pattern used by copy and making future backend-specific GEMM changes more localized.

## Validation

- `./format.sh`
- `PYTHONPATH=$(pwd) python -m compileall -q tilelang/backend tilelang/tileop/gemm`
- CUDA GEMM smoke test, 64x64x64, max diff 0.007762908935546875
- CPU GEMM smoke test, 64x64x64, max diff 5.7220458984375e-06

## Summary by CodeRabbit

* **New Features**
  * Added optimized GEMM backends for CPU, CUDA, and ROCm with hardware-aware selection and multiple specialized implementations.

* **Refactor**
  * Introduced a pluggable, priority-based GEMM resolver so backends can register and be chosen at runtime.
  * Updated Python backend and tile op dispatch to route GEMM selection through the new resolver.

* **Chores**
  * Removed legacy low-level GPU query bindings in favor of the resolver-based API.


LeiWang1999:
@regression-perf

PR #2169, opened 2026-05-08
[Example] Add CLC-pipelined 2-CTA GEMM example for sm100

ighoshsubho:
Adds `gemm_clc_persistent_2cta_pipelined_clc` next to the existing single-stage CLC kernel. It pipelines the CLC tile-id handshake across `clc_stages` slots so the next tile's `clusterlaunchcontrol.try_cancel` can issue while the current tile is still being computed.

the scheduler runs `clc_stages` ahead of the consumer. Slot s is  written by scheduler iter s, s+clc_stages, ... and read by consumer iter  s+1, s+1+clc_stages, ... — non-overlapping. The `schedule_finished`  arrive count is set to 5 (consumer arrives only); the scheduler does not arrive on it. This breaks the circular dependency that would otherwise deadlock for clc_stages > 1.

Numbers on B200, bf16, baseline = single-stage CLC kernel. Run the example to reproduce; it sweeps clc_stages ∈ {2, 3, 4} per shape:                                                                                                                                           
                                                                                                                                                                                                              
  | M=N=K  | base   | pipe(2) | pipe(3) | pipe(4)   | torch |                                                                                                                                                 
  |--------|-------:|--------:|--------:|----------:|------:|                                                                                                                                                 
  | 4096   | 1478   | 1441    | 1355    | 1416      | 1437  |                                                                                                                                                 
  | 8192   | 1579   | 1597    | 1386    | 1410      | 1380  |                                                                                                                                                 
  | 16384  | 1375   | 1470    | 1565    | **1588**  | 1421  |

(All TFLOPS.) Pipelined wins at 16384³ both vs the baseline (+8%) and  vs cuBLAS (+20%). At smaller shapes the single-stage scheduler already overlaps with compute well enough that pipelining doesn't pay off. 

## Summary by CodeRabbit

* **New Features**
  * Added a configurable pipelined GEMM kernel with multi-stage scheduling for improved throughput.

* **Refactor**
  * Replaced the example entrypoint to run multi-size correctness checks and full performance benchmarks.

* **Tests**
  * Added automated correctness validation across multiple matrix sizes and a reference bfloat16 matmul for comparisons, including TFLOPS reporting against a PyTorch baseline.

### DIFF
--- a/examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py
+++ b/examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py
@@ -370,19 +370,19 @@ if __name__ == "__main__":
         a = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
         b = torch.randn(K, N, device="cuda", dtype=torch.bfloat16)
         ref = ref_program(a, b)
+        flops = 2 * M * N * K
 
         c = gemm_clc_persistent_2cta(a, b, *base_args, group_size)
         torch.testing.assert_close(c, ref, rtol=1e-2, atol=1e-2)
         ms_base = do_bench(lambda a=a, b=b: gemm_clc_persistent_2cta(a, b, *base_args, group_size), backend="event")
 
-        c2 = gemm_clc_persistent_2cta_pipelined_clc(a, b, *base_args, 3, group_size)
-        torch.testing.assert_close(c2, ref, rtol=1e-2, atol=1e-2)
-        ms_pipe = do_bench(lambda a=a, b=b: gemm_clc_persistent_2cta_pipelined_clc(a, b, *base_args, 3, group_size), backend="event")
-
         ms_torch = do_bench(lambda a=a, b=b: a @ b, backend="event")
-        flops = 2 * M * N * K
-        print(
-            f"M=N=K={M:<6}  base: {flops / (ms_base / 1e3) / 1e12:6.1f} TFLOPS  "
-            f"pipe(clc=3): {flops / (ms_pipe / 1e3) / 1e12:6.1f} TFLOPS  "
-            f"torch: {flops / (ms_torch / 1e3) / 1e12:6.1f} TFLOPS"
-        )
+        print(f"M=N=K={M:<6}  base: {flops / (ms_base / 1e3) / 1e12:6.1f} TFLOPS  torch: {flops / (ms_torch / 1e3) / 1e12:6.1f} TFLOPS")
+
+        for clc in (2, 3, 4):
+            c2 = gemm_clc_persistent_2cta_pipelined_clc(a, b, *base_args, clc, group_size)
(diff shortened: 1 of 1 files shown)


Rachmanino:
Hi @ighoshsubho, thanks for your contribution! 

I'm the author of the original CLC kernel. I remember the original CLC kernel achieves ~1700TFLOPs @(8192, 8192, 8192) on B200. Besides, the torch result reported in this script is actually inaccurate. In fact, GEMM on B200 significantly suffers from power issues, thus the 2nd kernel to run in the example will have a severe performance degration. You can validate this by running the torch kernel only (I remember it was about 1720T, not sure). 

Back to your kernel, could you please shed more light on the difference compared to the original one? Thanks!

ighoshsubho:
> Back to your kernel, could you please shed more light on the difference compared to the original one? Thanks!

yeah sure, the point is to issue the next tile's `clusterlaunchcontrol.try_cancel`  while the current tile is still computing, so the CLC handshake latency is hidden. The single-stage baseline only overlaps the CLC handshake with the tile's compute; this version overlaps it with `clc_stages` tiles in flight.

- So scheduler runs `clc_stages - 1` ahead: at iter `k` it writes slot `k % clc_stages`; consumers (TMA / MMA / epilogue) at iter `k` read the slot scheduler wrote at iter `k - 1`. So slot `s` is written at  `s, s + clc_stages, ...` and read at `s + 1, s + 1 + clc_stages, ...`, which are non-overlapping.
- `schedule_finished` arrive count is `5` (not 7); the scheduler does NOT arrive on it. Per cycle, the 5 consumer arrives (producer 2 + MMA 1 + epilogue 2) flip the parity alone. This breaks the circular dependency that would otherwise deadlock for `clc_stages > 1`.



> In fact, GEMM on B200 significantly suffers from power issues

I will try with cupti backend, also yeah b200 does suffer with performance on GEMMs due to power issues, I will again try some tests and share you some results. I was getting >1700 for clc=3 a day ago on same config

cc: @Rachmanino

Rachmanino:
I roughly understand your point. Thanks again!

ighoshsubho:
| shape   | torch  | base   | pipe(clc=2) | pipe(clc=3) | pipe(clc=4) |                                                                                                                                     
  |---------|-------:|-------:|------------:|------------:|------------:|                                                                                                                                     
  | 4096³   | **1514**   | 1419   | 1425        | 1388        | 1404        |                                                                                                                                     
  | 8192³   | 1637   | 1657 | 1571      | 1563        | **1660**        |                                                                                                                                     
  | 16384³  | 1564   | 1543   | **1672**    | 1636        | 1582        | 

After couple on bench, it did clock near 1700. Let me know if you try it on your side and find something different.

cc: @Rachmanino @LeiWang1999

LeiWang1999 (approved):

PR #2171, opened 2026-05-08
[Feature][Blackwell] Add SM120 T.float4_e2m1fn FP4 GEMM support.

TerminusAkivili:
## Summary

This PR adds SM120 fragment-MMA GEMM support for semantic `T.float4_e2m1fn`
operands.

TileLang programs continue to declare FP4 operands as `T.float4_e2m1fn`.
For the SM120 performance path, lowering maps those semantic operands onto the
hidden `T.float4_e2m1_unpacked` byte carrier internally. Users do not need to
write FP4 GEMM operands as `uint8` tensors in TileLang programs.

Supported SM120 GEMM combinations:

| A dtype | B dtype | Accumulator | SM120 path |
| --- | --- | --- | --- |
| `T.float4_e2m1fn` | `T.float4_e2m1fn` | `T.float32` | FP4 x FP4 fragment MMA |
| `T.float8_e4m3fn` | `T.float4_e2m1fn` | `T.float32` | A8W4 fragment MMA |
| `T.float4_e2m1fn` | `T.float8_e4m3fn` | `T.float32` | W4A8 fragment MMA |

## Design Goals

- Preserve semantic TileLang signatures: FP4 tensors stay `T.float4_e2m1fn` at
  the language level.
- Use `T.float4_e2m1_unpacked` as the hidden physical carrier for the SM120 FP4
  GEMM path.
- Align the SM120 FP4 performance path with the byte-carrier / ordinary
  `ldmatrix` model used by #2182.
- Add SM120 FP4/A8W4/W4A8 support without changing existing int4/uint4
  `ldmatrix` behavior.
- Reject unsupported FP4 MMA K tile shapes early instead of generating kernels
  that silently skip a K tail.

## Design

The key separation is between the public dtype and the physical carrier:

- TileLang program signatures use `T.float4_e2m1fn`.
- SM120 GEMM lowering uses hidden `custom[float4_e2m1_unpacked]8` shared/local
  carrier buffers where the hardware path needs byte slots.
- Runtime tensors for this SM120 path may use byte-compatible storage with
  logical `(M, K)` / `(N, K)` shapes.
- The FFI binder accepts byte-compatible runtime tensors for semantic SM120 FP4
  operands while preserving the public `T.float4_e2m1fn` handle dtype.
- The SM120 MMA path shifts FP4 payload bits only for FP4 operands before
  calling the CuTe SM120 F8/F6/F4 atom.

### CUDA Templates

- Add SM120 `cute::SM120_16x8x32_TN` dispatch for FP4xFP4, FP8xFP4, and
  FP4xFP8 into FP32.
- Apply the FP4 operand register shift only to operands that are actually FP4.
- Bridge TileLang FP4 template types to CuTe FP4 types while keeping existing
  packed helper types.

### CUDA Lowering

- Lower semantic SM120 FP4 GEMM operands through hidden
  `float4_e2m1_unpacked` carrier buffers where required by the performance path.
- Emit ordinary `tl::ptx_ldmatrix_x*` for the hidden unpacked-carrier path.
- Keep packed FP4 fallback vector load/store safe for odd or not-proven-even
  logical offsets by lowering those cases through per-lane nibble helpers.
- Preserve shared-memory alias information for CuTeDSL codegen.
- Preserve FP4 storage casts when layout lowering changes the physical carrier
  dtype.

### Python Lowering

- Use `T.float4_e2m1_unpacked` as the hidden local/shared carrier for semantic
  SM120 FP4 operands.
- Build SM120 FP4 shared layouts through an unpacked-carrier view while keeping
  public buffer dtypes semantic.
- Validate mixed A/B dtypes explicitly for A8W4 and W4A8.
- Reject `T.gemm` block K values that are not divisible by the selected
  instruction K tile.

### Examples

- Add mainline-style SM120 examples for FP4xFP4 and A8W4:
  - `examples/gemm_fp4/example_gemm_fp4_sm120.py`
  - `examples/gemm_fp4/example_gemm_a8w4_sm120.py`
- The examples keep semantic `T.float4_e2m1fn` kernel signatures and use
  byte-compatible host tensors only as an interoperability detail.

### Tests

- Add focused SM120 lowering coverage in
  `testing/python/language/test_tilelang_language_float4_e2m1_unpacked_gemm.py`.
- The tests check that semantic FP4 GEMM uses hidden unpacked shared/local
  carriers, ordinary `tl::ptx_ldmatrix_x*`, and no `tl::ptx_ldmatrix_b4x16` on
  the main performance path.

## Why The Review Fixes Matter

FP4 storage has two concerns that should not be conflated: the user-facing dtype
and the physical carrier used by the hardware path. The hidden
`float4_e2m1_unpacked` carrier lets SM120 GEMM keep the semantic
`T.float4_e2m1fn` API while matching the ordinary-ldmatrix byte-carrier
performance model used by #2182.

For packed fallback cases, FP4 byte storage is byte-addressed while logical FP4
elements are nibble-addressed. A vector reinterpret load/store is safe only when
the logical base offset is known to be even. If the offset is odd, or if codegen
cannot prove it is even, vectorized byte reinterpretation can read or write the
wrong nibble without producing a compilation error. This PR routes those cases
through per-lane nibble helpers.

SM120 FP4/A8W4/W4A8 MMA consumes K in fixed `m16n8k32` instruction chunks.
Allowing a `block_K` such as 48 would execute only the representable K=32
portion and miss the K tail. This PR turns that silent numerical error into an
explicit unsupported-shape error.

## Validation

Local SM120 validation used an RTX PRO 6000 / compute capability 12.0
environment.

Build and focused examples:

```bash
cmake --build build -j$(nproc)
PYTHONPATH=$PWD${PYTHONPATH:+:$PYTHONPATH} python examples/gemm_fp4/example_gemm_fp4_sm120.py
PYTHONPATH=$PWD${PYTHONPATH:+:$PYTHONPATH} python examples/gemm_fp4/example_gemm_a8w4_sm120.py
```

Generated CUDA and TIR were inspected for the expected SM120 FP4 markers:

```text
public TIR handles: float4_e2m1fn
internal carrier: custom[float4_e2m1_unpacked]8
ordinary ldmatrix: tl::ptx_ldmatrix_x*
not present on the main performance path: tl::ptx_ldmatrix_b4x16
SM120 m16n8k32 FP4/A8W4/W4A8 dtype dispatch
```

Focused tests:

```bash
PYTHONPATH=$PWD${PYTHONPATH:+:$PYTHONPATH} python -m pytest \
  testing/python/language/test_tilelang_language_float4_e2m1_unpacked_dtype.py \
  testing/python/language/test_tilelang_language_float4_e2m1_unpacked_gemm.py -q
```

Observed focused-test results:

```text
testing/python/language/test_tilelang_language_float4_e2m1_unpacked_dtype.py: 5 passed
testing/python/language/test_tilelang_language_float4_e2m1_unpacked_gemm.py: 2 passed
```

Performance comparison against #2182 used the same SM120 GEMM shapes for
FP4xFP4, A8W4, and W4A8, with 6 shapes, 2 `block_K` settings, and 9 repeats per
point. No block-scale cases were included.

Latency delta is this PR divided by #2182 minus 1:

```text
mean geomean:
FP4xFP4  +0.41%
A8W4     +0.55%
W4A8     +0.75%
all      +0.57%

median geomean:
FP4xFP4  +1.37%
A8W4     -1.25%
W4A8     +0.23%
all      +0.11%
```

Equivalent TOPS delta is #2182 latency divided by this PR latency minus 1:

```text
mean geomean:
FP4xFP4  -0.41%
A8W4     -0.55%
W4A8     -0.74%
all      -0.57%

median geomean:
FP4xFP4  -1.35%
A8W4     +1.27%
W4A8     -0.23%
all      -0.11%
```

The overall result is effectively performance-neutral versus #2182: mean TOPS
is about 0.57% lower, while median TOPS is about 0.11% lower.

## Notes And Non-Goals

- Mixed A8W4/W4A8 dispatch is selected from explicit FP8/FP4 dtype pairs.
- `float4_e2m1_unpacked` is a hidden physical carrier for the SM120 path, not
  the public GEMM dtype users are expected to write.
- `uint8` remains a runtime storage/interoperability detail for byte-compatible
  host tensors, not the public TileLang FP4 GEMM dtype.
- Existing int4/uint4 `ldmatrix` offset behavior stays on the existing path.

### DIFF
--- a/examples/gemm_fp4/example_gemm_a8w4.py
+++ b/examples/gemm_fp4/example_gemm_a8w4.py
@@ -1,0 +1,94 @@
+import torch
+import tilelang
+import tilelang.language as T
+
+
+FP4_E2M1_TO_FLOAT = [
+    0.0,
+    0.5,
+    1.0,
+    1.5,
+    2.0,
+    3.0,
+    4.0,
+    6.0,
+    -0.0,
+    -0.5,
+    -1.0,
+    -1.5,
+    -2.0,
+    -3.0,
+    -4.0,
+    -6.0,
+]
+
+
+def calc_diff(x, y):
+    x, y = x.double(), y.double()
+    denominator = (x * x + y * y).sum()
+    sim = 2 * (x * y).sum() / denominator
+    return 1 - sim
+
+
+def fp4_uint8_to_float(tensor):
+    lut = torch.tensor(FP4_E2M1_TO_FLOAT, dtype=torch.float32, device=tensor.device)
+    return lut[(tensor.to(torch.uint8) & 0x0F).to(torch.int64)]
+
+
+@tilelang.jit(
+    target="cuda",
+    pass_configs={
+        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
+        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
+    },
+)
+def matmul(A, B, block_M, block_N, block_K, out_dtype=T.float32, accum_dtype=T.float32):
+    M, N, K = T.const("M, N, K")
+
+    A: T.Tensor((M, K), T.float8_e4m3fn)
+    B: T.Tensor((N, K), T.float4_e2m1fn)
+    C = T.empty((M, N), out_dtype)
+
+    with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128) as (bx, by):
+        A_shared = T.alloc_shared((block_M, block_K), T.float8_e4m3fn)
+        B_shared = T.alloc_shared((block_N, block_K), T.float4_e2m1fn)
+        C_local = T.alloc_fragment((block_M, block_N), accum_dtype)
+
+        T.clear(C_local)
+        for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=3):
+            T.copy(A[by * block_M, k * block_K], A_shared)
+            T.copy(B[bx * block_N, k * block_K], B_shared)
+            T.gemm(A_shared, B_shared, C_local, transpose_B=True)
+
+        T.copy(C_local, C[by * block_M, bx * block_N])
+
+    return C
+
+
+def test_gemm_a8w4(M, N, K):
+    a = torch.randn(M, K, dtype=torch.float16, device="cuda").to(torch.float8_e4m3fn)
+    b = torch.randint(0, 16, (N, K), dtype=torch.uint8, device="cuda")
+
+    c = matmul(a, b, 128, 128, 128)
+    ref_c = a.to(torch.float32) @ fp4_uint8_to_float(b).T
+
+    diff = calc_diff(c, ref_c)
+    print(f"diff: {diff}")
+    assert diff < 1e-3
+
+
+def main():
+    test_gemm_a8w4(1024, 1024, 1024)
+
+
+def run_regression_perf():
+    M, N, K = 4096, 4096, 4096
+    kernel = matmul.compile(M=M, N=N, K=K, block_M=128, block_N=128, block_K=128)
+    profiler = kernel.get_profiler(tilelang.TensorSupplyType.Integer)
+    if torch.version.hip is None:
+        return profiler.do_bench(backend="cupti")
+    return profiler.do_bench()
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/gemm_fp4/example_gemm_fp4.py
+++ b/examples/gemm_fp4/example_gemm_fp4.py
@@ -1,0 +1,94 @@
+import torch
+import tilelang
+import tilelang.language as T
+
+
+FP4_E2M1_TO_FLOAT = [
+    0.0,
+    0.5,
+    1.0,
+    1.5,
+    2.0,
+    3.0,
+    4.0,
+    6.0,
+    -0.0,
+    -0.5,
+    -1.0,
+    -1.5,
+    -2.0,
+    -3.0,
+    -4.0,
+    -6.0,
+]
+
+
+def calc_diff(x, y):
+    x, y = x.double(), y.double()
+    denominator = (x * x + y * y).sum()
+    sim = 2 * (x * y).sum() / denominator
+    return 1 - sim
+
+
+def fp4_uint8_to_float(tensor):
+    lut = torch.tensor(FP4_E2M1_TO_FLOAT, dtype=torch.float32, device=tensor.device)
+    return lut[(tensor.to(torch.uint8) & 0x0F).to(torch.int64)]
+
+
+@tilelang.jit(
+    target="cuda",
+    pass_configs={
+        tilelang.PassConfigKey.TL_DISABLE_TMA_LOWER: True,
+        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
+    },
+)
+def matmul(A, B, block_M, block_N, block_K, out_dtype=T.float32, accum_dtype=T.float32):
+    M, N, K = T.const("M, N, K")
+
+    A: T.Tensor((M, K), T.float4_e2m1fn)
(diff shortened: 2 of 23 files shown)


TerminusAkivili:
Hi @LeiWang1999, no rush at all. Feel free to check it whenever it's convenient for you. I'd love your feedback. Thank you!

Hale423:
Oh, seems like this PR overlaps with part of my PR #2182.  Just wanted to clarify that there was no intent to duplicate the work as it is carried over from the earlier FP4 branch (didn't notice this pr the moment I create my new one).  I'm totally okey to coordinate scope if any feedback received  from maintainers, thanks for your work.

TerminusAkivili:
Thanks @Hale423  for the clarification! I took a closer look at the SM120 overlap, and I think the two PRs are taking slightly different directions.
My understanding is that #2182 uses a `uint8` carrier for the SM120 FP4 path and maps that into FP4/F8F6F4 MMA lowering, which gives it a more byte-aligned fast path. #2171 keeps the TileLang-facing API semantic: kernels use `T.float4_e2m1fn` / `T.float8_e4m3fn`, while packed storage, shared layout, ldmatrix, and fragment handling stay inside lowering/codegen.
In local default-shape testing, #2182 is faster for FP4xFP4 by around 8-9% on average. That seems expected because both A and B pay the FP4 shared-layout/copy overhead in #2171’s semantic `b4x16_p64` path, while #2182’s byte-carrier path is closer to the packed fast path. For A8W4 the gap is much smaller, close to parity on larger shapes, since only one operand is FP4.
Happy to coordinate scope based on maintainer feedback. One possible follow-up would be to keep #2171’s semantic public API while adding an internal byte-carrier-style fast path for FP4xFP4.

TerminusAkivili:
The downside is that it introduces FP4xFP4-specific logic across layout, copy analysis, pipeline planning, codegen, and CUDA helpers, which increases maintenance cost and expands the scope of this PR.
A cleaner long-term direction may be a more general packed low-bit dtype storage lowering mechanism, where semantic dtype, packed global storage, shared-memory carrier/layout, and MMA load path are described in a unified way. But that would touch a much larger surface area.
So I think which direction to take, and how to structure it, should depend on the maintainers’ feedback and guidance.

Hale423:
Confirmed, feel free to implement any idea on SM120, I'm willing to coordinate scope, thanks for sharing!

TerminusAkivili:
@Hale423 I also tried a separate RFC version that keeps the public API as `T.float4_e2m1fn`, but internally uses a byte-carrier lowering path similar to #2182. Its benchmark results are roughly on par with #2182.
The branch/commit is here for reference:
 https://github.com/TerminusAkivili/tilelang/commit/b8e2818ffaf63b190732b6e129f6f1f96658edc6
Since that version touches a larger surface area, I’m not planning to include it in #2171 before maintainer feedback. If we also count that direction, I think the core scopes are no longer conflicting.

Hale423:
> @Hale423 I also tried a separate RFC version that keeps the public API as `T.float4_e2m1fn`, but internally uses a byte-carrier lowering path similar to #2182. Its benchmark results are roughly on par with #2182. The branch/commit is here for reference: [TerminusAkivili@6c118bc](https://github.com/TerminusAkivili/tilelang/commit/6c118bc45ce643c60408aed705c4baf279bcbc1f) Since that version touches a larger surface area, I’m not planning to include it in #2171 before maintainer feedback. If we also count that direction, I think the core scopes are no longer conflicting.

Got it, thanks for your clarification

LeiWang1999:
Looks interesting, but for f8f6f4, I think we need to introduce a hidden type, float4_e2m1_unpacked – that's exactly what we're working on. Thanks.

TerminusAkivili:
@LeiWang1999  Thanks for the update and for working on this. I’ll keep an eye on the progress.

TerminusAkivili:
I’d appreciate it if you could review this PR when you have time. Thanks! @LeiWang1999

PR #2179, opened 2026-05-11
[ROCm] Try to fix ROCm CI error 

zihaomu:
This PR works around a ROCm-specific `tvm-ffi` / `torch-c-dlpack-ext` tensor marshalling issue in TileLang.

  On ROCm, the torch C DLPack fast path can corrupt tensor metadata when TileLang calls generated TVM FFI kernels. This shows up as packed API errors such as:

  ```text
  ndim expected 2, but got 512
  ```
  The root issue is likely in the ROCm path of `torch-c-dlpack-ext` or `tvm-ffi` tensor exchange, not in individual TileLang kernels. This PR does not fix that dependency-level bug; it avoids the broken path inside TileLang.

  ## Changes

  - Disable the torch C DLPack loader on ROCm during TileLang import.
  - For ROCm FP8 tensors, export the byte view through DLPack and restore the expected TileLang dtype with _create_view.
  - Simplify atomic test reference computation from Python GPU scalar loops to A.sum(dim=0).

  ## Testing

  - ROCm CI passed:
    Test for Python 3.12 with Nightly-ROCm-7.2
  - Local ROCm-equivalent test run:
    664 passed, 540 skipped

### DIFF
--- a/testing/python/language/test_tilelang_language_atomic.py
+++ b/testing/python/language/test_tilelang_language_atomic.py
@@ -26,16 +26,9 @@ def run_atomic_add(K, M, N, block_M, block_N, dtype=T.float32):
     kernel = atomic_add_program(K, M, N, block_M, block_N, dtype=dtype)
     import torch
 
-    def ref_program(A, B):
-        for k in range(K):
-            for i in range(M):
-                for j in range(N):
-                    B[i, j] += A[k, i, j]
-
     A = torch.randn(K, M, N, dtype=getattr(torch, dtype)).cuda()
     B = torch.zeros(M, N, dtype=getattr(torch, dtype)).cuda()
-    ref_B = B.clone()
-    ref_program(A, ref_B)
+    ref_B = B + A.sum(dim=0)
     kernel(A, B)
     torch.testing.assert_close(B, ref_B, atol=1e-3, rtol=1e-3)
 
@@ -59,16 +52,9 @@ def run_atomic_memory_order(K, M, N, block_M, block_N, dtype=T.float32):
     kernel = atomic_memory_order_program(K, M, N, block_M, block_N, dtype=dtype)
     import torch
 
-    def ref_program(A, B):
-        for k in range(K):
-            for i in range(M):
-                for j in range(N):
-                    B[i, j] += A[k, i, j]
-
     A = torch.randn(K, M, N, dtype=getattr(torch, dtype)).cuda()
     B = torch.zeros(M, N, dtype=getattr(torch, dtype)).cuda()
-    ref_B = B.clone()
-    ref_program(A, ref_B)
+    ref_B = B + A.sum(dim=0)
     kernel(A, B)
     torch.testing.assert_close(B, ref_B, atol=1e-3, rtol=1e-3)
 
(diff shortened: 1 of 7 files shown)


LeiWang1999:
@zihaomu Awesome!

SiriusNEO:
@zihaomu Thanks! This issue has been bothering us for a while!

testing/python/language/test_tilelang_language_atomic.py
```diff
@@ -59,16 +52,9 @@ def run_atomic_memory_order(K, M, N, block_M, block_N, dtype=T.float32):
...
-        for k in range(K):
-            for i in range(M):
-                for j in range(N):
-                    B[i, j] += A[k, i, j]
-
     A = torch.randn(K, M, N, dtype=getattr(torch, dtype)).cuda()
     B = torch.zeros(M, N, dtype=getattr(torch, dtype)).cuda()
-    ref_B = B.clone()
-    ref_program(A, ref_B)
+    ref_B = B + A.sum(dim=0)
```
zihaomu, line 57: Replaced the Python scalar reference loop with B + A.sum(dim=0).

This keeps the same expected result, but avoids many tiny GPU scalar operations during test setup, which were unstable under ROCm + pytest-xdist. The TileLang kernel behavior is unchanged.

tilelang/jit/adapter/tvm_ffi.py
```diff
@@ -25,6 +25,28 @@
...
+        dtype
+        for name in ("float8_e4m3fn", "float8_e4m3fnuz", "float8_e5m2", "float8_e5m2fnuz")
+        if (dtype := getattr(torch, name, None)) is not None
+    )
+
+
+_TORCH_FLOAT8_DTYPES = _torch_float8_dtypes()
+
+
+def _adapt_rocm_float8_tensor(arg: Any, expected_dtype: Any):
```
zihaomu: After disabling `torch-c-dlpack-ext`, the fallback DLPack path cannot handle FP8 tensors and reports: "float8 types are not supported by dlpack".

This change exports ROCm FP8 tensors as an `int8` DLPack byte view, then uses `_create_view` to restore the expected FP8 dtype on the TVM side.

tilelang/jit/adapter/cutedsl/wrapper.py
```diff
@@ -612,6 +612,7 @@ def _generate_cubin_if_needed({cubin_gen_params}):
       "torch.float8_e4m3fnuz": cutlass.Float8E4M3FN,
       "torch.float8_e4m3fn": cutlass.Float8E4M3FN,
       "torch.float8_e5m2": cutlass.Float8E5M2,
+      "torch.float4_e2m1fn_x2": cutlass.Float4E2M1FN,
```
zihaomu: Follow-up to [#2174](https://github.com/tile-ai/tilelang/pull/2174): TileLang now returns PyTorch's real FP4 storage dtype, `torch.float4_e2m1fn_x2`. CuTeDSL maps runtime `str(tensor.dtype)` to CUTLASS types, but this key was missing, causing
  `KeyError: 'torch.float4_e2m1fn_x2'` in the DeepSeek V4 FP4 example.

  This adds the missing bridge:

  `torch.float4_e2m1fn_x2 -> cutlass.Float4E2M1FN

zihaomu:
Try to fix the CuTeDSL CI errors in [PR2187](https://github.com/tile-ai/tilelang/pull/2187) instead of here. This PR should not affect the CuTe part.

tilelang/jit/adapter/tvm_ffi.py
LeiWang1999: Thanks but I think we should introduce another pr for this fix?

LeiWang1999 (approved):

Issue #2202, opened 2026-05-14
[BUG] T.While is broken for target="c" / CPU mode


PicoCreator:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.9

### Problem description

The use of T.While loops within tilelang, currently causes compilation issues for CPU / C based backend. But not for GPU based backends

### Reproducible example code

The Python snippets:

```python
from __future__ import annotations

import sys
from pathlib import Path

import tilelang
import tilelang.language as T

def test_cpu_kernel_source_generation_with_while_reproduces_issue() -> None:
    @T.prim_func
    def main(flag: T.Buffer((1,), "int32"), out: T.Buffer((1,), "int32")):
        with T.Kernel(1, is_cpu=True):
            state = T.alloc_fragment((1,), "int32")
            state[0] = 0

            with T.While(state[0] == 0):
                state[0] = flag[0]

            out[0] = state[0]

    # This currently aborts or fails during CPU lowering instead of compiling cleanly.
    tilelang.compile(main, target="c")
```

### Traceback

```pytb
F                                                                                                        [100%]
=================================================== FAILURES ===================================================
________________________ test_cpu_kernel_source_generation_with_while_reproduces_issue _________________________

    def test_cpu_kernel_source_generation_with_while_reproduces_issue() -> None:
        @T.prim_func
        def main(flag: T.Buffer((1,), "int32"), out: T.Buffer((1,), "int32")):
            with T.Kernel(1, is_cpu=True):
                state = T.alloc_fragment((1,), "int32")
                state[0] = 0
    
                with T.While(state[0] == 0):
                    state[0] = flag[0]
    
                out[0] = state[0]
    
        # This currently aborts or fails during CPU lowering instead of compiling cleanly.
>       tilelang.compile(main, target="c")

test/tl/device_runtime/test_tilelang_cpu_while_repro.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.13/site-packages/tilelang/jit/__init__.py:115: in compile
    return cached(
[...]

PR #2203, opened 2026-05-14
[BugFix] cpu with while loop compile error

PicoCreator:
Original issue reported at: https://github.com/tile-ai/tilelang/issues/2202

## Summary
This PR fixes a CPU lowering bug where `T.While(...)` inside `T.Kernel(..., is_cpu=True)` could leak the synthetic fallback thread variable `v_thread` into host/device splitting.

Previously, the CPU `c` target could incorrectly treat `v_thread` as an undefined device parameter, which caused compilation failures during packed API generation or later C wrapper generation.

The fix canonicalizes the synthetic CPU fallback `v_thread` to `0` in the CPU split-device path before device parameter collection, preventing it from becoming part of the generated CPU kernel ABI.

### `src/transform/split_host_device.cc`
- Added a small CPU-only canonicalization pass:
  - `CPUFallbackThreadVarCanonicalizer`
- Applied it only when:
  - `device_target->kind->name == "c"`
- This ensures the synthetic fallback `v_thread` is rewritten to a constant before undefined-var analysis collects device function parameters.

### `testing/python/cpu/test_tilelang_cpu_while_repro.py`
- Added a dedicated regression test for CPU `T.While` lowering
- Improved the test docstring to explain:
  - the historical failure mode
  - the role of synthetic `v_thread`
  - the expected success condition
- Made the success condition explicit by asserting that:
  - `tilelang.compile(main, target="c")` returns a non-`None` compiled object

## Root cause
For CPU kernels, earlier lowering may introduce a synthetic degenerate thread variable `v_thread` to model thread-like structure even when no real thread binding exists.

When a kernel body contained `T.While(...)`, that synthetic variable could survive into `SplitHostDevice`. Since device parameters are inferred from undefined variable analysis there, `v_thread` was incorrectly promoted into the generated CPU device-kernel signature.

That led to failures such as:
- packed API free-variable errors
- generated C wrapper/device signature mismatch

## Why this fix is low risk
- The change is explicitly gated to the CPU `c` backend:
  - `if (device_target->kind->name == "c")`
- It does not alter CUDA / ROCm / Metal / WebGPU / LLVM device lowering paths
- The final fix is localized to host/device splitting, where the incorrect parameter promotion actually occurred
- A temporary broader workaround in `MakePackedAPI` was removed, so the final patch stays narrowly scoped

### Repro test
```bash
python -m pytest testing/python/cpu/test_tilelang_cpu_while_repro.py -q
```
Result:
- `1 passed`

### CPU regression suite
```bash
python -m pytest testing/python/cpu -q
```
Result:
- `15 passed, 15 warnings`

### Before
Lowered CPU IR could produce a kernel like:
```python
def main_kernel(flag, out, v_thread):
```
and the host wrapper would call it with a synthetic `v_thread`.

### After
The lowered CPU kernel signature is:
```python
def main_kernel(flag, out):
```
with no leaked synthetic thread parameter.

## AI Acknowledgement
While the bug was found in our experiments in our dev environments.
AI models were used to subsequently isolate, replicate, and fix the issue, and generate the PR summary above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CPU kernel compilation failures when using while loops in CPU-targeted kernels, improving reliability for complex CPU kernel code.

* **Tests**
  * Added a regression test to ensure CPU kernel compilation succeeds for kernels containing while loops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2203)

### DIFF
--- a/src/transform/split_host_device.cc
+++ b/src/transform/split_host_device.cc
@@ -51,14 +51,14 @@ namespace tir = tvm::tir;
 namespace {
 
 class CPUFallbackThreadVarCanonicalizer : public tir::StmtExprMutator {
- public:
+public:
   static tir::Stmt Rewrite(tir::Stmt stmt) {
     CPUFallbackThreadVarCanonicalizer canonicalizer;
     return canonicalizer(std::move(stmt));
   }
 
- private:
-  PrimExpr VisitExpr_(const tir::VarNode* op) final {
+private:
+  PrimExpr VisitExpr_(const tir::VarNode *op) final {
     if (op != nullptr && op->name_hint == "v_thread") {
       return tir::make_zero(op->dtype);
     }
@@ -66,7 +66,7 @@ class CPUFallbackThreadVarCanonicalizer : public tir::StmtExprMutator {
   }
 };
 
-}  // namespace
+} // namespace
 
 // This pass traverses the AST, split the target function into host part and
 // device part and copies all assume attribute statements to the device side.
--- a/src/transform/split_host_device.cc
+++ b/src/transform/split_host_device.cc
@@ -50,8 +50,21 @@ namespace tir = tvm::tir;
 
 namespace {
 
+/*! \brief Rewrite the synthetic CPU fallback thread variable to a constant.
+ *
+ * Some CPU lowering paths materialize a degenerate `v_thread` variable to
+ * reuse thread-oriented transformations even though the generated CPU kernel
+ * has no real thread binding.  If that synthetic variable survives until
+ * host/device splitting, undefined-variable analysis may incorrectly treat it
+ * as a device-kernel parameter.
+ *
+ * This mutator canonicalizes that CPU-only fallback variable to `0` before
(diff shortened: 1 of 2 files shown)


LeiWang1999:
Hi @PicoCreator , thanks for your contribution. but I have a question why does WhileNode relate to vthread?

LeiWang1999:
I see, it's by design and a minor issue for our layout inference systems, I'll take a look for this pr then and thanks for your report and fix

LeiWang1999:
I tried to push this cleanup directly to your branch, but I do not have write permission to featherless-ai-integrations/tilelang, so GitHub rejected the push. I will open a separate PR with this cleanup:)

PicoCreator:
@LeiWang1999 - thanks for the fix & merge. 

There was no particular reason for the vthread/while relation, I just presumed it wasn't used. 

On my side, I'm just using "cpu mode" as a means to validate/unit-test tilelang logic, before moving to larger GPU-based tests.

PR #2185, opened 2026-05-11
[Transform] Rewrite MergeSharedMemoryAllocations with per-epoch liveness

TensorGlue-IEIT:
<html>
<body>
[...]

PR #2204, opened 2026-05-14
[Fix] Remove "stop on other gen" heuristic in kill-point reorder

### LINKED ISSUE #2199: [BUG] Running sparse_mla_bwd.py in tilelang 0.1.9 produces NaNs
### Required prerequisites

0.1.9

### System information

3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] linux
0.1.9
2.7.0a0+79aa17489c.nv25.04

### Problem description

- Install tilelang 0.1.9
- Clone tilelang's codebase and Checkout tag v0.1.9
- Run examples/deepseel_v32/sparse_mla_bwd.py

It cannot pass the tensor similarity test. And there are almost all Nans in `tl_dq` and `tl_dkv`.

Change the 
```
tilelang.PassConfigKey.TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE: True,
```
to 
```
 tilelang.PassConfigKey.TL_ENABLE_AGGRESSIVE_SHARED_MEMORY_MERGE: False,
```
in the tl.jit argument of bwd fixes this, though it would use more shared memory. And I want to know what breaks it

### Reproducible example code

-

### Traceback

```pytb
ERROR: dq similarity check failed, diff=nan (threshold=1.00e-04)
  File "/workspace/inference/chenfeiyu/repos/github/tilelang/examples/deepseek_v32/sparse_mla_bwd.py", line 373, in <module>
    test_sparse_mla_bwd(B=1, S=4096, SKV=8192, H=64, HKV=1, DQKV=576, DV=512, topk=2048, dtype=torch.bfloat16, check_correctness=True)
  File "/workspace/inference/chenfeiyu/repos/github/tilelang/examples/deepseek_v32/sparse_mla_bwd.py", line 313, in test_sparse_mla_bwd
    assert_tensors_similar(tl_dq, ref_dq, eps=1e-4, name="dq")
  File "/workspace/inference/chenfeiyu/repos/github/tilelang/examples/deepseek_v32/utils.py", line 310, in assert_tensors_similar
    assert False  # noqa: B011
           ^^^^^
AssertionError
```

### Expected behavior

It passes the test

### Additional context

_No response_

Rachmanino:
The "stop on other gen" early-stop condition in the kill-point reorder (added in 3a956e0e) truncates buffer liveness when another shared buffer's first use (gen) appears inside a loop body. This causes the arena packer to overlap buffers that are still needed in subsequent loop iterations, producing NaN results.

In sparse_mla_bwd, Q_shared is loaded once before the pipelined loop and read in every iteration. The heuristic stops Q_shared's kill at acc_dkv_shared's gen (inside the loop), shortening Q_shared's liveness to not cover the full loop body. The arena packer then assigns acc_dkv_shared to Q_shared's memory region. When acc_dkv_shared is written, it overwrites Q_shared — the next iteration reads garbage and produces NaN.

Timeline of the underlying issue:

1. Before 3a956e0e: the old arena algorithm could not detect that Q_shared and O_shared in Flash Attention had disjoint lifetimes, so they were allocated separately (192KB, exceeding A100's 164KB shared memory limit).

2. 3a956e0e added the "stop on other gen" heuristic to shorten Q_shared's kill so the old arena could see the disjoint lifetimes and merge Q/O (160KB). However, this heuristic incorrectly shortens liveness for buffers used across loop iterations, breaking sparse_mla_bwd.

3. f7ba45d8 replaced the arena algorithm with LinearScanPack, which natively detects disjoint lifetimes from the [start, end) intervals — the "stop on other gen" heuristic became redundant.

4. This commit removes the now-unnecessary heuristic. The existing `level == gen_level` condition in the scan loop is sufficient to stop at the correct scope boundary.

Verified:
- sparse_mla_bwd with aggressive merge: correctness passes
- Flash Attention Q_shared/O_shared still share offset 0 (160KB)
- Flash Attention default config: all checks pass

Closes #2199

## Summary by CodeRabbit

* **Refactor**
  * Adjusted shared-memory handling so deallocation placement respects statement-scope boundaries and analyses now treat shared-buffer accesses as tracked definitions/uses for more accurate dataflow.

* **Tests**
  * Strengthened an existing sparse MLA test to run correctness checks for both small and larger cases.
  * Added a CUDA-only dequantize-GEMM test targeting Hopper (sm_90) for BF16/mxfp4 scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2204?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/deepseek_v32/test_tilelang_example_deepseek_v32.py
+++ b/examples/deepseek_v32/test_tilelang_example_deepseek_v32.py
@@ -39,9 +39,7 @@ def test_example_sparse_mla_fwd_pipelined():
 @tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_example_sparse_mla_bwd():
     sparse_mla_bwd.test_sparse_mla_bwd(S=256, SKV=512, H=64, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=True)
-    sparse_mla_bwd.test_sparse_mla_bwd(
-        S=256, SKV=512, H=128, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=True
-    )  # test for large H
+    sparse_mla_bwd.test_sparse_mla_bwd(S=256, SKV=512, H=128, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=True)  # test for large H
 
 
 if __name__ == "__main__":
--- a/examples/deepseek_v32/test_tilelang_example_deepseek_v32.py
+++ b/examples/deepseek_v32/test_tilelang_example_deepseek_v32.py
@@ -38,9 +38,9 @@ def test_example_sparse_mla_fwd_pipelined():
 @tilelang.testing.requires_cuda
 @tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_example_sparse_mla_bwd():
-    sparse_mla_bwd.test_sparse_mla_bwd(S=256, SKV=512, H=64, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=False)
+    sparse_mla_bwd.test_sparse_mla_bwd(S=256, SKV=512, H=64, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=True)
     sparse_mla_bwd.test_sparse_mla_bwd(
-        S=256, SKV=512, H=128, HKV=1, DQKV=576, DV=512, topk=256, check_correctness=False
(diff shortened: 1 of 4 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
pending as #2185

PR #2229, opened 2026-05-20
[BugFix] Fix sparse int8 default metadata dtype

TerminusAkivili:
## Summary

- Fix `compress()` so int8 inputs default to int32 sparse metadata.
- Keep explicit `meta_dtype` overrides unchanged.
- Cover default metadata dtype selection in the existing compress test flow.

## Motivation

`torch_compress()` defaults int8 metadata to int32 to match the CUTLASS convention, but `compress()` previously used the global int16 default when `meta_dtype` was omitted.

This made the default int8 path inconsistent with the reference compressor and the documented metadata layout.

## Kernel

The issue shows up in the default int8 sparse GEMM setup:

```python
import torch
import tilelang.language as T
from tilelang.utils.sparse import compress, get_e_factor


M, N, K = 64, 64, 64
A_int8 = torch.empty((M, K), device="cuda", dtype=torch.int8)
B_int8 = torch.empty((K, N), device="cuda", dtype=torch.int8)

A_sparse, E = compress(A_int8)  # default metadata dtype

e_dtype = T.int32
e_factor = get_e_factor(T.int8, e_dtype)


@T.prim_func
def int8_sparse_gemm(
    A_sparse: T.Tensor((M, K // 2), T.int8),
    E: T.Tensor((M, K // e_factor), e_dtype),
    B: T.Tensor((K, N), T.int8),
    C: T.Tensor((M, N), T.int32),
):
    with T.Kernel(1, 1, threads=128):
        A_shared = T.alloc_shared((M, K // 2), T.int8)
        E_shared = T.alloc_shared((M, K // e_factor), e_dtype)
        B_shared = T.alloc_shared((K, N), T.int8)
        C_frag = T.alloc_fragment((M, N), T.int32)

        T.copy(A_sparse, A_shared)
        T.copy(E, E_shared)
        T.copy(B, B_shared)
        T.clear(C_frag)
        T.gemm_sp(A_shared, E_shared, B_shared, C_frag)
        T.copy(C_frag, C)
```

Before this patch, `compress(A_int8)` produced int16 metadata by default, while the int8 sparse GEMM path expects int32 metadata.

## Summary by CodeRabbit

* **Tests**
  * Expanded compression tests to include int8 inputs with unspecified metadata dtype and verify both compression paths produce metadata with the resolved dtype.

* **Bug Fixes**
  * When metadata dtype is unspecified for int8 tensors, it now defaults to a 32-bit integer dtype, ensuring consistent metadata across implementations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/testing/python/utils/test_compress_utils.py
+++ b/testing/python/utils/test_compress_utils.py
@@ -81,10 +81,18 @@ def generate_dense_input(N, trans_A, trans_B, in_dtype, seed=0):
     return A, B
 
 
+def _default_meta_dtype(dtype):
+    return T.int32 if dtype == T.int8 else T.int16
+
+
 def _test_compress(dtype, meta_dtype):
     A, B = generate_dense_input(64, in_dtype=dtype.as_torch(), trans_A=False, trans_B=False)
-    sp_tl, meta_tl = compress(A, meta_dtype=meta_dtype.as_torch())
-    sp_ref, meta_ref = torch_compress(A, meta_dtype=meta_dtype.as_torch())
+    torch_meta_dtype = meta_dtype.as_torch() if meta_dtype is not None else None
+    actual_meta_dtype = meta_dtype if meta_dtype is not None else _default_meta_dtype(dtype)
+    sp_tl, meta_tl = compress(A, meta_dtype=torch_meta_dtype)
+    sp_ref, meta_ref = torch_compress(A, meta_dtype=torch_meta_dtype)
+    assert meta_tl.dtype == actual_meta_dtype.as_torch()
+    assert meta_ref.dtype == actual_meta_dtype.as_torch()
     # NOTE: in case that there are multiple zeros, the case might fail occasionally
     # if we directly compare the compressed sparse values
     program = matmul(
@@ -99,8 +107,8 @@ def _test_compress(dtype, meta_dtype):
         dtype,
         dtype,
         T.int32 if dtype == T.int8 else T.float32,
-        meta_dtype,
-        get_e_factor(dtype, meta_dtype),
+        actual_meta_dtype,
+        get_e_factor(dtype, actual_meta_dtype),
         0,
         128,
     )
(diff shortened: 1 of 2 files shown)


LeiWang1999:
cc @botbw

testing/python/utils/test_compress_utils.py
```diff
@@ -122,8 +130,10 @@ def _test_compress(dtype, meta_dtype):
         (T.int8, T.int8),
         (T.int8, T.int16),
         (T.int8, T.int32),
+        (T.int8, None),
```
botbw: This will be a duplicate test of `(T.int8, T.int32)`?

testing/python/utils/test_compress_utils.py
```diff
@@ -122,8 +130,10 @@ def _test_compress(dtype, meta_dtype):
...
         (T.float16, T.int8),
         (T.float16, T.int16),
+        (T.float16, None),
```
botbw: Same here with `(T.float16, T.int16)`

botbw (changes requested):
@TerminusAkivili Thanks for correcting this. Yuh it's indeed a default setting inconsistency with `torch_compress`. For the test case, since all available `dtype` and `metadata_dtype` combinations are already tested, I think no need to add an extra one for default setting. LGTM other than that.

cc: @LeiWang1999

LeiWang1999:
failed tests are not relevant to this pr, and overall lgtm @TerminusAkivili would you mind resolving comments and we can then let it in!

TerminusAkivili:
@LeiWang1999 Acknowledged. Rebased and kept the diff limited to the sparse metadata default path.

botbw:
@TerminusAkivili One more reduntant case `(T.int8, None)` (duplicate of `(T.int8, T.int32)`)

TerminusAkivili:
Comments resolved. Thanks!

testing/python/utils/test_compress_utils.py
botbw: @TerminusAkivili I think we can directly restore the test file? As no test case was added.

PR #2234, opened 2026-05-20
[Pipeline] Replay scalar binds in pipelined stages

LeiWang1999:
## Summary

- Fix software pipeline injection for flat scalar `Bind` statements that are used by multiple pipelined stages.
- Prevent generated CUDA from referencing scalar temporaries outside their lexical scope after pipeline prologue, body, and epilogue rewriting.
- Add a transform regression that checks scalar binds are replayed into each stage block that uses them.

## Changes

- Collect scalar `Bind` statements from the ordered pipeline body before emitting the rewritten pipeline.
- Skip standalone scalar `Bind` blocks as executable pipeline stages, and replay the required bindings at the start of each consumer block using that block's normalized access index.
- Preserve dependency ordering among scalar bindings so bindings that depend on earlier scalar bindings are replayed first.
- Cover the failure mode with a regression where a scalar index is consumed by both copy and store stages.

## Validation

- `python -m pytest testing/python/transform/test_tilelang_transform_Inject_software_pipeline.py -q`
- `python /nfs-df/prod/data/permanent/wanglei/tilelang/debug/0520_bug/mqa_index_distill_mlp_dynamic.py > /tmp/tilelang_mqa_index_distill_dynamic_restore.log 2>&1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure scalar bindings are correctly replayed per pipeline stage so stage-local scalar values are preserved and emitted where used across software-pipelined loops.

* **Tests**
  * Added regression test verifying per-stage scalar binding replay in the software-pipeline transformation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/transform/inject_pipeline.cc
+++ b/src/transform/inject_pipeline.cc
@@ -711,6 +711,7 @@ public:
     for (const auto &[block, anno] : pipeline_info_) {
       ordered_stmts_.Set(anno.order, block);
     }
+    CollectScalarBindings();
 
     // Step 2: Emit the pipeline prologue, body and epilogue.
     Optional<Integer> pipeline_num_stages =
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #2237, opened 2026-05-21
[Pipeline] Support scalar bind-free pipeline annotations

LeiWang1999:
## Summary

- Distinguish replayable scalar `Bind` aliases from `Bind` statements that represent pipeline-dependent loads.
- Keep bind-free pipeline annotations for value aliases, including read-only `BufferLoad` aliases such as `idx = Ids[i]`.
- Count `Bind` statements that read pipeline-written buffers as scheduled statements, because those loads have real producer/consumer dependencies.

## Changes

- Detect buffers written inside a pipeline body and only filter/replay scalar `Bind` statements whose reads do not touch those buffers.
- Preserve compatibility with legacy annotations that include replayable `Bind` slots by filtering those slots before scheduling.
- Validate scheduled `Bind` dependencies so scalar values without storage versioning stay in the same stage as their consumers and are ordered before them.
- Align `PipelinePlanning` and `InjectSoftwarePipeline` around the same replayable-bind model.
- Document the distinction between value aliases and materialized producers, including why `x = A[i]` can be replayed when `A` is not written by the pipeline, while `x = S[i]` must be scheduled when `S` is pipeline-written.
- Add regression coverage for read-only `BufferLoad` bind replay and pipeline-written-buffer bind scheduling.

## Validation

- `python -m pytest testing/python/transform/test_tilelang_transform_pipeline_planning.py -q`
- `python -m pytest testing/python/transform/test_tilelang_transform_Inject_software_pipeline.py -q`
- `python /nfs-df/prod/data/permanent/wanglei/tilelang/debug/0520_bug/mqa_index_distill_mlp_dynamic.py > /tmp/tilelang_mqa_index_distill_dynamic_replayable_bind.log 2>&1`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -34,6 +34,7 @@ programming_guides/overview
 programming_guides/language_basics
 programming_guides/instructions
 programming_guides/control_flow
+programming_guides/software_pipeline
 programming_guides/python_compatibility
 programming_guides/autotuning
 programming_guides/type_system
--- a/docs/programming_guides/control_flow.md
+++ b/docs/programming_guides/control_flow.md
@@ -99,6 +99,10 @@ for ko in T.Pipelined(T.ceildiv(K, BK), num_stages=3):
     T.gemm(A_s, B_s, C_f)             # stage: compute
 ```
 
+For manual `stage` / `order` annotations and the rule that scalar `Bind`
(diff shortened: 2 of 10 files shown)


LeiWang1999:
@regression-perf

Issue #2196, opened 2026-05-13
RFC: migrate tilelang from TileLang/tvm to apache/tvm + apache/tvm-ffi


megacpp:
# RFC: migrate tilelang from TileLang/tvm to apache/tvm + apache/tvm-ffi

We have a downstream TileLang branch running on current apache/tvm plus
apache/tvm-ffi, while still carrying the TileLang fixes we needed for CUDA,
HIP, Apple Metal, and MLX interop.

This RFC is a request to the TileLang maintainers:

1. Please confirm whether the current TileLang/tvm submodule fork is an
   intentional long-term fork, or just a historical pin.
2. If the fork is not intentionally load-bearing, we would like to upstream
   our migration work and help TileLang move to apache/tvm + apache/tvm-ffi.
3. Independently of that larger decision, we can split out the useful fixes
   we found during the migration into small PRs, so the team can take value
   without accepting the whole stack at once.

Working proof-of-concept repos:

- tilelang fork: https://github.com/DatasunriseOU/tilelang
- tvm fork: https://github.com/DatasunriseOU/tvm
- tvm-ffi fork: https://github.com/DatasunriseOU/tvm-ffi

The short version: the migration is real and buildable in our branch, but the
first upstream PR should be staged carefully. The current TileLang/tvm history
is more nuanced than "a frozen fork": main is frozen, but the TileLang
submodule actually points into tilelang_main, which carries local TVM
changes. This RFC reflects that chronology explicitly.

---

## Verified chronology

I re-checked the repository state on 2026-05-13 after fetching
tile-ai/tilelang, TileLang/tvm, apache/tvm, DatasunriseOU/tvm,
apache/tvm-ffi, and DatasunriseOU/tvm-ffi.

### tile-ai/tilelang -> TileLang/tvm

Upstream tile-ai/tilelang@main still declares:

text
3rdparty/tvm -> https://github.com/TileLang/tvm

The submodule commit pinned by tile-ai/tilelang@main is:

text
0e15b274bce8b46f971abf5ac390e844aa6acee5
2026-04-16 Fix duplicate __weakref__ declaration in derived_object wrappers (#33)

That commit is not on the frozen TileLang/tvm main line. It is on
TileLang/tvm's default branch, tilelang_main.

### TileLang/tvm branch split

TileLang/tvm has two relevant lines:

- TileLang/tvm main
  - commit: 3e8fe753800d55e843c441f302a6a90b26dd22b9
  - date: 2025-08-23
[...]

PR #2239, opened 2026-05-21
Fix Metal local var and bf16 codegen

cklxx:
Refs #2196.

Summary:
- Normalize Metal `local.var` buffers to Metal-supported thread-local storage before device codegen.
- Legalize Metal bf16 following TVM semantics: store bf16 as `uint16`, load through fp32 boundary conversion, and use round-to-nearest-even f32-to-bf16 bit conversion on stores/casts.
- Avoid native MSL `bfloat`, because it requires Metal 3.1 while current TVM Metal compile/runtime paths target Metal 2.3.
- Add source-codegen and macOS runtime regressions for local-var conditional stores and bf16 GEMM.

Validation:
- `python -m pytest testing/python/metal -q`

Implementation note:
Directly running TVM `BF16ComputeLegalize` at this late Metal device-codegen stage can create promoted shared-buffer handles without matching TileLang allocations, so this PR adapts the same TVM storage/conversion semantics at Metal load/store and cast boundaries.


oraluben (changes requested):
It's a virtual issue that local var is not supported. Would you mind rebase your code?

PR #2240, opened 2026-05-21
[CUDA][IR] Preserve dynamic shared memory aliases

LeiWang1999:
## Summary

- Preserve named shared-memory aliases when dynamic shared allocations are merged instead of rewriting every access to the merged allocation.
- Keep thread sync analysis aware of alias byte offsets so disjoint aliases do not create false shared-memory conflicts.

## Changes

- Emit `Bind` statements that define each original shared allocation as a `handle_add_byte_offset` alias of `buf_dyn_shmem`.
- Track `shared.dyn` alias offsets in the thread sync planner and include the corresponding element offset in load, store, and pointer-access conflict ranges.
- Add a CUDA codegen test covering named aliases and avoiding repeated direct casts from `buf_dyn_shmem`.

## Example

Before this change, accesses after dynamic shared-memory merging were printed directly against the merged allocation:

```cpp
*(uint4*)(((half_t*)buf_dyn_shmem) + a_offset) = ...;
*(uint4*)(((half_t*)buf_dyn_shmem) + b_offset) = ...;
```

After this change, the original shared buffers are preserved as aliases of the merged dynamic allocation, and later accesses use those alias names:

```cpp
extern __shared__ __align__(1024) uchar buf_dyn_shmem[];
void* A_shared = ((void*)((char*)buf_dyn_shmem + 0));
void* B_shared = ((void*)((char*)buf_dyn_shmem + 8192));

*(uint4*)(((half_t*)A_shared) + a_index) = ...;
*(uint4*)(((half_t*)B_shared) + b_index) = ...;
```

The sync planner still reasons about physical shared-memory locations by folding alias byte offsets into element indices. For example:

```text
A_shared[tx] -> tx + 0
B_shared[tx] -> tx + 4096  // 8192 bytes / sizeof(half_t)
```

This keeps `A_shared` and `B_shared` in the same `shared.dyn` allocation for dependency analysis while avoiding false conflicts between disjoint regions.

## Validation

- `./format.sh`
- `PYTHONPATH=$(pwd):$PYTHONPATH python -m pytest testing/python/components/test_cuda_shared_memory_alias_codegen.py -q`

## Summary by CodeRabbit

* **New Features**
  * Shared-memory merging adds an alias-preservation mode with a public option to emit per-buffer handle aliases (enabled by default for non-WebGPU targets).

* **Bug Fixes**
  * Thread-storage sync and access planning now account for shared-memory handle aliases so memory accesses and offsets are correct.

* **Tests**
  * Added/updated tests to verify CUDA codegen emits named pointer aliases and to robustly extract shared-memory offsets in both alias-preserving and merged-buffer outputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2240?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/transform/merge_shared_memory_allocations.cc
+++ b/src/transform/merge_shared_memory_allocations.cc
@@ -472,6 +472,44 @@ public:
   }
 
 private:
+  std::vector<Stmt> MakeAliasBindings() const {
+    struct AliasInfo {
+      const VarNode *var{nullptr};
+      PrimExpr byte_offset;
+    };
+
+    std::vector<AliasInfo> aliases;
+    aliases.reserve(buffer_byte_offsets_.size());
+    for (const auto &pair : buffer_byte_offsets_) {
+      if (shmem_allocs_.count(pair.first) == 0) {
+        continue;
+      }
+      aliases.push_back(AliasInfo{pair.first, pair.second});
+    }
+
+    std::sort(aliases.begin(), aliases.end(),
+              [](const AliasInfo &lhs, const AliasInfo &rhs) {
+                const auto *lhs_offset = lhs.byte_offset.as<IntImmNode>();
+                const auto *rhs_offset = rhs.byte_offset.as<IntImmNode>();
+                if (lhs_offset != nullptr && rhs_offset != nullptr &&
+                    lhs_offset->value != rhs_offset->value) {
+                  return lhs_offset->value < rhs_offset->value;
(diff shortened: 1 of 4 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2252, opened 2026-05-23
[Metal] M5 Cooperative Tensor T.gemm

oraluben:
Needs https://github.com/tile-ai/tvm/pull/44

## Summary

This PR extends the existing Metal backend with a Metal 4 cooperative tensor path for `T.gemm`.

The Metal backend already supported simdgroup-based GEMM lowering. This PR adds cooperative tensor as a new fast path on supported Apple GPUs, while keeping simdgroup as the compatibility path for older devices and non-Metal-4 targets.

## Motivation

Metal 4 cooperative tensor exposes Apple's tensor-core-like matrix compute path. On supported hardware, it provides a substantially faster GEMM implementation than the existing simdgroup path.

This PR adds that path to TileLang so Metal `T.gemm` can use the newer hardware capability while preserving the existing simdgroup implementation for compatibility.

## Design Notes

Although cooperative tensor is conceptually the Metal-side counterpart of CUDA tensor core programming, the programming model is not a direct CUDA clone.

As a practical approximation for CUDA reviewers, Apple GPUs expose a less CUDA-like split between register and threadgroup storage; both are backed by a more hardware-managed on-chip memory system. Because of that, explicit threadgroup staging is not automatically a faster path than feeding cooperative tensor operands directly.

This PR therefore keeps CUDA-shaped shared staging as a compatibility path, but optimizes the direct cooperative-tensor path as the Metal fast path. `T.gemm` remains the frontend abstraction, and Metal-specific instruction choice stays inside the Metal backend.

## What Changed

At a high level, this PR adds:

- cooperative tensor lowering for Metal GEMM;
- Metal codegen support for cooperative tensor source emission;
- target capability guarding so cooperative tensor source is only generated when supported/requested;
- preservation of the existing simdgroup path;
- pass-level compatibility so Metal-specific storage scopes do not leak into generic TVM assumptions;
- tests for both the new cooperative tensor path and the existing simdgroup path.

Detailed lowering rules and implementation notes are documented separately in the Metal compiler internals doc.

## Impact on TileLang

The main TileLang-level impact is that Metal now has a dedicated high-performance GEMM path that reflects Metal's own matrix programming model.

In particular:

- `T.gemm` remains the user-facing abstraction.
- Existing simdgroup kernels continue to work.
- Cooperative tensor support is gated by target/runtime capability.
- CUDA-style shared staging remains supported for compatibility, but is not assumed to be the default Metal performance model.
- Future Metal schedules, such as shared-staging bypass or MLX-style variants, can be added behind the same backend boundary without changing the frontend API.

## Compatibility

This PR is intended to be backward-compatible for existing Metal users.

The existing simdgroup path is still present and tested. The new cooperative tensor path is only used when the target/runtime capability allows it, so building TileLang with a newer SDK should not force all Metal kernels to require Metal 4.

One important compatibility check is that this PR has been validated on GitHub Actions with macOS 26 and M1 hardware. That environment exposes the newer SDK at build time but does not support cooperative tensor in hardware, so passing there verifies that the backend correctly falls back to the simdgroup path on unsupported devices.

## Testing

Test coverage includes:

- Metal simdgroup fallback / old path;
- Metal cooperative tensor codegen;
- runtime cooperative tensor correctness where supported;
- source-level checks to avoid pulling cooperative tensor dependencies into simdgroup-only kernels.

Validated locally with:

- `pip install .`
- `python -m pytest testing/python/metal/ -q -x`
- `python -m pre_commit run --all-files`

GitHub Actions additionally validated the macOS 26 + M1 fallback case described above.

## Summary

- Added Metal 4 cooperative-tensor support for `T.gemm`, including lowering, code generation, target/runtime guards, and simdgroup fallback.
- Updated TVM transformation passes for `metal.cooperative_tensor` storage scopes.
- Added cooperative-tensor intrinsics, swizzle support, benchmarks, documentation, and comprehensive codegen/correctness tests.
- Updated the TVM submodule pin.

## C++ style / lint notes

- C++ changes touch areas covered by `docs/developer_guide/cpp_style.md`, including public APIs, FFI interfaces, and backend/runtime shims.
- CI includes the **C++ API Style Audit (warning only)** step. Any findings are advisory and should be evaluated separately from correctness, build, and test failures.
- No audit output was provided; TLCPP003/TLCPP004 warnings should not block merging unless they introduce a concrete API, FFI, or maintainability risk.


src/transform/storage_rewrite.cc
```diff
@@ -56,6 +56,14 @@ using runtime::StorageScope;
 using namespace tirx;
 using namespace ffi;
 
+namespace {
+
+bool IsMetalCooperativeTensorBuffer(const Buffer &buffer) {
```
SiriusNEO: Can we put this utils under the `metal/` folder?

tilelang/utils/language.py
```diff
@@ -133,6 +133,12 @@ def is_metal_simdgroup(buffer: BufferLikeType) -> bool:
     return buffer.scope() == "metal.simdgroup"
 
 
+def is_metal_cooperative_tensor(buffer: BufferLikeType) -> bool:
```
SiriusNEO: Same

SiriusNEO (dismissed):

PR #2253, opened 2026-05-23
Add SM120 dense block-scaled MMA and essential support for SageAttention3

sepcnt:
## Summary

This PR adds SM120 block-scale MMA support and the TileLang compiler/runtime features needed to implement SageAttention3 FP4 raw-core kernels in pure TileLang. It introduces a generic `T.ptx_mma_blockscale` intrinsic for `mxf8f6f4` and `mxf4nvf4`, adds the SM120 SageAttention3 FP4 example kernel, and fixes reduce/storage-sync lowering needed by the new path and existing language tests.

## Changes

- Add `tl.ptx_mma_blockscale` builtin and Python frontend bindings.
- Add CUDA codegen support for lowering block-scale MMA calls through `tl::mma_sync_blockscale`.
- Add `mma_blockscale.h` with SM120 block-scale MMA dispatcher specializations.
- Add PTX-side validation metadata for supported block-scale MMA configs:
  - `mxf8f6f4`, `m16n8k32`, `scale_vec=1`, `ue8m0`
  - `mxf4nvf4`, `m16n8k64`, `scale_vec=2`, `ue8m0`
  - `mxf4nvf4`, `m16n8k64`, `scale_vec=4`, `ue4m3`
- Add `PrintMMABlockScaleAssembly` helpers and shared dtype/string mapping for block-scale MMA validation.

- Add `T.pack_b16` for packing four b16 values into a global store.
- Add `examples/sage_attention_sm120` with a pure TileLang SM120 SageAttention3 FP4 raw-core kernel and helper macros.
- Extend `T.ptx_ldmatrix` with optional matrix shape/dtype metadata and add CUDA lowering for `m8n16.x4.shared.b8x16.b4x16_p64`.
- Add `T.lds32` and CUDA shared-memory 32-bit load helpers for packed scale/FP4 data paths.
- Add strict `T.copy(..., prefer_instruction="ldsm"|"stsm")` handling so matrix-copy requests fail clearly instead of silently falling back.
- Add `T.tma_copy(..., tma_element_dtype=...)` and improve TMA descriptor projection, folded base offsets, `elem_offset` handling, and transaction-byte accounting.
- Preserve explicit pack lanes when reshaping packed byte/sub-byte layouts.
- Add `role_scoped=True` allocation support so branch-local register arrays can stay inside warp-specialized producer/consumer roles.
- Improve CUDA codegen for scoped `pragma_unroll_factor`, aligned narrow local allocations, and FP4 local `address_of`.

- Update `ReduceLowerer` batch handling to clamp requested batch size to the physical per-thread output element count before divisibility checks.
- Use float32 temporary accumulators for fp16/bf16 `sum` and `abssum`, then cast back to the destination dtype.
- Avoid packed fp16/bf16 sum reducers where low-precision pairwise accumulation can exceed existing reduce-test tolerances.
- Remove allocation-wide alias ranges from thread-storage-sync disjointness proofs so pointer/buffer conflict checks stay conservative.

- Add handle byte offset support in TMA descriptor parsing for nvrtc

## Summary by CodeRabbit

* **New Features**
  * Added SM120 block‑scaled MMA intrinsic with scale‑vector support and runtime dispatch.
  * Frontend/TIR/TileLang APIs exposed for the new intrinsic and a pack_b16(v0, v1) helper.
* **Chores**
  * CUDA backend and codegen updated to emit the block‑scale MMA assembly and wire the intrinsic.
  * Backend builtin registered so the intrinsic is available to the compiler stack.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2253?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Rachmanino:
Thanks for your contribution! A recent pr introduce support for tcgen5 fp4 blockscaled gemm, and I personally recommend following similar API standard:)

tilelang/language/loop.py
```diff
@@ -211,6 +225,10 @@ def serial(
     step : PrimExpr
         The step size of the iteration.
 
+    unroll_factor : int
```
Rachmanino: Why we need to introduce this instead of `T.unroll`

tilelang/language/allocate.py
```diff
@@ -49,43 +52,63 @@ def alloc_shared(shape: ShapeType, dtype: DType, scope="shared.dyn") -> Buffer:
...
 @overload
-def alloc_var(dtype: DType, scope: str = "local.var", *, init: PrimExpr | int | float | None = None) -> Buffer: ...
-
-
-def alloc_var(dtype: DType, *args, scope: str = "local.var", init: PrimExpr | int | float | None = None) -> Buffer:
+def alloc_var(
+    dtype: DType,
+    scope: str = "local.var",
+    *,
+    init: PrimExpr | int | float | None = None,
```
Rachmanino, line 121: Is there any better ways to control the allocation position (better automatically analyzed)? We can discuss with @LeiWang1999

tilelang/language/tir/op.py
Rachmanino, line 1: better to provide a higher-level API, like existing `T.tcgen5_gemm_blockscaled` for tcgen5, rather than a thin wrapped ptx intrinsic.

src/tl_templates/cuda/instruction/mma_blockscaled.h
Rachmanino, line 1: I personally think that it would be better to move these instructions to `tl_templates/cuda/gemm_sm120.h` or somewhere else

tilelang/language/loop.py
sepcnt: In certain cases, `nvcc` may unroll loops by default, which can be harmful when there is significant register spilling. In addition, TileLang aggressively inlines expressions, making the issue even worse.

examples/sage_attention_sm120/sageattn3_fp4.py
```diff
@@ -0,0 +1,700 @@
+"""SageAttention3 FP4 raw-core TileLang kernel for NVIDIA SM120."""
+
+from __future__ import annotations
+
+import tilelang
+import tilelang.language as T
+from tilelang.carver.arch import driver
+
+from examples.sage_attention_sm120.tir_helpers import (
+    mma_m16n32k64_blockscale_f32,
```
Rachmanino, line 10: As I mentioned before, compared to using these thin PTX intrin wrappers, consider supporting automatically lowering from tile level operation `T.gemm` and `T.copy` to mma atom and ldmatrix/stmatrix.

testing/python/issue/test_tilelang_issue_ws_simt_copy_full_producer_extent.py
Rachmanino, line 1: why remove this?

PR #2257, opened 2026-05-25
Fix SM70 BF16 fallback lowering

cklxx:
## Summary
- skip Volta fragment-to-fragment dtype copies during layout inference and lower them through shared-memory staging
- add SM70 BF16 MMA fallback that value-converts BF16 fragment registers to FP16 inputs while keeping FP32 accumulation
- add V100-only regression coverage for fragment staging and BF16 SS/RS GEMM compilation

## Validation
- local: `python3 -m py_compile testing/python/kernel/test_tilelang_kernel_sm70_fragment_copy.py`
- local: `git diff --check HEAD~1..HEAD`
- V100: `python -S` pytest harness on `testing/python/kernel/test_tilelang_kernel_sm70_fragment_copy.py` -> `2 passed`
- V100 ARLE AOT: `batch_decode_paged_hd128.py`, q_heads=40, kv_heads=8, cuda_arch=70 -> emitted `/tmp/arle_sm70_decode_hd128_q40_kv8/batch_decode_paged_hd128_q40_kv8_sm70.c`

## Notes
- Ampere/Hopper paths do not use `mma_sm70.h`, so the BF16 fallback is SM70-pinned.
- This is intended for Volta compatibility, not a claim of BF16 numeric equivalence to native BF16 tensor cores.

## Summary by CodeRabbit

* **New Features**
  * Volta-specific optimization for fragment-to-fragment copies when source and destination dtypes differ: uses a shared-memory staging step to ensure correct and efficient transfers.
  * BF16 input handling on SM70: BF16 fragments are routed through an FP16 MMA path while preserving FP32 accumulation.

* **Tests**
  * Added SM70-targeted tests exercising fragment copy, fragment-cast behaviors, and the BF16→FP16 MMA fallback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2257?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


LeiWang1999:
please fix the lint :)

cklxx:
Heads-up: the `Test for Python 3.12 with CUDA-12.8 (on self-hosted-nvidia)` failure is **pre-existing on `main`**, not introduced by this PR. The same three tests fail on the latest main CI run (https://github.com/tile-ai/tilelang/actions/runs?branch=main&workflow=CI):

- `python/jit/test_tilelang_jit_cutedsl.py::test_gemm_jit_kernel`
- `python/jit/test_tilelang_jit_cutedsl.py::test_cutedsl_kernel_multi_stream`
- `python/jit/test_tilelang_jit_cutedsl.py::test_cutedsl_dynamic_shape`

All three error with `tvm.error.InternalError: Unresolved call ir.Op(... name="tirx.handle_add_byte_offset", ...)` in the CuTeDSL codegen path (`CodeGenTileLangCuTeDSL::VisitExpr_`). This PR only touches:

- `src/backend/cuda/op/copy.cc` — Volta fragment-to-fragment dtype-converting copy
- `src/tl_templates/cuda/instruction/mma_sm70.h` — BF16 → FP16 MMA fallback
- `testing/python/kernel/test_tilelang_kernel_sm70_fragment_copy.py` — new sm_70 regression test

None of these touch the CuTeDSL codegen.

src/backend/cuda/op/copy.cc
```diff
@@ -615,6 +629,10 @@ Stmt Copy::Lower(const CopyNode &op, const LowerArgs &T,
         << "Got target=" << T.target;
     return LowerCluster(op, T, analyzer);
   }
+  if (copy_inst == CopyInst::kNormal &&
```
LeiWang1999, line 632: Thanks, for what reason we should introduce this change?

cklxx: This is needed to make the Volta BF16 fallback path lower correctly, not as a general copy optimization.

The concrete failing pattern is a dtype-changing fragment copy before the next GEMM, e.g. FP32 accumulation/softmax output narrowed to BF16/FP16 fragment, then used as the A operand of another `T.gemm`. On sm_70, lowering that as a direct `local.fragment -> local.fragment` SIMT copy couples the source producer layout with the destination Volta MMA operand layout. Those layouts are not compatible on V100, so layout inference fails before nvcc.

The shared-memory staging breaks that layout dependency: `src fragment -> shared(dst dtype) -> dst fragment`, with a shared sync between the two normal copies. Then the destination fragment can be assigned the layout required by the following Volta MMA. This is paired with the `mma_sm70.h` BF16->FP16 MMA fallback because Volta has no native BF16 tensor-core MMA.

The predicate is intentionally narrow: Volta target only, both buffers are `local.fragment`, and src/dst dtype differs. sm_80+ paths are untouched.

cklxx: Short version: this is the missing lowering step before the SM70 BF16 MMA fallback can be used.

`mma_sm70.h` makes `BF16 x BF16 -> FP32` compile on Volta by value-converting BF16 fragment registers to FP16 MMA inputs. But the attention kernel reaches that MMA through a preceding dtype-changing fragment copy (`FP32 fragment -> BF16 fragment`). On sm_70, lowering that copy directly couples the producer fragment layout with the destination Volta-MMA operand layout, and layout inference fails before codegen.

The shared-memory staging breaks that layout coupling. It is not a general optimization; it is gated to Volta + `local.fragment -> local.fragment` + dtype mismatch, so native BF16 sm_80+ paths are unaffected.

LeiWang1999:
I see, thanks for your contributions, but this solution feels inelegant to me.

For dtype combinations that are not supported by SM70 MMA, a cleaner direction may be to extend/add a `GemmFMA` path rather than modifying the MMA implementation to emulate unsupported BF16 behavior. In other words, unsupported layouts or dtype combinations should probably be handled at the GEMM lowering level, not by special-casing the MMA template.

LeiWang1999:
closed for now, we should open another pr for that `GemmFMA` if you want to extend T.gemm to support FMA :)

PR #2272, opened 2026-05-25
[Bugfix][TMA] Drop buffer_oob gate from 1D bulk-copy eligibility

### LINKED ISSUE #2180: [BUG] TMA load unnecessary spliting
### Required prerequisites

0.1.9+cuda.gitbeef5cf4

### System information

0.1.9+cuda.gitbeef5cf4

### Problem description

Read 1D TMA as 2D TMA

### Reproducible example code

The Python snippets:

```python
import tilelang
import torch
@tilelang.jit
def gemm(A):
    M = T.dynamic('M')
    K = T.const('K')
    A: T.Tensor[[M, K], T.float32]
    with T.Kernel(M, threads=256) as pid:
        var = T.alloc_var(T.int32, init=0)
        a_shared = T.alloc_shared(K, dtype=T.float32)
        mbar = T.alloc_barrier(256)
        T.tma_copy(A[var, 0:K], a_shared, barrier=mbar)
    return None
# Get kernel object by calling compile()
A = torch.empty((23,1024), dtype=torch.float32, device='cuda')
kernel = gemm.compile(A)
# Print kernel source

```

### Traceback

```pytb
++
#if defined(_MSC_VER) && !defined(__clang__) && _MSC_VER < 1940
#define _tl_orig_alignas alignas
#define alignas(N) _tl_orig_alignas((N) <= 64 ? (N) : 64)
#include <cuda.h>
#undef alignas
#define alignas _tl_orig_alignas
#endif
#include <tl_templates/cuda/gemm.h>
#include <tl_templates/cuda/copy.h>
#include <tl_templates/cuda/reduce.h>
#include <tl_templates/cuda/ldsm.h>
#include <tl_templates/cuda/threadblock_swizzle.h>
#include <tl_templates/cuda/debug.h>
#ifdef ENABLE_BF16
#include <tl_templates/cuda/cuda_bf16_fallbacks.cuh>
#endif

extern "C" __global__ void gemm_kernel(__grid_constant__ const CUtensorMap A_desc, int M);
extern "C" __global__ void __launch_bounds__(256, 1) gemm_kernel(__grid_constant__ const CUtensorMap A_desc, int M) {
  __shared__ __align__(16) uint64_t mbar_mem[1];
  auto mbar = reinterpret_cast<Barrier*>(mbar_mem);
  int var = 0;
  extern __shared__ __align__(1024) float a_shared[];
    tl::prefetch_tma_descriptor(A_desc);
  }
    mbar[0].init(256);
  }
  tl::fence_barrier_init();
  __syncthreads();
  var = 0;
  if (tl::tl_shuffle_elect<256>()) {
    mbar[0].expect_transaction(4096);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[0])), 0, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[256])), 256, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[512])), 512, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[768])), 768, var);
  }
}
```

### Expected behavior

Should not split TMA into 4 parts and should use 1D TMA. The expected TMA size is not right as well.

### Additional context

_No response_

mygitljf:
Fixes #2180.

When i was looking at the issue, the 4-way split came from the 1D bulk-copy eligibility being gated on the same flag the descriptor-based 2D path uses, even though the 1D path doesn't actually need that gate. I narrowed it to a single safety check that only matters for the descriptor variant. While testing that on H100, i realised the gate was also catching legitimately out-of-bounds outer slices for free, so the second commit tightens the 1D eligibility check to keep those on the 2D descriptor path where the hardware clamp can do its job. Together with 4 regression tests (3 positive, 1 negative).

## Verification

Pre-patch generated source for the issue's repro (extracted from the codegen string):

```cpp
extern "C" __global__ void __launch_bounds__(256, 1)
gemm_kernel(__grid_constant__ const CUtensorMap A_desc, int M) {
  ...
  if (tl::tl_shuffle_elect<256>()) {
    mbar[0].expect_transaction(4096);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[  0])),   0, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[256])), 256, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[512])), 512, var);
    tl::tma_load(A_desc, mbar[0], (&(a_shared[768])), 768, var);
  }
}
```

Post-patch:

```cpp
extern "C" __global__ void __launch_bounds__(256, 1)
gemm_kernel(float* __restrict__ A, int M) {
  ...
  mbar[0].expect_transaction(4096);
  tl::tma_load((&(a_shared[0])), (&(A[(((int64_t)var) * (int64_t)1024)])), mbar[0], 4096);
}
```

A single 1D `cp.async.bulk` issue, no `CUtensorMap` descriptor — matches the expected behavior in the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined bulk-copy contiguity checks so certain 1D copy cases are correctly treated as contiguous.
  * Relaxed gating for layout-dependent TMA so descriptor-based TMA considerations occur in more valid layout cases.

* **Tests**
  * Added TileLang CUDA tests covering 1D TMA lowering for full-row variants and an out-of-bounds fallback that verifies use of the 2D descriptor path.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2272?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/cuda/op/copy_analysis.cc
+++ b/src/backend/cuda/op/copy_analysis.cc
@@ -227,6 +227,16 @@ bool CheckBulkCopy1D(const Buffer &global_tensor, const Buffer &shared_tensor,
                                   global_range[i]->min == 0,
                               arith::ProofStrength::kSymbolicBound)) {
         global_not_full_dim_encounter = true;
+        bool single_row = analyzer->CanProve(
+            global_range[i]->extent == 1, arith::ProofStrength::kSymbolicBound);
+        bool provably_in_bounds =
+            analyzer->CanProve(global_range[i]->min + global_range[i]->extent <=
+                                   global_tensor->shape[i],
+                               arith::ProofStrength::kSymbolicBound);
+        if (!single_row && !provably_in_bounds) {
+          global_is_contiguous = false;
+          break;
+        }
       }
     } else {
       if (!analyzer->CanProve(global_range[i]->extent == 1,
@@ -493,12 +503,6 @@ CopyFacts AnalyzeCopyFacts(const CopyNode &op, const CopyAnalysisContext &ctx) {
   const LayoutMap &layout_map =
       ctx.layout_map != nullptr ? *ctx.layout_map : empty_layout_map;
   bool is_cutedsl = TargetIsCuTeDSL(ctx.target);
-  // Issue #2180: only the descriptor-based 2D TMA path needs the OOB gate.
-  // The 1D bulk-copy path emits `cp.async.bulk`, which has the same OOB
-  // semantics as plain T.copy(); gating it on `buffer_oob` causes
-  // InferLayout to fall through to the 2D path for dynamic-outer-shape
-  // tensors and install a swizzle-shaped shared layout, which then forces
-  // Lower() into LowerBulk and triggers the 256-element splitting.
   facts.layout_dependent_tma_available = facts.has_layout_map && !is_cutedsl;
 
   if (facts.layout_dependent_tma_available) {
--- a/src/backend/cuda/op/copy_analysis.cc
+++ b/src/backend/cuda/op/copy_analysis.cc
@@ -493,8 +493,13 @@ CopyFacts AnalyzeCopyFacts(const CopyNode &op, const CopyAnalysisContext &ctx) {
   const LayoutMap &layout_map =
       ctx.layout_map != nullptr ? *ctx.layout_map : empty_layout_map;
   bool is_cutedsl = TargetIsCuTeDSL(ctx.target);
-  facts.layout_dependent_tma_available =
-      facts.has_layout_map && !is_cutedsl && !ctx.buffer_oob;
+  // Issue #2180: only the descriptor-based 2D TMA path needs the OOB gate.
+  // The 1D bulk-copy path emits `cp.async.bulk`, which has the same OOB
+  // semantics as plain T.copy(); gating it on `buffer_oob` causes
+  // InferLayout to fall through to the 2D path for dynamic-outer-shape
+  // tensors and install a swizzle-shaped shared layout, which then forces
+  // Lower() into LowerBulk and triggers the 256-element splitting.
+  facts.layout_dependent_tma_available = facts.has_layout_map && !is_cutedsl;
 
   if (facts.layout_dependent_tma_available) {
     facts.can_bulk_load_1d =
--- a/testing/python/language/test_tilelang_language_tma_1d.py
+++ b/testing/python/language/test_tilelang_language_tma_1d.py
@@ -46,10 +46,47 @@ def run_elementwise_add(M, N):
         assert "tma_load" in code and "CUtensorMap" in code
 
 
+def _lower_issue_2180_kernel(K, dtype):
+    M = T.dynamic("M")
+
+    @T.prim_func
+    def gemm(A: T.Tensor([M, K], dtype)):
+        with T.Kernel(M, threads=256):
+            var = T.alloc_var(T.int32, init=0)
+            a_shared = T.alloc_shared(K, dtype=dtype)
+            mbar = T.alloc_barrier(256)
+            T.tma_copy(A[var, 0:K], a_shared, barrier=mbar)
+
+    artifact = tilelang.lower(gemm, target={"kind": "cuda", "arch": "sm_90a"})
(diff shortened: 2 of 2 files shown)


src/backend/cuda/op/copy_analysis.cc
```diff
@@ -227,6 +227,16 @@ bool CheckBulkCopy1D(const Buffer &global_tensor, const Buffer &shared_tensor,
                                   global_range[i]->min == 0,
                               arith::ProofStrength::kSymbolicBound)) {
         global_not_full_dim_encounter = true;
+        bool single_row = analyzer->CanProve(
```
LeiWang1999, line 230: This `single_row` exception looks too permissive. It allows any extent-1 outer
  slice to use raw 1D bulk copy even when the row is provably out of bounds, e.g.
  `A[H_full, 0:K]` or `A[-1, 0:K]`.

  For 1D `cp.async.bulk`, we no longer have the tensor-map descriptor/OOB
  handling, so a provably-OOB single row should not be eligible for
  `kBulkLoad1D`.

  Could we reject provably-OOB first, then only allow the single-row case when
  bounds are unknown or proven safe?

  Something like:

  ```cpp
  bool provably_oob =
      analyzer->CanProve(global_range[i]->min < 0,
                         arith::ProofStrength::kSymbolicBound) ||
      analyzer->CanProve(global_range[i]->min + global_range[i]->extent >
                             global_tensor->shape[i],
                         arith::ProofStrength::kSymbolicBound);

  if (provably_oob) {
    global_is_contiguous = false;
    break;
  }
```

File: src/backend/cuda/op/copy_analysis.cc [hunk 2]
```diff
@@ -493,8 +503,7 @@ CopyFacts AnalyzeCopyFacts(const CopyNode &op, const CopyAnalysisContext &ctx) {
   const LayoutMap &layout_map =
       ctx.layout_map != nullptr ? *ctx.layout_map : empty_layout_map;
   bool is_cutedsl = TargetIsCuTeDSL(ctx.target);
-  facts.layout_dependent_tma_available =
-      facts.has_layout_map && !is_cutedsl && !ctx.buffer_oob;
+  facts.layout_dependent_tma_available = facts.has_layout_map && !is_cutedsl;
```
LeiWang1999, line 506: This removes the `buffer_oob` gate for all layout-dependent 1D TMA eligibility,
  not just explicit `T.tma_copy`. That broadens the behavior for implicit
  `T.copy` as well, including shared->global stores.

  For #2180, we only need to relax the gate for explicit `T.tma_copy` where the
  user has asked for TMA semantics. For automatic `T.copy`, keeping the old
  conservative behavior seems safer unless we also add a complete 1D linear
  bounds check.

  Could this be scoped to explicit TMA?

  ```cpp
  facts.layout_dependent_tma_available =
      facts.has_layout_map && !is_cutedsl &&
      (!ctx.buffer_oob || facts.explicit_tma);
```

LeiWang1999 (changes requested):
Thanks for your valuable contribution, I left some comments :)

LeiWang1999:
Please add a negative test for a provably-OOB single-row slice, since the
  current `single_row` exception would appear to allow it:

  ```python
  T.tma_copy(A[H_full, 0:K], a_shared, barrier=mbar)

  This should not lower to pointer-form 1D tl::tma_load.

  It would also be useful to add/confirm a shared->global T.copy(..., A[var,
  0:K]) case so we know whether implicit stores are intentionally affected by
  this change.

LeiWang1999:
likely this is not a bug and has been resolved, closed :)

mygitljf:
Got it, that makes sense — the 2D descriptor path with HW clamp is the correct safety net when `var + 1 <= M` cannot be proven, so promoting to 1D for those cases would be moving correctness onto the user. 

If it'd be useful as a follow-up, I could put up a tiny doc PR adding an "1D TMA preconditions" note to the `T.tma_copy` docstring (or the TMA section under docs/) so the next personesn't go down this same path.

Thanks for your careful read!

PR #2281, opened 2026-05-26
[TIR][Transform] Revert per-epoch shared memory liveness

LeiWang1999:
## Summary

- Reverts `e84a0ee699c6a1b42cc2f24e28722dcd39343c4f`, which rewrote `MergeSharedMemoryAllocations` around per-epoch liveness.
- Restores the previous shared-memory reuse implementation after the new planner caused unsafe aliasing and excessive dynamic shared memory usage in existing kernels.

## Changes

- Remove the epoch graph and shared access analysis helpers introduced by the reverted change.
- Restore the earlier `merge_shared_memory_allocations` implementation and related thread sync behavior.
- Drop transform tests and the branch arena example that depended on the reverted epoch-graph planner.

## Validation

- `./format.sh`
- `PYTHONPATH=/weka-hg/prod/deepseek/permanent/wanglei/tilelang:$PYTHONPATH python testing/python/language/test_tilelang_language_atom_mma.py`
- `PYTHONPATH=/weka-hg/prod/deepseek/permanent/wanglei/tilelang:/weka-hg/prod/deepseek/permanent/wanglei/tilelang/examples/deepseek_v32:$PYTHONPATH python -m pytest examples/deepseek_v32/test_tilelang_example_deepseek_v32.py::test_example_sparse_mla_bwd -q -s`

## Notes

- The DeepSeek v32 sparse MLA backward test previously failed while setting dynamic shared memory to roughly 235 KB. With this revert, the test compiles and runs successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer shared-memory merging and reuse: certain shared buffers are now treated conservatively to avoid unsafe aliasing and sync issues.

* **Tests**
  * Added CUDA tests validating behavior with opaque extern calls and shared-memory sync.
  * Removed several older epoch/thread-sync test modules and one example demonstrating arena byte-inspection.

* **Chores**
  * Updated pass configuration keys (added aggressive shared-memory-merge option; removed a debug key).

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2281?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


LeiWang1999:
@TensorGlue-IEIT Sorry, we need to revert this PR for now because it breaks some kernels, for example testing/python/language/test_tilelang_language_atom_mma.py.

 The root issue is that the new planner reasons only about explicit shared-memory accesses in TIR, but MMA descriptors hide shared-memory dependencies behind handle values. Once a buffer address is encoded into a descriptor, later MMA
  instructions depend on that buffer even if they no longer mention it directly. The planner currently cannot model this hidden use, so it may end a buffer’s lifetime too early and illegally alias it with another shared buffer, causing
  operand corruption.

PR #2285, opened 2026-05-27
[TIR][Transform] Handle ragged SIMT copy partitioning

LeiWang1999:
## Summary

- Allow SIMT copy layout planning to use a smaller active thread set when the full block thread count would create a ragged non-bijective layout.
- Preserve full-block replication for parallel loops that touch local fragment buffers, so scalar or small fragments remain available across the block.
- Add regressions for both the 384-thread shared-copy case and the 96-thread fragment replication case.

## Changes

- Add active thread extent selection in the loop partitioner based on divisibility of the flattened loop size and vector width.
- Require active fragment partitions to divide the full block thread count, then replicate the base partition across the complete block.
- Teach parallel copy plan candidate selection to account for fragment full-block replication when choosing vector widths.
- Cover ragged shared-copy and fragment-replication layout inference cases in tests.

## Validation

- `./format.sh`
- `PYTHONPATH=$(pwd):${PYTHONPATH} python examples/deepseek_mhc/example_mhc_pre.py`
- `python -m pytest testing/python/transform/test_tilelang_transform_layout_inference.py -x`
- `python -m pytest testing/python/transform/test_tilelang_transform_lower_tile_op.py -x`

## Notes

- Shared/global copies may predicate inactive block threads when a smaller active thread set avoids a ragged layout.
- Fragment loops keep a layout over the full block by replicating a compatible active partition instead of predicating away fragment-owning threads.

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust thread/vector validation to avoid incompatible or ragged layouts; clearer failures and debug diagnostics when adjustments are needed.
  * Layout inference now correctly tracks bind-variable-based buffer accesses, improving fragment detection and preventing missed layouts.

* **Chores**
  * Loop-partitioning and planning enhanced to use analyzer-assisted proofs and a replication flag to ensure correct active-thread selection and full-block replication.

* **Tests**
  * Added tests for ragged thread counts and fragment-layout replication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2285?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/cuda/op/copy.cc
+++ b/src/backend/cuda/op/copy.cc
@@ -335,7 +335,7 @@ private:
   static Layout ComputeLinearLayout(const Buffer &shared_tensor);
 
   static void CollectFragmentLayouts(const PrimExpr &expr,
-                                     const Map<Var, PrimExpr> &let_var_to_expr,
+                                     const Map<Var, PrimExpr> &bind_var_to_expr,
                                      const LayoutMap &existing_layouts,
                                      PrimExpr thread_extent,
                                      Range thread_bounds,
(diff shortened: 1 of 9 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2286, opened 2026-05-27
Scalarize vectorized math intrinsics on HIP

kashif:
On HIP, lowering rules now match, which is enough for scalar math, but anything that vectorizes a transcendental still breaks. E.g. `exp2(g)` where `g` is a `float32x4`, and that blows up  at codegen:

```
tvm.error.InternalError: Unresolved call ir.Op(name="tirx.exp2")
    at target/source/codegen_c.cc:771  (CodeGenTileLangHIP -> CodeGenC)
```

The reason is that HIPMath/FloatSuffix only returns a name for scalar float/double. For a `float32x4,` they return an empty string, so `DispatchPureExtern` leaves the call untouched, and the source codegen later trips over the unlowered op. HIP has no `exp2(float4)`. `exp2f` is scalar-only (confirmed with hipcc; only `h2exp2` for `__half2` is packed), so there's nothing to lower it to directly.

This PR adds `DispatchPureExternScalarized<T>`: when the call is a vector, it pulls each lane out with `Shuffle::ExtractElement`, lowers it to the scalar extern, and re-packs the results with a `Shuffle`. Scalars take the existing `DispatchPureExtern` path unchanged. The float-math ops (exp, exp2, log, sqrt, sin, cos, …) are switched over to it; integer/warp/popcount ops are left alone. This mirrors what the CUDA backend already does inside `PrintCallExtern`, just done in the lowering rule instead.

Tested on gfx1150 (Radeon 890M, ROCm 7.2):  A build/run on a larger AMD GPU would be a good extra check.

## Summary by CodeRabbit

* **Bug Fixes / Improvements**
  * Improved HIP code generation for many floating-point math intrinsics on packed/vector data (trig, exp, log, pow, sqrt, fmod), ensuring vector inputs are scalarized per lane for correct lowering.
* **Tests**
  * Added a regression test that verifies vectorized math is lowered into per-lane scalar calls in generated HIP source.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2286?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/backend/rocm/codegen/intrin_rule_hip.cc
+++ b/src/backend/rocm/codegen/intrin_rule_hip.cc
@@ -14,35 +14,36 @@ namespace codegen {
 namespace intrin {
 // Add float suffix to the intrinsics, HIP fast math.
 using tirx::FLowerIntrinsic;
-using tirx::Shuffle;
 
 // HIP has no vectorized fp32 math builtins: exp2f etc. are scalar
-// (float exp2f(float)), with no exp2(float4)/float2 overload (only half2
-// has packed intrinsics like h2exp2). HIPMath/FloatSuffix, therefore return
-// no name for non-scalar dtypes, leaving the op unlowered. Scalarize: lower
-// each lane to the scalar extern and re-pack with a shuffle; scalars fall
-// through to DispatchPureExtern unchanged. (CUDA does the equivalent in
-// codegen's PrintCallExtern; doing it in the lowering rule keeps HIP simple.)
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@kashif would you mind appending a test for this fix？

kashif:
sure will do! I'll also test on an AMD CI too

LeiWang1999 (approved):

PR #2288, opened 2026-05-27
[Transform] Prefer full-thread loop partitioning

LeiWang1999:
## Summary

- Prefer full-block loop partitioning by reducing vector width before falling back to active-thread subsets.
- Preserve the ragged SIMT copy fix for thread counts that cannot be partitioned by the full block.
- Add coverage for the gather-like int8 fragment copy case that should keep the original full-thread strategy.

## Changes

- Split plan candidate selection into a full-thread search followed by an active-thread fallback.
- Restart fallback search from the initial vector size only when full-block partitioning has no valid choice.
- Add a regression that checks CUDA source generation stays on the 16-byte vector path for the 256-thread gather-shaped case.

## Validation

- `python -m pytest testing/python/transform/test_tilelang_transform_layout_inference.py -x`
- `python /nfs-df/prod/data/permanent/wanglei/tilelang/debug/0527_bug/gather.py`

## Notes

- Active-thread fallback remains available for genuinely ragged thread counts, such as the 384-thread SIMT copy case where no full-block vector width can evenly partition the loop.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced vectorization sizing strategy in parallel operations to better preserve original layout patterns and optimize fallback behavior.

* **Tests**
  * Added test case validating kernel code generation with vectorization strategy selection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2288?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -672,9 +672,10 @@ Fragment ParallelOpNode::ComputePlanCandidate(const LayoutInferArgs &T) const {
   // As the pass will do post processing to the layout
   auto maybe_remapped_root_ =
       IfBufferRemapLoopGenerator::run(root_, T.buffer_remap, T.layout_map);
-  int vector_size =
+  int initial_vector_size =
       GetVectorizeSize(maybe_remapped_root_, T.analyzer, T.layout_map);
-  DLOG(INFO) << "[PlanLoopPartition] vector_size = " << vector_size << '\n';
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2319, opened 2026-06-02
[BugFix][CuTeDSL] Complete TileKernels benchmark support

JayceSu98:
## Summary

This PR completes several CuTeDSL backend paths that are required for TileKernels workloads to compile and run with `TILELANG_TARGET=cutedsl`.

The changes are kept in TileLang's CuTeDSL backend and adapter layers. They are not TileKernels source workarounds, H100-specific special cases, or fixes that depend on a single CUDA/CUTLASS DSL version combination. TileKernels is the coverage source that exposed the missing backend behavior.

## Background

Running the TileKernels benchmark suite with the CuTeDSL backend on H100 exposed failures across packed quantization, MoE mapping, generated CUDA-template math helpers, and cubin wrapper generation.

Representative failures included:

- missing CuTeDSL lowering for TileLang/CUDA-style vote and shuffle helpers such as `shfl_xor`, `match_any`, and `popc`;
- invalid FP4/FP8 scalar and vector conversion paths, especially for packed/sub-byte storage;
- eager conditional lowering through `tl.where`, which evaluates both branches and can issue unsafe predicated loads;
- conditional branch dtype mismatches in generated CuTeDSL expressions;
- unsupported lane-wise load/store/copy patterns for sub-byte vectors;
- generated helper calls such as `__habs`, `__float2half_rz`, `log1p`, and `isfinite` not being available in `tilelang.contrib.cutedsl.math`;
- fake tensor construction in the cubin wrapper receiving zero-sized dynamic dimensions even though those tensors are compile-time placeholders.

These are generic CuTeDSL backend gaps. Any workload that emits the same TileLang/TIR shapes can hit them; TileKernels provides a dense reproducer set.

After the backend fixes were pushed, the ordinary CUDA CI job also exposed an unrelated flaky BF16 reduce test. The test seeded only at module import time, so `pytest-xdist` worker scheduling could change the random input consumed by `test_reduce[sum-bfloat16-64x128-f2f-t256-b4]`. One CI run sampled a small-reference BF16 sum where the expected packed-BF16 reduction error crossed the existing tolerance by a single element.

## Root Causes

- Predicated `BufferLoad` values can be legalized as load-or-zero expressions. Lowering them with eager `tl.where` is unsafe because both branches are evaluated before selection.
- Safe-memory legalization can bind the zero branch through a local zero-like variable, so the codegen must recognize zero-like variables as well as literal zero expressions.
- CuTeDSL does not support every scalar FP8/FP4 conversion form directly; scalar conversions need to bridge through aligned vector/tensor casts where required by CuTeDSL/NVGPU lowering.
- Sub-byte values such as `Float4E2M1FN` cannot be handled as ordinary lane-addressable scalar storage.
- The CuTeDSL Python backend needs to provide the CUDA-template math helper names that TileLang-generated code can call.
- CuTeDSL fake compact tensor construction requires positive placeholder extents, while runtime inputs may legitimately contain zero dynamic dimensions.
- The BF16 reduce test input depended on worker-local torch RNG state instead of the parameterized case itself, making the test sensitive to xdist ordering.

## Changes

- Add/lower the missing CuTeDSL helper operations needed by the TileLang backend for vote/shuffle-style codegen and generated math wrappers.
- Preserve lazy predicated-load semantics for load-or-zero paths instead of routing them through eager `tl.where`.
- Recognize zero-like variables emitted by safe-memory legalization.
- Add FP8 scalar bridge lowering through aligned vector/tensor casts.
- Keep packed/sub-byte vector stores in supported TensorSSA or packed-storage form instead of forcing unsupported lane-wise scalar accesses.
- Add dtype-aware CuTeDSL wrapper plumbing for generated module arguments and C++ expression rendering.
- Replace the fragile single-counter software grid barrier with a counter plus generation sense barrier.
- Clamp only cubin fake tensor placeholder shapes to positive extents so empty runtime inputs can still compile without changing actual runtime tensor metadata.
- Make the parameterized reduce test input schedule-independent by deriving a deterministic seed from the case parameters and using a local `torch.Generator`.

## Validation

Environment used for local validation:

- GPU: NVIDIA H100 PCIe, `sm_90`, 114 SMs
- CUDA: 13.2
- Python: 3.12.3
- PyTorch: 2.12.0+cu132
- CUTLASS DSL libs: `nvidia-cutlass-dsl-libs-base==4.5.0`
- TileLang target: `TILELANG_TARGET=cutedsl`
- TileKernels benchmark backend: event timing
- Parallelism: `pytest -n 2`

Validation commands/results:

- `ruff check tilelang/contrib/cutedsl tilelang/jit/adapter/cutedsl tilelang/jit/adapter/utils.py`
- `python -m py_compile tilelang/contrib/cutedsl/math.py`
- `python -m py_compile tilelang/jit/adapter/cutedsl/wrapper.py`
- `git diff --check`
- Original failing MoE packed-UE8M0 benchmark node: passed locally with `-n 2`
- Representative quant/cast-back benchmark failures: passed locally
- Full TileKernels benchmark suite with CuTeDSL backend: `10165 passed, 160 warnings in 2:44:58`
- `testing/python/language/test_tilelang_language_reduce.py` with `--numprocesses=8`: `52 passed, 1 skipped`
- Failed BF16 reduce node repeated 10 consecutive times: passed

## Summary by CodeRabbit

* **New Features**
  * Added math ops (log1p, abs2, max2, min2, isfinite), warp intrinsics (__shfl_xor_sync, __match_any_sync, __popc), and exposed sync_warp/runtime_assert.
  * Expanded tensor casting and export of cast_tensor (FP16 narrowing, Float4E2M1FN) and broader codegen support for FP8 and vectorized ops.

* **Bug Fixes**
  * Fixed grid-wide barrier to avoid missed releases; improved cubin/dtype handling and tolerant handling of missing dynamic tensor refs.

* **Refactor**
  * Type-preserving min/max and consistent cumulative init values.

* **Tests**
  * Made reduce tests deterministic by seeding input generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tile-ai/tilelang/pull/2319?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tilelang/contrib/cutedsl/grid_sync.py
+++ b/tilelang/contrib/cutedsl/grid_sync.py
@@ -43,33 +43,52 @@ def _find_gpu_module_from_ip():
 def _ensure_global_counter(loc=None):
     """Declare the global counter variable at gpu.module level if not already done.
 
-    Creates: @__grid_sync_ctr = internal global i32 0, addrspace(1)
+    Creates:
+    - @__grid_sync_ctr = internal global i32 0, addrspace(1)
+    - @__grid_sync_gen = internal global i32 0, addrspace(1)
     """
     gpu_mod = _find_gpu_module_from_ip()
     if gpu_mod is None:
         raise RuntimeError("Cannot find gpu.module in MLIR hierarchy for global variable declaration")
 
-    # Check if the global already exists
+    # Check if the globals already exist.
     module_body = gpu_mod.regions[0].blocks[0]
+    found_ctr = False
+    found_gen = False
     for op in module_body.operations:
         if op.name == "llvm.mlir.global":
             sym = op.attributes.get("sym_name")
-            if sym is not None and str(sym) == '"__grid_sync_ctr"':
-                return
+            if sym is not None:
+                if str(sym) == '"__grid_sync_ctr"':
+                    found_ctr = True
+                elif str(sym) == '"__grid_sync_gen"':
+                    found_gen = True
+    if found_ctr and found_gen:
+        return
 
     # Insert at the beginning of the module body
     ctx = mlir_ir.Context.current
     linkage_attr = mlir_ir.Attribute.parse("#llvm.linkage<internal>", ctx)
 
     with mlir_ir.InsertionPoint.at_block_begin(module_body):
-        llvm.GlobalOp(
-            T.i32(),
-            "__grid_sync_ctr",
-            linkage_attr,
-            value=mlir_ir.IntegerAttr.get(T.i32(), 0),
-            addr_space=1,
-            loc=loc,
-        )
+        if not found_gen:
+            llvm.GlobalOp(
+                T.i32(),
+                "__grid_sync_gen",
+                linkage_attr,
+                value=mlir_ir.IntegerAttr.get(T.i32(), 0),
+                addr_space=1,
+                loc=loc,
+            )
+        if not found_ctr:
+            llvm.GlobalOp(
+                T.i32(),
+                "__grid_sync_ctr",
+                linkage_attr,
+                value=mlir_ir.IntegerAttr.get(T.i32(), 0),
+                addr_space=1,
+                loc=loc,
+            )
 
 
 def sync_grid():
@@ -94,19 +113,28 @@ def _grid_sync_barrier(*, loc=None, ip=None) -> None:
     """
     _ensure_global_counter(loc=loc)
 
-    # Get address of the global counter (ptr in address space 1 = global memory)
+    # Get addresses of the global barrier state (ptr in address space 1 = global memory)
     counter_ptr = llvm.mlir_addressof(
         llvm.PointerType.get(1),
         "__grid_sync_ctr",
         loc=loc,
         ip=ip,
     )
+    generation_ptr = llvm.mlir_addressof(
+        llvm.PointerType.get(1),
+        "__grid_sync_gen",
+        loc=loc,
+        ip=ip,
+    )
 
     # All barrier logic in one inline PTX block:
-    # - Thread 0 check, grid dim computation, atomic increment, spin-wait, reset
+    # - Thread 0 check, grid dim computation, atomic increment, sense wait
+    # - The last arriving block resets the counter and advances generation.
+    #   Other blocks wait for generation to change, so they cannot miss release
+    #   if the counter is reset before they observe the final arrival count.
     llvm.inline_asm(
         None,
-        [counter_ptr],
+        [counter_ptr, generation_ptr],
         """
 {
     .reg .s32 %r_tid;
@@ -115,7 +143,10 @@ def _grid_sync_barrier(*, loc=None, ip=None) -> None:
     .reg .s32 %r_nctaid_z;
     .reg .s32 %r_num_blocks;
     .reg .s32 %r_arrived;
+    .reg .s32 %r_generation;
+    .reg .s32 %r_new_generation;
     .reg .pred %p_is_thread0;
+    .reg .pred %p_is_last;
     .reg .pred %p_done;
 
     // Check if this is thread 0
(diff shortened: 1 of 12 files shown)


LeiWang1999:
cc @cherichy @lucifer1004

lucifer1004:
@JayceSu98 Could you also fix the failed example please? Thanks!

tilelang/contrib/cutedsl/grid_sync.py
lucifer1004, line 1: Could you explain a bit why did you add the new `grid_sync_gen`?

cherichy: It is used in multi-round grid_sync. `grid_sync_gen` is just like the phase in mbarrier. Right?

JayceSu98: > Could you explain a bit why did you add the new `grid_sync_gen`?

Yes, this is essentially the same idea as the phase/generation field in an mbarrier-style protocol.

  The previous implementation only used a single global counter, `__grid_sync_ctr`, to count how many CTAs had reached the grid barrier. That is enough to detect the last arriving CTA for one simple barrier round, but
  it is not sufficient for a reusable barrier.

  The issue is that the counter has an ABA/lost-wakeup problem across multiple `sync_grid()` calls. In each round the counter goes from `0 -> num_blocks -> 0`. The last arriving CTA resets the counter to 0 before the
  next round can start. Other CTAs may not reliably distinguish whether they are observing the counter before release, after reset, or during a later barrier round. In the worst case, a CTA can miss the short-lived
  `counter == num_blocks` state, or a later round can reuse the same counter value and confuse waiters from the previous round.

  `__grid_sync_gen` fixes that by making the barrier release edge explicit. Each CTA captures the current generation before publishing its arrival. The last arriving CTA resets the counter and increments the generation.
  Non-last CTAs wait for the generation to change, rather than waiting on a transient counter value. This makes the release condition stable and separates different rounds of `sync_grid()`.

  So the roles are:

  - `__grid_sync_ctr`: counts CTA arrivals in the current barrier round.
  - `__grid_sync_gen`: identifies the current barrier generation/phase.
  - Last arriving CTA: resets the counter and advances the generation.
  - Other CTAs: wait until `generation != old_generation`.

  This was added together with the device-scope fences because the barrier also needs correct memory ordering. Since only thread 0 publishes the CTA arrival, all threads in the CTA need a device/global-scope fence
  before thread 0 publishes, otherwise another CTA could observe the barrier release before prior global writes from non-thread-0 lanes are visible. After thread 0 observes the grid-level release, the CTA-level sync and
  acquire-side fence make sure all threads return from `sync_grid()` only after the protected global writes are ordered.

  In short, `grid_sync_gen` is not just extra state. It turns the single-counter barrier into a reusable generation-based barrier and avoids counter-reset/lost-wakeup issues when `sync_grid()` is used more than once in
  the same kernel.

cherichy:
@JayceSu98 Thanks for the fixes. My main concern is that the `BufferLoad` and `if_then_else` lowering changes may affect generated code and performance. Could you share targeted before/after perf data for kernels that exercise these two paths, ideally with generated code comparison as well?

JayceSu98:
Thanks for raising this. I ran targeted checks for the two lowering paths rather than relying only on the full benchmark result.

A direct `origin/main` before/after number is not available for these affected TileKernels/CuTeDSL cases because the pre-PR backend does not compile/run them:

- `expand_to_fused_with_sf` fails before timing while preparing the packed-SF input:
  `TypeError: Dynamic symbolic var in_sf_stride refers to a non-tensor param at index 1`
- `per_token_cast` packed FP8 -> FP4 fails in CuTeDSL codegen:
  `Should not print a non-scalar type in CuTeDSL: float16x2`

To isolate the performance impact of the BufferLoad / conditional-lowering refinements themselves, I compared the first runnable benchmark-support commit against the final benchmark-fix head:

- before: `1aa67ed3` (`[BugFix] CuTeDSL backend support for TileKernels benchmarks`)
- after: `62786ddb` (`[BugFix] Avoid zero-sized fake tensors in CuTeDSL cubin generation`)
- environment: H100 PCIe / `sm_90`, `TILELANG_TARGET=cutedsl`, `TK_BENCHMARK_BACKEND=event`, single pytest worker, benchmark fixture default `rep=30`

| Case | Exercises | Before | After | Delta |
| --- | --- | ---: | ---: | ---: |
| `expand_to_fused_with_sf`, packed UE8M0 SF, `hidden=576`, `num_topk=2`, `num_experts=4`, `num_ep_ranks=64` | predicated/safe `T.copy` and conditional lowering | 17.44 us, 682.45 GB/s | 17.67 us, 673.28 GB/s | +1.3% time |
| `per_token_cast`, `in_dtype=e4m3`, `fmt=e2m1`, packed UE8M0 SF, `hidden=6144`, `x_block_size=(32,32)` | BufferLoad + zero conditional + sub-byte vector/TensorSSA path | 124.10 us, 597.64 GB/s | 113.56 us, 653.15 GB/s | -8.5% time |

For generated code comparison:

- In the `expand_to_fused_with_sf` generated kernel, the hot kernel body is unchanged in this before/after comparison. The only generated source difference is in fake compile tensor construction, where the final version wraps fake tensor shapes with `_positive_fake_shape(...)` before `make_fake_compact_tensor(...)`. That is outside the launched kernel body.

- In the `per_token_cast` generated kernel, the old code lowered the predicated load through eager `tl.where`, which materializes the load branch:

```python
_tsa = tl.make_rmem_tensor((1,), cutlass.Boolean)
_tsa[0] = (((blockIdx.x * 2) + i) < num_tokens)
_where = tl.where(
    _tsa.load(),
    tl.make_tensor_at_offset(x.iterator, ...).load(),
    tl.make_filled_tensor((16,), cutlass.Float8E4M3FN(broadcast_var)).load(),
)
tl.make_tensor_at_offset(...).store(_where)
```

The final code initializes the result tensor to zero and guards only the real load with an `if`, so the out-of-bounds load is not evaluated:

```python
_if_tensor = tl.make_rmem_tensor((16,), cutlass.Float8E4M3FN)
# lanes initialized to zero
if (((blockIdx.x * 2) + i) < num_tokens):
    _if_tensor.store(tl.as_tensor_ssa(tl.make_tensor_at_offset(x.iterator, ...).load()))
tl.make_tensor_at_offset(...).store(_store_tensor.load())
```

The same generated kernel also shows the Tensor/TensorSSA normalization change around vector casts, e.g.:

```python
tl.cast_tensor(_aload.load(), cutlass.Float16)
```

becomes:

```python
tl.cast_tensor(tl.as_tensor_ssa(_aload), cutlass.Float16)
```

So the change is not adding a generic extra branch to every load. It is targeted at the `zero branch + BufferLoad branch` pattern produced by safe-memory/predicated copy lowering. In the targeted measurements above, the safe-copy kernel is effectively flat, and the packed FP8 -> FP4 path improves because it avoids the eager conditional load path and uses the TensorSSA form CuTeDSL expects.

LeiWang1999:
cc @cherichy

cherichy (approved):
Thanks @JayceSu98 for your contribution, these fixes can be merged when CI passes.

JayceSu98:
@LeiWang1999  Could you please re-run or approve the remaining CI jobs when convenient? I’ll follow up if any failure still reproduces.

PR #2322, opened 2026-06-03
[BugFix] Complete CuTeDSL support for TileKernels full functional tests

JayceSu98:
## Summary

This PR completes the CuTeDSL backend fixes needed for the TileKernels full functional correctness suite on H100.

This branch currently contains 8 commits relative to `origin/main`. The first 3 are the benchmark-oriented CuTeDSL fixes from the earlier TileKernels benchmark work. The last 5 commits are the new full-functional correctness fixes.

Commits on this branch:

- `1aa67ed3` `[BugFix] CuTeDSL backend support for TileKernels benchmarks`
- `f90559de` `[BugFix] Complete CuTeDSL support for TileKernels benchmarks`
- `62786ddb` `[BugFix] Avoid zero-sized fake tensors in CuTeDSL cubin generation`
- `f83c2c0c` `[BugFix] Complete CuTeDSL scan lowering for TileKernels`
- `cb07934d` `[BugFix] Resolve CuTeDSL dynamic shapes from live tensors`
- `7bdf491a` `[BugFix] Permit dead optional CuTeDSL stride symbols`
- `fc084420` `[BugFix] Fix CuTeDSL optional ABI and early-return codegen`
- `d4ddd297` `[BugFix] Promote narrow CuTeDSL shifts before wide casts`

## Background

TileKernels exercises a wider set of CuTeDSL code paths than the existing TileLang unit tests. After the benchmark-oriented CuTeDSL support, the full TileKernels functional suite still exposed correctness and lowering gaps around prefix scans, dynamic symbolic shape resolution, optional tensor ABIs, early-return control flow, mutable vector operands, and integer shift promotion for packed quantization formats.

These fixes are not H100-only workarounds, TileKernels source specializations, or version-specific shims. The H100/CUDA/CuTeDSL environment is the reproducer that exposes the missing backend coverage. The underlying problems are semantic mismatches between TileLang TIR lowering and the current CuTeDSL Python/codegen path.

### Complete scan lowering for TileKernels

TileKernels MoE routing uses `T.cumsum` over shared buffers with more than one warp-sized segment. The previous CuTeDSL `CumSum1D` path only handled a single warp-local scan, so expert offsets reset at later 32-element chunks and routing positions overlapped.

This PR aligns the CuTeDSL scan helpers with the CUDA inclusive scan behavior by scanning 32-element segments with carry propagation. The same helper is used for `CumSum1D`, `CumMax1D`, and 2D row/column scan paths. The codegen side also adds shared index linearization support for non-flat buffer accesses and lowers `tl.match_any_sync` through the existing CuTeDSL warp intrinsic.

### Resolve dynamic shapes from live tensors

The CuTeDSL adapter previously recorded only the first shape/stride source for each dynamic symbol. That is unsafe when the first source belongs to an optional tensor that is `None` at runtime. In TileKernels `reduce_fused`, this caused `num_tokens` to resolve to `0` even though another live tensor carried the correct shape, resulting in an invalid zero-grid launch.

The adapter now records all candidates for a dynamic symbol and resolves runtime values from the first candidate backed by a live `torch.Tensor`. Missing shape symbols still fail clearly instead of silently using an unsafe value.

### Permit dead optional stride ABI symbols

Some optional tensor variants keep stride-only ABI symbols even after the optional tensor has been pruned out of the lowered device kernel. These symbols are ABI placeholders rather than values used for output allocation or launch dimensions.

This PR keeps shape symbols strict, but allows stride-only dead optional symbols to use a zero placeholder so the generated host call signature remains consistent without weakening dynamic shape correctness.

### Fix optional ABI handling and early-return codegen

TileKernels `top2_sum_gate` exposed three additional issues:

- Optional tensors compiled out by static flags could still leave host ABI dynamic shape arguments.
- The early-return rewriter only handled a bare `thread_return`, missing the common pattern where stores occur before the return.
- Vector binary codegen reused SSA aliases for operands containing mutable `BufferLoad` nodes, so later stores were not reflected in subsequent vector expressions.

The fix preserves strict output allocation, allows dead host ABI dynamic placeholders where appropriate, preserves pre-return stores before guarding fallthrough statements, and avoids SSA alias reuse for vector operands that contain buffer loads.

### Promote narrow shifts before widened casts

TileKernels E5M6 quantization packs truncated FP16 values into `uint32` words using expressions such as `T.cast(half_u16[i] << 20, T.uint32)`. CUDA C++ applies integer promotion before the final cast, but CuTeDSL strongly typed integer wrappers previously shifted in the narrow `uint16` type and then widened, dropping the high bits.

When a narrow integer shift is immediately cast to a wider integer type, CuTeDSL codegen now promotes the left operand to the target integer type before shifting. This preserves the intended packed-bit construction and fixes the downstream E5M6 cast-back/per-token quantization corruption.

## Validation

Environment:

- H100 CUDA backend, `sm_90`
- `CUDA_VISIBLE_DEVICES=0` maps to `NVIDIA H100 PCIe` in the local validation environment
- `TILELANG_TARGET=cutedsl`
- `TILELANG_DISABLE_CACHE=1`
- Local TileLang build from this branch
- TileKernels functional suite

Focused TileLang/TileKernels validation covered:

- CuTeDSL scan/codegen tests
- CuTeDSL dynamic symbolic shape adapter regressions
- TileKernels MoE `get_fused_mapping`, `reduce_fused`, `expand_to_fused`, and `top2_sum_gate` representative failures
- TileKernels E5M6 quantization focused cases

Full TileKernels functional run with `-n 2` completed with:

```text
7 failed, 36736 passed, 10165 skipped
```

The remaining 7 failures were all `torch.OutOfMemoryError` in large `test_swiglu_backward_and_per_token_cast` configurations under two-worker concurrency. They were rerun on a single H100 worker and all passed:

```text
7 passed, 16 warnings in 114.34s
```

This indicates the remaining full-run failures are H100 memory pressure under concurrent execution, not CuTeDSL correctness failures.

## Release Notes

* **New Features**
  * Added math operations: `log1p`, `abs2`, `max2`, `min2`, and `isfinite`.
  * Introduced warp-level primitives: `__shfl_xor_sync`, `__match_any_sync`, and `__popc`.
  * Added `sync_warp` and `runtime_assert` utilities for device synchronization and assertions.

* **Improvements**
  * Enhanced FP8 and FP4 floating-point type support with improved casting and materialization.
  * Improved JIT compilation's dynamic shape and stride resolution for tensors.
  * Optimized grid synchronization with generation-based barrier protocol.
  * Extended reduction operations with improved cumulative sum and max implementations.

### DIFF
--- a/src/cuda/codegen/codegen_cutedsl.cc
+++ b/src/cuda/codegen/codegen_cutedsl.cc
@@ -54,6 +54,31 @@ static bool ContainsLoopBreak(const Stmt &stmt) {
   return det.found;
 }
 
+static bool IsThreadReturnEvaluate(const Stmt &stmt) {
+  const auto *eval = stmt.as<EvaluateNode>();
+  if (eval == nullptr) {
+    return false;
+  }
+  const auto *call = eval->value.as<CallNode>();
+  return call != nullptr && call->op.same_as(builtin::thread_return());
+}
+
+std::optional<PrimExpr> GetConditionalThreadReturnCondition(const Stmt &stmt) {
+  const auto *if_node = stmt.as<IfThenElseNode>();
+  if (if_node == nullptr || if_node->else_case) {
+    return std::nullopt;
+  }
+  if (IsThreadReturnEvaluate(if_node->then_case)) {
+    return if_node->condition;
+  }
+  if (const auto *seq = if_node->then_case.as<SeqStmtNode>();
+      seq != nullptr && seq->seq.size() == 1 &&
+      IsThreadReturnEvaluate(seq->seq[0])) {
+    return if_node->condition;
+  }
+  return std::nullopt;
+}
+
 // The threshold of the loop extent to use cutlass.range_constexpr
 // Higher values would lead to DSLOptimizationWarning:
 // This static loop has 128 iterations, which may be very slow to compile,
@@ -146,6 +171,7 @@ CodeGenTileLangCuTeDSL::CodeGenTileLangCuTeDSL() {
 
 void CodeGenTileLangCuTeDSL::InitFuncState_(const PrimFunc &f) {
   raw_pointer_vars_.clear();
+  zero_like_vars_.clear();
   CodeGenTileLangPY::InitFuncState_(f);
 }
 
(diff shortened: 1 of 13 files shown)


LeiWang1999:
Please focus on #2319 :) closed.

PR #2343, opened 2026-06-05
[Transform] Validate fragment write owner compatibility

LeiWang1999:
## Summary

- Reject `T.Parallel` fragment writes when the loop layout can execute non-owner threads for a fragment buffer element.
- Stabilize parallel layout inference by preserving first-access buffer order when choosing and validating inferred layouts.
- Prevent fp4 scale fragments from compiling into layouts that silently overwrite per-channel local scale bytes.

## Changes

- Track fragment buffer accesses in first-seen order instead of iterating an `unordered_map` during inference.
- Validate fragment writes with bidirectional owner-thread containment so buffer owners are covered and loop writers are all legal owners.
- Add a layout inference regression test for an explicitly annotated scalar-channel fragment layout that used to pass and overwrite local storage.

## Validation

- `python debug/repro_fp4_explicit_layout.py` (expected `Layout infer conflict`)
- `PYTHONUNBUFFERED=1 timeout 45s python debug/cache.py` (timed out after multiple `correctness test passed` iterations, with no assertion or traceback)

## Risk

- This intentionally makes some previously accepted fragment write layouts fail at compile time when loop writer threads are not exactly compatible with the buffer layout owner threads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened write-side compatibility validation for fragment buffers to ensure stricter correctness checks.

* **Tests**
  * Added validation test to detect invalid fragment writes from non-owner threads, improving error detection.

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -257,6 +257,7 @@ void ParallelOpNode::RecordBufferAccess(const Buffer &buffer,
     BufferAccessInfo info;
     info.indices = indices;
     it = indice_map_.emplace(buffer, std::move(info)).first;
+    access_order_.push_back(buffer);
   }
   if (is_write) {
     it->second.is_write = true;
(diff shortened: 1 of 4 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2345, opened 2026-06-05
[BugFix] Reject T.alloc_barrier() on pre-Hopper targets with a clear error

### LINKED ISSUE #2344: [BUG][Fuzzer][ice-on-invalid-code] T.alloc_barrier() emits sm_90-only code on pre-Hopper targets (sm_86), failing nvcc
### Required prerequisites

0.1.9

### System information

- Installed via: pip
- Python: 3.9, linux
- TileLang: 0.1.9
- PyTorch: 2.8.0+cu128
- CUDA / nvcc: 12.8
- GPU: NVIDIA A10G — compute capability **8.6 (sm_86, Ampere)**

### Problem description

`T.alloc_barrier()` is accepted by the frontend on any target, but codegen
unconditionally emits Hopper-only (sm_90) primitives — the `Barrier` type,
`tl::tl_shuffle_elect`, and `tl::fence_barrier_init` — with no compute-capability
check. On sm_86 these don't exist, so the kernel fails at `nvcc` instead of being
rejected up front. It looks like a missing static target-arch guard in the
barrier lowering / codegen path.

### Reproducible example code

A single `T.alloc_barrier()` call in an otherwise empty kernel is enough:

```python
import tilelang
import tilelang.language as T

@T.prim_func
def main(C: T.Tensor((1,), "float32")):
    with T.Kernel(1, threads=1) as bx:
        bar = T.alloc_barrier(1)   # the entire trigger
        C[0] = 0.0

mod = tilelang.compile(main, out_idx=-1, target="cuda")  # fails on sm_86
```

Generated CUDA:

```cpp
__shared__ __align__(16) uint64_t bar_mem[1];
auto bar = reinterpret_cast<Barrier*>(bar_mem);
  bar[0].init(1);
}
tl::fence_barrier_init();
```

### Traceback

```pytb
tvm_kernels.cu(14): error: identifier "Barrier" is undefined
tvm_kernels.cu(15): error: namespace "tl" has no member "tl_shuffle_elect"
tvm_kernels.cu(18): error: namespace "tl" has no member "fence_barrier_init"
```

### Expected behavior

A static guard on target compute capability: either reject `T.alloc_barrier()`
with a clear error when the target is `< sm_90`, or lower it to a pre-Hopper
fallback. The failure should surface at the TileLang level, not as undefined
identifiers from nvcc.

### Additional context

A PR that does early reject is prepared.

Hughshine:
## Summary

`T.alloc_barrier()` is lowered by `LowerSharedBarrier` to hardware mbarrier code
— the cutlass `Barrier` type, `tl::tl_shuffle_elect`, and
`tl::fence_barrier_init` — all of which only exist on sm_90+. On a pre-Hopper
target (e.g. sm_86) the frontend accepted the call silently and the failure only
surfaced later inside `nvcc` (before this fix):

```
tvm_kernels.cu(14): error: identifier "Barrier" is undefined
tvm_kernels.cu(15): error: namespace "tl" has no member "tl_shuffle_elect"
tvm_kernels.cu(18): error: namespace "tl" has no member "fence_barrier_init"
```

This adds a static compute-capability guard so the failure surfaces at the
TileLang level with an actionable message instead.

Fixes #2344.

## Root cause

`CUDAPassPipelineBody` runs `LowerSharedBarrier()` unconditionally, with no check
on the target architecture. The primitives it emits resolve to sm_90+ only
constructs — `Barrier` (`cutlass::arch::ClusterTransactionBarrier`) and
`fence_barrier_init` in `src/tl_templates/cuda/barrier.h`, and `tl_shuffle_elect`
(via `cute::elect_one_sync`) in `intrin.h` — so the generated code does not
compile on older targets.

## Fix

A static guard in `tilelang/cuda/pipeline.py`, immediately before
`LowerSharedBarrier`: if the target is `< sm_90` **and** the module actually uses
a shared barrier, raise a `ValueError` that names the offending primitives and
points to `__syncthreads()` / named barriers as the pre-Hopper alternative.

- `have_mbarrier(target)` — a new capability helper in `tilelang.contrib.nvcc`,
  alongside the existing `have_tma` / `have_pdl`; true only for cuda targets with
  compute major `>= 9`.
- `_module_has_shared_barrier(mod)` — walks each function's blocks and checks
  `buffer.scope()` of the `alloc_buffers` for `shared.barrier` /
  `shared.cluster_barrier` (the same way `LowerSharedBarrier` detects them), so
  kernels without a barrier are unaffected.

This is intentionally a reject (option a), not a pre-Hopper lowering fallback
(option b): `T.alloc_barrier()` is a Hopper feature the compiler normally
inserts itself, and the immediate problem is the missing capability check. A
fallback lowering can be a follow-up.

## Test plan

- [x] Added `testing/python/target/test_tilelang_alloc_barrier_target_guard.py`:
  - `test_alloc_barrier_rejected_on_pre_hopper_target` (`@requires_cuda`) —
    compiling a minimal `T.alloc_barrier(1)` kernel to an `sm_86` target raises
    `ValueError` mentioning `requires sm_90` / `sm_86` / the offending primitive.
  - `test_alloc_barrier_allowed_on_hopper_target`
    (`@requires_cuda_compute_version_ge(9, 0)`) — ensures sm_90 still compiles
    (no regression); skipped on pre-Hopper runners.
- [x] `ruff check` / `ruff format --check` clean on the changed files.
- [x] Ran the new test file on an sm_86 box (A10G): the reject test passes
  (full `tilelang.compile` raises the guard `ValueError`); the sm_90 test is
  skipped as expected on non-Hopper hardware. No regression on the sibling
  `target/` codegen tests.
- [x] `have_mbarrier` returns the right answer across sm_75/80/86/89 (reject)
  and sm_90/90a/100/100a (allow).

## Summary by CodeRabbit

* **New Features**
  * Added validation that prevents using shared/barrier allocation on CUDA targets lacking hardware support and provides a clear error indicating the required compute capability.

* **Tests**
  * Added tests verifying barrier-allocation rejection on older CUDA architectures, successful compilation without barriers, and successful barrier allocation on newer architectures.

### DIFF
--- a/testing/python/target/test_tilelang_alloc_barrier_target_guard.py
+++ b/testing/python/target/test_tilelang_alloc_barrier_target_guard.py
@@ -1,0 +1,71 @@
+import pytest
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+
+
+CUDA_SM86_TARGET = {"kind": "cuda", "arch": "sm_86"}
+CUDA_SM90_TARGET = {"kind": "cuda", "arch": "sm_90"}
+
+
+def _barrier_kernel():
+    """A minimal kernel whose only special feature is a shared barrier."""
+
+    @T.prim_func
+    def main(C: T.Tensor((1,), "float32")):
+        with T.Kernel(1, threads=1) as _:
+            bar = T.alloc_barrier(1)  # noqa: F841
+            C[0] = 0.0
+
+    return main
+
+
+def _no_barrier_kernel():
+    """The same minimal kernel without any barrier, used as a control."""
+
+    @T.prim_func
+    def main(C: T.Tensor((1,), "float32")):
+        with T.Kernel(1, threads=1) as _:
+            C[0] = 0.0
+
+    return main
+
+
+def _compile(target, kernel_factory=_barrier_kernel):
+    """Compile a freshly built kernel for ``target`` with caching disabled."""
+    tilelang.disable_cache()
+    try:
+        return tilelang.compile(kernel_factory(), out_idx=-1, target=target)
+    finally:
+        tilelang.enable_cache()
+
+
+@tilelang.testing.requires_cuda
+def test_alloc_barrier_rejected_on_pre_hopper_target():
+    """T.alloc_barrier() must be rejected with a clear error on sm_86 instead of
+    emitting sm_90-only code that fails later inside nvcc."""
+    with pytest.raises(ValueError) as exc_info:
+        _compile(CUDA_SM86_TARGET)
+
+    message = str(exc_info.value)
+    assert "requires sm_90" in message
+    assert "sm_86" in message
+    # Names the offending primitives so users know what to remove.
+    assert "tl_shuffle_elect" in message or "Barrier" in message
+
+
+@tilelang.testing.requires_cuda
+def test_no_barrier_compiles_on_pre_hopper_target():
+    """The guard must not fire for pre-Hopper kernels that use no barrier."""
+    _compile(CUDA_SM86_TARGET, kernel_factory=_no_barrier_kernel)
+
+
+@tilelang.testing.requires_cuda_compute_version_ge(9, 0)
+def test_alloc_barrier_allowed_on_hopper_target():
+    """The guard must not regress sm_90, where T.alloc_barrier() is valid."""
+    _compile(CUDA_SM90_TARGET)
+
+
+if __name__ == "__main__":
+    tilelang.testing.main()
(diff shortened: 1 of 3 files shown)


Hughshine:
@coderabbitai Thanks! Skipping the `major, _` nitpick on `have_mbarrier` (nvcc.py:607) on purpose: I'm keeping `major, minor = parse_compute_version(...)` to stay byte-for-byte consistent with the adjacent capability helpers in the same file — `have_tma`, `is_hopper`, and especially `have_pdl` right above it all bind `major, minor` even though only `major` is used. The repo's configured `ruff check` doesn't flag it, so this is style-only, and matching the neighbors reads better here than making this one helper differ.

LeiWang1999 (approved):

PR #2349, opened 2026-06-07
[Transform] Respect fragment write owner layouts

LeiWang1999:
## Summary

- Preserve whether a parallel layout source buffer comes from a write access so fragment writes can inherit existing owner-thread layouts.
- Fix a layout inference failure where scalar reducer accumulators with a thread-0 owner layout were rejected after stricter fragment write owner validation.
- Add a regression test covering scalar reduce accumulation followed by a thread-0 read.

## Changes

- Track whether `source_buffer` was selected from a write access in `ParallelOpNode::InferLayout`.
- Allow write-derived source buffers to compute the loop layout even when common layout propagation is disabled for constant-index fragments.
- Keep the stricter bidirectional fragment write owner compatibility check intact.

## Validation

- `python debug/dflash.py`

## Release Notes

* **Bug Fixes**
  * Improved layout propagation accuracy in parallel operations to better handle write access scenarios.

* **Tests**
  * Added test coverage for scalar reduction with accumulator layout inference.

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -422,6 +422,7 @@ LayoutMap ParallelOpNode::InferLayout(const LayoutInferArgs &T,
 
   // Step 1: try to infer loop's partition from a source fragment
   Buffer source_buffer, read_source_buffer;
+  bool source_buffer_is_write = false;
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #2358, opened 2026-06-08
[JIT] Include closure free vars in frontend cache key

hhy3:
tilelang now fail with code like this
```python
import torch
import tilelang
import tilelang.language as T


def build(N):
    @tilelang.jit
    def make_kernel():
        M = T.dynamic("M")

        @T.prim_func
        def copy_kernel(
            x: T.Tensor[(M, N), T.float32],
            y: T.Tensor[(M, N), T.float32],
        ):
            with T.Kernel(M, threads=128) as pid:
                for j in T.Parallel(N):
                    y[pid, j] = x[pid, j]

        return copy_kernel

    return make_kernel()


def run(N):
    kernel = build(N)
    x = torch.randn(4, N, device="cuda", dtype=torch.float32)
    y = torch.empty(4, N, device="cuda", dtype=torch.float32)
    kernel(x, y)
    torch.cuda.synchronize()


run(128)
run(256)
```

### DIFF
--- a/testing/python/cache/test_tilelang_kernel_cache_atomic_save.py
+++ b/testing/python/cache/test_tilelang_kernel_cache_atomic_save.py
@@ -264,6 +264,34 @@ def test_jit_frontend_cache_hit_skips_tir_elaboration(monkeypatch):
     assert "frontend_cached_kernel" in calls[0][0]["function"]
 
 
+def test_frontend_cache_key_includes_closure_vars():
+    import tilelang
+    import tilelang.language as T
+
+    def build(N):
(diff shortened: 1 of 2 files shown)


LeiWang1999:
Thanks @hhy3  for your contribution! It's quite interesting that this issue was triggered. Essentially, it's because we had a layer of fast cache (though not the fastest) shared among kernels, using the function's arguments as keys. But in hindsight, this level of cache is actually unnecessary. Thank you for your contribution — I will make a more fundamental fix.

PR #2361, opened 2026-06-09
[Backend] Cleanup Metal Codegen, split AsyncCopy lowering and common target utils

SiriusNEO:
## Summary by CodeRabbit

* **New Features**
  * CUDA: PTX async-copy lowering exposed as a CUDA-specific transform.
  * ROCm: Async-copy injection pass added for AMD GPUs.

* **Changes**
  * Backend target utilities reorganized into per-backend modules (CPU/CUDA/Metal/ROCm) and FFI surface reduced.
  * Old public transform now a deprecated alias pointing to the CUDA-specific entrypoint.

* **Tests**
  * Tests updated to call the CUDA-specific async-copy transform.


SiriusNEO:
@regression-perf

PR #2362, opened 2026-06-09
[Fix] Bind statement missing role marker

ppppqp:
## Problem
Fixes a compiler crash that occurs when scalar temporaries are introduced inside a pipelined loop. I ran into it when playing around tilelang puzzle 10-dequantization with the following pattern:
```python
for i, j in T.Parallel(BLOCK_K, BLOCK_N // 2):
    B_low = T.cast(B_local[i, j] & 0x0F, A_dtype) - 8.0
    B_high = T.cast((B_local[i, j] >> 4) & 0x0F, A_dtype) - 8.0
    B_dequantized[i, 2 * j] = B_low
    B_dequantized[i, 2 * j + 1] = B_high
```

Gives me this compile error:
```
  File "/home/Documents/workspace/tilelang/src/cuda/
  transform/multi_version_buffer_rewriter.cc", line 131, in
  tvm::tl::Role tvm::tl::WarpSpecializedRoleMarker_::GetRole(const
  tvm::tirx::StmtNode*) const
      ICHECK(it != map_.end())

  tvm.error.InternalError: Check failed: (it != map_.end()) is
  false:  Cannot find role for stmt: tirx.Bind
```

The root cause is that `B_low = ...` is lowered into `tirx.Bind` statements, but not correctly classified as consumer by `WarpSpecializedRoleMarker_` so that  `GetRole()` later crashes, since it can not find the entry in the map.

## Fix
Marking the bind statement as `Consumer` because it reads value but does not perform producer-side pipeline work.
Also patched a few other nodes that seem reasonable to me (bodyless leaf statement), so that the following does not break inside pipeline

- `T.alloc_buffer`
- `T.decl_buffer`
- `T.Assert`

## Test
Added a minimal reproducable test case that crashes on main, but passes after the patch.

## Summary by CodeRabbit

* **Tests**
  * Added CUDA tests for two bind-style kernels to validate pipelined shared-memory copies and producer-driven shared scalar use.

* **Refactor**
  * More precise classification of multi-versioned shared buffers via enhanced scalar-dependency propagation.
  * Improved producer-consumer movability analysis for safer, more accurate loop-prefix optimizations.


LeiWang1999 (dismissed):
I think this is a reasonable short-term workaround, but semantically `Bind` is not always a consumer. A `Bind` can also appear in a producer stage as an auxiliary scalar definition, for example computing an index or pointer used by a later global-to-shared copy.

The current change works because `Bind` itself does not write the producer buffer, and the later `copy_stage` check can still classify the enclosing statement as producer when there is an actual shared write / TMA load. So treating `Bind` as consumer avoids the missing-role crash without usually breaking producer detection.

Longer term, I think we should introduce a `Neutral` role. Statements like pure scalar `Bind`, `AllocBuffer`, `DeclBuffer`, and some wrappers should not force the enclosing statement to become consumer or both.

Then the aggregation logic can ignore neutral statements and let real buffer accesses decide whether the stage is producer or consumer.

ppppqp:
Thanks for the detailed explanation! Agreed this is not semantically accurate and should be a short term fix.

testing/python/language/test_tilelang_language_let.py
```diff
@@ -36,5 +36,26 @@ def test_bind_kernel():
...
+
+@tilelang.testing.requires_cuda
+def test_producer_bind_kernel():
+    kernel_source = producer_bind_kernel.get_kernel_source()
+    assert "tl::tma_load" in kernel_source
+
+
 if __name__ == "__main__":
-    tilelang.testing.main()
+    # tilelang.testing.main()
```
ppppqp: revert?

LeiWang1999:
@regression-perf

LeiWang1999 (approved):

PR #2364, opened 2026-06-09
[CUDA] Add SM120 NVF4 block-scale MMA support

qqq-tao:
## [CUDA] Add SM120 NVFP4 block-scale MMA support

This PR adds the first TileLang GEMM path for SM120(a) (GeForce Blackwell,
e.g. RTX 5090) NVFP4 block-scaled tensor cores:

```
packed FP4 A/B  +  UE4M3 scale factors (blockscaled_chunk_kmajor)
  -> T.mma_gemm_blockscaled
  -> mma.sync.aligned.m16n8k64.row.col.kind::mxf4nvf4.block_scale.scale_vec::4X
```

### Design (shaped by review feedback — thanks @Rachmanino)

- **One high-level op**: `T.mma_gemm_blockscaled(A, B, C, SFA, SFB, ...)`,
  following the same scale-factor model as `T.tcgen05_gemm_blockscaled`
  (users pass scale tensors, `k_start`, K granularity; lowering derives the
  addressing). The only new low-level intrinsic is `ptx_mma_block_scale`,
  mirroring the existing `ptx_mma`.
- **No separate emitter**: `TensorCoreIntrinEmitter` gains an
  `is_blockscaled` mode (aligned with how TCGEN05 integrates blockscaling);
  micro-pipeline selection happens inside the lowering.
- **Simple example**: `examples/gemm_sm120/sm120_nvfp4_blockscaled_gemm.py`
  is a non-persistent tiled kernel using plain `T.Kernel` / `T.Pipelined` /
  `T.copy` — no handwritten warp specialization, no special compile flags.
  A hand-written warp-specialized performance kernel (1 producer + 2
  consumer warp groups) lives in
  `maint/gemm/benchmark_sm120_nvfp4_blockscaled_gemm.py`.
- **Scale staging is a library helper**: users keep an explicit
  `T.copy_ue4m3_scale_tile(SFA, SFA_shared, by, ko)` call (the scale operand
  stays visible in the kernel), but the tile addressing of the packed layout
  lives in one place; kernels with custom staging schedules (like the maint
  warp-specialized kernel) reuse the lower-level
  `T.ue4m3_scale_tile_source_coords`. Folding the staging into the op's
  lowering entirely is listed as follow-up work.
- **`tilelang.quantize` stays device-helper territory**: the library gains
  the NVFP4 format contract only — `_tl_*` device-side FP4/UE4M3
  encode/decode helpers usable from any kernel (mirroring the existing
  `_tir_*` converters), the `blockscaled_chunk_kmajor` packers/swizzles,
  and a torch reference quantizer. The tuned TileLang quantizer kernel
  lives in `maint/gemm/tilelang_nvfp4_quantizer.py` next to the benchmark
  that consumes it.

### Scale layout contract & interoperability

The scale operand uses the CUTLASS `BlockScaledBasicChunk` K-major layout,
exposed as `uint32 [MN, K/64]` (byte-identical to the canonical blocked SF
layout used by CUTLASS, CuTeDSL and cublasLt; the views convert zero-copy).
A contract test pins the packed bytes against an independently implemented
layout oracle. We validated interoperability by feeding the same device
bytes to this PR's kernels, CUTLASS (`CollectiveBuilder` +
`KernelTmaWarpSpecializedPingpong`), the CuTeDSL `blackwell_geforce`
example, and cuBLAS (`torch._scaled_mm` -> cublasLt): all produce
**bitwise-identical bf16 outputs** across shapes, including adversarial
position-encoded inputs covering all FP4 codes (negatives included) and
100+ distinct scale values.

### Supported instruction & future work

SM120 `mma` supports several narrow-precision kinds (PTX ISA, warp-level
`mma` with block scaling). This PR implements the NVFP4 variant only; the
config registry is keyed by `(kind, scale_vec, scale type)` so the others
can be added incrementally:

| PTX kind | shape | A/B types | scale_vec | scale type | status |
|---|---|---|---|---|---|
| `kind::mxf4nvf4` | m16n8k64 | e2m1 | `4X` | ue4m3 | **this PR (NVFP4)** |
| `kind::mxf4nvf4` / `kind::mxf4` | m16n8k64 | e2m1 | `2X` | ue8m0 | future (MXFP4) |
| `kind::mxf8f6f4` | m16n8k32 | e4m3/e5m2/e3m2/e2m3/e2m1 | `1X` | ue8m0 | future (MXFP8/6/4) |
| `kind::f8f6f4` | m16n8k32 | e4m3/e5m2/e3m2/e2m3/e2m1 | — | — | future (non-scaled) |

### Performance

Measured on GeForce RTX 5090 (driver 570.124.06, CUDA 12.8): CUDA-event
timing, 5 warmup + 50 iterations, median of 3 interleaved rounds at thermal
steady state (3 s preheat; round spread <= 1%), GPU held exclusively. All
implementations consume identical device bytes and produce bitwise-identical
outputs, so this is an apples-to-apples comparison. CUTLASS/CuTeDSL are shown
at their best tile config per shape; TileLang kernels use 128x128x256.

NVFP4 block-scaled GEMM throughput on GeForce RTX 5090

| M=N=K | TileLang example | TileLang WS (maint) | cuBLAS | CUTLASS best | CuTeDSL best |
|---|---|---|---|---|---|
| 1024 | 197 TFLOPS | 183 | 188 | 160 | 194 |
| 2048 | 608 | 641 | 612 | 603 | 654 |
| 4096 | 1061 | 1079 | 1118 | 1117 | 1080 |
| 8192 | 1328 | **1356** | 1298 | 1324 | 1306 |

The simple example stays within ~5% of the best implementation at every
size; the warp-specialized maint kernel is fastest overall at 8192^3.

### Core-pass changes are NVFP4-scoped

`merge_shared_memory_allocations.cc`: packed scalar NVFP4 shared buffers
are sized bit-exactly (two elements per byte) and their merged-arena byte
offsets convert back to logical element indices with the matching x2 rule.
Both special cases are keyed on `float4_e2m1fn` scalar only — every other
dtype (incl. int4/uint4) keeps bit-exact upstream behavior.
`copy.cc`: the TMA descriptor swizzle recast has an NVFP4-scalar-only
special case; other dtypes unchanged.

### Testing

- `testing/python/language/test_tilelang_language_nvf4_mma_block_scale.py`
  — fragment/lane mapping vs CuTe, codegen, correctness (constant & varying
  scales), API-surface contract tests.
- `testing/python/quantize/test_tilelang_quantize_nvfp4_scale_layout.py`
  — scale-layout packers vs independent oracles (incl. the CuTeDSL blocked
  layout), quantizer round-trip and error bounds.
- `testing/python/language/test_tilelang_sm120_nvfp4_example_cli.py`
  — runs the example end-to-end on SM120.
- Note: CI's self-hosted runner is not SM120, so the runtime kernels are
  exercised on local RTX 5090s (all tests above pass there); CI covers the
  compile/lowering paths. Happy to help wire an SM120 runner if available.

### Shape support

| constraint | rule |
|---|---|
| M | arbitrary (tail tiles handled: TMA zero-fill + predicated store) |
| N | multiple of 8 (16-byte-aligned bf16 rows — the same contiguous-dim rule CuTeDSL validates); N tails otherwise supported |
| M and N tails together | rejected with a clear error for now (a generic copy-lowering boundary bug is tracked separately) |
| K | multiple of 256 (the fulltile package contract) |

The packed scale source is zero-padded to full 128-row tiles by the
`tilelang.quantize` packers, matching the ceil convention
CUTLASS/cuBLAS/CuTeDSL use for the blocked SF tensor.

### Non-goals / follow-ups

- Batched (L > 1) GEMM, `block_K=128` tiles, and a persistent tile
  scheduler for the wide-MN/short-K regime are follow-up work.
- Automatic scale staging inside the `T.mma_gemm_blockscaled` lowering
  (removing the explicit `T.copy_ue4m3_scale_tile` call) is follow-up work.
- Other SM120 mma kinds per the table above.
- The Metal CI job failure is repo-wide (same error on unrelated branches
  since 2026-07-08, e.g. `fix_code`) and unrelated to this PR.


Rachmanino:
Really appreciate your contribution! I think it's better to provide high-level API for blockscaled mma on sm120, like existing `T.tcgen05_gemm_blockscaled` for sm100 tcgen5? Besides, it may be better to move blockscaled mma instructions to `src/tl_templates/cuda/gemm_sm120.h`? Would love to discuss if you have better suggestions!

qqq-tao:
Thanks, that makes sense.

This PR is currently a low-level SM120 NVF4 MMA emitter. It checks that the new mma.sync instruction works, and that the fragment layout and scale-factor lane mapping are correct.

I agree that users should have a higher-level API, similar to the existing SM100 blockscaled GEMM API. The SM120 version should not copy the TCGEN05/TMEM behavior directly, but it can provide a similar user experience on top of the SM120 warp-level mma.sync path.

For this PR, I can move the SM120 blockscaled MMA helper into gemm_sm120.h, so it sits with the SM120 GEMM code. Then we can add a higher-level tile op API as a follow-up, unless you prefer to include that in this PR.

Rachmanino:
Nice work! May I ask whether we support TMA load for operands and SFs? If so, we may also illustrate that in our example. Also, I'm curious whether you've considered warp-specialization (either automatic version or handwritten) and its influence on the performance.

qqq-tao:
Updated status: this PR now includes TMA loading for packed FP4 A/B and SFA/SFB, plus a warp-specialized SM120a NVFP4 performance path. Please see the latest compact status note for current validation and performance numbers: https://github.com/tile-ai/tilelang/pull/2364#issuecomment-4857190005

qqq-tao:
Current SM120 NVFP4 block-scaled GEMM status after the latest cleanup/update:

This PR adds the first TileLang SM120a NVFP4 block-scaled GEMM path with:

- TMA loading for packed FP4 A/B operands and SFA/SFB scaling factors.
- Warp-specialized execution with 1 producer warp group and 2 consumer warp groups.
- A fixed promoted performance strategy: tile `128x128x256`, stages=2, `scale_layout=blockscaled_chunk_kmajor`.
- C epilogue through the current `reg -> smem -> gmem` path.
- CUTLASS `BlockScaledBasicChunk` K-major scale-source contract plus TileLang quantizer support for that packed scale layout.

Validation on `dev5090_m2`, head `1e1576fb`:

```text
pre-commit targeted files: passed
pytest focused SM120/quantizer/example set: 117 passed, 32 warnings
512^3 random A/B + random_binary scale correctness: passed
512^3 bf16_quantized input correctness: passed
8192^3 random_binary scale: 0.8498 ms, 1293.81 TFLOPS
```

Additional short-loop perf sanity check, using `random` A/B + `random_binary` scales, 5 warmup iterations and 50 measured iterations:

```text
TileLang: about 1.30-1.31 PFLOPS
CUTLASS 79a K256 no-verify perf rebuild: about 1.44-1.46 PFLOPS
TileLang / CUTLASS: about 0.90x
```

The remaining gap is not from missing outer-level TMA or warp specialization support. Our current understanding is that the next optimization stage needs deeper CUTLASS/CuTe-style consumer-side modeling inside TileLang lowering: A/B/SFA/SFB `copy_view`, register-fragment packaging, scale slot/package modeling, and stronger `OMMA.SF` issue-continuity control.

Durable PR status observed for this head:

```text
mergeable: true
pre-commit.ci - pr: success
CodeRabbit: success
```

I am not pinning GitHub Actions state here because it is time-varying; please use the PR checks tab for the current Actions run state.

qqq-tao:
@Rachmanino
Hi, I am pleased to tell you that my tilelang NVFP4 GEMM has reached 90% of CUTLASS's performance. I hope you could review the PR and give me more advice to get it finally merged into the mainstream of Tilelang. Thank you for your time.

tilelang/cuda/intrinsics/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .macro.mma_macro_generator import (  # noqa: F401
     TensorCoreIntrinEmitter,
     TensorCoreIntrinEmitterWithLadderTransform,
+    TensorCoreIntrinEmitterWithBlockScale,
```
Rachmanino: Better to extend the original TensorCoreIntrinEmitter for mma to support blockscaling mode, which aligns with the impl of TCGEN05MMAIntricEmitter.

examples/gemm_sm120/sm120_nvfp4_blockscaled_gemm.py
Rachmanino, line 1: Personally I think these 2 examples are a bit too complicated.

1. I think we can make use of the auto ws provided by TileLang instead of handwritting ws
2. For an illustrative example, i think non-persistent is sufficient
3. `T.sm120_store_full_c_fragment_panel32_tma_bf16` it is not very proper to expose such procedure as an API under `tilelang.language`, which usually only includes essential operators or intrinisics. A better alternative way it to consider how to lower to it automatically
4. For simplicity, kernel implementation, along with a simple benchmark result demonstration are sufficient for an example. We can remove stuff like finding nvcc/cmake/cutlass and so on, or consider moving them to the maint folder.

tilelang/language/tir/ir.py
Rachmanino, line 1: If i understand correctly, multiple intrinsics exposed here that started with `sm120_mma` only differ in their micro pipeline strategies (i.e. how ldmatrix and atom mma are composed into a tiled blockscaled mma?). I think instead of exposing so many intrinsics, a unified sm120_mma_blockscaled API containing a parameter which controls the micro pipeline impl is cleaner and easier to maintain

src/cuda/codegen/codegen_cuda.cc
```diff
@@ -32,6 +33,18 @@ using namespace ffi;
 
 namespace {
 
+bool EnableSM120CompactUnpackedFP4Shared() {
+  const char *value = std::getenv("TL_SM120_COMPACT_UNPACKED_FP4_SHARED");
```
Rachmanino: why we need TL_SM120_COMPACT_UNPACKED_FP4_SHARED and TL_SM120_MMA_RAW_UNPACKED_FP4_ACCESS_PTR

qqq-tao:
@Rachmanino Thank you for your review and valuable comments. I am fixing it.

Rachmanino:
Thanks for your detailed feedback, i'll review again then

maint/gemm/gemm_sm120/benchmark_sm120_nvfp4_blockscaled_gemm.py
Rachmanino, line 1: consider move all sm120 nvfp4 relevent files to a unique folder under `maint/gemm`

im0qianqian:
Hi, this is great work—thank you for contributing it to the open-source community.

I also wanted to share some potentially relevant news. We recently discovered an undocumented Blackwell capability: its Tensor Cores can execute INT4/E0M3 block-scaled MMA at the same throughput as NVFP4 E2M1.

NVIDIA does not currently expose this format through the official software stack, so it has to be enabled through post-processing of the generated cubin. Fortunately, the INT4/E0M3 path shares almost the entire GEMM implementation with NVFP4. The main additional component is an INT4/E0M3 quantization kernel.

In our 4-bit pre-training experiments, INT4/E0M3 also delivered substantially better model quality than FP4 E2M1. 
> UFP4 recipe: https://arxiv.org/abs/2606.20381

I happened to come across this PR, and thought this might be relevant. If you are also interested in Blackwell INT4/E0M3 block-scaled MMA, it could potentially be supported alongside the existing NVFP4 implementation as an additional feature.

---

## For SM120 / SM121

In NVFP4 `OMMA.SF.16864.F32.E2M1.E2M1.UE4M3.4X`, the SASS bit 78/79 selects the left/right format (0=E2M1, 1=E0M3). CUBIN post-processing unlocks native full-speed INT4/E0M3, including mixed MMA.

061D849C-C448-4C81-930F-7ACB533FCEBC

---

## Reference

- FlashRT W4A4 support: https://github.com/flashrt-project/FlashRT/pull/146

### For SM100/SM103/SM110

- X link (English): https://x.com/im0qianqian_/status/2075990122269712856
- zhihu link (Chinese): https://zhuanlan.zhihu.com/p/2059376150565089368

maint/gemm/gemm_sm120/benchmark_sm120_nvfp4_blockscaled_gemm.py
qqq-tao: Hi, I have solved all comments that you added. Welcome to review again! Thank you.❤️

Rachmanino:
I'll review this today, thanks @qqq-tao

qqq-tao:
Hi @im0qianqian, thanks for sharing this — I actually came across it a few
days ago. Really cool result, especially the model-quality gains from E0M3
in your pre-training runs.

That said, I think we should think twice before bringing the cubin
bit-patching technique into TileLang itself, mainly for an architectural
reason. TileLang's codegen boundary is CUDA C++ source — we emit .cu, hand
it to nvcc, and PTX → SASS → cubin all happens inside ptxas. The OMMA.SF bit
78/79 that selects E2M1 vs E0M3 only exists *after* ptxas has run, and
ptxas only accepts e2m1. So there is simply no point in TileLang's
compilation pipeline where we could inject E0M3 — the format switch has to
be a post-processing pass on the finished cubin/.so, which lives entirely
outside the compiler.

On top of that, the patch relies on an undocumented hardware behavior and a
format the official stack doesn't expose: it's invisible to nvdisasm/
cuobjdump (the patched instruction still disassembles as E2M1.E2M1), and a
hardcoded encoding offset can break silently across CUDA/driver/arch
versions. That's a real correctness and debuggability risk to carry inside
a compiler.

So my feeling is this fits much better as an opt-in external tool/extension
on top of TileLang — the GEMM path is already shared, as you note, so a
standalone cubin-patcher + INT4/E0M3 quantization kernel could reuse it
without pulling SASS-level hacking into the core. Happy to keep discussing
though; it's genuinely interesting work, and I'd love to see where the E0M3
pre-training direction goes.☺️

LeiWang1999 (approved):

LeiWang1999:
@regression-perf

LeiWang1999 (approved):

PR #2363, opened 2026-06-09
[JIT] Remove frontend disk cache

LeiWang1999:
## Summary

- Remove the frontend disk cache path that mapped Python JIT call forms directly to cached kernel entries.
- Ensure lazy JIT kernels are re-elaborated into `PrimFunc` before hitting the existing kernel cache, so closure-specialized shapes are keyed by the generated IR.

**This is a following pr of #2358 **
## Changes

- Deleted frontend cache key generation and load/store APIs from `tilelang.cache`.
- Removed frontend cache directory handling and `PrimFunc` sidecar persistence from `KernelCache`.
- Simplified JIT cache misses to compile through the normal `PrimFunc`-based cache path.
- Replaced frontend-cache tests with a closure-specialization regression covering `N=128` and `N=256` lazy kernels.

## Validation

- `pre-commit run --all-files`
- `python -m py_compile tilelang/jit/__init__.py tilelang/cache/__init__.py tilelang/cache/kernel_cache.py testing/python/cache/test_tilelang_kernel_cache_atomic_save.py`
- `python debug/reproduce_issue.py`
- `python -m pytest testing/python/cache/test_tilelang_kernel_cache_atomic_save.py -x`

## Release Notes

* **Tests**
  * Updated kernel cache test utilities and adjusted coverage.
  * Removed tests for serialized function loading; added test for specialization re-elaboration.

* **Refactor**
  * Removed frontend cache layer from kernel caching system.
  * Kernels now compile directly on cache miss without intermediate lookup.
  * Removed public cache APIs (`load_frontend_cached`, `store_frontend_cache`) to streamline architecture.

Issue #2172, opened 2026-05-08
[BUG] Int8 Matmul Wrong Answer (on 4090) + Compiler Internal Error (on certain shapes)


Triang-jyed-driung:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.9

### System information
```
addict                       2.4.0
aiohappyeyeballs             2.6.1
aiohttp                      3.13.5
aiosignal                    1.4.0
anaconda-anon-usage          0.7.6
anaconda-auth                0.14.2
anaconda-cli-base            0.8.2
annotated-types              0.6.0
anyio                        4.12.1
apache-tvm-ffi               0.1.11
archspec                     0.2.5
attrs                        26.1.0
beautifulsoup4               4.14.3
boltons                      25.0.0
brotlicffi                   1.2.0.0
certifi                      2026.4.22
cffi                         1.17.1
charset-normalizer           3.4.4
click                        8.2.1
cloudpickle                  3.1.2
conda                        26.3.2
conda-anaconda-telemetry     0.3.0
conda-anaconda-tos           0.2.2
conda-content-trust          0.3.1
conda-libmamba-solver        26.4.0
conda-package-handling       2.4.0
conda_package_streaming      0.12.0
contourpy                    1.3.3
cryptography                 45.0.7
cuda-bindings                13.2.0
cuda-pathfinder              1.5.4
cuda-python                  13.2.0
cuda-toolkit                 13.0.2
cycler                       0.12.1
datasets                     4.8.5
dill                         0.4.1
distro                       1.9.0
einops                       0.8.2
filelock                     3.29.0
fonttools                    4.62.1
frozendict                   2.4.6
frozenlist                   1.8.0
fsspec                       2026.2.0
gdown                        6.0.0
gram-newton-schulz           0.1.4
h11                          0.16.0
hf-xet                       1.5.0
httpcore                     1.0.9
httpx                        0.28.1
huggingface_hub              1.14.0
idna                         3.11
jaraco.classes               3.4.0
jaraco.context               6.1.0
jaraco.functools             4.4.0
jeepney                      0.7.1
Jinja2                       3.1.6
joblib                       1.5.3
jsonpatch                    1.33
jsonpointer                  3.1.1
keyring                      25.7.0
[...]

PR #2366, opened 2026-06-10
[BugFix][Transform] Demote illegal-width residual cp.async to a synchronous copy

### LINKED ISSUE #2365: [BUG][Fuzzer][ice-on-valid-code] Pipelined fp16/bf16 GEMM crashes at codegen when K = BLOCK_K + 1 (illegal 2-byte cp.async)
### Required prerequisites

```
0.1.11 (also reproduced on 0.1.9)
```

### System information
```
tilelang 0.1.11 (built from source, git 53adf03a) · torch 2.8.0+cu128 · CUDA 12.8 · Python 3.11 · NVIDIA A10G (sm_86)
```

### Problem description
A valid pipelined fp16 (or bf16) GEMM crashes during CUDA codegen when K leaves a 1-element residual K-tile, i.e. `K = BLOCK_K + 1` (e.g. `K=33`, `BLOCK_K=32`):

```
InternalError: Check failed: (IsValidCPAsyncTransferBytes(total_bytes)) is false:
tl::ptx_cp_async requires a final PTX byte width in {4, 8, 16}, but got 2
```

The residual copy is scalarized into per-element (`num_elems == 1`) `cp.async` calls of 2 bytes, which is not a PTX-legal `cp.async` width. `K` not being a multiple of `BLOCK_K` is a normal case, and the same GEMM with a plain `T.serial` K-loop compiles and runs correctly — only the `T.Pipelined` (cp.async) path crashes.

Trigger boundary: crashes when the K residual is exactly 1 fp16/bf16 element (2 bytes) — `K=33/BK=32`, `K=65/BK=64`, `K=17/BK=16`. Divisible K is fine; fp32 (4-byte residual) is fine; fp8 (1-byte residual) also crashes (`got 1`).

### Reproducible example code
```python
import tilelang
import tilelang.language as T
import torch

M, N, K = 128, 128, 33          # K = BLOCK_K + 1 -> 1-element residual K-tile
BM, BN, BK = 64, 64, 32

@T.prim_func
def main(A: T.Tensor((M, K), "float16"),
        As = T.alloc_shared((BM, BK), "float16")
        Bs = T.alloc_shared((BK, BN), "float16")
        Cl = T.alloc_fragment((BM, BN), "float32")
        T.clear(Cl)
        for k in T.Pipelined(T.ceildiv(K, BK), num_stages=2):
            T.copy(A[by * BM, k * BK], As)
            T.copy(B[k * BK, bx * BN], Bs)
            T.gemm(As, Bs, Cl)
        T.copy(Cl, C[by * BM, bx * BN])

kernel = tilelang.compile(main, out_idx=[2], execution_backend="cython")  # crashes here
a = torch.randn(M, K, device="cuda", dtype=torch.float16)
b = torch.randn(K, N, device="cuda", dtype=torch.float16)
c = kernel(a, b)
print("ok:", torch.allclose(c, a.float() @ b.float(), rtol=1e-2, atol=1e-1))
```

### Traceback
```python
tvm.error.InternalError: Check failed: (IsValidCPAsyncTransferBytes(total_bytes)) is false:
tl::ptx_cp_async requires a final PTX byte width in {4, 8, 16}, but got 2
  ... in CodeGenTileLangCUDA::VisitExpr_(const CallNode*)   # tl::ptx_cp_async branch
```

### Expected behavior
The kernel compiles and runs, matching `torch.matmul` (out-of-bounds K lanes treated as zero).

### Additional context
Likely a missing fallback in loop-vectorization legalization, not the codegen guard: `InjectPTXAsyncCopy` selects `cp.async` while the residual copy is still a vectorized 16-byte transfer; the residual's out-of-bounds predicate then cuts inside a thread's vectorized lane group (`...+vec < 1`), forcing `LegalizeVectorizedLoop` (`src/transform/loop_vectorize.cc`) to scalarize it to `num_elems == 1` (2-byte) `cp.async` calls. Nothing demotes the unrealizable `cp.async` to a synchronous copy, so it hits the codegen `ICHECK`. Happy to send a PR.

Hughshine:
## Summary
A valid pipelined fp16/bf16/fp8 GEMM whose K dimension leaves a 1-element residual K-tile
(`K = BLOCK_K + 1`, e.g. `K=33`, `BLOCK_K=32`) crashed during CUDA codegen:

```
InternalError: Check failed: (IsValidCPAsyncTransferBytes(total_bytes)) is false:
tl::ptx_cp_async requires a final PTX byte width in {4, 8, 16}, but got 2
```

The program is valid — the same GEMM with a plain `T.serial` K-loop compiles and runs correctly.

Fixes #2365

## Root cause
A `cp.async`'s legality depends on its enclosing vectorized loop supplying a legal total width.

1. `InjectPTXAsyncCopy` selects `cp.async` while the residual copy is still a vectorized 16-byte transfer; its width guard passes.
2. `LegalizeSafeMemoryAccess` adds an out-of-bounds predicate that, for a 1-element residual, cuts inside a thread's vectorized lane group (`...+vec < 1`).
3. `LegalizeVectorizedLoop` (`src/transform/loop_vectorize.cc`) must therefore scalarize the loop, collapsing the 16-byte `cp.async` into per-lane `num_elems == 1` calls of `dtype_bits/8` bytes — 2 B for fp16/bf16, 1 B for fp8, none in `{4, 8, 16}`.
4. Nothing demotes the now-unrealizable `cp.async`, so it reaches the codegen `ICHECK`.

The codegen check is a correct last-line guard; the gap is that the pass which removes the vectorization (and thereby invalidates the `cp.async`) did not legalize the copy it broke.

## Fix
In `loop_vectorize.cc`, when `VectorizeLoop` scalarizes (`vector_size == 1`), demote a `tl::ptx_cp_async` whose per-call width is byte-aligned but not in `{4, 8, 16}` to a synchronous masked store — the form the async injector would have emitted had it declined `cp.async`:

```
As[idx] = if_then_else(pred, A[idx], (dtype)0)
```

- The typed zero-fill replicates `cp.async`'s hardware zero-fill on a false predicate; without it the residual lane is uninitialized and a downstream `T.gemm` reads garbage.
- Runs before `LowerAccessPtr`, so the operands are still `tl::access_ptr` carrying a recoverable `BufferLoad`.
- Gated to the scalarized branch: copies that vectorize keep `num_elems == 1` here but get widened to a legal width downstream, so demoting them would strip `cp.async` off the legal fast path.
- Sub-byte dtypes (int4/int1/fp4: non-byte-aligned width) are not demoted — an element-granular copy is unsafe for packed nibbles — and still hit the codegen guard.

`LegalizeVectorizedLoop` is shared by the CUDA, ROCm and Metal pipelines (CuTeDSL uses the CUDA pipeline), so the fix is backend-agnostic.

## Test plan
Adds `testing/python/issue/test_tilelang_issue_2365.py`:
- fp16/bf16 `K=BLOCK_K+1` (BLOCK_K 16/32/64/128) compiles, runs, matches `torch.matmul`;
- a NaN-poison test that fails if the zero-fill is dropped;
- controls (divisible K, residual ≥ 2, fp32) stay correct;
- a perf guard asserting the divisible-K fast path still emits `cp_async_gs`.

Verified on NVIDIA A10G (sm_86), CUDA 12.8: the new test passes on the patched build and fails on the unpatched build with the `got 2` crash; existing GEMM and async-copy transform tests pass. fp8 cases are gated to sm_89+.

## Summary by CodeRabbit

* **Bug Fixes**
  * Convert unsupported asynchronous memory-copy cases into safe synchronous masked copies to preserve zero-fill semantics, avoid out-of-bounds garbage, and ensure correctness in scalarized/vectorized paths.

* **Tests**
  * Added GPU regression tests for pipelined GEMM residuals (fp16/bf16/float32 and fp8), verifying compile/run success, absence of NaNs, numerical agreement with PyTorch, and that legal-width async copies remain unchanged.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -1060,6 +1060,113 @@ For ParallelToSerial(const For &loop) {
   return Downcast<For>(rewriter(loop));
 }
 
+/*!
+ * \brief Demote a scalarized illegal-width cp.async to a synchronous masked
+ * copy.
+ *
+ * A residual copy whose out-of-bounds predicate depends on the vectorized lane
+ * (e.g. a 1-element K residual, "...+vec < 1") is scalarized to per-lane
+ * cp.async calls with num_elems == 1. For sub-16-byte dtypes that per-call
+ * width (fp16/bf16 = 2 B, fp8 = 1 B) is not a legal cp.async size {4, 8, 16}
+ * and would abort at codegen, so we demote it to a synchronous masked store:
+ *
+ *     As[idx] = if_then_else(pred, A[idx], (dtype)0)
+ */
+class IllegalCPAsyncToSyncDemoter : public StmtExprMutator {
+public:
+  /// Rewrite \p stmt, demoting any illegal-width cp.async it contains.
+  static Stmt Demote(Stmt stmt) {
+    IllegalCPAsyncToSyncDemoter demoter;
+    return demoter(std::move(stmt));
+  }
+
+private:
+  /// Recover the underlying BufferLoad (buffer + indices) from a
+  /// `tl::access_ptr(BufferLoad, extent, rw_mask)` argument, or null if \p arg
+  /// is not that form. At this pass cp.async still carries tl::access_ptr
+  /// (LowerAccessPtr has not run yet), so the BufferLoad is directly available.
+  static const BufferLoadNode *AsAccessPtrLoad(const PrimExpr &arg) {
+    const auto *call = arg.as<CallNode>();
+    if (call == nullptr || !call->op.same_as(tl::access_ptr())) {
+      return nullptr;
+    }
+    if (call->args.empty()) {
+      return nullptr;
+    }
+    return call->args[0].as<BufferLoadNode>();
+  }
+
+  /// Demote a single Evaluate(tl::ptx_cp_async(...)) of illegal width to a
+  /// synchronous masked store; leave every other statement unchanged.
+  Stmt VisitStmt_(const EvaluateNode *op) final {
+    const auto *call = op->value.as<CallNode>();
+    if (call == nullptr || !call->op.same_as(tl::ptx_cp_async())) {
+      return StmtExprMutator::VisitStmt_(op);
+    }
+    // args: [dst_access_ptr, src_access_ptr, num_elems, (predicate)]
+    if (call->args.size() != 3 && call->args.size() != 4) {
+      return StmtExprMutator::VisitStmt_(op);
+    }
+    const auto *num_elems_imm = call->args[2].as<IntImmNode>();
+    const BufferLoadNode *dst_load = AsAccessPtrLoad(call->args[0]);
+    const BufferLoadNode *src_load = AsAccessPtrLoad(call->args[1]);
+    if (num_elems_imm == nullptr || dst_load == nullptr ||
+        src_load == nullptr) {
+      // Not the recoverable tl::access_ptr form (e.g. already lowered, or a
+      // multi-element vectorized copy whose width is handled elsewhere). Leave
+      // it for the codegen guard.
+      return StmtExprMutator::VisitStmt_(op);
+    }
+
+    int64_t num_elems = num_elems_imm->value;
+    DataType dst_dtype = dst_load->buffer->dtype;
+    int64_t per_call_bits = num_elems * dst_dtype.bits() * dst_dtype.lanes();
+
+    // Demote only widths that are a whole number of bytes but not a legal
+    // cp.async size. Legal widths keep their cp.async; sub-byte (non-byte-
+    // aligned) widths are left to crash at the codegen guard, since an
+    // element-granular copy is unsafe for packed sub-byte dtypes (nibbles).
+    bool byte_aligned = (per_call_bits % 8 == 0);
+    bool legal = byte_aligned && IsValidCPAsyncTransferBytes(
+                                     static_cast<int>(per_call_bits / 8));
+    if (!byte_aligned || legal) {
+      return StmtExprMutator::VisitStmt_(op);
+    }
+
+    // An illegal width here is always num_elems == 1 (the injector only emits
+    // cp.async at a legal full width, so num_elems > 1 was filtered out above).
+    // The reconstruction copies a single element, so assert the invariant.
+    ICHECK_EQ(num_elems, 1)
+        << "illegal-width cp.async demotion expects num_elems == 1, got "
+        << num_elems;
+
+    PrimExpr loaded = BufferLoad(src_load->buffer, src_load->indices);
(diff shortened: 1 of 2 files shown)


Hughshine:
Thanks! Adopted the narrowed exception handling in `a283d003`: the broad `except Exception` in `test_pipelined_residual_gemm_compiles_and_correct` is now `except RuntimeError`. The crash we guard against (`tvm ... InternalError`) subclasses `RuntimeError`, so it is still caught, while `AssertionError` and any unexpected exceptions now propagate instead of being silently recorded as a compile crash.

Note: the failing "Run examples (CUDA-12.8)" job is unrelated to this change — it aborts at import of `examples/analyze/test_example_analyze.py` (step "Run examples"), before the CUDA test step (which is skipped). This PR only touches `src/transform/loop_vectorize.cc` and adds `testing/python/issue/test_tilelang_issue_2365.py`.

Hughshine:
Updated in `a4399e4c`:
- Narrowed the test's `except Exception` to `except RuntimeError` (per the review comment).
- Added docstrings to the new Python and C++ helpers so the docstring-coverage check is satisfied.

Re: the failing **Test for Python 3.12 with CUDA-12.8** job — this looks like a runner-environment issue unrelated to the change, not a regression from this PR:

- It fails in the **"Run examples"** step, aborting at *import* of `examples/analyze/test_example_analyze.py` (line 1, `Fatal Python error: Aborted` — a native abort during `import tilelang`, before any example runs). The actual CUDA **test** step is skipped, so the new test here is never reached.
- It reproduces identically across two unrelated pushes (`cc056ad4` → `a4399e4c`), and the diff only touches `src/transform/loop_vectorize.cc` plus a new `testing/python/issue/` file — nothing on the native import path.
- The same code merged into current `main` is healthy: the recent refactor PRs (#2363/#2361/#2357/#2350/#2323) all passed the CUDA test job when merged.
- The job log shows many `Failed to reflink ... Operation not supported (os error 95)` warnings from the uv cache, which suggests a stale/inconsistent dependency cache on the self-hosted runner.

Could a maintainer re-run the job (and/or clear the uv cache on that runner)? Happy to rebase or adjust if you'd prefer. Thanks!

Hughshine:
Note: this shares a root cause with #2172 — its `test(3,128,257)` / `test(2,128,129)` *compiler failed* cases are the same K = BLOCK_K + 1 residual issue (illegal-width cp.async, `got 1` for int8). That crash was never actually fixed (#2172 was closed citing the unrelated FP8-guard commit `33b6368`) and still reproduces on current `main`. This PR fixes it generally in `LegalizeVectorizedLoop` and also resolves the original int8 case. See #2365 for details.

*(Drafted with AI assistance.)*

LeiWang1999:
@regression-perf

LeiWang1999:
Thanks, but after thinking about it a bit more, I don’t think this is a fundamental fix, I'll make another pr for this fix.

PR #2369, opened 2026-06-10
[BugFix][CuTeDSL] Fix TileKernels scan, optional-shape, and e5m6 paths

JayceSu98:
## Summary

This PR fixes the CuTeDSL lowering/runtime gaps that blocked TileKernels full functional coverage on H100, plus small review-driven correctness hardening for the same touched CuTeDSL paths.

## TileKernels Blockers

These are the changes directly needed by the TileKernels full CuTeDSL run:

- **Scan lowering and buffer addressing**
  - Rework `CumSum1D`/`CumSum2D` to carry prefix state across 32-lane segments, matching the CUDA scan behavior used by MoE routing prefix calculations.
  - Route `CumMax1D`/`CumMax2D` through the same line-scan implementation so the CuTeDSL scan helpers stay consistent.
  - Add shared index linearization for non-flat buffer loads/stores, `address_of`/`reinterpret`, and atomic pointer lowering.

- **Dynamic shape and optional tensor ABI handling**
  - Track all candidate sources for each dynamic symbol and resolve runtime values from the first live tensor.
  - Keep missing shape symbols strict for output allocation and launch dimensions.
  - Allow zero placeholders only for dead optional stride/ABI symbols that no longer have a live tensor source after optional-argument pruning.
  - Handle optional host ABI symbols in `top2_sum_gate` variants where optional tensors are compiled out.

- **CuTeDSL control-flow and mutable-vector correctness**
  - Recognize `if cond: stores; thread_return()` and guard the fallthrough body correctly.
  - Avoid reusing SSA aliases for vector operands containing `BufferLoad`, so mutable local tensors are reloaded after intervening stores.

- **E5M6 bit packing**
  - Promote narrow integer shift operands before widening casts, matching CUDA C integer-promotion behavior for packed e5m6 construction.

## Review Hardening

These are not separate TileKernels blockers. They are correctness boundaries for the same generic TileLang/CuTeDSL changes above and were added in response to review:

- Shift promotion now requires matching signedness, avoiding logical right shifts for signed source values widened to unsigned destinations.
- Rank-1 explicit-stride buffer accesses no longer bypass linearization, while vectorized/no-stride one-index `Ramp` accesses keep the previous fast path.
- Reverse `cummax` no longer lets inactive lanes in a partial segment contribute zero to negative-valued tails. The regression uses `block_N=40, threads=64` so CUDA's scan template constraints are respected while CuTeDSL still covers the `active=8` tail case.

## Commit Map

| Commit | Purpose |
| --- | --- |
| `[BugFix] Complete CuTeDSL scan lowering for TileKernels` | Scan carry across segments, non-flat buffer linearization, `match_any_sync`, thread-tag fallback, and review hardening for rank-1 stride / reverse-cummax tail semantics. |
| `[BugFix] Resolve CuTeDSL dynamic shapes from live tensors` | Resolve repeated dynamic symbols from live tensors instead of the first optional/`None` candidate. |
| `[BugFix] Permit dead optional CuTeDSL stride symbols` | Allow only stride-only dead optional ABI placeholders to resolve to zero. |
| `[BugFix] Fix CuTeDSL optional ABI and early-return codegen` | Optional host ABI placeholders, `stores; thread_return()` lowering, and mutable vector reload correctness. |
| `[BugFix] Promote narrow CuTeDSL shifts before wide casts` | Preserve e5m6 packed bits for `uint16 << 20` style expressions, with signedness guard. |

## Validation Environment

Local validation was run with:

- GPU: NVIDIA H100 PCIe
- CUDA toolkit / NVCC: CUDA 13.2, `nvcc` 13.2.78
- `cutlass.__version__`: 4.5.0

CuTeDSL-focused validation used:

```bash
TILELANG_TARGET=cutedsl
TILELANG_DISABLE_CACHE=1
```

For the CUDA-backend regression check that guards the shared language test, `TILELANG_TARGET` was intentionally left unset so the default CUDA backend path was exercised.

## Validation

- `cmake --build build -j 32`
- `ruff check testing/python/target/test_tilelang_codegen_cutedsl_integer_shift.py testing/python/language/test_tilelang_language_scan.py tilelang/contrib/cutedsl/reduce.py`
- `git diff --check origin/main..HEAD`
- `pytest testing/python/target/test_tilelang_codegen_cutedsl_integer_shift.py -q --tb=short`: 2 passed
- `pytest testing/python/target/test_tilelang_codegen_cutedsl_scan.py testing/python/language/test_tilelang_language_scan.py -q --tb=short` with `TILELANG_TARGET=cutedsl`: 13 passed
- `pytest testing/python/language/test_tilelang_language_scan.py::test_cummax_smem_1d -q --tb=short` with default CUDA backend: 1 passed
- `pytest testing/python/language/test_tilelang_language_scan.py::test_cummax_smem_1d -q --tb=short` with `TILELANG_TARGET=cutedsl`: 1 passed
- TileKernels full CuTeDSL run on H100 with `-n 2`: remaining failures were 8 reference-path CUDA OOM cases in `swiglu_backward_and_per_token_cast`; rerunning those exact node ids with `-n 1` passed, classifying them as parallel memory-budget failures rather than CuTeDSL correctness failures.

## Summary by CodeRabbit

* **New Features**
  * Improved CuTeDSL codegen for multi-dimensional buffer indexing and safer conditional return emission.
  * Reworked CuTeDSL segmented warp scans for `cumsum`/`cummax` with better correctness in tail segments.

* **Bug Fixes**
  * Fixed integer shift lowering by promoting narrow shifts before casts and avoiding incorrect cross-signedness promotion.
  * Enhanced dynamic symbol resolution for optional tensors (shape/stride now resolves from live inputs or uses configured fallback).

* **Tests**
  * Added/extended adapter and reduction/codegen regression tests, including optional-symbol cases and additional `cummax` scenarios.

### DIFF
--- a/tilelang/contrib/cutedsl/reduce.py
+++ b/tilelang/contrib/cutedsl/reduce.py
@@ -115,7 +115,7 @@ def bar_sync_ptx(barrier_id, number_of_threads):
 
 
 # Import shuffle functions from warp module
-from .warp import __shfl_up_sync, __shfl_down_sync
+from .warp import __shfl_sync, __shfl_up_sync, __shfl_down_sync
 
 
 def _warp_prefix_sum_forward(val, lane, MASK=0xFFFFFFFF):
@@ -123,7 +123,7 @@ def _warp_prefix_sum_forward(val, lane, MASK=0xFFFFFFFF):
     Warp-level inclusive prefix sum (forward).
     Uses shfl.up to propagate values from lower lanes.
     """
-    # Unrolled loop for SEG=32: off = 1, 2, 4, 8, 16
+    # Unrolled loop for WARP_SIZE=32: off = 1, 2, 4, 8, 16
     n = __shfl_up_sync(MASK, val, 1)
     val = cutlass.select_(lane >= 1, val + n, val)
     n = __shfl_up_sync(MASK, val, 2)
@@ -142,18 +142,18 @@ def _warp_prefix_sum_reverse(val, lane, MASK=0xFFFFFFFF):
     Warp-level inclusive prefix sum (reverse).
     Uses shfl.down to propagate values from higher lanes.
     """
-    SEG = 32
-    # Unrolled loop for SEG=32: off = 1, 2, 4, 8, 16
+    WARP_SIZE = 32
+    # Unrolled loop for WARP_SIZE=32: off = 1, 2, 4, 8, 16
     n = __shfl_down_sync(MASK, val, 1)
-    val = cutlass.select_(lane < SEG - 1, val + n, val)
+    val = cutlass.select_(lane < WARP_SIZE - 1, val + n, val)
     n = __shfl_down_sync(MASK, val, 2)
-    val = cutlass.select_(lane < SEG - 2, val + n, val)
+    val = cutlass.select_(lane < WARP_SIZE - 2, val + n, val)
     n = __shfl_down_sync(MASK, val, 4)
-    val = cutlass.select_(lane < SEG - 4, val + n, val)
+    val = cutlass.select_(lane < WARP_SIZE - 4, val + n, val)
     n = __shfl_down_sync(MASK, val, 8)
-    val = cutlass.select_(lane < SEG - 8, val + n, val)
+    val = cutlass.select_(lane < WARP_SIZE - 8, val + n, val)
     n = __shfl_down_sync(MASK, val, 16)
-    val = cutlass.select_(lane < SEG - 16, val + n, val)
+    val = cutlass.select_(lane < WARP_SIZE - 16, val + n, val)
     return val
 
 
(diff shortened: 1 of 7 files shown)


tilelang/contrib/cutedsl/reduce.py
```diff
@@ -268,60 +341,24 @@ def run(self, src: cute.Pointer, dst: cute.Pointer, H, W):
         MASK = 0xFFFFFFFF
         tidx, _, _ = cute.arch.thread_idx()
         lane = tidx % self.SEG
-        row = tidx // self.SEG
+        item = tidx // self.SEG
+        tile = self.threads // self.SEG
```
cherichy: self.SEG is currently delegated in _scan_line_*, which is only used as warp_size in the class CumSum1D and 2D. Please consider change the name into WARP_SIZE.

cherichy (dismissed):
LGTM, Thanks for the contribution.

JayceSu98:
Addressed the scan naming review by renaming the CuTeDSL scan warp-width member from `self.SEG` to `self.WARP_SIZE` and updating the local scan helper constants/comments from `SEG` to `WARP_SIZE`. The behavior is unchanged; this is folded into the existing scan-lowering commit.

JayceSu98:
Updated the branch for the CuTeDSL host-codegen test review.

What changed:

- Added `_require_cutedsl()` before the three adapter tests that import `tilelang.jit.adapter.cutedsl.adapter` directly:
  - `test_cutedsl_adapter_resolves_dynamic_symbol_from_live_tensor_candidate`
  - `test_cutedsl_adapter_allows_optional_stride_symbol_without_live_tensor`
  - `test_cutedsl_adapter_allows_optional_abi_shape_symbol_without_live_tensor`

Why this is needed:

- These tests exercise CuTeDSL adapter internals and import `CuTeDSLKernelAdapter` directly.
- In environments without the CuTeDSL Python stack, the project convention is to skip via `_require_cutedsl()` instead of failing during import.
- Other tests in this module already follow that pattern through `_lower_cutedsl()` or an explicit `_require_cutedsl()` call, so these adapter regression tests should do the same.

I also simplified the small helper TIR programs used by these adapter tests:

- The tests only need dynamic symbols from function parameter shapes/strides so that `_process_dynamic_symbolic()` and `_resolve_dynamic_symbolic_value()` can be exercised.
- They do not need to test dynamic `T.Kernel(T.ceildiv(N, 64), ...)` launch lowering.
- After rebasing onto the latest `origin/main`, dynamic kernel launch materialization changed enough that those helper bodies became an unrelated dependency and failed before the adapter logic was reached.
- Replacing those helper bodies with `T.evaluate(0)` keeps the tests focused on adapter symbol resolution and avoids coupling them to dynamic launch lowering.

The update was folded into the existing commit history; no standalone review-fix commit was added.

Validated locally after installing the missing pytest/ruff tooling for `python3.12`:

- `python3.12 -m ruff format --check testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py tilelang/contrib/cutedsl/reduce.py`
- `python3.12 -m ruff check testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py tilelang/contrib/cutedsl/reduce.py`
- `python3.12 -m py_compile testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py tilelang/contrib/cutedsl/reduce.py`
- `git diff --check`
- `python3.12 -m pytest testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py::test_cutedsl_adapter_resolves_dynamic_symbol_from_live_tensor_candidate testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py::test_cutedsl_adapter_allows_optional_stride_symbol_without_live_tensor testing/python/jit/test_tilelang_jit_cutedsl_host_codegen.py::test_cutedsl_adapter_allows_optional_abi_shape_symbol_without_live_tensor -q`

Result: `3 passed, 10 warnings`.

PR #2387, opened 2026-06-12
[Transform] Materialize kernel launch per backend

LeiWang1999:
## Summary

- Refactor `T.Kernel` to emit a target-neutral launch nest that backend pipelines materialize after target binding.
- Add `T.ClusterKernel` for CUDA thread block cluster launches so plain `T.Kernel` can be shared across CPU and SIMT backends.
- Update tests, examples, maintenance scripts, and docs to use the new launch split where cluster dimensions are required.

## Changes

- Add `tl.MaterializeKernelLaunch`, which converts launch thread-binding loops into backend-specific forms.
- Run the materialization pass in CPU, CUDA, ROCm, and Metal pipelines with backend-appropriate lowering behavior.
- Remove the CPU-only kernel-frame attribute path and normalize launch thread handling through the shared `KernelLaunchFrame` API.
- Update cluster examples and tests to call `T.ClusterKernel` while leaving non-cluster kernels on `T.Kernel`.

## Validation

## Summary by CodeRabbit

* **New Features**
  * Cluster-oriented kernel launch support exposed as a public API (ClusterKernel).
  * Compilation now materializes kernel launches earlier in backend pipelines.

* **Documentation**
  * Examples and guides updated to use cluster-aware kernel launches and reflect new launch patterns.

* **API Changes**
  * Kernel helper simplified; cluster-oriented launches should use ClusterKernel going forward.

* **Tests**
  * Tests and demos updated to exercise cluster-aware launches and the new launch-materialization pass.

### DIFF
--- a/docs/programming_guides/cluster_tma.md
+++ b/docs/programming_guides/cluster_tma.md
@@ -3,7 +3,7 @@
 This page describes two advanced data-movement features that are available on
 NVIDIA Hopper (SM90) and later: **TMA multicast** and **SM-to-SM cluster
 copy**. Both features are exposed through extensions to the existing `T.copy`
-operator and require a kernel launched with thread block cluster, i.e., with `cluster_dims != (1, 1, 1)`.
+operator and require a kernel launched with `T.ClusterKernel`, i.e., with `cluster_dims != (1, 1, 1)`.
 
 Requirements:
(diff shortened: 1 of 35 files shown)


LeiWang1999:
@regression-perf

PR #2406, opened 2026-06-16
[BugFix] Fix warp-specialized register over-subscription

Rachmanino:
The auto-injected setmaxnreg pair was hard-coded to dealloc=24/alloc=240, which only fits the canonical 1-producer + 2-consumer warpgroup layout (<=384 threads: 128\*24 + 256\*240 = 64512 max regs/SM). Wider warp-specialized layouts over-subscribe the register file and the consumer's setmaxnreg.inc deadlocks (observed as a GPU hang on sm_120 Blackwell, while sm_89 Ada is unaffected since it never takes the TMA+WS path).

When no explicit reg hint is given, derive the producer/consumer thread split from the warp-specialization guard and pick a consumer request so producer_threads*dec + consumer_threads*inc <= 64512:
  - 1P + 2C (384 threads): keep 24/240 (no warning)
  - 2P + 2C (512 threads): 24/224 (+ warning)
  - 1P + 3C (512 threads): 24/160 (+ warning)
  - any other layout: skip setmaxnreg entirely (+ warning) rather than risk over-subscribing the register file

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined CUDA warp-group register allocation to dynamically select optimal configurations instead of using generic defaults

* **Improvements**
  * Expanded support for additional warp-group layout configurations
  * Enhanced warning messages for register reallocation scenarios

### DIFF
--- a/src/cuda/transform/annotate_warp_group_reg_alloc.cc
+++ b/src/cuda/transform/annotate_warp_group_reg_alloc.cc
@@ -251,14 +251,85 @@ private:
             (has_explicit_hints || !has_simt_copy)) {
           int final_dec_reg = has_explicit_hints ? dec_reg : 24;
           int final_inc_reg = has_explicit_hints ? inc_reg : 240;
-          dec_reg_stmt =
-              Evaluate(Call(DataType::Handle(), set_max_nreg(),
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

Issue #2368, opened 2026-06-10
[Question] Is end-to-end execution for the LLVM target supported?


Witherstrike:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Questions

Hi, I am trying to use the LLVM target to compile and execute a `PrimFunc` on the CPU. 

While lowering the function via `tilelang.lower` works as expected and successfully generates the kernel source, attempting to compile it using `tilelang.compile` fails with a `ValueError`. It seems that the LLVM target is currently not supported or recognized inside `device_codegen` during the compilation phase. 

Additionally, when switching back to the default execution backend (`cython`), a similar `ValueError` is raised when creating the adapter inside `_process_buffer_device`.

Is end-to-end execution for the LLVM target currently supported, or is this a known limitation/missing feature?

### Reproduction Code
```
import tilelang
import tilelang.language as T
from tilelang.utils.target import determine_target

def tl_mul_relu_1d_mem(N: int, BLOCK_N: int):
    dtype = T.float16

    @T.prim_func
    def main(
        A: T.Tensor((N,), dtype),
        B: T.Tensor((N,), dtype),
        C: T.Tensor((N,), dtype),
    ):
        with T.Kernel(N // BLOCK_N, is_cpu=True) as bx:
            A_f = T.alloc_fragment(BLOCK_N, dtype)
            B_f = T.alloc_fragment(BLOCK_N, dtype)
            T_f = T.alloc_fragment(BLOCK_N, dtype)

            T.copy(A[BLOCK_N * bx], A_f)
            T.copy(B[BLOCK_N * bx], B_f)
            
            for i in T.Parallel(BLOCK_N):
                T_f[i] = A_f[i] * B_f[i]
                T_f[i] = T.if_then_else(T_f[i] > 0, T_f[i], 0)
            
            T.copy(T_f, C[BLOCK_N * bx])
    
    return main


def test():
    N = 1024 * 4096
    BLOCK_N = 1024

    func = tl_mul_relu_1d_mem(N=N, BLOCK_N=BLOCK_N)

    target={"kind":"llvm","tag":"","keys":["cpu"],"mtriple":"x86_64-pc-linux-gnu"}
    target = determine_target(target, return_object=True)

    # this works
    """ with target:
        ker = tilelang.lower(func, target=target)
[...]

PR #2409, opened 2026-06-17
[Feature] Add LLVM backend support

Witherstrike:
Enable LLVM as a TileLang backend in CMake and route llvm targets through TVM LLVM codegen during lowering.

Fix tvm_ffi output tensor device inference by deriving the output device from tensor inputs, and add LLVM codegen/compile regression tests for matmul, T.gemm, T.copy, and T.While.

Related issue:
https://github.com/tile-ai/tilelang/issues/2368#issuecomment-4690493542

## Summary by CodeRabbit

* **New Features**
  * Added an LLVM backend as a selectable compilation target, including environment-variable based configuration.

* **Tests**
  * Added new LLVM-focused GEMM tests (code generation checks, JIT compilation with correctness validation, and copy-path coverage), plus an LLVM regression test for `While` kernel source generation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,11 +250,12 @@ else()
 endif()
 
 # Configs
-set(TILELANG_BACKENDS CUDA ROCM METAL)
+set(TILELANG_BACKENDS CUDA ROCM METAL LLVM)
 
 set(TILELANG_BACKEND_DOC_CUDA "Enable CUDA backend (ON/OFF/or CUDA SDK path)")
 set(TILELANG_BACKEND_DOC_ROCM "Enable ROCm backend (ON/OFF/or ROCm SDK path)")
 set(TILELANG_BACKEND_DOC_METAL "Enable Metal backend")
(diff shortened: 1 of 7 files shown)


LeiWang1999 (commented):
Thanks for adding this. The change is useful because it connects the existing CPU/LLVM pipeline to the actual `tvm_ffi` compile path by routing `target.kind.name == "llvm"` through `target.build.llvm`, and the new tests cover the main end-to-end cases well.

  Could you please add two follow-ups before merge?

  1. Add a short doc update for LLVM support, especially that `USE_LLVM` is OFF by default and users need to enable it explicitly, e.g. `cmake .. -DUSE_LLVM=ON` or `USE_LLVM=1 cmake ..`. It would also help to mention that the underlying TVM build must have LLVM enabled.

  2. Add LLVM requirement guards to the new tests, e.g. `@tilelang.testing.requires_llvm` or an equivalent file-level marker, so these tests are skipped cleanly when TileLang/TVM is built without LLVM.

Witherstrike:
Done.

LeiWang1999 (approved):

PR #2420, opened 2026-06-18
fix(gemm_sm100): fix intermittent hang in pipelined CLC scheduler GEMM

### LINKED ISSUE #2410: [CUDA backend] Intermittent hang in SM100 CLC scheduler GEMM
## Symptom

The CUDA backend SM100 CLC scheduler GEMM can intermittently hang during repeated execution or benchmarking. The hang is random and may not reproduce every run.

## Repro

Full example:

```bash
TILELANG_TARGET=cuda TILELANG_DISABLE_CACHE=1 \
python examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py
```

Focused stress repro candidate:

```bash
TILELANG_TARGET=cuda TILELANG_DISABLE_CACHE=1 \
python debug/clc_bench_repro.py \
  --m 8192 --n 8192 --k 8192 \
  --clc 3 \
  --repeat 64 \
  --sync-each \
  --keep-outputs \
  --progress
```

## Notes

The hang appears to be random, but stress testing may expose it. We suspect a subtle issue in the CLC scheduler implementation.

ighoshsubho:
Fixes #2410

## Problem

`gemm_clc_persistent_2cta_pipelined_clc` intermittently hangs under repeated execution / benchmarking (e.g. 8192^3 with many reps and a sync between each). The hang is timing-dependent and does not reproduce on every run.

## Cause

A cross-warp read/write race on the CLC schedule slots, exposed by pipelining.

In each consumer warp (TMA, MMA, epilogue) the order was:

```python
T.mbarrier_wait_parity(schedule_arrived[s_cons], c_cons & 1)
if tx == ...:
    T.mbarrier_arrive(schedule_finished[s_cons], 0)   # signals "slot free"
if schedule_valid[s_cons] == 0:                       # ...but reads slot AFTER
    break
tile_id = ... schedule_tile_id[s_cons] ...            # ...and reads slot AFTER
```

The consumer arrives schedule_finished before reading schedule_valid / schedule_tile_id. That arrival is exactly what unblocks the CLC scheduler warp to reuse (overwrite) the same slot. With clc_stages > 1 the scheduler runs
ahead and its next try_cancel is already in flight, so the overwrite can land between the consumer's arrive and its read. Different consumer warps then read different schedule_valid for the same slot -> some break, some don't -> mismatched barrier arrival counts ->  deadlock.

The non-pipelined variant has the same ordering but survives because the unhidden try_cancel latency leaves a large gap between the arrive and the overwrite. Pipelining exists precisely to hide that latency, which collapses the protective gap and exposes the race -> hence the intermittency.

## Fix

Each consumer warp now captures schedule_valid / schedule_tile_id into thread-local registers immediately after the wait, and only then arrives schedule_finished. The break check and tile_id are taken from the registers, so the slot is fully read before it can be reused. No barrier counts or pipelining structure change; the latency-hiding benefit of clc_stages is
preserved.

## Verification

- Correctness: torch.testing.assert_close passes for clc=2,3,4 at 4096³ / 8192³ / 16384³.
- Stress: survives 200 relaunch+sync reps per (shape, clc) (--stress 200).

## Summary by CodeRabbit

* **New Features**
  * Added `--stress` command-line argument to enable repeated stress testing iterations on GEMM operations with synchronized CUDA execution between runs.

* **Refactor**
  * Optimized scheduler status handling in GEMM pipeline execution for improved performance.

### DIFF
--- a/examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py
+++ b/examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py
@@ -224,6 +224,10 @@ def gemm_clc_persistent_2cta_pipelined_clc(
         clc_result = T.alloc_shared((clc_stages, 4), "uint32", scope="shared")
         schedule_valid = T.alloc_shared((clc_stages,), "int32")
         schedule_tile_id = T.alloc_shared((clc_stages,), "int32")
+        # Reading the schedule into thread local registers before arriving schedule_finished,
+        # so the pipelined scheduler can't overwrite the slot mid-read (causing intermittent hang).
+        local_valid = T.alloc_local((1,), "int32")
+        local_tile_id = T.alloc_local((1,), "int32")
 
         tx = T.get_thread_binding()
         cta_id = T.block_rank_in_cluster()
@@ -236,15 +240,17 @@ def gemm_clc_persistent_2cta_pipelined_clc(
 
                 if work_iter > 0:
                     T.mbarrier_wait_parity(schedule_arrived[s_cons], c_cons & 1)
+                    local_valid[0] = schedule_valid[s_cons]
+                    local_tile_id[0] = schedule_tile_id[s_cons]
                     if tx == 0:
                         T.mbarrier_arrive(schedule_finished[s_cons], 0)
-                    if schedule_valid[s_cons] == 0:
+                    if local_valid[0] == 0:
                         break
 
                 tile_id = T.if_then_else(
                     work_iter == 0,
                     block_id // 2,
-                    schedule_tile_id[s_cons],
+                    local_tile_id[0],
                 )
                 bx, by = get_swizzled_block_idx(tile_id, group_size, m_clusters, cta_id)
 
@@ -270,9 +276,10 @@ def gemm_clc_persistent_2cta_pipelined_clc(
 
                 if work_iter > 0:
                     T.mbarrier_wait_parity(schedule_arrived[s_cons], c_cons & 1)
+                    local_valid[0] = schedule_valid[s_cons]
                     if tx == 32:
                         T.mbarrier_arrive(schedule_finished[s_cons], 0)
-                    if schedule_valid[s_cons] == 0:
+                    if local_valid[0] == 0:
                         break
 
                 T.mbarrier_wait_parity(tmem_empty[work_iter & 1], ((work_iter // 2) & 1) ^ 1)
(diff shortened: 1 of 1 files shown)


VitalyAnkh:
I tested this PR head (`0a8f2ee561d1e95b01e4e99ddafc050311b19bc3`) in a local worktree with its own native build artifacts:

```text
/home/vitalyr/dev/ai/tilelang_pr_2420
```

The exact example fails correctness before I can get a meaningful no-hang result from repeated execution. This is the stderr tail from the failing run:

```text
Loading tilelang libs from dev root: /home/vitalyr/dev/ai/tilelang_pr_2420/build
  File "/home/vitalyr/dev/ai/tilelang_pr_2420/examples/gemm_sm100/gemm_tcgen5mma_ws_clc.py", line 391, in <module>
    torch.testing.assert_close(c, ref, rtol=1e-2, atol=1e-2)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vitalyr/dev/ai/tilelang/.venv/lib/python3.13/site-packages/torch/testing/_comparison.py", line 1631, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 393161 / 67108864 (0.6%)
Greatest absolute difference: 464.0 at index (3824, 5016) (up to 0.01 allowed)
Greatest relative difference: 2352.0 at index (4418, 6949) (up to 0.01 allowed)
exit
```

So this PR does not pass the concrete validation path for #2410 in my environment.

One important testing detail: this example needs to be run through its actual `if __name__ == "__main__"` path, ideally repeatedly in subprocesses. Pytest collection/import checks are not enough to validate this failure mode. I opened #2423 with the minimal schedule-state lifetime fix and included the 5 rounds x 50 exact-example validation command there.

ighoshsubho:
> The exact example fails correctness before I can get a meaningful no-hang result from repeated execution. This is the stderr tail from the failing run:

Good approach yeah, I find this similar to gemm sm100 from quack where they use 13 arrivals counting the leaders and the process looks clean because of schedule published is being waited on by consumers and also before arriving schedule finised we read the tile id, my pr was a little unconventional in sense that it uses thread local registers to store tile id before consumers arrive on schedule finished, it worked on my env, dumping cuda can explain the real issue but anyways since your PR looks clean and better I am closing this, thanks for contributing to this fix. 

cc: @VitalyAnkh

PR #2424, opened 2026-06-19
[BugFix] Fix grouped reduce_sum over-counts on straddle layout

hhy3:
## Problem
  A `T.reduce_sum` over a reshaped fragment (`[TILE_X, K]` → `[TILE_X, G, gran_k]`, reduce `dim=2`) silently over-counts by the
  AllReduce width when the group layout straddles a thread tile — i.e. `G` is not a power of two. Example: `K=1280, gran_k=32, G=40`
  (tiled as 4 float4 lanes × 8 threads) sums each group **8×**.

  ## Root cause
  The `collapse_dead_residue` pass in `CompressIterator` must drop a dead thread-residue split from the local-loop trip count. The invariance proof it relies on (a straddle floordiv like `(i*320 + j*8 + rv%8) // 128 == (i*320 + j*8) // 128`) depended on whatever facts the shared analyzer had already accumulated, making the collapse **order-dependent**: the first reduce lowered missed the dead residue and over-counted, while a later reduce (e.g. `absmax`) with an identical index collapsed correctly.

  ## Fix
  Bind every input iter range plus the rebuilt reduce var into the analyzer before the collapse loop, so the invariance proof is
  deterministic.

  ## Test
  Adds `test_reduce_sum_grouped_straddle_overcount_regression` at the minimal trigger (`TILE_X=2, K=192, gran_k=32, G=6, threads=128`):
   pre-fix `out == ref × AllReduce_width`, post-fix `out == ref`. Verified FAIL on `origin/main`, PASS with the fix.

## Summary
- Fixed `T.reduce_sum` over-counting for reshaped fragment reductions with straddle layouts by making thread-owned residue collapse deterministic in reduction ownership planning.
- Refactored reduction lowering to precompute and reuse a shared ownership plan for local indices and per-thread reduction steps.
- Added a CUDA regression test covering the minimal failing case (`tile_m=2`, `hidden=192`, `group=6`, `group_k=32`).

## Testing
- Added/updated regression coverage in `testing/python/language/test_tilelang_language_reduce.py`.

## C++ style / lint notes
- This PR modifies C++ code in `src/backend/common/op/reduce.h`, which is within the scope of `docs/developer_guide/cpp_style.md` (formatting/naming/header conventions for TileLang-owned headers under `src/`).
- No new build/correctness issues were implied by the change itself.
- If the PR introduces new `TLCPP003`/`TLCPP004`-type style or public-API surface warnings, they would be advisory only; the repo’s **“C++ API Style Audit (warning only)”** CI step is the relevant place to verify any such findings.

### DIFF
--- a/src/backend/common/op/reduce.h
+++ b/src/backend/common/op/reduce.h
@@ -395,6 +395,28 @@ TryRemoveThreadOwnedFactor(const PrimExpr &expr, const IterVar &iter_var,
   return std::make_pair(new_expr, new_iter_var);
 }
 
+inline void CheckThreadOwnedFactorProjectable(const PrimExpr &index_expr,
+                                              const Var &reduce_var,
+                                              int64_t thread_owned_factor,
+                                              arith::Analyzer *analyzer) {
+  if (thread_owned_factor <= 1) {
+    return;
+  }
+
+  PrimExpr factor = make_const(reduce_var.dtype(), thread_owned_factor);
+  PrimExpr masked_reduce_var = FloorDiv(reduce_var, factor) * factor;
+  PrimExpr simplified_index = analyzer->Simplify(index_expr);
+  PrimExpr projected_index = analyzer->Simplify(
+      Substitute(simplified_index, {{reduce_var, masked_reduce_var}}));
+
+  ICHECK(analyzer->CanProveEqual(projected_index, simplified_index))
+      << "ReduceOp cannot lower a layout where a source index depends on a "
+         "thread-owned reduce factor: src_index="
+      << simplified_index << ", projected_src_index=" << projected_index
+      << ", reduce_var=" << reduce_var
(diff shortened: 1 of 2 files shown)


ppppqp:
I think the fix makes sense. Wondering if it's cleaner to put the fix in `ComputeReducerLayout` or `ReduceLowerer` since it's specific to reduction, but that can introduce more work

hhy3:
@ppppqp updated

ppppqp:
LGTM! Will defer to @LeiWang1999 for final review

LeiWang1999:
@regression-perf

LeiWang1999:
Thanks for your contributions! It's likely current fix is a patch that we first generate a wrong ir and then fix it, let me take a look on this issue :)

LeiWang1999 (approved):

PR #2438, opened 2026-06-22
[BugFix] Fix make_int sign-extending negative int8 lanes

### LINKED ISSUE #2427: [BUG][Fuzzer][wrong-code] `T.fill` of an `int8` buffer with a negative value silently clobbers 3 of every 4 elements to `-1` instead of writing the value to all of them
### Required prerequisites

0.1.11

### System information

- Installed from the `0.1.11` wheel (source pinned to tag `v0.1.11`, commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`).
- Python 3.13.13, Linux.
- GPU: NVIDIA H100 80GB HBM3 (sm_90). The defect is in a device helper that runs on any CUDA arch.

```python
import sys, tilelang, torch
print(sys.version.split()[0], sys.platform)                              # 3.13.13 linux
print(tilelang.__version__)                                              # 0.1.11
print(torch.__version__, torch.version.cuda)                             # 2.11.0+cu130 13.0
```

### Problem description

**TL;DR:** the four-`int8`-into-one-`int` packing helper `make_int(signed char, …)` omits a per-lane `& 0xFF` mask, so each negative lane is sign-extended to 32 bits before the `<<`/`|` and its high `1`-bits flood the other lanes — a negative `T.fill` on `int8` silently writes 1 lane correctly and clobbers the other 3 to `-1`. One-line fix: mask each lane before shifting.

`T.fill(buf, v)` on an `int8` buffer with a negative scalar `v` writes only 1 of every 4 elements with the intended value and silently sets the other 3 of every 4 to `-1` (`0xFF`); the kernel compiles and runs with no error and returns wrong numbers, deterministically.

For an 8×128 = 1024-element `int8` buffer, `v = -7, -8, -100, -128` each give 768/1024 elements wrong (the only values present are `[v, -1]`). `v = -1` is correct (its bit pattern is `0xFF` in every byte, so the corruption is a no-op) and all non-negative `v` are correct. The pattern is exactly "lane 0 survives, lanes 1–3 become `0xFF`": the fill is vectorized four `int8` at a time, and only the first of each group of four keeps its value. The single varying factor that flips correct↔wrong is the sign of `v` — same `T.fill`, same `int8` dtype, same vectorized lowering, both shared and fragment scope (controls below).

The generated CUDA for `v = -7` materializes the scalar as a runtime local `signed char broadcast_var = (signed char)-7;` and packs four copies into one 32-bit store via `make_int(broadcast_var, broadcast_var, broadcast_var, broadcast_var)`. `make_int(-7,-7,-7,-7)` returns `0xFFFFFFF9` (bytes `[F9, FF, FF, FF]`) instead of the intended `0xF9F9F9F9` (`[F9, F9, F9, F9]`) — exactly "1 lane = `-7`, 3 lanes = `-1`".

### Reproducible example code

```python
import tilelang
tilelang.disable_cache()
import tilelang.language as T
import torch


def run(val, M=8, N=128, scope="shared"):
    @tilelang.jit
    def f():
        O = T.empty((M, N), "int8")
            if scope == "shared":
                s = T.alloc_shared((M, N), "int8")
            else:
                s = T.alloc_fragment((M, N), "int8")
            T.fill(s, val)
            T.copy(s, O)
        return O
    k = f.compile()
    out = k().cpu()
    exp = torch.full((M, N), val, dtype=torch.int8)
    nbad = int((out != exp).sum().item())
    return torch.equal(out, exp), nbad, torch.unique(out).tolist(), k


# Defect: negative int8 fill clobbers 3 of every 4 lanes to -1.
for v in [5, 0, -1, -7, -8, -100, -128]:
    eq, nbad, uniq, _ = run(v)
    print(f"int8 v={v:5d}: {'OK   ' if eq else 'WRONG'} nbad={nbad:4d}/1024 unique={uniq}")

# Control: fragment scope behaves identically (the helper is scope-independent).
eq, nbad, uniq, _ = run(-7, scope="fragment")
print(f"fragment int8 v=-7: {'OK' if eq else 'WRONG'} nbad={nbad} unique={uniq}")

# The packing call is visible in the generated CUDA:
_, _, _, k = run(-7)
for line in k.get_kernel_source().splitlines():
    if "make_int" in line or "broadcast_var" in line:
        print("CODEGEN:", line.strip())
```

Output (stable across repeated runs):

```
int8 v=    5: OK    nbad=   0/1024 unique=[5]
int8 v=    0: OK    nbad=   0/1024 unique=[0]
int8 v=   -1: OK    nbad=   0/1024 unique=[-1]
int8 v=   -7: WRONG nbad= 768/1024 unique=[-7, -1]
int8 v=   -8: WRONG nbad= 768/1024 unique=[-8, -1]
int8 v= -100: WRONG nbad= 768/1024 unique=[-100, -1]
int8 v= -128: WRONG nbad= 768/1024 unique=[-128, -1]
fragment int8 v=-7: WRONG nbad=768 unique=[-7, -1]
CODEGEN: signed char broadcast_var = (signed char)-7;
CODEGEN: *(int*)(s_local_cast + (i * 4)) = make_int(broadcast_var, broadcast_var, broadcast_var, broadcast_var);
```

Non-negative `v` and the single value `v = -1` are exact; every other negative `v` loses 3/4 of its lanes — isolating the defect to the sign-extension in the four-`int8` packing helper, not to `T.fill`, the copy, or the harness.

### Traceback

```pytb
```

### Expected behavior

`T.fill(buf, v)` should write every element of an `int8` buffer to `v`, for negative `v` exactly as for non-negative `v` — i.e. match `torch.full((M, N), v, dtype=torch.int8)` for all `v` in `[-128, 127]`. This is grounded in TileLang's own behavior: non-negative `v` and `v = -1` of the identical program are exact, and the codegen already ships the correct masking — the constant-`int8` `BroadcastNode` fast-path masks each lane with `& 0xFF` before the shift/OR ([`codegen_cuda.cc#L4957-L4985`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L4957-L4985)), producing `0xF9F9F9F9` for `v = -7`. The project's intent is therefore that four `int8` lanes pack without sign-bit cross-contamination; the runtime helper `make_int` is the one place that omits the mask.

### Additional context

**Root cause.** The vectorized `int8` fill packs four lanes into one 32-bit `int` using the device helper `make_int(signed char, …)` at [`src/tl_templates/cuda/common.h#L136-L139`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L136-L139):

```cpp
TL_DEVICE int make_int(signed char x0, signed char x1, signed char x2,
                       signed char x3) {
  return (x3 << 24) | (x2 << 16) | (x1 << 8) | x0;
}
```

By C++ integer promotion each `signed char` is promoted to `int` before the shift/OR. A negative lane such as `-7` promotes to `0xFFFFFFF9` (sign-extended); the low term `x0 = 0xFFFFFFF9` has its upper 24 `1`-bits OR over every other lane, so the result is `0xFFFFFFF9` = bytes `[F9, FF, FF, FF]` rather than `[F9, F9, F9, F9]`. Only lane 0's low byte survives; lanes 1–3 are clobbered to `0xFF`. This is directly demonstrable in C: `make_int(-7,-7,-7,-7) == 0xFFFFFFF9`. `v = -1` "works" only because `0xFF` in all four bytes is unaffected by the sign-bit flooding.

**Why this helper is reached for `fill`.** TileLang emits `int` for `int8` with `lanes == 4` to avoid char-pack instructions ([`codegen_cuda.cc#L891-L900`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L891-L900)), so the vector constructor name resolves to `make_int`. The masked fast-path at [`codegen_cuda.cc#L4957-L4985`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L4957-L4985) is gated on `as_const_int(op->value)` returning non-null, but in the fill path the scalar is materialized as a runtime local (`signed char broadcast_var = (signed char)-7;`), so `as_const_int` returns `nullptr` and that path is skipped. Control falls through to the generic vector constructor at [`codegen_cuda.cc#L5176-L5186`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L5176-L5186), which emits `make_int(broadcast_var, broadcast_var, broadcast_var, broadcast_var)` — invoking the unmasked signed-char helper (confirmed by the generated CUDA above).

<details>
<summary>Lane-by-lane evidence and the wider-vector / unsigned overloads</summary>

| `v` | byte pattern | `make_int` result | result |
|-----|--------------|-------------------|--------|
| `5` | `0x05` | `0x05050505` | correct |
| `0` | `0x00` | `0x00000000` | correct |
| `-1` | `0xFF` | `0xFFFFFFFF` | correct (all-`0xFF` is sign-extension-invariant) |
| `-7` | `0xF9` | `0xFFFFFFF9` | wrong — lanes 1–3 become `0xFF` |
| `-128` | `0x80` | `0xFFFFFF80` | wrong — lanes 1–3 become `0xFF` |

The `make_int4(signed char × 16)` overload ([`common.h#L152-L164`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L152-L164)) delegates to `make_int` per group of four and inherits the same bug, which is why a 16×128 buffer (wider vector store) shows the same 3/4 corruption (`nbad=1536/2048`). The companion `make_uint(unsigned char, …)` ([`common.h#L177`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L177)) is correct because `unsigned char` zero-extends (`make_uint(0xF9 ×4) == 0xF9F9F9F9`), so the defect is specific to the signed-char overload.

</details>

**Suggested fix.** Mask each lane to 8 bits before shifting, so promotion to `int` cannot leak sign bits across lanes — at [`common.h#L138`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L136-L139):

```cpp
return ((x3 & 0xFF) << 24) | ((x2 & 0xFF) << 16) | ((x1 & 0xFF) << 8) | (x0 & 0xFF);
```

This matches the masked constant path the codegen already uses at [`codegen_cuda.cc#L4957-L4985`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L4957-L4985). The same `& 0xFF` mask would apply to the `make_int4(signed char × 16)` overload ([`common.h#L152-L164`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L152-L164)) since it delegates to `make_int`; the `make_uint` overload ([`common.h#L177`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L177)) is already correct, though the same mask there is harmless and defends future signed callers. With the helper masked, the generic runtime-broadcast fallback at [`codegen_cuda.cc#L5176-L5186`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L5176-L5186) becomes correct too, so no codegen change is strictly required. Verified: with the mask applied to `make_int`, the negative-`int8` fills go from 768/1024 wrong (`unique=[-7, -1]`) to 0/1024 wrong (`unique=[-7]`), and all non-negative `v` and `v = -1` remain bit-exact.

This is distinct from #1103 "Incorrect Code Generation for Vectorized Operations", which is a `make_longlong4` arg-count compile error (it fails to build) in the same `make_*` packing-helper family, not a silent sign-extension miscompile of negative `int8` fills. I searched the open and closed issue tracker and did not find an existing report of this defect.

Chennesxu:
## Summary

`T.fill` of an `int8` buffer with a negative scalar silently writes only 1 of every 4 elements with the intended value and sets the other 3 of every 4 to `-1` (`0xFF`). It compiles and runs with no error. Non-negative values are fine; `-1` is also fine because its bytes are already all `0xFF`.

```text
int8 dest, value = -7

before fix: [-7, -1, -1, -1, ...]
after fix : [-7, -7, -7, -7, ...]
```

## Root cause

A vectorized int8 `T.fill` packs four lanes into one 32-bit store via `make_int(v, v, v, v)` in `src/tl_templates/cuda/common.h`:

```cpp
return (x3 << 24) | (x2 << 16) | (x1 << 8) | x0;
```

Each `signed char` is integer-promoted to `int` before the shift/OR. A negative value sign-extends, for example `-7 -> 0xFFFFFFF9`, so the high 1-bits of the lower lanes flood the upper bytes through the OR. As a result, `make_int(-7, -7, -7, -7)` yields `0xFFFFFFF9`, whose bytes are `[F9, FF, FF, FF]`, i.e. `[-7, -1, -1, -1]`. Only the lowest lane survives. `-1` coincidentally looks correct because its bytes are already `0xFF`.

  ## Fix

  Build the 32-bit pattern from explicit unsigned bytes so the sign-extension bits cannot leak across lanes:

  ```cpp
  const unsigned int b0 = static_cast<unsigned char>(x0);
  const unsigned int b1 = static_cast<unsigned char>(x1);
  const unsigned int b2 = static_cast<unsigned char>(x2);
  const unsigned int b3 = static_cast<unsigned char>(x3);
  return static_cast<int>((b3 << 24) | (b2 << 16) | (b1 << 8) | b0);
  ```

  The constant int8x4 broadcast path in `src/cuda/codegen/codegen_cuda.cc` already takes the low byte before packing; only the runtime `make_int` helper was missing it. `make_int2` / `make_int4(char)` delegate to `make_int`; `make_uint*` take `unsigned char` and therefore do not sign-extend; and `make_int4(short...)` uses struct construction instead of byte packing.

## Tested

* Added `testing/python/language/test_tilelang_language_clear.py::test_fill_int8_negative`, which fills int8 buffers with `-7`, `-128`, `-1`, and `5`, then checks that every element equals the expected value.
* Verified locally on a TITAN RTX (sm_75, Turing): before the fix, negative fills produced 768/1024 wrong elements; after the fix, all elements are correct.

Fixes #2427

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CUDA integer packing to properly handle negative values without sign-extension interference.

* **Tests**
  * Added test coverage validating negative 8-bit integer operations in CUDA.

### DIFF
--- a/src/tl_templates/cuda/common.h
+++ b/src/tl_templates/cuda/common.h
@@ -137,10 +137,16 @@ TL_DEVICE unsigned __pack_nv_bfloat162(const bfloat16_t x, const bfloat16_t y) {
   return (v1 << 16) | v0;
 }
 
-// Pack four char values.
+// Pack four char values. Build the 32-bit pattern from unsigned bytes: a
+// negative signed char would otherwise sign-extend and flood the other lanes
+// through the OR.
 TL_DEVICE int make_int(signed char x0, signed char x1, signed char x2,
                        signed char x3) {
-  return (x3 << 24) | (x2 << 16) | (x1 << 8) | x0;
+  const unsigned int b0 = static_cast<unsigned char>(x0);
+  const unsigned int b1 = static_cast<unsigned char>(x1);
+  const unsigned int b2 = static_cast<unsigned char>(x2);
+  const unsigned int b3 = static_cast<unsigned char>(x3);
+  return static_cast<int>((b3 << 24) | (b2 << 16) | (b1 << 8) | b0);
 }
 
 // Pack eight char values.
--- a/testing/python/language/test_tilelang_language_clear.py
+++ b/testing/python/language/test_tilelang_language_clear.py
@@ -81,5 +81,27 @@ def test_shared_fill_cast_zero_uses_st_bulk():
     assert torch.allclose(out, torch.zeros_like(out))
 
 
+@tilelang.testing.requires_cuda
+def test_fill_int8_negative():
+    M, N = 8, 128
+
+    def program(value):
+        @T.prim_func
+        def main(out: T.Tensor((M, N), "int8")):
+            with T.Kernel(1, threads=128):
+                smem = T.alloc_shared((M, N), "int8")
+                T.fill(smem, value)
+                T.copy(smem, out)
+
+        return main
+
+    for value in (-7, -128, -1, 5):
+        kernel = tilelang.compile(program(value), out_idx=[0])
+        out = kernel()
+        torch.cuda.synchronize()
+        ref = torch.full((M, N), value, dtype=torch.int8, device="cuda")
+        torch.testing.assert_close(out, ref)
+
+
 if __name__ == "__main__":
     test_matmul()


LeiWang1999 (commented):
could we make the byte-packing arithmetic explicitly unsigned here?

  The `& 0xFF` masks fix the sign-extension bug, but `((x3 & 0xFF) << 24)` is still a signed `int` shift. Since this helper is constructing a 32-bit bit pattern, using unsigned intermediates would make the intent clearer and avoid relying on signed left-shift
  behavior when the top byte sets the sign bit.

  ```cpp
  const unsigned int b0 = static_cast<unsigned char>(x0);
  const unsigned int b1 = static_cast<unsigned char>(x1);
  const unsigned int b2 = static_cast<unsigned char>(x2);
  const unsigned int b3 = static_cast<unsigned char>(x3);
  return static_cast<int>((b3 << 24) | (b2 << 16) | (b1 << 8) | b0);
```

LeiWang1999 (approved):

Issue #2431, opened 2026-06-22
[Feature Request] Raise warning when vectorized code failed.


MARD1NO:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

When the code in T.vectorized loop fall back to serial, TileLang should raise a warning

example code:
```
import torch
import tilelang
from tilelang import language as T
from typing import Optional
import os


@tilelang.jit(
    pass_configs={
        tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True,
        tilelang.PassConfigKey.TL_DISABLE_OUT_OF_BOUND_WARNING: True,
        tilelang.PassConfigKey.TL_DISABLE_THREAD_STORAGE_SYNC: True,
    },
)
def get_rmsnorm_kernel(
    hidden_size: int
):  # fmt: off
    # Kernel config
    warp_size = 32
    num_threads = 128

    # Runtime symbols
    num_tokens = T.dynamic('num_tokens')

    num_tokens_per_block = num_threads // warp_size
    grid_num = T.ceildiv(num_tokens, num_tokens_per_block)
    elems_per_thread = hidden_size // warp_size
    vec_size = 8
    assert elems_per_thread % vec_size == 0, 'hidden_size must be divisible by warp_size * vec_size.'
    load_iter_num = elems_per_thread // vec_size

    @T.prim_func
    def rms_norm_kernel(
        hidden_states: T.Tensor[(num_tokens, hidden_size), T.bfloat16],
        weight: T.Tensor[(hidden_size,), T.float32],
        out: T.Tensor[(num_tokens, hidden_size), T.bfloat16],
    ):
        with T.Kernel(grid_num, threads=num_threads) as pid:
            thread_idx = T.get_thread_binding()
            token_idx = thread_idx // warp_size
            global_token_idx = token_idx + pid * num_tokens_per_block
            lane_idx = thread_idx % warp_size
            warp_stride = warp_size * vec_size

            if global_token_idx >= num_tokens:
                T.thread_return()
                
            hidden_states_local = T.alloc_local((elems_per_thread,), dtype=T.float32)
            norm_weight_local = T.alloc_local((elems_per_thread,), dtype=T.float32)

            powsum = T.alloc_var(dtype=T.float32)
            powsum = 0.0 

            for load_iter in T.serial(load_iter_num): 
[...]

PR #2439, opened 2026-06-22
[Transform] Warn on vectorized loop serial fallback

LeiWang1999:
## Summary

- Add a warning when `T.vectorized` is legalized into a serial loop because no valid vectorization plan is available.
- Make silent vectorization fallbacks easier to diagnose, especially for scalar accumulator updates inside vectorized loops.

## Changes

- Use the vectorization planner result in `LegalizeVectorizedLoop` to detect non-trivial `T.vectorized` loops whose vector size collapses to 1.
- Reuse the planned vector size as the `VectorizeLoop` hint so the planner is not run twice.
- Avoid warning for trivial extent-1 vectorized loops.

## Validation

- Manual `LegalizeVectorizedLoop` repro with a multi-statement `T.vectorized` loop and scalar `T.alloc_var` accumulator, confirming the warning and serial lowering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of vectorized loop lowering with enhanced fallback behavior when vectorization cannot be applied.
  * Added diagnostic warnings to identify cases where vectorized loops are converted to serial execution due to compatibility constraints.

### DIFF
--- a/src/transform/legalize_vectorized_loop.cc
+++ b/src/transform/legalize_vectorized_loop.cc
@@ -24,6 +24,7 @@
 
 #include "support/check.h"
 #include <tvm/ir/cast.h>
+#include <tvm/runtime/logging.h>
 #include <tvm/s_tir/utils.h>
 #include <tvm/tirx/builtin.h>
 #include <tvm/tirx/op.h>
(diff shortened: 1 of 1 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

LeiWang1999:
resolved issue #2431

PR #2448, opened 2026-06-24
[DEBUG] PassConfig for printing compile commands

bucket-xv:
## Summary
- Added a new pass config flag, `tl.print_device_compile_command`, to control printing the device compiler command used for CUDA compilation.
- Wired the flag through the C++ builtin/operator attribute keys and the Python pass config enum.
- Updated `compile_cuda` to emit the NVCC command when this pass config is enabled.

## C++ style / lint notes
- This PR touches C++ code in `src/op/builtin.cc` and `src/op/builtin.h`, which are covered by the project’s C++ style guidance in `docs/developer_guide/cpp_style.md`.
- The change appears to be a small API/config addition rather than a style-rule change; no blocking correctness, build, or test concerns are indicated from the summary.
- The “C++ API Style Audit (warning only)” CI step may surface advisory warnings for the new exported config key, but any TLCPP003/TLCPP004 findings here should be treated as warning-only unless they reveal a concrete API or maintainability issue.

### DIFF
--- a/tilelang/transform/pass_config.py
+++ b/tilelang/transform/pass_config.py
@@ -71,6 +71,10 @@ class PassConfigKey(str, Enum):
     TL_ENABLE_PTXAS_VERBOSE_OUTPUT = "tl.enable_ptxas_verbose_output"
     """Enable ptxas verbose output. Default: False"""
 
+    TL_PRINT_DEVICE_COMPILE_COMMAND = "tl.print_device_compile_command"
+    """Print the device compiler (e.g. nvcc) command line used to compile the
+    generated kernel. Useful for debugging compilation issues. Default: False"""
+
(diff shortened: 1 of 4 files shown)


tilelang/transform/pass_config.py
```diff
@@ -71,6 +71,10 @@ class PassConfigKey(str, Enum):
     TL_ENABLE_PTXAS_VERBOSE_OUTPUT = "tl.enable_ptxas_verbose_output"
     """Enable ptxas verbose output. Default: False"""
 
+    TL_PRINT_DEVICE_COMPILE_COMMAND = "tl.print_device_compile_command"
```
SiriusNEO, line 74: I think we can put this option in the `env.py`? There is a section for "Kernel Build options"

PR #2455, opened 2026-06-25
[BugFix] Fix T.Persistent dropping tiles when last dim is not a multiple of group_size

### LINKED ISSUE #2433: [BUG][Fuzzer][wrong-code] `T.Persistent` silently drops a band of output tiles instead of producing all of them when `n_blocks` isn't a multiple of 8
### Required prerequisites

0.1.11

### System information

- Installed from source (commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`, tag `v0.1.11`).
- Python 3.13, Linux.
- GPU: NVIDIA H100 80GB HBM3 (sm_90). The defect is in the architecture-independent persistent-loop index lowering (`PersistentFor`, `src/ir.cc`), not in any Hopper-specific path.

```python
import sys, tilelang, torch
print(sys.version.split()[0], sys.platform)                              # 3.13.13 linux
print(tilelang.__version__)                                              # 0.1.11
print(torch.__version__, torch.version.cuda)                             # 2.11.0+cu130 13.0
```

### Problem description

**TL;DR:** `T.Persistent` decodes each flat tile index back to a `(bx, by)` tile coordinate using `truncdiv(n_blocks, 8)` for the column-group count, which drops the remainder when `n_blocks` isn't a multiple of 8 — so the last `n_blocks % 8` columns of output tiles are never emitted and are left stale. Pure index-decode miscompile, no error. Fix direction: `truncdiv` → `ceildiv` + a coordinate-range guard.

A persistent GEMM written with the shipped `examples/gemm/example_gemm_persistent.py` pattern — `T.Persistent([ceildiv(M, block_M), ceildiv(N, block_N)], sm_num, block_id)` ([`example_gemm_persistent.py#L58`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/gemm/example_gemm_persistent.py#L58)) — at `M=512, N=2304, K=512`, `block_M=128, block_N=256` (so `m_blocks=4`, `n_blocks=9`, `group_size = min(8, 9) = 8`) compiles and runs with no error, but the three output tiles `(bx, by) = (1, 8), (2, 8), (3, 8)` — the entire `by = 8` last-column remainder band — are never produced and hold stale memory, for **any** `sm_num` (the trigger is divisibility alone; controls below). A divisor shape such as `N=2048 → n_blocks=8` is unaffected (Control A below). (Mechanism in Additional context.)

### Reproducible example code

(Use a real `.py` file — TileLang's eager front end calls `inspect.getsource`, so `python -c '…'` trips an unrelated `OSError`. This exercises the `T.Persistent` primitive **alone** — no GEMM, no shared memory, no pipeline, no tolerance: each persistent block increments `Visited[bx, by]` for every tile it visits, and we compare the visited set directly to the trivial full-grid reference. Correct: every tile visited exactly once → `Visited == all-ones`. Run as `python min_persistent.py <m_blocks> <n_blocks> <sm_num>`.)

```python
import sys
import tilelang
tilelang.disable_cache()
import tilelang.language as T
import torch

M_BLOCKS = int(sys.argv[1]) if len(sys.argv) > 1 else 4
N_BLOCKS = int(sys.argv[2]) if len(sys.argv) > 2 else 9
SM_NUM   = int(sys.argv[3]) if len(sys.argv) > 3 else 3


@tilelang.jit
def persistent_visit(Visited, sm_num):
    MB, NB = T.const("MB, NB")
    Visited: T.Tensor((MB, NB), T.int32)
    with T.Kernel(sm_num, threads=1) as block_id:
        for bx, by in T.Persistent([MB, NB], sm_num, block_id):
            Visited[bx, by] += 1


def run(m_blocks, n_blocks, sm_num):
    g = min(8, n_blocks)
    kernel = persistent_visit.compile(MB=m_blocks, NB=n_blocks, sm_num=sm_num)
    V = torch.zeros(m_blocks, n_blocks, device="cuda", dtype=torch.int32)
    kernel(V)
    ref = torch.ones(m_blocks, n_blocks, device="cuda", dtype=torch.int32)
    dropped = sorted((int(i), int(j)) for i, j in (V == 0).nonzero(as_tuple=False).tolist())
    over    = sorted((int(i), int(j), int(V[i, j])) for i, j in (V > 1).nonzero(as_tuple=False).tolist())
    ok = bool(torch.equal(V, ref))
    waves = -(-(m_blocks * n_blocks) // sm_num)
    print(f"m_blocks={m_blocks} n_blocks={n_blocks} group_size={g} "
          f"divides={'yes' if n_blocks % g == 0 else 'NO'} sm_num={sm_num} waves={waves}")
    print("dropped (Visited==0):", dropped)
    print("overflow (Visited>1):", over)
    print("PASS" if ok else "FAIL: persistent visited != full grid")


run(M_BLOCKS, N_BLOCKS, SM_NUM)
```

Output (deterministic across runs):

```
$ python min_persistent.py 4 9 3        # DEFECT: group_size=8 does NOT divide n_blocks=9, waves=12
m_blocks=4 n_blocks=9 group_size=8 divides=NO sm_num=3 waves=12
dropped (Visited==0): [(1, 8), (2, 8), (3, 8)]
overflow (Visited>1): []
FAIL: persistent visited != full grid

$ python min_persistent.py 4 8 3        # CONTROL A: divisor (8 divides 8), same code path
m_blocks=4 n_blocks=8 group_size=8 divides=yes sm_num=3 waves=11
dropped (Visited==0): []
overflow (Visited>1): []
PASS

$ python min_persistent.py 4 9 36       # CONTROL B: single wave (sm_num>=n_tiles, waves=1)
m_blocks=4 n_blocks=9 group_size=8 divides=NO sm_num=36 waves=1
dropped (Visited==0): [(1, 8), (2, 8), (3, 8)]
overflow (Visited>1): []
FAIL: persistent visited != full grid
```

The dropped band `{(1,8),(2,8),(3,8)}` is exactly the wrong-tile band of the persistent GEMM, with no overflow (never a duplicate). Control A (divisor `n_blocks=8`) PASSes and Control B (`sm_num=36`, single wave) still drops the same band — so the failure is divisibility alone, not the wave count. (Delta-debug: the smallest still-defective shape is `m_blocks=2 n_blocks=9`, dropping just `{(1,8)}`; `m_blocks=1 n_blocks=9` does not reproduce because the lone band tile `(0,8)` happens to be reached.)

### Traceback

```pytb
No traceback — the kernel compiles and runs to completion; the result is silently wrong (a stale output-tile band) and deterministic.
```

### Expected behavior

`T.Persistent(domain, grid, block_id)` should enumerate **every** tile in `domain` exactly once across all persistent waves, for any `grid` and any `domain` — including when the last-dimension extent is not a multiple of the internal `group_size`. The persistent GEMM should therefore produce all `m_blocks * n_blocks` output tiles and match the `torch` reference (0 wrong elements) for every domain shape, independent of `grid`.

TileLang's own behavior is the precedent: the divisor shapes `n_blocks ∈ {1..8}` write all tiles correctly at the same `sm_num` (Control A), so only the non-divisor grouped decode regresses. `T.Persistent` is the documented mechanism of the shipped `example_gemm_persistent.py`, whose whole purpose is to enumerate every output tile exactly once. The example's own non-primitive fallback (`use_persistent_primitive=False`) is correct precisely because it manually guards each tile with `if bx * block_M < M and by * block_N < N:` ([`example_gemm_persistent.py#L73`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/gemm/example_gemm_persistent.py#L73)) — but that guard only skips *over*-range tiles; it cannot recover a band the primitive's decode never enumerates.

### Additional context

**Root cause.** [`PersistentFor`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L140) builds a grouped mixed-radix domain to decode the flat persistent index into tile coordinates, but that domain's cardinality can be smaller than the true tile count when `group_size` does not divide the last extent:

- `domain_size = m_blocks * n_blocks` is the true count, and `waves = ceildiv(domain_size, wave_size)` is computed from it ([`#L152`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L152)), so the outer loop genuinely steps the flat index `rem` over `[0, domain_size)`.
- `group_size = min(group_size, domain.back())` ([`#L154`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L154)) caps `group_size` at `n_blocks` but does nothing about `n_blocks % group_size != 0`.
- The grouped domain is `[truncdiv(n_blocks, group_size), m_blocks, group_size]` ([`#L165-L170`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L165-L170)); with a truncating division its product (`32`) is strictly less than `domain_size` (`36`) for the repro.
- The mixed-radix decode ([`#L177-L184`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L177-L184)) reconstructs the column coordinate as `by = idxs[0] * group_size + idxs[last]` ([`#L205-L207`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L205-L207)). With `idxs[0] ∈ {0}` and `idxs[last] ∈ [0, 8)`, `by` only ever ranges over `[0, 8)`, so the `by = 8` remainder column is never emitted.
- The per-iteration break guard ([`#L186-L190`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L186-L190)) is keyed to `domain_size = 36`, so the flat indices in `[32, 36)` still run and the decode overflows to out-of-range `by ∈ {9, 10, 11}`.

This is the primitive lowering itself, not the caller's index math — the example calls `T.Persistent` correctly ([`#L58`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/gemm/example_gemm_persistent.py#L58)). The `truncdiv` grouped decode has been present since the persistent loop was introduced in #563 ("Introduce Persistent Loop and Update GEMM Example"), so this affects every `T.Persistent` use with a non-multiple-of-8 last dimension since that feature landed. The shipped `example_gemm_persistent.py` does not trip it because its defaults (`M=N=K=4096`/`8192` with `block_N` 128/256) make `n_blocks` a multiple of 8, where `group_size = 8` divides exactly.

<details>
<summary>Full divisibility sweep (n_blocks 1..12, m_blocks=4) — dropped iff group_size ∤ n_blocks</summary>

A tile band is dropped **iff `group_size = min(8, n_blocks)` does NOT divide `n_blocks`**; the dropped tiles are always the last-dimension remainder column band `by ∈ [ (n_blocks // group_size) * group_size, n_blocks )` — always a dropped band, never a duplicate. Swept at `m_blocks = 4`, `sm_num = 3` (multi-wave), plus a single-wave control at `sm_num = 36`:

| `n_blocks` | `group_size` | divides? | `sm_num` | `waves` | dropped columns `by` |
|-----------:|-------------:|:--------:|:--------:|:-------:|:---------------------|
| 1 | 1 | yes | 3 | 2 | none |
| 2 | 2 | yes | 3 | 3 | none |
| 3 | 3 | yes | 3 | 4 | none |
| 4 | 4 | yes | 3 | 6 | none |
| 5 | 5 | yes | 3 | 7 | none |
| 6 | 6 | yes | 3 | 8 | none |
| 7 | 7 | yes | 3 | 10 | none |
| 8 | 8 | yes | 3 | 11 | none |
| **9** | 8 | **no** | 3 | 12 | **drops `{8}`** |
| **10** | 8 | **no** | 3 | 14 | **drops `{8, 9}`** |
| **11** | 8 | **no** | 3 | 15 | **drops `{8, 9, 10}`** |
| **12** | 8 | **no** | 3 | 16 | **drops `{8, 9, 10, 11}`** |
| **9** | 8 | **no** | **36** | **1** | **drops `{8}` (single wave!)** |

The rows `n_blocks ∈ {1..8}` are safe because `min(8, n_blocks) == n_blocks` divides trivially; the first non-divisor `n_blocks = 9` is the first failure, and the remainder band grows by one column per unit past the largest multiple of `group_size` below `n_blocks`. The number of waves is irrelevant: the truncated 32-element grouped space is stepped over the full `[0, 36)` flat range in both the single- and multi-wave cases.

</details>

**Note (unrelated path).** With `num_stages >= 2` (the warp-specialized path) this shape independently hits a separate compile-time assert (`ProducerConsumerWS: failed to replace pipeline loop`); `num_stages=0` isolates the tile-skip miscompile cleanly, away from that unrelated WS issue.

**Suggested fix.** Make the grouped decode space cover the full last-dimension extent so its cardinality is at least the true tile count, and skip the padded tail. Roughly:

1. Use `ceildiv(domain.back(), group_size)` for `grouped_domain[0]` ([`#L165-L170`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/ir.cc#L165-L170)) so the grouped product `>= domain_size` and the remainder band is enumerated (`[ceildiv(9, 8), 4, 8] = [2, 4, 8]`, product `64`).
2. Because `ceildiv` can now reconstruct padded coordinates with `by >= n_blocks`, add a per-tile coordinate-range guard (`bx < domain[0] && by < domain.back()`) around the body so the padded tail is skipped — mirroring the guard the example's non-primitive fallback already uses ([`#L73`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/gemm/example_gemm_persistent.py#L73)).
3. The wave loop must then iterate over the full padded grouped space (its larger cardinality), not over `domain_size`, or the tail tiles beyond the first padded group are still skipped.

Verified on a from-source build: with all three together, the repro's wrong tiles `[(1,8),(2,8),(3,8)]` / 11980 wrong elems become `[]` / 0 wrong elems (PASS), and a divisor shape (`n_blocks = 8`) still PASSes. Applying only points 1–2 (keeping the wave loop bounded by `domain_size`) under-fixes: the grouped space `64 > 36`, so iterating only `[0, 36)` recovers just one of the dropped tiles and still drops `(1,8),(2,8),(3,8)` — point 3 is required. A numerical regression test with a non-divisor `n_blocks` would help guard this.

I searched the open and closed issue tracker and did not find an existing report of this defect.

RuneFang:
## [Bugfix] PersistentFor: stop dropping the trailing tile band when `n_blocks % group_size != 0`

Fixes #2433.

### Summary

`T.Persistent` silently drops a band of output tiles whenever the persistent loop's last domain dimension isn't a multiple of `group_size` (which defaults to `8`). On the issue's exact shape (`M=512, N=2304, K=512, block=128x256` →
`4×9` tile grid) the entire `by=8` column was never visited, leaving NaN columns in the output.

This PR fixes the IR construction in `PersistentFor` (`src/ir.cc`) and adds a primitive-level reproducer/regression test that doesn't depend on the warp-specialized GEMM path.

### The fix (three coupled changes)

1. `truncdiv → ceildiv` so the grouped space covers every real tile 
2. Iterate the wave loop and the early-break check over the **padded**grouped cardinality `padded_domain_size = grouped_domain`. Without this, padded slots steal flat-index positions from real tiles in earlier waves and just shift the "dropped band" symptom around.
3. Wrap the user body in an `if (last_coord < domain.back())` guard so the padded tail is skipped at execution time.

### Testing

image
image

## Summary
- Fixed `T.Persistent` loop grouping in `src/ir.cc` so the final dimension uses ceil-based padding instead of truncation, preventing trailing tile columns from being silently skipped.
- Updated the wave loop/cardinality and `loop_break` sequencing to use the padded grouped size, and added an execution-time guard (`last_coord < domain.back()`) so padded tiles are not processed.
- Added a primitive-level regression test for issue `#2433` (`testing/python/issue/test_tilelang_issue_2433.py`) that reproduces the missing-tile/divisibility failure without relying on the warp-specialized GEMM path.

## Testing
- `pytest testing/python/issue/test_tilelang_issue_2433.py`
- `python -m pytest testing/python/issue/ -v --tb=short`

## C++ style / lint notes
- This PR touches C++ (`src/ir.cc`), which falls under `docs/developer_guide/cpp_style.md` (“This guide applies to C++ source under `src/`…”). It does not introduce a new public/API/FFI surface.
- CI note: the “C++ API Style Audit (warning only)” step is not expected to be affected by this behavioral fix; any TLCPP003/TLCPP004 findings (if they appear) should be treated as advisory unless the PR introduces a concrete API/maintainability risk.
- Warning-only signals from tests: observed pytest output includes `PytestUnknownMarkWarning` for `gpu`/`cuda` and some deprecation warnings (`T.symbolic`→`T.dynamic`, `T.Buffer`→`T.Tensor`, `tl.disable_tma_lower` deprecation). These do not indicate build/correctness issues introduced by this PR.

### DIFF
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -143,9 +143,20 @@ ForFrame PersistentFor(const Array<PrimExpr> &domain, const PrimExpr &wave_size,
     domain_size *= domain[i];
   }
 
-  auto waves = ceildiv(domain_size, wave_size);
-  auto loop_var = Var("w", waves.dtype());
   group_size = min(group_size, domain[domain.size() - 1]);
+  Array<PrimExpr> grouped_domain;
+  grouped_domain.push_back(ceildiv(domain[domain.size() - 1], group_size));
+  for (int i = 0; i < domain.size() - 1; ++i) {
+    grouped_domain.push_back(domain[i]);
+  }
+  grouped_domain.push_back(group_size);
+  PrimExpr padded_domain_size = grouped_domain[0];
+  for (int i = 1; i < grouped_domain.size(); ++i) {
+    padded_domain_size *= grouped_domain[i];
+  }
+
+  auto waves = ceildiv(padded_domain_size, wave_size);
+  auto loop_var = Var("w", waves.dtype());
   Array<Var> coord_vars;
 
   for (int i = 0; i < domain.size(); ++i) {
@@ -155,13 +166,6 @@ ForFrame PersistentFor(const Array<PrimExpr> &domain, const PrimExpr &wave_size,
     n->vars.push_back(coord);
     n->doms.push_back(Range(make_const(dtype, 0), domain[i]));
   }
-
-  Array<PrimExpr> grouped_domain;
-  grouped_domain.push_back(truncdiv(domain[domain.size() - 1], group_size));
-  for (int i = 0; i < domain.size() - 1; ++i) {
-    grouped_domain.push_back(domain[i]);
-  }
-  grouped_domain.push_back(group_size);
 
   n->f_make_for_loop = [=](const Array<Var> &vars, const Array<Range> &doms,
                            const Array<Optional<PrimExpr>> &steps,
@@ -176,17 +180,20 @@ ForFrame PersistentFor(const Array<PrimExpr> &domain, const PrimExpr &wave_size,
       rem = truncdiv(rem, grouped_domain[i]);
     }
     idxs.Set(0, rem);
-
+    PrimExpr last_coord =
+        idxs[0] * group_size + idxs[grouped_domain.size() - 1];
+    PrimExpr in_range = last_coord < domain[domain.size() - 1];
     auto out_if = tvm::tirx::IfThenElse(
-        domain_size <= (loop_var * wave_size + index),
+        padded_domain_size <= (loop_var * wave_size + index),
         tvm::tirx::Evaluate(
             tvm::tirx::Call(DataType::Handle(), tvm::tl::loop_break(), {})),
         Stmt());
+    Stmt guarded_body = tvm::tirx::IfThenElse(in_range, body, Stmt());
 
     arith::Analyzer analyzer;
-    Stmt new_body = body;
+    Stmt new_body = guarded_body;
     if (analyzer.CanProveGreaterEqual(waves, 2)) {
-      new_body = SeqStmt({out_if, body});
+      new_body = SeqStmt({out_if, guarded_body});
     }
     Optional<PrimExpr> step =
         !steps.empty() ? steps[0] : Optional<PrimExpr>(std::nullopt);
@@ -196,9 +203,7 @@ ForFrame PersistentFor(const Array<PrimExpr> &domain, const PrimExpr &wave_size,
     for (int i = 0; i < vars.size() - 1; ++i) {
       outer = SeqStmt({tirx::Bind(vars[i], idxs[i + 1]), outer});
     }
-    outer = SeqStmt({tirx::Bind(vars[vars.size() - 1],
-                                idxs[0] * group_size + idxs[vars.size()]),
-                     outer});
+    outer = SeqStmt({tirx::Bind(vars[vars.size() - 1], last_coord), outer});
     return outer;
   };
 
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -143,9 +143,11 @@ ForFrame PersistentFor(const Array<PrimExpr> &domain, const PrimExpr &wave_size,
     domain_size *= domain[i];
   }
 
-  group_size = min(group_size, domain[domain.size() - 1]);
+  PrimExpr last_extent = domain[domain.size() - 1];
+  group_size =
+      max(make_const(group_size.dtype(), 1), min(group_size, last_extent));
   Array<PrimExpr> grouped_domain;
-  grouped_domain.push_back(ceildiv(domain[domain.size() - 1], group_size));
+  grouped_domain.push_back(ceildiv(last_extent, group_size));
   for (int i = 0; i < domain.size() - 1; ++i) {
     grouped_domain.push_back(domain[i]);
   }
--- a/testing/python/issue/test_tilelang_issue_2433.py
+++ b/testing/python/issue/test_tilelang_issue_2433.py
@@ -1,0 +1,93 @@
+"""Regression for tile-ai/tilelang issue #2433."""
+
+import torch
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+
+
+@tilelang.jit
+def _persistent_visit_kernel(MB, NB, sm_num):
+
+    @T.prim_func
+    def kernel(Visited: T.Tensor((MB, NB), T.int32)):
+        with T.Kernel(sm_num, threads=1) as block_id:
+            for bx, by in T.Persistent([MB, NB], sm_num, block_id):
+                Visited[bx, by] += 1
+
+    return kernel
+
+
+@tilelang.jit
+def _persistent_visit_kernel_3d(D0, D1, D2, sm_num):
+
+    @T.prim_func
+    def kernel(Visited: T.Tensor((D0, D1, D2), T.int32)):
+        with T.Kernel(sm_num, threads=1) as block_id:
+            for i, j, k in T.Persistent([D0, D1, D2], sm_num, block_id):
+                Visited[i, j, k] += 1
+
+    return kernel
+
+
+def run_persistent_visit(m_blocks: int, n_blocks: int, sm_num: int):
+    kernel = _persistent_visit_kernel(m_blocks, n_blocks, sm_num)
+    visited = torch.zeros(m_blocks, n_blocks, device="cuda", dtype=torch.int32)
+    kernel(visited)
+    expected = torch.ones_like(visited)
+    assert torch.equal(visited, expected), (
+        f"T.Persistent visited grid mismatch for "
+        f"m_blocks={m_blocks}, n_blocks={n_blocks}, sm_num={sm_num}: "
+        f"missing={(visited == 0).nonzero(as_tuple=False).tolist()}, "
+        f"overflow={(visited > 1).nonzero(as_tuple=False).tolist()}"
+    )
+
+
+def run_persistent_visit_3d(d0: int, d1: int, d2: int, sm_num: int):
+    kernel = _persistent_visit_kernel_3d(d0, d1, d2, sm_num)
+    visited = torch.zeros(d0, d1, d2, device="cuda", dtype=torch.int32)
+    kernel(visited)
+    expected = torch.ones_like(visited)
+    assert torch.equal(visited, expected), (
+        f"T.Persistent 3D visited grid mismatch for "
+        f"({d0},{d1},{d2}), sm_num={sm_num}: "
+        f"missing={(visited == 0).nonzero(as_tuple=False).tolist()}, "
+        f"overflow={(visited > 1).nonzero(as_tuple=False).tolist()}"
+    )
+
+
+@tilelang.testing.requires_cuda
+def test_issue_2433():
+    """Reproduce the failure table from the issue and the divisor controls.
+
+    With group_size defaulting to min(8, n_blocks), only the rows where
+    n_blocks % group_size != 0 trip the bug; the divisor rows act as
+    controls to make sure the fix doesn't regress them.
+    """
+    # Divisor controls (must always pass).
+    for n_blocks in (1, 2, 4, 8):
+        run_persistent_visit(m_blocks=4, n_blocks=n_blocks, sm_num=3)
+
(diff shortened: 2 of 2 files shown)


RuneFang:
@LeiWang1999  
Would you be willing to help me review it? when you have time 🙏

RuneFang:
@LeiWang1999  https://github.com/tile-ai/tilelang/pull/2461 CI failed，I think this is a random failure. I've already submitted a separate PR to fix it. and  I'd like to put them in one PR. https://github.com/tile-ai/tilelang/actions/runs/28152838420/job/83583983878

chengyupku:
LGTM, merged

PR #2467, opened 2026-06-27
[BugFix] Skip source-compilation options when exporting LLVM module

penguin-wwy:
Running the LLVM backend tests on macOS produces a cache-save compilation error, even though the kernel itself compiles and runs correctly:
```
PYTHONPATH=$(pwd) python -m pytest testing/python/llvm/test_tilelang_llvm_gemm.py -s
The test passes (kernel executes correctly), but the disk-cache save fails with:
[TileLang:tilelang.cache.kernel_cache:ERROR] Error during atomic cache save Compilation error:
/var/folders/.../lib0.o:1:1: error: source file is not valid UTF-8
/var/folders/.../lib1.o:1:1: error: source file is not valid UTF-8
/var/folders/.../devc.o:1:1: error: source file is not valid UTF-8
...
```

The failing command (from the log):
```
/usr/bin/g++ -shared -fPIC -undefined dynamic_lookup -o ...so \
  -x objective-c++ -g -std=gnu++17 -I...torch/include \
  lib0.o lib1.o devc.o
```
The binary .o object files are being parsed as source text.

The error is triggered by the interaction between tilelang's cache-save path and TVM's export_library:

- target=llvm: TVM's export_library emits binary .o object files and constructs a link command (-shared -fPIC -undefined dynamic_lookup) to combine them into a .so. No source compilation is involved.
- tilelang on macOS: _get_compile_args() unconditionally returns -x objective-c++ (originally intended for the Metal codegen path, which emits C source).

## Summary
- Fixed macOS kernel cache export failures for the LLVM backend by ensuring LLVM export/link does not receive Metal/source-compilation flags (e.g., `-x objective-c++`).
- Refactored cache export argument handling:
  - Renamed `_get_compile_args` → `_get_source_compile_args`.
  - Added `_get_export_link_args`.
  - Updated `KernelCache._safe_write_executable(...)` to accept explicit `export_kwargs` and pass them directly to `Executable.export_library` (so source vs export args are no longer inferred inside the atomic write helper).
- Updated `TVMFFIKernelCache` to select `export_kwargs` based on the kernel’s normalized **host target** (`kernel.artifact.target_host.kind.name`), avoiding using the device target to decide host export behavior.
  - Tests cover that LLVM host export does not include `"options"` while non-LLVM host export uses the source compile args.
- Extended artifact metadata so export can reliably use host target info:
  - Added `target` and `target_host` fields to `CompiledArtifact`.
  - Updated `lower()` to populate/pass `target` and `target_host` when host codegen is enabled/disabled.

## Regression coverage
- Added/updated atomic-save tests in `testing/python/cache/test_tilelang_kernel_cache_atomic_save.py` to verify:
  - `KernelCache._safe_write_executable` forwards explicit `export_kwargs` unchanged.
  - `TVMFFIKernelCache` chooses export kwargs from the normalized host target kind (LLVM vs `c`).

## C++ style / lint notes
- This PR does not touch C++ code or the C++ style guide in `docs/developer_guide/cpp_style.md` (it only changes Python cache/JIT logic and Python tests).
- The “C++ API Style Audit (warning only)” CI step is not expected to be affected, and no TLCPP003/TLCPP004 warning risks are introduced by this change.


LeiWang1999:
Thanks @penguin-wwy , would be better to add a regression test for this change :)

penguin-wwy:
> Thanks @penguin-wwy , would be better to add a regression test for this change :)

Done

penguin-wwy:
@LeiWang1999 would appreciate a review :)

LeiWang1999:
Thanks, I just pushed a small follow-up to make the fix closer to the underlying issue.

Instead of checking `kernel.target.kind.name == "llvm"` inside `_safe_write_executable()`, the updated version separates source compile args from export/link args. `_safe_write_executable()` is now just an atomic export helper, and the TVM FFI cache chooses export kwargs based on the normalized host target. This avoids using the device target to infer host export behavior, which could miss cases like CUDA device target with LLVM host export :)

LeiWang1999 (approved):

LeiWang1999:
Let me know if you think these changes make sense then we can let this pr in.

penguin-wwy:
> Let me know if you think these changes make sense then we can let this pr in.

Great, distinguishing between compiler options and linker options does indeed get closer to the essential issue.

PR #2469, opened 2026-06-28
[BugFix] Fix LayoutInference divide-by-zero on non-power-of-two broadcast

### LINKED ISSUE #2394: [BUG][Fuzzer][ice-on-valid-code] LayoutInference crashes with an internal "Divide by zero" instead of compiling — column-broadcast into a 2-D `T.Parallel` when the tile width isn't a power of two
### Required prerequisites

0.1.11+cuda.gita4399e4c

### System information

NVIDIA A10G (sm_86), Python 3.11. This is a `LayoutInference` (compile-time) crash — it happens before
codegen, so the CUDA/PyTorch runtime versions are not involved.

### Problem description

A kernel that broadcasts a 1-D fragment across the columns of a 2-D loop (`x[i, j] = d_local[j] * 2.0`)
compiles fine with tile width `BN = 32`, but crashes the compiler with an internal divide-by-zero when
`BN = 48` — nothing else changed. The trigger is simply that `BN` is not a power of two:

| `BN` | result |
|------|--------|
| 16, 32, 64, 128 (powers of two) | compiles & runs |
| 24, 40, 48, 96 (not powers of two) | `InternalError: Check failed: pb->value != 0 : Divide by zero` |

The crash happens during `LayoutInference` (at compile time), inside TVM's constant folder.

### Reproducible example code

```python
M, N = 256, 192
BM, BN = 64, 48          # BN=48 (not a power of two) -> crashes; 32 or 64 compile fine

@tilelang.jit
def k(D_in: T.Tensor((N,), T.bfloat16), Out: T.Tensor((M, N), T.bfloat16)):
        d_local  = T.alloc_fragment((BN,), T.float32)
        d_shared = T.alloc_shared((BN,), T.bfloat16)
        x  = T.alloc_fragment((BM, BN), T.float32)
        xs = T.alloc_shared((BM, BN), T.bfloat16)
        T.copy(D_in[bx * BN], d_shared)
        T.copy(d_shared, d_local)
        for i, j in T.Parallel(BM, BN):
            x[i, j] = d_local[j] * 2.0      # 1-D fragment broadcast along the COLUMN index
        T.copy(x, xs)
        T.copy(xs, Out[by * BM, bx * BN])

D = torch.randn(N, dtype=torch.bfloat16, device='cuda')
Out = torch.zeros(M, N, dtype=torch.bfloat16, device='cuda')
k(D, Out)
```

### Traceback

```pytb
File ".../tilelang/engine/phase.py", line 191, in LowerAndLegalize
  ...
  tvm::tl::ParallelOpNode::CompleteBufferFragment(...)
  tvm::tl::infer_fragment_index(...) / DivideUnusedIterators / get_unused_iters
  tvm::arith::TryConstFold<tvm::tir::FloorMod>(...)
tvm.error.InternalError: Check failed: pb->value != 0 (0 vs. 0) : Divide by zero
```

### Expected behavior

It should compile, like the power-of-two widths do — a valid layout exists for `BN = 48` (confirmed: with
the fix below it compiles and the result is numerically exact). At minimum it should give a clear error,
not an internal divide-by-zero.

### Additional context

**The intuition.** With 128 threads, the element→thread map is taken `% 128` so the thread index wraps
into `[0, 128)` — correct and necessary. To finish the layout, inference computes a "leftover" size =
`(total positions in this dim) / (positions already claimed)`. For `BN = 48` the claimed size is
recorded from the `% 128` as its upper bound **128** (times a stride of 4 → **512**), instead of the
true count of distinct values, **12**. But the dim only has **48** positions, and 512 > 48 is
impossible for "already claimed", so the leftover comes out `48 / 512 = 0`. That `0` is then used as a
divisor (a modulus) further on → `x % 0` → abort. It should have been `48 / 48 = 1` (the same leftover
the working `BN = 32` case produces). The root mistake is recording the size as the modulus 128 rather
than the actual 12; the `/0` is a downstream symptom.

(Why 128 and not 12: the size is stamped from the structure `% 128` — its result range `[0,128)` — and
the narrower fact that the input only reaches 48, so the values can't exceed 12, isn't propagated. 128
is a safe over-estimate that's normally harmless, but here it lands in a divisor. That stamping happens
in the bundled TVM (`3rdparty/tvm`, a submodule), not in TileLang, which is why the fix below is on the
TileLang side.)

**In code,** in [`get_unused_iters`](https://github.com/tile-ai/tilelang/blob/a4399e4c9934edd4e34bde89628eba9377548b10/src/layout/utils.cc#L76):
the per-thread index is `(_i // 4) % 128` over `_i ∈ [0, 48)`. `_i // 4` has 12 distinct values but the
split records its extent as the modulus **128**, so `expected_lower_factor = 4 * 128 = 512` exceeds
`mark->extent = 48`, and the leftover iterator's extent is
`FloorDiv(mark->extent, expected_lower_factor) = FloorDiv(48, 512) = 0`
([utils.cc:122](https://github.com/tile-ai/tilelang/blob/a4399e4c9934edd4e34bde89628eba9377548b10/src/layout/utils.cc#L122)).
That zero-extent iterator is carried downstream (through `infer_fragment_index` →
`NormalizeIterMapToExpr`) where its extent becomes a modulus — `x % 0` — and `ICHECK(divisor != 0)`
aborts. (Power-of-two widths divide evenly, so the size is never over-counted.)

A minimal fix is to clamp the claimed span to the mark extent in the matched-split branch of
`get_unused_iters`:
```cpp
PrimExpr covered = splits[j]->lower_factor * splits[j]->extent;
if (analyzer->CanProve(covered > mark->extent)) covered = mark->extent;
expected_lower_factor = covered;
```
Verified: with this, `BN ∈ {40, 48, 96}` compile and the broadcast is numerically exact (matches `D*2`),
power-of-two widths are unchanged, and GEMM / `T.reduce_sum` still pass (no regression in the shared
`get_unused_iters`). The clamp rests on an invariant that always holds here — a claimed span can't exceed
the source size — so it stays in TileLang and needs no TVM change; the alternative (fixing the modulus-as-
extent in the TVM submodule) is broader and touches code shared well beyond this path.

Chennesxu:
## Summary

  `LayoutInference` can crash with an internal `Divide by zero` when lowering a valid column-broadcast fragment access whose tile width is not a power of two.

  ```python
  d_local = T.alloc_fragment((BN,), T.float32)
  x = T.alloc_fragment((BM, BN), T.float32)

  for i, j in T.Parallel(BM, BN):
      x[i, j] = d_local[j] * 2.0
  ```

  ## Root cause

  `get_unused_iters` tracks the portion of an iterator already covered by existing `IterSplitExpr`s using:

  ```cpp
  expected_lower_factor = splits[j]->lower_factor * splits[j]->extent;
  ```

  For non-power-of-two tile widths, iter split normalization may over-approximate the covered span. For example, an iterator with extent `48` may be represented by a split span that covers up to `64`.

  The old logic only treated exact equality as full coverage:

  ```cpp
  expected_lower_factor == mark->extent
  ```

  So when the covered span became larger than the mark extent, it was treated as incomplete coverage. `get_unused_iters` then synthesized a leftover split with zero extent, which later normalized into a divide-
  by-zero.

  ## Fix

  Keep the split-chain progression intact, but treat proven over-coverage as full coverage at the final leftover check:

  ```cpp
  bool covers_full_iter =
      analyzer->CanProveEqual(expected_lower_factor, mark->extent) ||
      analyzer->CanProve(expected_lower_factor > mark->extent);
  ```

  If the consumed span is equal to or greater than the original iterator extent, there is no remaining iterator space to synthesize, so `get_unused_iters` does not emit a zero-extent leftover split.

  The fix also simplifies the newly advanced `expected_lower_factor` before the final coverage check.

  ## Tested

  - Added regression coverage in `testing/python/transform/test_tilelang_transform_layout_inference.py`:
    - host-lowering crash regression with `block_n = 24, 40, 48, 64, 96`
    - CUDA numerical regression with `block_n = 32, 48, 96`, checking that the column broadcast matches `D * 2`
  - Verified locally on a TITAN RTX host:
    - `testing/python/transform/test_tilelang_transform_layout_inference.py`
    - `testing/python/layout`

  Fixes #2394

## Summary
- Fixed a `LayoutInference` compile-time crash in leftover-iterator handling by treating proven over-coverage as full coverage, preventing zero-extent leftover splits that could lead to a divide-by-zero during lowering of valid non-power-of-two column-broadcast fragments (issue `#2394`).
- Added regression tests covering both lowering and CUDA numerical correctness for column-broadcast fragment kernels across several non-power-of-two `block_n` values (including newly added `24` and `40`), and verified that numerical output matches the reference broadcasted result.

## C++ style / lint notes
- This PR touches C++ layout inference logic in `src/layout/utils.cc`, so it is relevant to `docs/developer_guide/cpp_style.md`.
- No public API/FFI changes were introduced.
- The “C++ API Style Audit (warning only)” CI step should still be expected to run for this C++ change; this PR does not introduce any specific new warning-only findings based on the provided diff.

### DIFF
--- a/src/layout/utils.cc
+++ b/src/layout/utils.cc
@@ -111,12 +111,17 @@ Array<IterSplitExpr> get_unused_iters(const IterMark &mark,
     } else {
       used[j] = true;
       i++;
-      expected_lower_factor = splits[j]->lower_factor * splits[j]->extent;
+      expected_lower_factor =
+          analyzer->Simplify(splits[j]->lower_factor * splits[j]->extent);
     }
   }
-  bool match_full_iter =
-      analyzer->CanProveEqual(expected_lower_factor, mark->extent);
-  if (!match_full_iter) {
+  // Iter split normalization may over-approximate the original mark span.
+  // Treat over-coverage as fully covered instead of synthesizing a zero-extent
+  // leftover iterator, which later normalizes into division by zero.
+  bool covers_full_iter =
+      analyzer->CanProveEqual(expected_lower_factor, mark->extent) ||
+      analyzer->CanProve(expected_lower_factor > mark->extent);
+  if (!covers_full_iter) {
     results.emplace_back(
         mark, expected_lower_factor,
         analyzer->Simplify(FloorDiv(mark->extent, expected_lower_factor)), 1);
--- a/testing/python/transform/test_tilelang_transform_layout_inference.py
+++ b/testing/python/transform/test_tilelang_transform_layout_inference.py
@@ -223,7 +223,7 @@ def _column_broadcast_fragment_kernel(block_n):
 
 
 @tilelang.testing.requires_cuda
-@pytest.mark.parametrize("block_n", [32, 48, 96])
+@pytest.mark.parametrize("block_n", [24, 32, 40, 48, 96])
 def test_column_broadcast_fragment_values(block_n):
     # Numerical regression for issue #2394: the column broadcast must match D*2.
     kernel = _column_broadcast_fragment_kernel(block_n)
--- a/testing/python/transform/test_tilelang_transform_layout_inference.py
+++ b/testing/python/transform/test_tilelang_transform_layout_inference.py
@@ -4,6 +4,7 @@ import tilelang as tl
 import tilelang.language as T
 import tilelang.testing
 import pytest
+import torch
 
 auto_target = tvm.target.Target(determine_target("auto"))
 
@@ -168,6 +169,70 @@ def test_static_ragged_copy_allows_1024_elements_384_threads():
     assert "B[((i * 384) + ((int)threadIdx.x))]" in kernel_source
     assert "(((int)threadIdx.x) >> 7)) < 8" in kernel_source
     assert "threadIdx.x) < 128" not in kernel_source
+
+
+@pytest.mark.parametrize("block_n", [24, 40, 48, 64, 96])
+def test_column_broadcast_fragment_tile_width_lowers(block_n):
+    # Regression for issue #2394: LayoutInference used to synthesize a zero-extent
+    # leftover iterator for non-power-of-two column broadcasts, then divide by zero.
+    m, n = 256, block_n * 4
+    block_m = 64
+
+    @T.prim_func
+    def main(D_in: T.Tensor((n,), T.bfloat16), Out: T.Tensor((m, n), T.bfloat16)):
+        with T.Kernel(T.ceildiv(n, block_n), T.ceildiv(m, block_m), threads=128) as (bx, by):
+            d_local = T.alloc_fragment((block_n,), T.float32)
+            d_shared = T.alloc_shared((block_n,), T.bfloat16)
+            x = T.alloc_fragment((block_m, block_n), T.float32)
+            xs = T.alloc_shared((block_m, block_n), T.bfloat16)
+
+            T.copy(D_in[bx * block_n], d_shared)
+            T.copy(d_shared, d_local)
+            for i, j in T.Parallel(block_m, block_n):
+                x[i, j] = d_local[j] * 2.0
+            T.copy(x, xs)
+            T.copy(xs, Out[by * block_m, bx * block_n])
+
+    with tvm.target.Target(auto_target):
+        artifact = tl.lower(main, target=auto_target, enable_device_compile=False)
+
+    assert artifact.kernel_source
+
+
+@tl.jit(out_idx=[1])
+def _column_broadcast_fragment_kernel(block_n):
+    m, n = 256, block_n * 4
+    block_m = 64
+
+    @T.prim_func
+    def main(D_in: T.Tensor((n,), T.float32), Out: T.Tensor((m, n), T.float32)):
+        with T.Kernel(T.ceildiv(n, block_n), T.ceildiv(m, block_m), threads=128) as (bx, by):
+            d_local = T.alloc_fragment((block_n,), T.float32)
+            d_shared = T.alloc_shared((block_n,), T.float32)
+            x = T.alloc_fragment((block_m, block_n), T.float32)
+            xs = T.alloc_shared((block_m, block_n), T.float32)
+
+            T.copy(D_in[bx * block_n], d_shared)
+            T.copy(d_shared, d_local)
+            for i, j in T.Parallel(block_m, block_n):
+                x[i, j] = d_local[j] * 2.0
+            T.copy(x, xs)
+            T.copy(xs, Out[by * block_m, bx * block_n])
+
+    return main
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("block_n", [32, 48, 96])
+def test_column_broadcast_fragment_values(block_n):
+    # Numerical regression for issue #2394: the column broadcast must match D*2.
+    kernel = _column_broadcast_fragment_kernel(block_n)
+
+    d = torch.arange(block_n * 4, device="cuda", dtype=torch.float32)
+    out = kernel(d)
+    expected = d.unsqueeze(0).expand(256, -1) * 2.0
+
+    assert torch.equal(out, expected)
 
 
 if __name__ == "__main__":


LeiWang1999:
@regression-perf

Chennesxu:
> # Performance Regression Test Report
> Triggered by: @LeiWang1999 Workflow run: https://github.com/tile-ai/tilelang/actions/runs/28357306283
> 
> ## Results
> File	Original Latency	Current Latency	Speedup
> example_topk	30.0392	39.0637	0.768981
> example_mha_sink_bwd_bhsd_sliding_window	0.0379896	0.038633	0.983346
> example_dequant_gemm_bf16_mxfp4_hopper	0.355491	0.359938	0.987647
> example_vertical_slash_sparse_attn	0.15908	0.160987	0.988158
> example_mhc_pre	0.14348	0.144572	0.992446
> example_mha_bwd_bhsd	0.0297975	0.0299544	0.994759
> sparse_mla_fwd_pipelined	0.0583219	0.0586148	0.995003
> example_warp_specialize_gemm_copy_1_gemm_0	0.0176288	0.0177094	0.995448
> example_mha_fwd_varlen	0.0329942	0.0331447	0.995458
> example_gqa_sink_bwd_bhsd_sliding_window	0.0178338	0.0178992	0.996344
> example_warp_specialize_gemm_barrierpipe_stage2	0.0281582	0.0282548	0.99658
> example_warp_specialize_gemm_copy_0_gemm_1	0.0267214	0.026785	0.997624
> example_fusedmoe_tilelang	0.0967872	0.0969987	0.99782
> block_sparse_attn_tilelang	0.00697268	0.00698751	0.997877
> example_per_token_cast_to_fp8	0.00662436	0.00663558	0.998309
> example_warp_specialize_gemm_softpipe_stage2	0.0176443	0.0176708	0.998498
> topk_selector	0.0416608	0.0417234	0.998498
> example_tilelang_nsa_fwd	0.00546951	0.00547652	0.99872
> example_linear_attn_fwd	0.0280643	0.0280914	0.999034
> example_dequant_gemm_w4a8	3.56174	3.56416	0.99932
> example_mhc_post	0.106125	0.10619	0.999387
> example_group_per_split_token_cast_to_fp8	0.00770348	0.00770683	0.999566
> example_gemm	0.0167398	0.0167459	0.999637
> example_elementwise_add	0.112905	0.112945	0.999647
> example_tilelang_gemm_splitk_vectorize_atomicadd	0.788525	0.788798	0.999654
> example_dequant_gemm_bf16_fp4_hopper	0.390943	0.391054	0.999715
> example_tilelang_nsa_decode	0.00561493	0.00561624	0.999768
> example_mha_sink_fwd_bhsd	0.0127041	0.0127035	1.00005
> example_mha_inference	0.0619983	0.0619922	1.0001
> example_tilelang_block_sparse_attn	0.0073949	0.00739343	1.0002
> example_gemv	0.202586	0.202541	1.00022
> example_gqa_fwd_bshd	0.0497286	0.0497125	1.00032
> example_gemm_intrinsics	0.0229871	0.0229702	1.00074
> example_dequant_gemv_fp16xint4	0.0271149	0.0270935	1.00079
> example_mha_fwd_bshd	0.0189009	0.0188847	1.00086
> fp8_lighting_indexer	0.0233325	0.0233123	1.00087
> sparse_mla_fwd	0.0806824	0.0806118	1.00088
> example_tilelang_sparse_gqa_decode_varlen_mask	0.0128955	0.0128836	1.00092
> example_gqa_sink_bwd_bhsd	0.0289764	0.0289477	1.00099
> example_mha_fwd_bhsd	0.00909485	0.00908223	1.00139
> example_mha_sink_fwd_bhsd_sliding_window	0.0124697	0.0124522	1.0014
> example_mha_bwd_bshd	0.0290566	0.0290134	1.00149
> example_dynamic	0.483545	0.482529	1.00211
> example_gqa_decode	0.0411603	0.0410599	1.00245
> example_tilelang_sparse_gqa_decode_varlen_indice	0.0117414	0.0117107	1.00262
> example_mla_decode	0.320815	0.319869	1.00296
> example_convolution	0.764008	0.761738	1.00298
> example_tilelang_gemm_splitk	0.771553	0.769098	1.00319
> example_gqa_bwd	0.032652	0.0325309	1.00372
> example_blocksparse_gemm	0.0132627	0.0131798	1.00629
> example_tilelang_gemm_fp8_2xAcc	0.0770575	0.0765476	1.00666
> example_tilelang_gemm_fp8	0.232787	0.23123	1.00673
> example_mha_sink_bwd_bhsd	0.0507325	0.050223	1.01014
> sparse_mla_bwd	0.232304	0.22961	1.01173
> example_dequant_gemm_fp4_hopper	0.714361	0.705705	1.01226
> example_linear_attn_bwd	0.118427	0.116846	1.01354
> example_convolution_autotune	0.747126	0.73418	1.01763
> example_gqa_bwd_tma_reduce_varlen	0.033734	0.032957	1.02358
> ## Artifacts
> * regression_result.png (speedup plot) is attached as a workflow artifact. Download it from the workflow run page above.

The `example_topk` change in the report looks more like measurement noise than a code regression. This PR only modifies the leftover-iterator check in `get_unused_iters`, and the new branch (`frontier > extent`) only affects inputs that reach the over-coverage case from #2394. For kernels that already compiled fine, including `example_topk`, I would not expect this change to alter the generated code.

I ran a controlled local A/B test using the repository's perf regression entry point (`examples/topk/regression_topk_tilelang.py` -> `example_topk.run_regression_perf()`): same machine, both `main` and this PR rebuilt, GPU idle before each set, 10 runs each with `do_bench`.

**Test environment:** NVIDIA TITAN RTX (Turing, sm_75), driver 550.144.03, CUDA 12.4, PyTorch 2.6.0+cu124.

`main` (10 runs, ms):

```text
[105.4102, 105.0767, 99.6066, 98.5510, 100.3664,
 100.6388, 100.2405, 98.4862, 104.2080, 97.8045]
mean=101.0389  median=100.3034  std=2.6819  spread=7.8%
```

This PR (10 runs, ms):

```text
[105.1217, 98.3787, 96.5558, 96.9236, 106.1212,
 102.5140, 98.6105, 97.1820, 103.9076, 99.1520]
mean=100.4467  median=98.8812  std=3.4328  spread=9.9%
```

`main / PR = 101.0389 / 100.4467 = 1.006×`. The difference between the two means is much smaller than each set's standard deviation, so this looks consistent with normal run-to-run variation, and the ~30% delta from CI (30.04 → 39.06 ms) does not reproduce locally.

LeiWang1999:
@regression-perf

LeiWang1999 (approved):

PR #2470, opened 2026-06-28
[Feature] add lower-trace support for debugging

erhsh:
# IR Lower Trace — Feature Overview

> Corresponding commit: `66b1a7ad [feat] add lower-trace support for debugging`

## 1. One-line Summary

A **zero-intrusion compilation debugging tool**: once enabled, it automatically captures the IR before and after every pass (and the final codegen) in the compilation pipeline, renders a per-pass diff in the terminal and as a self-contained HTML report, and supports an "edit the generated codegen source → recompile with edits injected" workflow for verification.

## 2. Why It's Needed

TileLang programs are lowered through a sequence of compiler passes, each of which may rewrite the IR. When the final result is unexpected, it is hard to tell *which* pass changed the IR incorrectly, *which* backend stage a pass belongs to, or *what* the final generated C/CUDA/HIP source looks like — let alone quickly patch the generated kernel for experimentation without touching the TileLang source. This tool closes those observability gaps: **off → zero overhead; on → the whole pipeline is visible**.

## 3. Implementation

Located under [tilelang/tools/lower_trace/](tilelang/tools/lower_trace/), the tool consists of three parts: hook installation & record collection (core), unified diff computation (diff), and self-contained HTML rendering (html).

- **Three transparent hooks** (all monkey-patched, fully restorable): intercept `Pass.__call__` to capture before/after IR for each pass; intercept `PassPipeline.lower` (new arch) / phase functions (legacy arch, auto-discovered via AST scan + bytecode inspection) to tag each pass with its pipeline stage; intercept codegen FFIs to capture the final TIR → source step.
- **Runtime records**: passes are appended at runtime rather than pre-registered, so conditionally skipped passes leave no phantom slots.
- **Crash-safe**: the HTML report is flushed incrementally after every pass (with a section cache, O(n) total cost), so partial results survive even a killed process.
- **Multi-run accumulation**: repeated compilations in the same process are tagged with `run2_`/`run3_` prefixes and merged into a single report for side-by-side comparison.
- **Codegen edit-recompile**: three cooperating files (baseline / working copy / latest output) are compared each run to decide `PATCHED` / `REGENERATED` / `SYNCED` / `CONFLICT`, so user edits are never silently overwritten (requires a source-compiling backend: `nvrtc` / `cython` / `cutedsl`).
- **Compatibility**: works with both new and legacy TVM FFI, both `inspect_source`/`get_source`; path components are sanitized against directory escape; disk writes are best-effort (failures warn, never abort compilation).

## 4. How to Use

**Environment variable (simplest):**

```bash
TL_LOWER_TRACE=1 python3 my_script.py            # HTML report
TL_LOWER_TRACE=terminal python3 my_script.py     # terminal diff only
TL_LOWER_TRACE=both python3 my_script.py         # terminal + HTML
```

`TL_LOWER_TRACE_DIR` sets the artifact root (default `./tmp/lower_trace_dir`); the report is at `<dir>/<script>/report.html`.

**Programmatic API:**

```python
from tilelang.tools import lower_trace as lt

# Trace the entire pipeline (equivalent to the env var)
lt.enable(mode="both")
kernel = tilelang.compile(func)

# Or a one-shot diff over a fixed pass chain (no global hook installed)
lt.lower_trace(func, [pass_a, pass_b], mode="both", html_path="diff.html")
```

`enable()` is idempotent (optional `mode`/`trace_dir`/`codegen_output`); `reset()` clears records but keeps the hook (per-kernel reports in one process); `disable()` restores all hooks.

**edit-recompile flow:** run once with tracing on → edit the on-disk `codegen.cpp` → rerun to compile the edited source; if a TileLang source change regenerates codegen that conflicts with your edits, the working copy is auto-backed-up to `.bak` for recovery.

**HTML report features:** sidebar with per-pass navigation + status dots (changed / no-op / failed / codegen) + line stats, stage filtering, side-by-side diff (word-level highlighting, collapsible context), `j`/`k` keyboard navigation, dark/light theme, and failed-pass error boxes showing the exception alongside the pre-crash IR.

## 5. Impact on the Existing Framework (non-intrusive, easy to merge)

This tool is added in a **purely additive, non-intrusive** way — no existing framework behavior is changed, and merge risk is low:

- **Zero overhead when off:** when `TL_LOWER_TRACE` is unset, the loading logic in [tilelang/__init__.py](tilelang/__init__.py) is skipped entirely — no import, no patching, no runtime cost; [tilelang/tools/__init__.py](tilelang/tools/__init__.py) lazily loads the submodule via `__getattr__`, leaving the normal startup path untouched.
- **Observation-only when on, semantics unchanged:** every hook is a transparent wrapper that captures before/after IR and forwards the original call — pass behavior and return values are not altered. The only exception is source-only codegen FFIs, which return a proxy object to support edit-recompile; binary FFIs conservatively return the real module.
- **Fully restorable:** `disable()` restores `Pass.__call__`, `PassPipeline.lower`, and codegen FFIs to their originals and clears all state — no trace left.
- **Mostly new files:** changes are concentrated in the new `tilelang/tools/lower_trace/` module plus tests and docs; the only edits to existing source are an env-gated loading block and a lazy-load entry in `__init__.py` — no compilation-core logic is touched.
- **Tolerant persistence:** all disk writes are best-effort; failures only warn and never abort compilation.

In short, the feature provides full-pipeline observability without touching the framework's compilation core, is off by default with zero production impact, and is suitable for direct merge.

---

### Appendix: Artifact directory layout

```text
<TL_LOWER_TRACE_DIR>/
└── <script_name>/
    ├── report.html                       # symlink → latest run's report
    ├── codegen.cpp(.original/.latest)    # edit-recompile trio
    └── .run_records/run_<ts>_<pid>/
        ├── report.html
        ├── <phase>/00_<pass>_before.tir / _after.tir
        └── codegen/.._before.tir / .._after.cpp
```

### Summary
- Added a new `lower_trace` debugging workflow for TileLang lowering, with hook-based IR capture before/after passes, per-pass diffs, terminal output, and self-contained HTML reporting.
- Added support for codegen tracing and edit/recompile verification, including baseline/working/latest source tracking and conflict handling for user-edited generated files.
- Enabled import-time activation via `TL_LOWER_TRACE` and lazy module loading through `tilelang.tools`.
- Expanded docs to replace the older Pass Diff workflow with IR Lower Trace.
- Added comprehensive tests covering env/API behavior, HTML output, multi-run tracing, codegen patching, failure handling, and import-time activation.

### Testing
- Added `testing/python/debug/test_lower_trace.py` with broad regression coverage for the new tracing flow.


erhsh:
@LeiWang1999 please help review & merge, thx

docs/tutorials/debug_tools_for_tilelang.md
SiriusNEO, line 1: I think it's better to put this doc under `docs/tools`

SiriusNEO (changes requested):
Hi @erhsh , thanks for your contribution! This tool looks like an "advanced version" of Pass Diff. Could you rebase/solve the conflict first and we can move ahead with the review/merge

SiriusNEO:
Closed due to the new PR: https://github.com/tile-ai/tilelang/pull/2725

PR #2492, opened 2026-06-30
[BugFix][CUDA] Use PTX v4 atomics for fp16/bf16 atomic_addx4

### LINKED ISSUE #2478: [BUG][Fuzzer][wrong-code] `T.atomic_addx4` on a matched `float16`/`bfloat16` destination silently returns wrong numbers (and crashes at unaligned offsets) instead of doing a 4-element atomic add — no fp16/bf16 `AtomicAddx4` specialization
### Required prerequisites

0.1.11

### System information

- Python 3.13.13, linux; PyTorch 2.11.0+cu130.
- Reproduced on **sm_90 (H100), sm_89 (L40S), and sm_86 (A10G)** — the behavior differs by arch (matrix below). The missing-specialization root cause is in `atomic.h`, a header shared by all targets, but the generic `AtomicAddx4` template forks at `__CUDA_ARCH_LIST__ >= 900`, so the **face** of the defect is arch-dependent.

### Problem description

`T.atomic_addx4` on a matched 16-bit (`float16`/`bfloat16`) source and destination has **no fp16/bf16 specialization** in `atomic.h`, so it binds to the generic `float4`-hardcoded `AtomicAddx4` template (which first turns the value into a `float4` via `ToFloat4`). Because the template body is hardcoded to `float`, the same missing specialization shows **three different faces depending on the target architecture**:

- **sm_90 (H100):** binds to the `atomicAdd(reinterpret_cast<float4*>(ref), …)` branch, which accepts the `float4`-typed value. It compiles and runs with no error and **silently returns wrong numbers** (the 8-byte 4×16-bit destination is reinterpreted as one 16-byte `float4`). The `float4` `atomicAdd` also requires a 16-byte-aligned destination, so at a misaligned 16-bit offset (e.g. element 4) the launch instead **aborts with `cudaErrorMisalignedAddress`** — same root cause, alignment picks the face.
- **pre-sm_90 (sm_89, sm_86):** takes the fallback branch, which calls the same-type-only `AtomicAddx4Scalar(T* ref, T x, …)` with the already-`float`-ized value — i.e. `AtomicAddx4Scalar(half_t*, float, …)`, whose template deduction fails. So fp16/bf16 `atomic_addx4` is a **compile error**, not a silent miscompile, on these arches.

`float32` is correct on every arch and offset (4 elements genuinely are a `float4`). `T.atomic_addx4(B[0], A[0])` with `A`,`B` both `float16` emits, for every dtype alike, `AtomicAddx4((&(B[0])), (&(A[0])));` — no 16-bit specialization is selected.

| arch (tested) | `float32` | fp16 / bf16 `atomic_addx4` |
|---|---|---|
| sm_90 — H100, 0.1.11 | correct | **silent miscompile** (aligned) / **`cudaErrorMisalignedAddress` crash** (misaligned) |
| sm_89 — L40S, 0.1.11 | correct | **compile error** (`AtomicAddx4Scalar` deduction fails, `atomic.h:680`→`:225`) |
| sm_86 — A10G, 0.1.9 | correct | **compile error** (same) |

All three rows were run on that silicon this session; no cell is inferred.

<details>
<summary>Observed vs expected — silent miscompile at aligned offset (fresh H100 run)</summary>

`A=[1,2,3,4]`, destination pre-filled `B0=[10,20,30,40]` (so a dropped/garbage add is visible; an all-zero destination would hide the fp16 "dropped add"). Correct result = `B0 + threads*A`.

```
=== dtype=float32 threads=1 ===   got: [11,22,33,44]   ref: [11,22,33,44]   MATCH
=== dtype=float32 threads=2 ===   got: [12,24,36,48]   ref: [12,24,36,48]   MATCH
=== dtype=float16 threads=1 ===   got: [10,20,30,40]   ref: [11,22,33,44]   *** MISMATCH (add silently DROPPED)
=== dtype=float16 threads=2 ===   got: [10,20,30,40]   ref: [12,24,36,48]   *** MISMATCH
=== dtype=bfloat16 threads=1 ===  got: [589824,22,2031616,44]            ref: [11,22,33,44]   *** MISMATCH (garbage)
=== dtype=bfloat16 threads=2 ===  got: [3.4e10,24,1.37e11,48]            ref: [12,24,36,48]   *** MISMATCH
```

The corruption flips on **element width**, not rounding: the fp16 inputs (`1,2,3,4`) are exactly representable, yet the add is dropped — two fp16 lanes are reinterpreted as one `float32`, so the destination `[10,20]` reads as `≈1.3e8` and the addend `[1,2]` as `≈2.0`; `1.3e8 + 2.0` rounds back to `1.3e8` (the addend is below the float32 ULP at that magnitude), so the bits are unchanged. The bf16 lanes reinterpret to differently-scaled floats → garbage.
</details>

<details>
<summary>Second face of the same defect (sm_90) — <code>cudaErrorMisalignedAddress</code> crash at a misaligned offset</summary>

Both the silent face above and this crash face are **sm_90** (the `atomicAdd(float4*)` branch); pre-sm_90 never reaches either because it fails to compile. A `float4` `atomicAdd` needs a 16-byte-aligned destination; a 16-bit element at byte offset `2*k` is 16-byte-aligned only when `k % 8 == 0`. So element offset `0,8,16,…` takes the silent path; any other (e.g. element 4 → byte 8) **aborts** with `cudaErrorMisalignedAddress`. `float32` never crashes (4 floats span 16 bytes from any aligned base).

```
[float32  off=0/4/8] OK
[float16  off=0]  OK   (silent path — aligned)
[float16  off=8]  OK   (silent path — aligned)
[float16  off=4]  *** EXCEPTION  CUDA error: misaligned address
[bfloat16 off=4]  *** EXCEPTION  CUDA error: misaligned address
```

Note: a CUDA misaligned-address error is *sticky* — once it fires every later launch in the same process re-reports it, so each case above was run in its own process.
</details>

<details>
<summary>Control — matched-dtype <code>x2</code> and scalar are correct; only <code>x4</code> is broken</summary>

Same matched fp16/bf16, same pre-filled destination:

| call | fp16 | bf16 | path in `atomic.h` |
|------|------|------|--------------------|
| `T.atomic_addx4` (this bug) | **wrong** (dropped) | **wrong** (garbage) | generic `float4` template — no 16-bit specialization |
| `T.atomic_addx2` | MATCH | MATCH | real `AtomicAddx2(half_t*)` / `(bfloat16_t*)` (`half2`/`__nv_bfloat162` atomicAdd) |
| scalar `T.atomic_add` | MATCH | MATCH | per-element `atomicAdd` |

```
x2     float16: got=[11,22,30,40]  ref=[11,22,30,40]  MATCH
scalar float16: got=[11,22,33,44]  ref=[11,22,33,44]  MATCH
x2     bfloat16:got=[11,22,30,40]  ref=[11,22,30,40]  MATCH
scalar bfloat16:got=[11,22,33,44]  ref=[11,22,33,44]  MATCH
```

The kernel is well-formed and fp16/bf16 atomics work in general — only the packed-`x4` path is wrong. This rules out "fp16/bf16 atomics are unsupported" and isolates the defect to the missing `x4` 16-bit specialization.
</details>

<details>
<summary>Root cause — the generic <code>float4</code> template and the arch fork</summary>

`atomic.h` has no fp16/bf16 `AtomicAddx4` overload, so the frontend's `AtomicAddx4(half_t*/bfloat16_t* ref, …)` binds to the generic `float4`-hardcoded template at [`atomic.h#L623-L639`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L623-L639):

```cpp
float4 add_val = ToFloat4(val);                        // L626: *reinterpret_cast<const float4*>(val) — reads 16B from an 8B source
atomicAdd(reinterpret_cast<float4 *>(ref), add_val);   // L628: 8B/4-elem 16-bit dst treated as one 16B float4; needs 16B alignment
```

`ToFloat4` is the reinterpret at [`#L586`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L586); the `float4` `atomicAdd` at [`#L628`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L628) is both the wrong-numbers site (aligned) and the 16-byte-alignment crash site (misaligned). This is compounded by the frontend: `atomic.py::atomic_addx4` computes `return_type = "float4" if "float" in str(dst.dtype).lower() else "handle"` ([`atomic.py#L353`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/atomic.py#L353)), which classifies `"float16"`/`"bfloat16"` as the `float4` path with no legality guard. By contrast the `x4` is correct for `float32` (a real `float4`), and `x2` is correct for fp16/bf16 because it *does* ship 16-bit specializations ([`#L482`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L482), [`#L538`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L538)).

The generic template forks on arch: the `atomicAdd(float4*)` form above is the `__CUDA_ARCH_LIST__ >= 900` branch. Pre-sm_90 takes the fallback at [`#L676-L682`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L676-L682), which routes the `float`-typed value into `AtomicAddx4Scalar(T* ref, T x, …)` at [`#L225`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L225) — calling it `AtomicAddx4Scalar(half_t*, float, …)`, whose single-type-`T` deduction fails. So on pre-sm_90 the same missing specialization surfaces as a **compile error** rather than a silent miscompile (verified on sm_89/sm_86 — matrix above).
</details>

### Reproducible example code

Use a real `.py` file — TileLang's eager front end calls `inspect.getsource`, so `python -c` trips an unrelated `OSError`. Run on an sm_90 GPU.

```python
import tilelang
tilelang.disable_cache()
import tilelang.language as T
import torch

def make(dtype, nthreads):
    @T.prim_func
    def main(A: T.Tensor((8,), dtype), B: T.Tensor((8,), dtype)):
        with T.Kernel(1, threads=nthreads) as _:
            T.atomic_addx4(B[0], A[0])
    return main

def run(dtype, nthreads):
    kernel = tilelang.compile(make(dtype, nthreads), target="cuda")
    tdt = {"float16": torch.float16, "bfloat16": torch.bfloat16, "float32": torch.float32}[dtype]
    A  = torch.tensor([1, 2, 3, 4, 0, 0, 0, 0], dtype=tdt, device="cuda")
    B0 = torch.tensor([10, 20, 30, 40, 0, 0, 0, 0], dtype=tdt, device="cuda")  # pre-filled so a dropped add shows
    B = B0.clone()
    kernel(A, B); torch.cuda.synchronize()
    ref = B0.float().clone(); ref[:4] += nthreads * A[:4].float()   # exact: small ints
    got = B.float()
    print(f"{dtype} threads={nthreads}: got={got.tolist()[:4]} ref={ref.tolist()[:4]} "
          f"{'MATCH' if torch.equal(got, ref) else '*** MISMATCH ***'}")

for dt in ["float32", "float16", "bfloat16"]:   # float32 MATCH; float16/bfloat16 MISMATCH
    for nt in [1, 2]:
        run(dt, nt)
```

<details>
<summary>Control (correct) and the misaligned-crash repro</summary>

Same matched fp16/bf16 through `x2` / scalar is correct, isolating the fault to `x4`:

```python
def make_x2(dtype):
    @T.prim_func
    def main(A: T.Tensor((8,), dtype), B: T.Tensor((8,), dtype)):
        with T.Kernel(1, threads=1) as _:
            T.atomic_addx2(B[0], A[0])          # real half2/bf162 specialization -> correct
    return main

def make_scalar(dtype):
    @T.prim_func
    def main(A: T.Tensor((8,), dtype), B: T.Tensor((8,), dtype)):
        with T.Kernel(1, threads=1) as _:
            for i in T.serial(4):
                T.atomic_add(B[i], A[i])        # per-element -> correct
    return main
```

Misaligned crash — run **one case per process** (the CUDA error is sticky):

```python
@T.prim_func
def main(A: T.Tensor((16,), "float16"), B: T.Tensor((16,), "float16")):
    with T.Kernel(1, threads=1) as _:
        T.atomic_addx4(B[4], A[4])              # byte offset 8 — NOT 16-byte aligned
kernel = tilelang.compile(main, target="cuda")
A = torch.ones(16, dtype=torch.float16, device="cuda")
B = torch.zeros(16, dtype=torch.float16, device="cuda")
kernel(A, B); torch.cuda.synchronize()          # -> AcceleratorError: CUDA error: misaligned address
```
</details>

### Traceback

```pytb
At an aligned offset there is **no traceback** — the kernel runs to completion and the result is silently wrong. At a misaligned 16-bit offset the launch aborts:


AcceleratorError: CUDA error: misaligned address
```

### Expected behavior

`T.atomic_addx4` on a matched `float16`/`bfloat16` source and destination should perform a correct 4-element 16-bit atomic add (`B[k] += A[k]` for `k in 0..3`), **or** the frontend should reject the unsupported dtype at compile time — not silently reinterpret the lanes as a `float4` (aligned), nor abort with `cudaErrorMisalignedAddress` (misaligned).

This is grounded in TileLang's own behavior: the project already ships genuine 16-bit specializations for the `x2` packed atomic (`AtomicAddx2(half_t*)` [`#L482`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L482), `AtomicAddx2(bfloat16_t*)` [`#L538`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L538)), and the control shows those are bit-correct on the same matched dtypes. `x4` is the wider sibling of `x2` and already works for `float32`; the 16-bit case should be symmetric.

### Additional context

**Suggested fix (proposed, not built).** Mirror the `x2` design. Preferably both:

1. **Add genuine fp16/bf16 `AtomicAddx4` specializations** in `atomic.h`, symmetric with `AtomicAddx2(half_t*)`/`(bfloat16_t*)`: a 4-element 16-bit add maps to two `half2`/`__nv_bfloat162` `atomicAdd`s (`ref+0`, `ref+2`). A `half2`/`__nv_bfloat162` `atomicAdd` needs only **4-byte** alignment, which an fp16/bf16 element offset always satisfies — so this one change fixes **both** faces: correct accumulation at aligned offsets *and* no 16-byte requirement, so no `cudaErrorMisalignedAddress` crash.
2. **Add a frontend legality guard** in `atomic.py::atomic_addx4` ([`#L353`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/atomic.py#L353)) so a non-32-bit element dtype is rejected (or scalarized) rather than classified as the `float4` path.

A regression test should compare **numerics** (not just assert the string `AtomicAddx4` is emitted) with a **pre-filled** destination. I have not built or verified either change.

<details>
<summary>Dedup — distinct from previously reported issues</summary>

I scanned all open and closed `tile-ai/tilelang` issues; none concerns the packed-`x4` 16-bit template gap.

- **#2382** (our prior report) is a **scalar** `T.atomic_add` with a **dtype mismatch** (fp32 src → fp16/bf16 dst) that the auto-vectorizer promotes to `x2` with no `src==dst` guard. This report is different: an **explicit** `T.atomic_addx4` with **matched** dtypes (no vectorizer, no mismatch), root cause = the missing `AtomicAddx4` fp16/bf16 specialization. Proof: the matched-dtype `x2` path is *correct* here (it has a real specialization), while `x4` is wrong — so #2382's vectorizer `src==dst` guard would not touch this path.

The `cudaErrorMisalignedAddress` crash is **not** a separate bug — it is the alignment-dependent face of the same missing specialization, closed by the same fix, so it is folded into this report.
</details>

JayceSu98:
## Summary

Fixes #2478.

This PR fixes CUDA `T.atomic_addx4` for `float16` and `bfloat16` destinations. The original failing path could select the generic `float4` atomic implementation for four 16-bit lanes, which is both the wrong element width and too strict on alignment for offset-4 half/bfloat16 destinations.

After the latest review, the Hopper path now uses native PTX vector atomics instead of simulating x4 with two x2 atomics:

- `atom.global.v4.f16.add.noftz` for fp16
- `atom.global.v4.bf16.add.noftz` for bf16

The PTX form follows NVIDIA PTX `atom` documentation: vector atomics are global-memory operations, 16-bit floating vector atomics require `.noftz`, and vector atomics require `sm_90+`. Pre-sm90 targets keep the existing two-`AtomicAddx2` fallback for correctness because native vector atomics are not available there.

## Changes

- Add sm90+ CUDA `AtomicAddx4(half_t*, ...)` and `AtomicAddx4(bfloat16_t*, ...)` overloads backed by native PTX v4 atomics.
- Support the existing memory-order variants for the new v4 helpers: relaxed, release-like, acquire, and acq_rel-like.
- Keep the existing fp32 `AtomicAddx4` path unchanged.
- Add a regression test covering `T.atomic_addx4` on fp16 and bf16 at offsets 0 and 4.

## Validation

- `pre-commit run --files src/tl_templates/cuda/atomic.h testing/python/language/test_tilelang_language_atomic.py`
- H100 / sm90, cache disabled:
  - `CUDA_VISIBLE_DEVICES=0 TILELANG_DISABLE_CACHE=1 TILELANG_CACHE_DIR=/tmp/tilelang-pr2492-cache-h100-docsyntax PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -B -m pytest testing/python/language/test_tilelang_language_atomic.py::test_atomic_addx4_16bit -q -s`
  - Result: `1 passed`
- A100 / sm80, cache disabled, fallback path:
  - `CUDA_VISIBLE_DEVICES=1 TILELANG_DISABLE_CACHE=1 TILELANG_CACHE_DIR=/tmp/tilelang-pr2492-cache-a100-docsyntax PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -B -m pytest testing/python/language/test_tilelang_language_atomic.py::test_atomic_addx4_16bit -q -s`
  - Result: `1 passed`

## Performance

Microbenchmark on H100 PCIe / sm90 with TileLang cache disabled, `N=65536`, `threads=128`, CUDA event median timing, comparing native v4 `T.atomic_addx4` against an equivalent two-`T.atomic_addx2` split implementation:

| dtype | native v4 | split 2x addx2 | speedup |
| --- | ---: | ---: | ---: |
| fp16 | 0.075136 ms | 0.184832 ms | 2.460x |
| bf16 | 0.076576 ms | 0.186816 ms | 2.440x |

## Summary
- Added `AtomicAddx4` support for `half_t` and `bfloat16_t` CUDA destinations in `src/tl_templates/cuda/atomic.h`.
- Implemented native SM90+ 4-wide PTX atomic add paths for fp16/bf16, with a pre-SM90 fallback that composes two existing `AtomicAddx2` calls.
- Kept the existing float32 `float4` path unchanged.
- Expanded CUDA regression coverage to verify `T.atomic_addx4` on prefilled fp16/bf16 tensors at offsets `0` and `4`, alongside the existing float32 coverage.

## Testing
- Added a new CUDA/compute-version-gated test for fp16 and bf16 `atomic_addx4`.
- Updated the existing `atomic_addx4` test to run directly instead of remaining TODO-only.

## C++ style / lint notes
- This PR touches C++-adjacent CUDA template code in `docs/developer_guide/cpp_style.md` scope.
- No clear build or correctness issues are introduced from a style/lint perspective.
- The added low-level atomic specializations may surface warning-only items in the “C++ API Style Audit (warning only)” step, but nothing here appears merge-blocking.

### DIFF
--- a/src/tl_templates/cuda/atomic.h
+++ b/src/tl_templates/cuda/atomic.h
@@ -168,6 +168,72 @@ TL_DEVICE void tl_atomic_add_v2_bf16(unsigned short &ret_x,
   }
 }
 
+TL_DEVICE void
+tl_atomic_add_v4_f16(unsigned short &ret_x, unsigned short &ret_y,
+                     unsigned short &ret_z, unsigned short &ret_w,
+                     unsigned long long addr, unsigned short val_x,
+                     unsigned short val_y, unsigned short val_z,
+                     unsigned short val_w, int memory_order) {
+  if (IsRelaxedMemoryOrder(memory_order)) {
+    asm volatile(
+        "atom.global.v4.f16.add.noftz {%0,%1,%2,%3}, [%4], {%5,%6,%7,%8};"
+        : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+        : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+        : "memory");
+  } else if (IsReleaseLikeMemoryOrder(memory_order)) {
+    asm volatile("atom.release.gpu.global.v4.f16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  } else if (IsAcquireMemoryOrder(memory_order)) {
+    asm volatile("atom.acquire.gpu.global.v4.f16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  } else if (IsAcqRelLikeMemoryOrder(memory_order)) {
+    asm volatile("atom.acq_rel.gpu.global.v4.f16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  }
+}
+
+TL_DEVICE void
+tl_atomic_add_v4_bf16(unsigned short &ret_x, unsigned short &ret_y,
+                      unsigned short &ret_z, unsigned short &ret_w,
+                      unsigned long long addr, unsigned short val_x,
+                      unsigned short val_y, unsigned short val_z,
+                      unsigned short val_w, int memory_order) {
+  if (IsRelaxedMemoryOrder(memory_order)) {
+    asm volatile(
+        "atom.global.v4.bf16.add.noftz {%0,%1,%2,%3}, [%4], {%5,%6,%7,%8};"
+        : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+        : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+        : "memory");
+  } else if (IsReleaseLikeMemoryOrder(memory_order)) {
+    asm volatile("atom.release.gpu.global.v4.bf16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  } else if (IsAcquireMemoryOrder(memory_order)) {
+    asm volatile("atom.acquire.gpu.global.v4.bf16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  } else if (IsAcqRelLikeMemoryOrder(memory_order)) {
+    asm volatile("atom.acq_rel.gpu.global.v4.bf16.add.noftz {%0,%1,%2,%3}, "
+                 "[%4], {%5,%6,%7,%8};"
+                 : "=h"(ret_x), "=h"(ret_y), "=h"(ret_z), "=h"(ret_w)
+                 : "l"(addr), "h"(val_x), "h"(val_y), "h"(val_z), "h"(val_w)
+                 : "memory");
+  }
+}
+
 TL_DEVICE void tl_atomic_add_v2_f32(float &ret_x, float &ret_y,
                                     unsigned long long addr, float val_x,
                                     float val_y, int memory_order) {
@@ -595,6 +661,66 @@ AtomicAddx2Ret(bfloat16_t *ref, src_type *val,
         tl_atomic_detail::UnpackBits16<__nv_bfloat16>(ret_val_y_cast));
   }
 }
+#endif
+
+#if (defined(__CUDA_ARCH_LIST__) && (__CUDA_ARCH_LIST__ >= 900))
+template <typename SrcType>
+TL_DEVICE void AtomicAddx4(half_t *ref, SrcType *val,
+                           int memory_order = int(cuda::memory_order_relaxed)) {
+  half2 add_val_lo = ToHalf2(val);
+  half2 add_val_hi = ToHalf2(val + 2);
+  unsigned short add_val_x_cast = tl_atomic_detail::PackBits16(add_val_lo.x);
+  unsigned short add_val_y_cast = tl_atomic_detail::PackBits16(add_val_lo.y);
+  unsigned short add_val_z_cast = tl_atomic_detail::PackBits16(add_val_hi.x);
+  unsigned short add_val_w_cast = tl_atomic_detail::PackBits16(add_val_hi.y);
+  unsigned long long ref_addr = reinterpret_cast<unsigned long long>(ref);
+  unsigned short ret_val_x_cast;
+  unsigned short ret_val_y_cast;
+  unsigned short ret_val_z_cast;
+  unsigned short ret_val_w_cast;
+  tl_atomic_detail::tl_atomic_add_v4_f16(
+      ret_val_x_cast, ret_val_y_cast, ret_val_z_cast, ret_val_w_cast, ref_addr,
+      add_val_x_cast, add_val_y_cast, add_val_z_cast, add_val_w_cast,
+      memory_order);
+}
+#else
+template <typename SrcType>
+TL_DEVICE void AtomicAddx4(half_t *ref, SrcType *val,
+                           int memory_order = int(cuda::memory_order_relaxed)) {
+  AtomicAddx2(ref, val, memory_order);
+  AtomicAddx2(ref + 2, val + 2, memory_order);
+}
+#endif
+
+#if (defined(__CUDA_ARCH_LIST__) && (__CUDA_ARCH_LIST__ > 750))
+#if (defined(__CUDA_ARCH_LIST__) && (__CUDA_ARCH_LIST__ >= 900))
+template <typename SrcType>
+TL_DEVICE void AtomicAddx4(bfloat16_t *ref, SrcType *val,
+                           int memory_order = int(cuda::memory_order_relaxed)) {
+  __nv_bfloat162 add_val_lo = ToBfloat162(val);
+  __nv_bfloat162 add_val_hi = ToBfloat162(val + 2);
+  unsigned short add_val_x_cast = tl_atomic_detail::PackBits16(add_val_lo.x);
+  unsigned short add_val_y_cast = tl_atomic_detail::PackBits16(add_val_lo.y);
+  unsigned short add_val_z_cast = tl_atomic_detail::PackBits16(add_val_hi.x);
+  unsigned short add_val_w_cast = tl_atomic_detail::PackBits16(add_val_hi.y);
+  unsigned long long ref_addr = reinterpret_cast<unsigned long long>(ref);
+  unsigned short ret_val_x_cast;
+  unsigned short ret_val_y_cast;
+  unsigned short ret_val_z_cast;
+  unsigned short ret_val_w_cast;
+  tl_atomic_detail::tl_atomic_add_v4_bf16(
+      ret_val_x_cast, ret_val_y_cast, ret_val_z_cast, ret_val_w_cast, ref_addr,
+      add_val_x_cast, add_val_y_cast, add_val_z_cast, add_val_w_cast,
+      memory_order);
+}
+#else
+template <typename SrcType>
+TL_DEVICE void AtomicAddx4(bfloat16_t *ref, SrcType *val,
+                           int memory_order = int(cuda::memory_order_relaxed)) {
+  AtomicAddx2(ref, val, memory_order);
+  AtomicAddx2(ref + 2, val + 2, memory_order);
+}
+#endif
 #endif
 
 template <typename T> TL_DEVICE float2 ToFloat2(T *val) {
--- a/testing/python/language/test_tilelang_language_atomic.py
+++ b/testing/python/language/test_tilelang_language_atomic.py
@@ -205,6 +205,35 @@ def run_atomic_addx4(M, N, block_M, block_N):
 
 
 @tilelang.jit
+def atomic_addx4_16bit_program(dtype, offset, nthreads):
+    @T.prim_func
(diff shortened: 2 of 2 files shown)


LeiWang1999 (changes requested):
Please use Atomic v4 instead of the simulated implementation, as the latter can lead to poor performance.
checkout https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-atom

LeiWang1999 (changes requested):
Please use Atomic v4 instead of the simulated implementation, as the latter can lead to poor performance.
checkout https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-atom

JayceSu98:
Addressed the latest review.

The sm90+ fp16/bf16 `AtomicAddx4` path now uses native PTX vector atomics with the documented vector form:

- `atom.global.v4.f16.add.noftz`
- `atom.global.v4.bf16.add.noftz`

I kept the existing two-`AtomicAddx2` fallback only for pre-sm90 targets, since PTX vector atomics require sm90+. On H100, the native v4 path benchmarks about 2.46x faster for fp16 and 2.44x faster for bf16 than the split two-x2 implementation in the included microbenchmark notes.

LeiWang1999 (approved):

PR #2493, opened 2026-06-30
[BugFix][CUDA] Fix signed int2 LOP3 decode

### LINKED ISSUE #2479: [BUG][Fuzzer][wrong-code] `decode_i2s_to_i4s` silently decodes signed int2 as unsigned (sign-correction branch is an empty `if constexpr`) — every negative int2 loses its sign
### Required prerequisites

0.1.11

### System information

- Python 3.13, linux
- GPU: **NVIDIA L40S** (compute capability 8.9, sm_89). The defect is in a device-side decode C-string selected by `get_lop3_intrin_group`; it is host/arch-independent (the `lop3.b32` it uses is available sm_75+).

### Problem description

`get_lop3_intrin_group(out_dtype="int4", source_format="int", source_bit=2)` returns the fast-decode function `decode_i2s_to_i4s`, which is supposed to decode a **signed** 2-bit value into int4 (sign-extending / bias-correcting it). It does **not**: the signed-specific correction is written but **entirely commented out**, so the function emits the raw (unsigned) 2-bit pattern. Every negative int2 silently decodes to the wrong (non-negative) value, with no error.

The signed decoder produces **byte-identical** output to the unsigned decoder `decode_i2u_to_i4s` — which is the whole bug: a signed decode that equals the unsigned decode has dropped the sign of every value.

The `TODO` marks the *implementation* as unfinished, but the *entry point* treats it as a fully supported config: [`get_lop3_intrin_group`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/lop3.py#L1091) accepts `(out_dtype="int4", source_format="int", source_bit=2)` with **no error or `NotImplementedError`** — it raises only for an invalid `source_format`, an invalid `out_dtype`, or signed-with-zeros, none of which this hits — and returns `decode_i2s_to_i4s` like any other decoder. So the unfinished signed path ships as a *silently wrong* supported configuration rather than a rejected one; that gap between "marked incomplete in the body" and "accepted as complete at the boundary" is what makes it a silent miscompile rather than a harmless `TODO`.

<details>
<summary>Root cause — the <code>isSigned</code> correction is an empty block</summary>

`decode_i2s_to_i4s` dispatches to `decode_i2b_to_i4s<…, isSigned=true>` ([`lop3.py#L1077-L1081`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/lop3.py#L1077-L1081)). In that template the `isSigned` branch — the only thing that distinguishes signed from unsigned — is an empty block with the correction commented out ([`lop3.py#L1068-L1074`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/lop3.py#L1068-L1074)):

```cpp
asm volatile("lop3.b32 ...");          // extract raw 2-bit field into the i4 nibble (zero-extend)
if constexpr (isSigned)
{
    // TODO(lei): uint4 sub should be enhanced.
    // 0x03 0x03 0x03 0x03
    // i4s[i] = (((i4s[i] << 1) | i4s[i]) << 1) | i4s[i];
}
```

`MEDIAN_NUM` is computed as `isSigned ? 0x33333333 : 0` ([`lop3.py#L1059`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/lop3.py#L1059)) but is then never used. So `isSigned=true` and `isSigned=false` produce identical code. The author's `TODO(lei)` shows the signed path was knowingly left unfinished, but it ships and is selected for signed int2 input rather than erroring.

Differential control — the same packed int2 input decoded by the signed (`decode_i2s_to_i4s`) and the unsigned (`decode_i2u_to_i4s`) function produces **byte-identical output for every input** (no lane-layout assumptions — just signed-output vs unsigned-output). Expected: the signed decode should sign-extend each 2-bit value (e.g. `-2` → int4 `-2`, `-1` → int4 `-1`), differing from the unsigned decode for every negative input.
</details>

### Reproducible example code

Note: use a real `.py` file (eager JIT calls `inspect.getsource`; `python -c` trips an unrelated `OSError`). `tilelang.disable_cache()`.

```python
tilelang.disable_cache()
from tilelang.quantize import get_lop3_intrin_group

N = 16
def build(source_format):
    info = get_lop3_intrin_group(out_dtype="int4", source_format=source_format, source_bit=2, storage_dtype="int8")
    csrc, fname = info["c_source"], info["func_name"]
    @T.prim_func
    def main(I: T.Tensor((N // 4,), "int8"), O: T.Tensor((N // 2,), "int8")):
        with T.Kernel(1, threads=1) as _:
            T.import_source(csrc)
            i2 = T.alloc_local((N // 4,), "int8"); o4 = T.alloc_local((N // 2,), "int8")
            for k in T.serial(N // 4): i2[k] = I[k]
            T.call_extern("handle", fname, T.address_of(i2[0]), T.address_of(o4[0]), N)
            for k in T.serial(N // 2): O[k] = o4[k]
    return tilelang.compile(main, target="cuda")

ks, ku = build("int"), build("uint")          # signed vs unsigned decoder
I = torch.randint(-128, 127, (N // 4,), dtype=torch.int8, device="cuda")
Os = torch.zeros(N // 2, dtype=torch.int8, device="cuda"); Ou = torch.zeros_like(Os)
ks(I, Os); ku(I, Ou); torch.cuda.synchronize()
print(Os.cpu().tolist()); print(Ou.cpu().tolist())
assert not torch.equal(Os, Ou), "signed decode == unsigned decode: sign correction is dead"  # FAILS
```

The assert fails: signed and unsigned outputs are identical.

### Traceback

```pytb
silent
```

### Expected behavior

`decode_i2s_to_i4s` should sign-extend each signed 2-bit value into int4 (so its output differs from the unsigned decoder for negative inputs), OR — until the signed path is implemented — `get_lop3_intrin_group(out_dtype="int4", source_format="int", source_bit=2)` should **raise** rather than silently return a decoder that ignores signedness. The sibling signed paths that ARE implemented (`decode_i2s_to_i8s` subtracts a median; `decode_i2s_to_f16` uses a signed magic) set the precedent that the signed int2 path is meant to apply a correction.

### Additional context

**Suggested fix (proposed, not built).** Implement the commented-out signed correction in `decode_i2b_to_i4s`'s `isSigned` branch (the `MEDIAN_NUM = 0x33333333` is already computed for this purpose), so signed int2 is mapped into the signed int4 range; or, conservatively, make `get_lop3_intrin_group` reject `(out_dtype=int4, source_format=int, source_bit=2)` with a clear "signed int2→int4 fast-decode not implemented" error until it is. Either removes the silent wrong result.

JayceSu98:
Fixes #2479

## Summary
- Implement the signed correction in `decode_i2b_to_i4s<..., isSigned=true>`.
- Sign-extend every extracted raw int2 lane inside its destination int4 nibble by propagating bit 1 into bits 2 and 3.
- Keep the unsigned `decode_i2u_to_i4s` path unchanged.
- Add CUDA regression coverage for signed int2 to int4 LOP3 decode using a payload that contains both negative raw int2 patterns.

## Problem
`get_lop3_intrin_group(out_dtype="int4", source_format="int", source_bit=2)` returns `decode_i2s_to_i4s`, so the signed int2 to int4 path is exposed as a supported fast decoder.

Inside the shared `decode_i2b_to_i4s` template, however, the `isSigned` branch was effectively empty. As a result, the signed decoder emitted the same raw unsigned int4 nibbles as `decode_i2u_to_i4s`. Raw int2 patterns with sign bit set, such as `0b10` and `0b11`, therefore decoded as non-negative values instead of signed int4 values.

## Fix
After the LOP3 extraction places each raw 2-bit lane in the low bits of an int4 nibble, the signed path now does nibble-local sign extension:

- mask the int2 sign bit in each nibble with `0x22222222`;
- copy that bit into the nibble bit2 and bit3 positions;
- leave the unsigned path byte-for-byte unchanged.

This matches the scalar int-to-int conversion semantics for signed packed integers while preserving the existing LOP3 layout.

## Regression Test
The new test compiles both signed and unsigned int2 to int4 LOP3 decoders. It decodes the same packed int2 payload through both paths, derives the expected signed int4 values from the unsigned raw lanes using two-bit sign extension, and verifies:

- signed output matches the sign-extended int4 expectation;
- signed output is no longer byte-identical to unsigned output;
- the payload covers raw `0b10` and `0b11`, so both negative int2 patterns are exercised.

## Validation
- `python -m ruff check tilelang/quantize/lop3.py testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py`
- `python -m ruff format --check tilelang/quantize/lop3.py testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py`
- `git diff --check`
- `pre-commit run --files tilelang/quantize/lop3.py testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py`
- Remote CUDA: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py -q -s` passed

### DIFF
--- a/testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py
+++ b/testing/python/issue/test_tilelang_lop3_i2s_to_i4s_decode.py
@@ -1,0 +1,84 @@
+"""Regression coverage for signed int2 to int4 LOP3 decode."""
+
+import torch
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+from tilelang.quantize import get_lop3_intrin_group
+
+
+N = 16
+I2_STORAGE_BYTES = N // 4
+I4_STORAGE_BYTES = N // 2
+
+
+def _build_lop3_i2_to_i4_decode(source_format):
+    lop3 = get_lop3_intrin_group(
+        out_dtype=T.int4,
+        source_format=source_format,
+        source_bit=2,
+        storage_dtype=T.int8,
+        with_scaling=False,
+        with_zeros=False,
+    )
+    source = lop3["c_source"]
+    func_name = lop3["func_name"]
+
+    @T.prim_func
+    def main(I: T.Tensor((I2_STORAGE_BYTES,), T.int8), O: T.Tensor((I4_STORAGE_BYTES,), T.int8)):
+        with T.Kernel(1, threads=1):
+            T.import_source(source)
+            i2 = T.alloc_local((I2_STORAGE_BYTES,), T.int8)
+            i4 = T.alloc_local((I4_STORAGE_BYTES,), T.int8)
+            for k in T.serial(I2_STORAGE_BYTES):
+                i2[k] = I[k]
+            T.evaluate(T.call_extern("handle", func_name, T.address_of(i2[0]), T.address_of(i4[0]), N))
+            for k in T.serial(I4_STORAGE_BYTES):
+                O[k] = i4[k]
+
+    return main
+
+
+def _pack_i2_values(values):
+    values = torch.tensor(values, dtype=torch.uint8).reshape(-1, 4)
+    packed = values[:, 0] | (values[:, 1] << 2) | (values[:, 2] << 4) | (values[:, 3] << 6)
+    return packed.contiguous().view(torch.int8).cuda()
+
+
+def _unpack_i4_values(packed, signed):
+    packed = packed.to(torch.uint8)
+    low = (packed & 0x0F).to(torch.int8)
+    high = (packed >> 4).to(torch.int8)
+    if signed:
+        low = (low << 4) >> 4
+        high = (high << 4) >> 4
+    return torch.stack((low, high), dim=-1).reshape(-1)
+
+
+@tilelang.testing.requires_cuda
+@tilelang.testing.requires_cuda_compute_version_ge(7, 5)
+def test_lop3_i2s_to_i4s_sign_extends_int2():
+    qweight = _pack_i2_values([0, 1, 2, 3] * 4)
+
+    signed_kernel = tilelang.compile(_build_lop3_i2_to_i4_decode(T.int), target="cuda")
+    unsigned_kernel = tilelang.compile(_build_lop3_i2_to_i4_decode(T.uint), target="cuda")
+
+    signed = torch.empty(I4_STORAGE_BYTES, dtype=torch.int8, device="cuda")
+    unsigned = torch.empty_like(signed)
+    signed_kernel(qweight, signed)
+    unsigned_kernel(qweight, unsigned)
+    torch.cuda.synchronize()
+
+    unsigned_values = _unpack_i4_values(unsigned, signed=False)
+    signed_values = _unpack_i4_values(signed, signed=True)
+    expected_signed = (unsigned_values << 6) >> 6
+
+    torch.testing.assert_close(signed_values, expected_signed, rtol=0, atol=0)
+    assert not torch.equal(signed, unsigned)
+    assert torch.any(unsigned_values == 2)
+    assert torch.any(unsigned_values == 3)
+
+
+if __name__ == "__main__":
+    tilelang.testing.main()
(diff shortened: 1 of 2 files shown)


LeiWang1999:
thanks, but closed as this was by design.

PR #2496, opened 2026-06-30
[Autotune] Support per-config pass_configs in autotuning

Da1L8-X:
## Summary

Add support for specifying per-config `pass_configs` in the AutoTuner, allowing each tuning configuration to apply different compiler pass settings (e.g., enabling/disabling warp specialization per config). This enables more fine-grained exploration of compiler optimization combinations during autotuning.

### Core Feature
- Each config in `configs` can now include a `pass_configs` key that overrides the global `pass_configs` for that specific compilation
- Per-config `pass_configs` is merged on top of global `pass_configs` (per-config takes priority)
- Supports both API mode (`AutoTuner.run()`) and decorator mode (`@autotune + @tilelang.jit`)

### Code Quality Improvements
- Extract `_merge_pass_configs_into_compile_args()` helper to deduplicate pass_configs merge logic across `grouped_compile.py` and `tuner.py`
- Replace `json.dumps` uniformity check with direct equality comparison (avoids unnecessary serialization overhead in large-scale tuning)
- Extract `_make_jit_compile_func()` factory method to unify lazy/eager `jit_compile` closure logic

### Bug Fix
- Fix decorator-mode cache isolation: use `compile()` directly when `per_config_pass_configs` is present to avoid JIT cache key collision (previously all configs with different pass_configs would share one cached result)

## Usage Example

```python
configs = [
    {"block_M": 128, "block_N": 128, "block_K": 32, "num_stages": 2, "thread_num": 128,
     "pass_configs": {PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True}},
    {"block_M": 128, "block_N": 128, "block_K": 32, "num_stages": 2, "thread_num": 128,
     "pass_configs": {PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: False}},
]

## Summary
- Added per-config `pass_configs` support to AutoTuner in both API mode and decorator/JIT flows.
- Per-config `pass_configs` are merged on top of global `pass_configs`, with per-config values taking priority, and `pass_configs` is treated as a reserved non-tunable key so it isn’t interpreted as an autotuning parameter.
- Updated grouped compilation to apply per-config `pass_configs` during compilation: when configs’ `pass_configs` differ, it bypasses grouped lowering/codegen and compiles each config separately; when uniform, it merges once and proceeds with grouped compilation.
- Extracted/centralized `pass_configs` merge logic and refactored JIT compilation plumbing to unify lazy/eager `jit_compile` closure behavior, including decorator-mode cache isolation fixes by using `compile()` directly when per-config `pass_configs` are present to avoid cache key collisions.
- Changed the grouped uniformity check for `pass_configs` from `json.dumps` comparison to direct equality.
- Added a GEMM autotuning example demonstrating per-config `pass_configs` usage (including `PassConfigKey.TL_DISABLE_WARP_SPECIALIZED`), plus CUDA-only tests covering per-config behavior, mixed global-vs-per-config handling, merge correctness, and decorator precedence/override.

## C++ style / lint notes
- This PR does not touch C++ code, FFI surfaces, or CI/lint/tooling configurations related to C++ style.
- No changes were made to `docs/developer_guide/cpp_style.md`, and no new TLCPP003/TLCPP004 warning-only issues are expected from this change.

### DIFF
--- a/tilelang/autotuner/grouped_compile.py
+++ b/tilelang/autotuner/grouped_compile.py
@@ -6,6 +6,7 @@ so tuner.py can stay focused on orchestration.
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 from collections.abc import Callable
 
@@ -19,6 +20,8 @@ from tilelang.jit.adapter import TVMFFIKernelAdapter
 from tilelang.jit.kernel import JITKernel
 from tilelang.transform import PassConfigKey
 
+logger = logging.getLogger(__name__)
+
 CompileUnitResult = tuple[int, dict[str, Any], JITKernel | None, Exception | None]
 
 
@@ -26,11 +29,7 @@ def _merge_pass_configs_into_compile_args(
     base: CompileArgs,
     per_config_pass_configs: dict[str, Any] | None,
 ) -> CompileArgs:
-    """Merge per-config pass_configs over global CompileArgs defaults.
-
-    If *per_config_pass_configs* is None or empty, returns *base* unchanged.
-    Otherwise returns a new CompileArgs with merged pass_configs.
-    """
+    """Merge per-config pass_configs over global CompileArgs defaults."""
     if not per_config_pass_configs:
         return base
     merged = dict(base.pass_configs or {})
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tilelang/autotuner/grouped_compile.py [hunk 1]
```diff
@@ -37,7 +60,34 @@ def compile_grouped_unit_tvm_ffi(
...
+        # Fall back to per-config compilation when pass_configs differ
+        unit_results: list[CompileUnitResult] = []
+        for (idx, config_arg), per_pc in zip(unit_items, per_config_pass_configs_list, strict=True):
+            try:
+                effective_compile_args = _merge_pass_configs_into_compile_args(compile_args, per_pc)
+                program = elaborate_func(**config_arg)
+                jit_kernel = effective_compile_args.compile_program(program)
+                unit_results.append((idx, config_arg, jit_kernel, None))
+            except Exception as e:
+                unit_results.append((idx, config_arg, None, e))
```
LeiWang1999: Ignoring this exception may mask underlying compilation issues. Is this exception handling truly necessary here?

LeiWang1999 (commented):
Thanks for your contribution, the latest additions read very much like AI‑generated code, it's ok but the code style with excessive defensive programming that makes it hard to follow. Would you be open to streamlining it so we keep only the essential checks?

Da1L8-X:
@LeiWang1999 
Thanks for the review! I've updated the code based on your comments. Additionally, I've rewritten the group compile logic so that cases with the same pass config can be compiled together. Could you please take another look?

tilelang/autotuner/tuner.py
```diff
@@ -1250,6 +1285,38 @@ class AutoTuneImpl(Generic[_P, _T]):
 
     def __post_init__(self):
         self._tuner_cache = {}
+        self._pass_configs_lock = threading.Lock()
```
SiriusNEO, line 1297: Why we need a thread lock here?

tilelang/autotuner/tuner.py
```diff
@@ -1250,6 +1263,38 @@ class AutoTuneImpl(Generic[_P, _T]):
...
+                        return self.jit_impl.compile(*args, **merged)
+                    finally:
+                        self.jit_impl.pass_configs = original_pass_configs
+            else:
+                if mode == "lazy":
+                    return self.jit_impl(*args, **kwargs, __tune_params=config_arg)
+                else:
+                    merged = dict(kwargs)
+                    merged.update(config_arg)
+                    return self.jit_impl.compile(*args, **merged)
```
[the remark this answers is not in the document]
SiriusNEO: @Da1L8-X Could you help to check this comment

tilelang/autotuner/tuner.py
```diff
@@ -1303,28 +1370,14 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> JITKernel | _T:
         if key not in self._tuner_cache:
 
             def jit_elaborate(**config_arg):
+                config_arg.pop("__pass_configs__", None)
```
SiriusNEO: Maybe we can define the "__pass_configs__" dict key as a constant string in this file

tilelang/autotuner/tuner.py
Da1L8-X: self.jit_impl is a shared instance, the jit_compile closure is created once and then called concurrently from ThreadPoolExecutor workers. Since jit_impl.compile() reads self.pass_configs internally, and we temporarily mutate it for per-config overrides, the lock serializes all compilation paths to prevent one thread's temporary mutation from being observed by another.

tilelang/autotuner/tuner.py
Da1L8-X: done

Da1L8-X:
@SiriusNEO Thanks for the review, I've updated the code, please help review again.

tilelang/autotuner/tuner.py
```diff
@@ -591,17 +618,26 @@ def compile_unit(unit_items: list[tuple[int, dict[str, Any]]]):
                     unit_results.append((idx, config_arg, None, e))
             return unit_results
 
-        compile_units: list[list[tuple[int, dict[str, Any]]]] = []
+        compile_units: list[tuple[list[tuple[int, dict[str, Any]]], dict | None]] = []
```
SiriusNEO: Some type annotations look overly long; perhaps we need to introduce type aliases within this compile unit.

SiriusNEO (dismissed):
LGTM. Just one suggestion on type annotation

tilelang/autotuner/tuner.py
SiriusNEO: (And other "unit" things in this file)

Da1L8-X: done

PR #2499, opened 2026-06-30
[FEATURE] Add block-causal attention for dLLM example

perkyfever:
## Summary

- Replaced the fixed block-causal attention implementation with TileLang forward and backward kernels parameterized by supported `dllm_block_size` values.
- Added variable-length packed-sequence support using `cu_seqlens`, including forward/backward kernels, PyTorch wrappers, reference implementations, and validation tests.
- Updated the test module to cover fixed-shape and variable-length attention across all supported block sizes.
- Removed kernel caching, benchmarking, regression, and related helper routines from the fixed-shape example.


LeiWang1999:
please fix the lint :)

examples/block_causal_attention/block_causal_attention.py
```diff
@@ -0,0 +1,593 @@
...
+def _diffusion_allowed(q_abs, k_abs, half_len: int, mask_block: int):
+    q_clean = T.if_then_else(q_abs >= half_len, 1, 0)
+    k_clean = T.if_then_else(k_abs >= half_len, 1, 0)
+    q_local = T.if_then_else(q_abs >= half_len, q_abs - half_len, q_abs)
+    k_local = T.if_then_else(k_abs >= half_len, k_abs - half_len, k_abs)
+    q_block = q_local // mask_block
+    k_block = k_local // mask_block
+
+    same_region = T.if_then_else(q_clean == k_clean, 1, 0)
+    block_diagonal = T.if_then_else(q_block == k_block, 1, 0) * same_region
```
Godofnothing: Should it be block_diagonal only for noisy regions?

In the current implementation it seems to be block_diagonal both for noisy and clean part.

perkyfever: This code implements the block-wise mask. Based on this [picture](https://habrastorage.org/r/w1560/getpro/habr/upload_files/0e7/3ac/2c6/0e73ac2c689b4dcaa658abfd325a294e.png) (left part) it should be only noisy.

Right now the code also enables block-diagonal part for the clean region, but those positions are allowed by clean block-causal masking anyway, so it's effectively redundant rather than wrong.

Though, the masking logic could be clearer if gated to noisy part only. Thanks for pointing this out!

perkyfever:
I will be fixing the remaining concerns soon :)

Godofnothing (commented):
-

examples/block_causal_attention/block_causal_attention.py
Godofnothing: @perkyfever 
I see, I thought the original intent was to have a mask like this
изображение
But if the intended mask is shown on the left pic you attached, then it should be correct.

examples/block_causal_attention/block_causal_attention.py
```diff
@@ -0,0 +1,590 @@
+import os
+
+import torch
+import tilelang
+import tilelang.language as T
+
+LOG2_E = 1.4426950408889634
+
+_FAST_MATH = os.environ.get("BLOCK_CAUSAL_TL_FAST_MATH", "0") == "1"
```
LeiWang1999: not too elegant to read an environment variable in example in my opnion.

examples/block_causal_attention/block_causal_attention.py
```diff
@@ -0,0 +1,590 @@
...
+
+_TL_DTYPE = {torch.bfloat16: "bfloat16", torch.float16: "float16"}
+_FWD_CACHE = {}
+_BWD_CACHE = {}
+
+
+def _tl_dtype(tensor: torch.Tensor) -> str:
+    if tensor.dtype not in _TL_DTYPE:
+        raise TypeError(f"unsupported dtype {tensor.dtype}; use bfloat16 or float16")
+    return _TL_DTYPE[tensor.dtype]
```
LeiWang1999: please use `T.dtype(tensor.dtype)`

examples/block_causal_attention/block_causal_attention.py
```diff
@@ -0,0 +1,590 @@
...
+
+            T.copy(dv, dv_shared)
+            T.copy(dk, dk_shared)
+            T.copy(dv_shared, dV[bz, by * block : (by + 1) * block, bx, :])
+            T.copy(dk_shared, dK[bz, by * block : (by + 1) * block, bx, :])
+
+    return dkv_kernel
+
+
+_build_fwd = tilelang.jit(out_idx=[3, 4], pass_configs=_FWD_PASS_CFG)(_fwd_template)
```
LeiWang1999: not too elegant as well, please follow another examples as possible.

LeiWang1999 (changes requested):

perkyfever:
# Block-causal (dLLM) attention example

## Motivation

Diffusion / block-diffusion LLMs (e.g. BD3-LM) train by attending over a sequence that is a **noisy half followed by a clean half**, under a specific block-causal mask. That mask is **block-sparse**, so a dense implementation (SDPA / a full attention mask, or an uncompiled FlexAttention mask) does a lot of wasted work.

## What it does

Two self-contained example modules:

- `block_causal_attention.py` -- fixed-length (`[batch, seq_len, heads, dim]`) forward + backward.
- `block_causal_attention_varlen.py` -- the packed / variable-length (`cu_seqlens`) twin, sharing
  the mask helper and reference from the fixed module.

Highlights:

- **Arbitrary diffusion block size** `{1, 2, 4, 8, 16, 32, 64}`;
- **The dLLM mask** (within each half, tokens grouped into blocks of `dllm_block`): `block_diagonal`, `offset_causal` and `block_causal`. See the left image from [here](https://habrastorage.org/r/w1560/getpro/habr/upload_files/0e7/3ac/2c6/0e73ac2c689b4dcaa658abfd325a294e.png);
- **Sparse tile schedule** -- each query tile visits only the key tiles it can attend;
- **FlashAttention-style**: single-pass online-softmax forward + dQ query-parallel + dK/dV key-parallel + register-accumulated. No atomics => deterministic;
- fp16 and bf16, validated against a PyTorch reference (forward output and gradients).

## Files

- `block_causal_attention.py` / `block_causal_attention_varlen.py` -- kernels + torch reference.
- `test_block_causal_attention.py` -- correctness tests (all block sizes, fixed + varlen).

## Benchmarks

### Setup

Tested on 1xH100 for bf16 dtype against FlexAttention (max-autotune `torch.compile` & Triton backend) + `BlockMask` (max-autotune `torch.compile`).

### Fixed-length kernel

`seq_len=8192, heads=32, head_dim=192, batch=2`:

| dllm_block | fwd+bwd (ours) | fwd+bwd (flex) | speedup |
|---|---|---|---|
| 1  | 13.42 | 25.47 | 1.90x |
| 8  | 13.55 | 25.97 | 1.92x |
| 32 | 13.27 | 26.01 | 1.96x |
| 64 | 13.51 | 26.06 | 1.93x |

### Varlen kernel

Total of 16k tokens, `dllm_block=32`:

| docs | fwd+bwd (ours) | fwd+bwd (flex) | speedup |
|---|---|---|---|
| 2048x8 | 6.66 | 10.06 | 1.51x |
| 4096x4 | 10.92 | 15.43 | 1.41x |
| 8192x2 | 19.33 | 26.15 | 1.35x |

### Summary

* **1.90-1.96x** faster than FlexAttention on fwd+bwd (fixed-length, uniform across block size);
* **1.35-1.51x** faster than FlexAttention on fwd+bwd for packed inputs.

perkyfever:
Hi @LeiWang1999,

I believe, all the concerns raised so far seem to have been addressed. As suggested, I explored other examples and tried to align my implementation with them. I have also added more context in the comment above.

Please let me know if there’s anything else to be updated.

LeiWang1999 (approved):

PR #2502, opened 2026-07-01
[Enhancement] Fix T.assume and loop bounds to eliminate redundant boundary checks

LJC00118:
### Summary
- Updated the TVM submodule pointer.
- Simplified `InjectAssumes` so `T.assume` is treated as an axiom without the previous local-variable read warning path.
- Added a regression test covering `T.assume` on a mutable `T.alloc_var` index and verifying redundant bounds checks are eliminated.
- Removed the old `InjectAssumes` transform test file.

### C++ style / lint notes
- This PR touches C++ code in `src/transform/inject_assumes.cc`, so the rules in `docs/developer_guide/cpp_style.md` are relevant.
- No new public C++ API is introduced, and the change appears to be an internal cleanup rather than a style-risky API/FFI adjustment.
- The warning-only “C++ API Style Audit” CI step may still report advisory findings, but nothing in this change suggests a new blocking correctness or build issue.


LeiWang1999 (approved):

LJC00118:
@regression-perf

PR #2505, opened 2026-07-02
[Pipeline] Fix physical async wait counts

LeiWang1999:
## Summary

- Fix async pipeline wait-count calculation when a logical pipeline stage is split into multiple physical `cp.async.commit_group` groups.
- Preserve fine-grained async issue and wait behavior by waiting only for the physical commit group that a consumer actually reads.
- Avoid generating an extra synthetic pipeline scope block when the rewritten pipeline does not own local allocations.

## Changes

- Track rewritten async statements back to their physical commit group and compute `wait_group` counts from the physical FIFO distance instead of only the logical pipeline stage distance.
- Remove the hard-coded wait-relaxation cap so fallback wait relaxation does not clamp valid physical wait counts.
- Guard barrier initialization updates for cases where pipeline rewriting no longer produces a synthetic scope block.
- Add a regression test covering three independent async producer groups with interleaved consumers, expecting steady-state waits to keep the unrelated groups in flight.

## Validation

- `PYTHONPATH=$PWD python -m pytest testing/python/transform/test_tilelang_transform_Inject_software_pipeline.py -q`

## Risk

- Kernels with several physical async producer groups per logical stage may now keep more `cp.async` groups in flight. This is intentional for fine-grained waiting, but performance should still be checked on affected kernels.

### Summary
- Fixes async software-pipeline wait-group computation so `tl_pipelined_wait_group`/`wait_group` is derived from the physical async FIFO/`cp.async` commit-group distance (including cases where one logical stage is split into multiple physical `commit_group`s), not just the logical stage distance.
- Removes the hard-coded cap on relaxed static async wait values, preventing valid fallback wait counts from being clamped.
- Adds guards to keep barrier initialization metadata consistent when pipeline rewriting no longer creates a synthetic scope block.
- Adds a regression test covering three independent async producer groups with interleaved consumers, asserting that unrelated groups remain in flight while the tail-drain wait sequence stays steady-state.

### C++ style / lint notes
- This PR updates C++ transformation logic in `src/transform/inject_pipeline.cc`, but does not appear to touch `docs/developer_guide/cpp_style.md`.
- No build-breaking or correctness-related C++ style issues are indicated by the changes themselves.
- The “C++ API Style Audit (warning only)” CI step may still emit advisory findings, but nothing in this PR suggests a new blocking API/FFI/maintainability risk.

### DIFF
--- a/src/transform/inject_pipeline.cc
+++ b/src/transform/inject_pipeline.cc
@@ -1545,6 +1545,8 @@ private:
 
   struct AsyncStateGlobal {
     BufferSet dst_buffers;
+    BufferCommitGroupMap buffer_to_commit_group;
+    int commit_group_count{0};
     Optional<PrimExpr> producer_head{PrimExpr(-1)};
 
     bool writes(const Buffer &buffer) const {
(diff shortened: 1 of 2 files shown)


LeiWang1999:
local test and performance regression can pass.

PR #2508, opened 2026-07-02
[BugFix][CUDA] Round-to-nearest-even for scalar int -> bfloat16 cast

### LINKED ISSUE #2483: [BUG][Fuzzer][wrong-code] `T.Cast` of an integer to `bfloat16` silently truncates toward zero instead of round-to-nearest-even — int→`float16` and `float32`→bf16 round correctly
### Required prerequisites

0.1.11

### System information

- Python 3.13, linux.
- Reproduced on **L40S (sm_89), H100 (sm_90), and B200 (sm_100)** — the defect is in `CastNode` codegen, host/arch-independent.

### Problem description

`T.Cast` of an integer to `bfloat16` lowers to a plain C-style `(bfloat16_t)(int)` cast, which **rounds toward zero** (truncates the bits beyond bf16's 8-bit significand) instead of round-to-nearest-even. Any integer whose magnitude needs more than 8 significant bits converts to the wrong bf16 value — always the representable step **closer to zero** (`259 → 258`, `33000 → 32768`, `−259 → −258`). It compiles and runs with no error: a silent, deterministic 1-ULP-toward-zero bias. The sibling int→`float16` cast and the `float32`→bf16 cast both round to nearest correctly, so this is an inconsistency specific to the int→bf16 rounding mode.

<details>
<summary>Root cause — int scalar cast falls through to a plain C-cast; the bf16 RNE path is below it</summary>

In [`CodeGenTileLangCUDA::VisitExpr_(const CastNode*)`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L1492) the first branch is:

```cpp
// Emit simple C-style type conversion for scalar casts without custom rounding.
if (from_ty.is_scalar() && cast_round.empty())
  return CodeGenC::VisitExpr_(op, os);                 // codegen_cuda.cc:1523
```

An `int` source is scalar and carries no `round` annotation, so int→bf16 takes this fallthrough and emits `(bfloat16_t)(int)` — `bfloat16_t` is `cutlass::bfloat16_t`, whose `int` constructor converts via truncation. The dedicated `float32`→bf16 round-to-nearest path (`__float22bfloat162_rn`, [`codegen_cuda.cc#L1593`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L1593)) sits *after* this fallthrough, so an integer source never reaches it. Emitted CUDA for the repro: `Ob[0] = ((bfloat16_t)Al[0]);`.
</details>

### Reproducible example code

Note: use a real `.py` file (eager JIT calls `inspect.getsource`; `python -c` trips an unrelated `OSError`). `tilelang.disable_cache()`.

```python
tilelang.disable_cache()
N = 32
@T.prim_func
def main(A: T.Tensor((N,), "int32"), O: T.Tensor((N,), "bfloat16"), Oh: T.Tensor((N,), "float16")):
    with T.Kernel(1, threads=N) as _:
        Al = T.alloc_fragment((N,), "int32")
        Ob = T.alloc_fragment((N,), "bfloat16"); Ol = T.alloc_fragment((N,), "float16")
        T.copy(A, Al)
        for i in T.Parallel(N):
            Ob[i] = T.Cast("bfloat16", Al[i])     # suspect: int -> bf16
            Ol[i] = T.Cast("float16",  Al[i])     # control: int -> fp16 (rounds correctly)
        T.copy(Ob, O); T.copy(Ol, Oh)

A = torch.zeros(N, dtype=torch.int32, device="cuda"); A[0], A[1], A[2] = 259, 33000, -259
mod = tilelang.compile(main, target="cuda")
O  = torch.zeros(N, dtype=torch.bfloat16, device="cuda")
Oh = torch.zeros(N, dtype=torch.float16,  device="cuda")
mod(A, O, Oh); torch.cuda.synchronize()
print(O[:3].float().tolist())                     # int->bf16 TileLang: [258.0, 32768.0, -258.0]
print(A[:3].to(torch.bfloat16).float().tolist())  # int->bf16 torch RNE: [260.0, 33024.0, -260.0]
print(Oh[:3].float().tolist())                    # int->fp16 control:  [259.0, 32992.0, -259.0]  (correct)
```

Every wrong bf16 value is the step **closer to zero** — the signature of round-toward-zero. The int→fp16 control in the same kernel rounds correctly (for fp16-inexact values it matches torch RNE), isolating the defect to the int→bf16 rounding mode rather than int→float casting in general.

### Traceback

```pytb
No traceback — the kernel compiles and runs to completion; the int→bf16 result is silently off by up to one bf16 ULP toward zero, deterministically.
```

### Expected behavior

`T.Cast(int → bfloat16)` should round to nearest even, matching IEEE default rounding and what the int→`float16` path and the `float32`→bf16 path already do. This is grounded in the project's own behavior: in the identical kernel the int→fp16 cast rounds to nearest, and `float32`→bf16 uses `__float22bfloat162_rn`; only the int→bf16 cast truncates — an inconsistency between the float casts, not a design choice.

### Additional context

**Suggested fix (proposed, not built).** Route an integer→bf16 cast through round-to-nearest (e.g. int→float→bf16 via the existing RNE path, as the int→float16 cast effectively does) rather than the plain `(bfloat16_t)(int)` truncating constructor reached by the scalar-cast fallthrough at [`codegen_cuda.cc#L1523`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L1523).

**Scope (verified).** The defect is **source-dtype-driven (integer source → bf16), not site-driven**: the same `(bfloat16_t)(int)` lowering is reached from elementwise `T.Cast`, a cast inside `T.copy`, and reduce/scan output-store casts — all truncate identically, so it is one bug. `float32`→bf16 is correct (RNE) at every site, so a `T.gemm` bf16 epilogue (fp32 accumulate → bf16 store) takes the RNE path and is **not** affected; the bug bites integer-source casts (quantization, integer index/scale → bf16).

Reproduced on L40S (sm_89), H100 (sm_90), and B200 (sm_100), TileLang 0.1.11, deterministic; the int→fp16 control rounds correctly on each. I did not build the fix.

**Provenance.** The `round`-annotation-gated scalar-cast fallthrough in its current form ([`codegen_cuda.cc#L1523`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L1523)) was shaped by [#2260](https://github.com/tile-ai/tilelang/pull/2260) ("Add stochastic rounding cast for f32→fp8/fp4", 2026-05-28), which added the `cast_round`/`sat`/`rbits` machinery. But the underlying behavior — a plain scalar `int→bf16` cast falling through to `CodeGenC`'s `(bfloat16_t)(int)` truncating conversion — predates that annotation work; #2260 added the rounding hooks for f32→fp8/fp4 and did not introduce (or fix) the integer-source truncation. So the defect is introduced **no later than #2260**; an exact origin is in the earlier CastNode lowering and is unverified — not pinned to a single PR.

RuneFang:
Fixes #2483

## Summary

Scalar `Ob[i] = T.Cast("bfloat16", Al[i])` currently rounds toward zero instead of round-to-nearest-even. one bf16 ULP lead different result  

| Input (int32) | Expected (bf16, RNE) | Before | After |
|---|---|---|---|
| `259`   | `260`    | `258`    | `260` |
| `33000` | `33024`  | `32768`  | `33024` |
| `-259`  | `-260`   | `-258`   | `-260` |

## Root cause

The scalar `int/uint -> bfloat16` cast falls through to the generic scalar branch `CodeGenTileLangCUDA::VisitExpr_(CastNode)` and lowers to a plain C-style cast:

```cpp
((bfloat16_t)int_value)   // == cutlass::bfloat16_t(int)
```

`cutlass::bfloat16_t`'s integer constructor takes the `from_32_bit_integer` path (`bits >> 16`), which **truncates toward zero** — it doesn't apply RNE. Any integer whose magnitude needs more than 8 significand bits gets biased by up to one ULP toward zero.

## Fix

In `CodeGenTileLangCUDA::VisitExpr_(CastNode)`, add an early branch that routes scalar `int/uint -> bf16` through an explicit `(float)` step so we hit `bfloat16_t(float)`:

```cpp
  if (from_ty.is_scalar() &&
      (target_ty.is_bfloat16() || target_ty.is_float16()) &&
      (from_ty.is_int() || from_ty.is_uint()) && cast_round.empty()) {
    os << "((";
    this->PrintType(target_ty, os);
    os << ")(float)(" << PrintExpr(op->value) << "))";
    return;
  }
```

## Test
```
 python -m pytest testing/python/issue/test_tilelang_issue_2483.py -v
```
before
image
after
image

### Summary
- Fixed CUDA codegen for scalar integer/unsigned-integer casts to `bfloat16`/`float16`: when the cast has an empty `round` hint, it now forces an explicit `(float)` conversion in the emitted C++ so the conversion uses round-to-nearest-even instead of truncating toward zero.
- Added regression coverage in `testing/python/issue/test_tilelang_issue_2483.py` for:
  - `int32 -> bfloat16` (matches Torch RNE)
  - `int64 -> bfloat16` (matches Torch RNE)
  - `int64 -> float16` (compiles and matches Torch)
  - `float32 -> bfloat16` (guard to ensure existing rounding remains correct)
- Updated `testing/python/jit/test_tilelang_jit_diagnostics.py` to adjust how `TILELANG_JIT_DIAGNOSTICS` is enabled and to disable cache for the affected test (workaround for unrelated JIT diagnostics/timeout/cache-miss flakiness).

Test (suggested)
- `pytest testing/python/issue/test_tilelang_issue_2483.py`

### C++ style / lint notes
- Does not touch rules documented in `docs/developer_guide/cpp_style.md`.
- No additional correctness/build/test changes introduced purely for style; any `C++ API Style Audit (warning only)` findings would be advisory (this PR is focused on cast-codegen correctness, not public API/FFI/maintainability).

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -1536,6 +1536,15 @@ void CodeGenTileLangCUDA::VisitExpr_(const CastNode *op, std::ostream &os) {
   bool cast_sat = get_bool_anno("sat", true);
   Optional<PrimExpr> cast_rbits = get_expr_anno("rbits");
 
+  if (from_ty.is_scalar() &&
+      (target_ty.is_bfloat16() || target_ty.is_float16()) &&
+      (from_ty.is_int() || from_ty.is_uint()) && cast_round.empty()) {
+    os << "((";
+    this->PrintType(target_ty, os);
+    os << ")(float)(" << PrintExpr(op->value) << "))";
+    return;
+  }
+
   // Emit simple C-style type conversion for scalar casts without custom
   // rounding.
   if (from_ty.is_scalar() && cast_round.empty())
--- a/testing/python/issue/test_tilelang_issue_2483.py
+++ b/testing/python/issue/test_tilelang_issue_2483.py
@@ -1,0 +1,124 @@
+import torch
+import tilelang
+import tilelang.testing
+import tilelang.language as T
+
+tilelang.disable_cache()
+
+# The three headline values from issue #2483. Each is one bf16 ULP away
+# from a bf16-representable neighbour, so RNE and truncate-toward-zero
+# produce different results:
+#   RNE  :  259 -> 260, 33000 -> 33024, -259 -> -260
+#   trunc:  259 -> 258, 33000 -> 32768, -259 -> -258
+ISSUE_2483_VALUES = [259, 33000, -259]
+ISSUE_2483_BF16_RNE = [260.0, 33024.0, -260.0]
+
+
+def test_int32_to_bf16_rounds_to_nearest_even():
+    """Faithful port of the issue #2483 reproducer."""
+    N = len(ISSUE_2483_VALUES)
+
+    @T.prim_func
+    def main(
+        A: T.Tensor((N,), "int32"),
+        O: T.Tensor((N,), "bfloat16"),
+        Oh: T.Tensor((N,), "float16"),
+    ):
+        with T.Kernel(1, threads=N) as _:
+            Al = T.alloc_fragment((N,), "int32")
+            Ob = T.alloc_fragment((N,), "bfloat16")
+            Ol = T.alloc_fragment((N,), "float16")
+            T.copy(A, Al)
+            for i in T.Parallel(N):
+                Ob[i] = T.Cast("bfloat16", Al[i])  # int -> bf16
+                Ol[i] = T.Cast("float16", Al[i])  # int -> fp16
+            T.copy(Ob, O)
+            T.copy(Ol, Oh)
+
+    kernel = tilelang.compile(main)
+
+    a = torch.tensor(ISSUE_2483_VALUES, dtype=torch.int32, device="cuda")
+    o_bf16 = torch.zeros(N, dtype=torch.bfloat16, device="cuda")
+    o_fp16 = torch.zeros(N, dtype=torch.float16, device="cuda")
+    kernel(a, o_bf16, o_fp16)
+    torch.cuda.synchronize()
+
+    got = o_bf16.float().tolist()
+
+    assert got == ISSUE_2483_BF16_RNE, f"int->bf16 must round-to-nearest-even; got {got}, expected {ISSUE_2483_BF16_RNE}"
+
+    # Cross-check against torch's native int->bf16 (which is RNE).
+    ref = a.to(torch.bfloat16)
+    tilelang.testing.torch_assert_close(o_bf16, ref, rtol=0, atol=0)
+
+    ref_fp16 = a.to(torch.float16)
+    tilelang.testing.torch_assert_close(o_fp16, ref_fp16, rtol=0, atol=0)
+
+
+def test_int64_to_bf16_rounds_to_nearest_even():
+    N = len(ISSUE_2483_VALUES)
+
+    @T.prim_func
+    def main(A: T.Tensor((N,), "int64"), O: T.Tensor((N,), "bfloat16")):
+        with T.Kernel(1, threads=N) as _:
+            Al = T.alloc_fragment((N,), "int64")
+            Ob = T.alloc_fragment((N,), "bfloat16")
+            T.copy(A, Al)
+            for i in T.Parallel(N):
+                Ob[i] = T.Cast("bfloat16", Al[i])
+            T.copy(Ob, O)
+
+    kernel = tilelang.compile(main, out_idx=[1])
+
+    a = torch.tensor(ISSUE_2483_VALUES, dtype=torch.int64, device="cuda")
+    o = kernel(a)
+    ref = a.to(torch.bfloat16)
+    tilelang.testing.torch_assert_close(o, ref, rtol=0, atol=0)
+
+
+def test_int64_to_fp16_compiles_and_matches_torch():
+    N = len(ISSUE_2483_VALUES)
+
+    @T.prim_func
+    def main(A: T.Tensor((N,), "int64"), O: T.Tensor((N,), "float16")):
+        with T.Kernel(1, threads=N) as _:
+            Al = T.alloc_fragment((N,), "int64")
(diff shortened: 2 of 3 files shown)


RuneFang:
The two failures in testing/python/jit/test_tilelang_jit_diagnostics.py::test_jit_compile_reports_timeout_for_hanging_nvcc and test_kernel_cache_miss_compile_logs_context on CI are unrelated to this PR. I've pushed an additional commit as a quick hotfix to disable the cache in the affected test.
image

LeiWang1999:
surprised this is considered an issue. If so, why is this the default behavior in cutlass? most libraries that depend on cute may follow this behavior, so I think we should preserve it.

RuneFang:
> surprised this is considered an issue. If so, why is this the default behavior in cutlass? most libraries that depend on cute may follow this behavior, so I think we should preserve it.



 you are right ,default int->bf16 behavior in cutlass is RNZ，I think the CUTLASS behavior here might just be historical. But for an int → float(bf16) cast like T.Cast("bfloat16", A[i]), RNE (round-to-nearest, ties-to-even) is the default rounding mode recommended by IEEE 754 ，sorry I just found fragmentary mentions in this link  https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/intel-ieee-754-2008-binary-float-conform-lib-use.html   
image
 https://standards.ieee.org/ieee/754/6210/

This issue may be too complicated to discuss. I think I could close this PR.

PR #2514, opened 2026-07-03
[Transform][Layout] Avoid thread-indexed replicated fragment readback

LeiWang1999:
## Summary

- Prefer loop layout candidates that avoid thread-dependent physical indices for fragment accesses when publishing replicated fragments to shared or global memory.
- Prevent fully replicated fragment readback from lowering to `fragment[threadIdx.x]`, which can force register arrays into local memory.

## Changes

- Add a layout-candidate tie-breaker in `src/op/parallel.cc` that detects whether a candidate would make fragment physical indices depend on the CUDA thread variable.
- Keep fragment-only loops on the existing containment and replication heuristic, limiting the behavior change to shared/global store paths.
- Extend `testing/python/layout/test_tilelang_layout_inference.py` with coverage for both explicit `T.Parallel` readback and `T.copy` readback from fully replicated fragments.

## Validation

- `./format.sh`

### Summary
- Improved layout inference for replicated fragment readback in `src/op/parallel.cc` by preferring candidates that avoid thread-dependent physical fragment indices when lowering stores to shared/global memory.
- Added a new helper in `ParallelOpNode` to detect whether a candidate layout would make fragment physical indices depend on the CUDA thread variable.
- Extended CUDA layout inference tests to cover fully replicated fragment readback through both `T.Parallel` and `T.copy`, with assertions that the generated kernel guards readback to a single thread.

### C++ style / lint notes
- This PR touches C++ code, so it falls under the rules in `docs/developer_guide/cpp_style.md`.
- No correctness, build, or test regressions are indicated in the change summary.
- The change appears focused on behavior rather than style; any `C++ API Style Audit (warning only)` findings should be treated as advisory unless they expose a clear API/FFI/maintainability risk.

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -10,6 +10,7 @@
 #include <tvm/runtime/logging.h>
 
 #include <algorithm>
+#include <tvm/tirx/analysis.h>
 #include <tvm/tirx/op.h>
 
 #include "../layout/layout.h"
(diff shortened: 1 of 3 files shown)


LeiWang1999:
@regression-perf

PR #2585, opened 2026-07-10
[BugFix] Preserve guard identity in LoopUnswitching

### LINKED ISSUE #2539: [BUG][Fuzzer][wrong-code] `LoopUnswitching` silently collapses two same-form `if` guards on *different* buffers into one, giving wrong results
### Required prerequisites

0.1.11

### System information

NVIDIA A10G (Ampere, sm_86), CUDA 12.8, Python 3.11, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). The defect is in a target-independent TIR pass (`LoopUnswitching`), not GPU-specific.

### Problem description

`tl.LoopUnswitching` collapses two loop-invariant `if` guards into one whenever their conditions have the same structural *form*, even when they test **different buffers**. A second guard `if Q[bx] < 2` is merged with the first `if P[bx] < 2`, so `Q`'s body then runs under `P`'s condition instead of its own — a silently wrong result, no error.

The pass hoists the first loop-invariant guard out of the loop and rewrites the body with `IfBranchReplacer`, which replaces **every** `if` whose condition is `StructuralEqual()` to the hoisted one ([`loop_unswitching.cc#L417-L428`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_unswitching.cc#L417-L428)). That match treats `Q[bx] < 2` as equal to the hoisted `P[bx] < 2`, so the `Q` guard is dropped: `Q`'s body runs unconditionally in the then-branch (`P[bx] < 2`) and never in the else-branch (`P[bx] >= 2`), regardless of `Q[bx]`.

The pass is on by default in the CUDA lowering pipeline ([`cuda/pipeline.py#L187`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/pipeline.py#L187)). The trigger is the structural identity of the condition *form*: a guard of a different form (e.g. `if Q[bx] >= 2`) is not collapsed and the result is correct.

### Reproducible example code

```python
import torch
import tilelang
import tilelang.language as T


def make_kernel(M, K, disable_unswitch):
    cfg = {"tl.disable_loop_unswitching": True} if disable_unswitch else {}

    @tilelang.jit(pass_configs=cfg)
    def kernel(
        P: T.Tensor((M,), "int32"),
        Q: T.Tensor((M,), "int32"),
        OA: T.Tensor((M,), "int32"),
        OB: T.Tensor((M,), "int32"),
    ):
        with T.Kernel(M, threads=1) as bx:
            OA[bx] = 0
            OB[bx] = 0
            for k in T.serial(K):        # a loop is required — LoopUnswitching only fires inside one
                if P[bx] < 2:            # guard on buffer P
                    OA[bx] = 1           # "P body ran" flag
                if Q[bx] < 2:            # sibling guard on a DIFFERENT buffer Q
                    OB[bx] = 1           # "Q body ran" flag -> must follow Q, not P
    return kernel


M, K = 4, 2
P = torch.tensor([0, 0, 5, 5], dtype=torch.int32, device="cuda")  # active iff < 2
Q = torch.tensor([0, 5, 0, 5], dtype=torch.int32, device="cuda")  # active iff < 2

OB_default = torch.empty(M, dtype=torch.int32, device="cuda")
make_kernel(M, K, disable_unswitch=False)(P, Q, torch.empty(M, dtype=torch.int32, device="cuda"), OB_default)

OB_bypass = torch.empty(M, dtype=torch.int32, device="cuda")
make_kernel(M, K, disable_unswitch=True)(P, Q, torch.empty(M, dtype=torch.int32, device="cuda"), OB_bypass)

print("expected (OB follows Q<2):", (Q < 2).int().cpu().tolist())  # [1, 0, 1, 0]
print("default build            :", OB_default.cpu().tolist())     # [1, 1, 0, 0]  WRONG -> follows P
print("bypass  build            :", OB_bypass.cpu().tolist())      # [1, 0, 1, 0]  correct
```

Output (deterministic across runs):

```
expected (OB follows Q<2): [1, 0, 1, 0]
default build            : [1, 1, 0, 0]     # OB follows P=[0,0,5,5], not Q -> Q guard was replaced by P's
bypass  build            : [1, 0, 1, 0]     # correct once LoopUnswitching is disabled
```

`OB` (did the `Q` body run) should track `Q`, but in the default build it tracks `P` exactly — the `Q` guard has been collapsed into the `P` guard. Disabling only this pass restores the correct result, isolating the defect to `LoopUnswitching`. This `M=4, K=2` is already minimal: one block of each `P`/`Q` combination triggers it, and `K=1` does not trigger — a single-trip loop is simplified away before unswitching.

### Traceback

```pytb
No traceback — the compile and run succeed; the output is silently wrong and deterministic.
```

### Expected behavior

`OB` should equal `[1, 0, 1, 0]` (it tracks `Q < 2`) — the `Q` guard must keep gating its own body, since `P[bx] < 2` and `Q[bx] < 2` are different conditions (different buffers). TileLang already handles the *truly identical* case correctly: `test_multiple_identical_conditions` ([test#L294](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/testing/python/transform/test_tilelang_transform_loop_unswitching.py#L294)) hoists two `if cond[0] > 0` guards on the *same* buffer into one loop, which is correct because the conditions really are the same. The sibling case (`P` vs `Q`) should be equally correct.

### Additional context

**Root cause.** The `IfBranchReplacer::VisitStmt_` match ([line 419](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_unswitching.cc#L419)) treats two `BufferLoad`-based conditions as equal whenever their structural form matches, because `StructuralEqual::operator()` ([`structural_equal.h#L71`](https://github.com/apache/tvm-ffi/blob/3c35034fd1026011736e19a4e0e1ed0f22058c42/include/tvm/ffi/extra/structural_equal.h#L71), `map_free_vars = false`) does not distinguish the buffer data `Var`s inside the loads. The bare data `Var`s compare unequal, but once wrapped in a `BufferLoad` the distinction is lost, so `StructuralEqual()(P[bx] < 2, Q[bx] < 2)` returns `True`. Dumping the module after `tilelang.transform.LoopUnswitching()` for the minimal kernel shows the collapse — the hoisted then-branch adds both `A` and `B` unconditionally, and the else-branch never adds `B`:

```python
# AFTER tilelang.transform.LoopUnswitching()  (post-pass IR, abbreviated)
if P[bx] < 2:
    for k in range(2):
        OA[bx] = 1
        OB[bx] = 1          # <-- Q[bx] < 2 guard is gone; the Q body runs whenever P[bx] < 2
else:
    for k in range(2):
        T.evaluate(0)       # <-- Q body never runs when P[bx] >= 2, even if Q[bx] < 2
```

<details>
<summary>Per-block observed vs expected (default build), wrong iff P and Q disagree</summary>

| bx | P[bx] | Q[bx] | expected `OB` | got `OB` | what happened |
|----|-------|-------|---------------|----------|---------------|
| 0 | 0 (active) | 0 (active) | 1 | 1 | agree — correct |
| 1 | 0 (active) | 5 (inactive) | 0 | 1 | `Q` body ran under `P`'s true branch |
| 2 | 5 (inactive) | 0 (active) | 1 | 0 | `Q` body dropped in `P`'s false branch |
| 3 | 5 (inactive) | 5 (inactive) | 0 | 0 | agree — correct |

`OB` tracks `P` (`[1,1,0,0]`) instead of `Q` (`[1,0,1,0]`): correct exactly where `P` and `Q` agree.

</details>

**Suggested fix.** One option is to require that a candidate `if` and the hoisted `if` reference the *same* buffer object(s) — an identity check on the `Buffer`/data `Var`s used by the condition — before collapsing, so only guards that are genuinely the same predicate are merged. The same identity-sensitivity would apply to the `CallCheckerExcludingIf` match at [line 172](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_unswitching.cc#L172), which excludes calls inside `if`s by the same `StructuralEqual` condition match. A numerical regression test with a sibling guard on a different buffer would help, since the existing test only exercises identical conditions on the same buffer.

**Distinct from #1804.** [#1804](https://github.com/tile-ai/tilelang/issues/1804) ("[BUG] Loop unswitching pass causes buggy codegen", closed, fixed by [#1816](https://github.com/tile-ai/tilelang/pull/1816) merged 2026-02-09, "gate non-trivial else behind PassConfig") concerned unswitching on `blockIdx`/`threadIdx`-dependent predicates breaking synchronization — that fix gates *whether to hoist*. The defect here is in *which guards are collapsed once hoisting proceeds* (`IfBranchReplacer`'s `StructuralEqual` match); hoisting still occurs, and this reproduces on `v0.1.11` (which includes #1816), so that fix does not cover this path. I searched the open and closed tracker and found no existing report of the sibling-buffer collapse.

zyy3077:
## Issue

https://github.com/tile-ai/tilelang/issues/2539

## Summary

LoopUnswitching treated sibling guards as the same predicate when their
conditions were structurally equal. This is unsafe when two conditions have the
same arithmetic shape but read from different buffers, because the guards are
not the same runtime predicate.

## Example

```python
for k in T.serial(K):
    if P[bx] < 2:
        OA[bx] = 1
    if Q[bx] < 2:
        OB[bx] = 1
```

Before this fix, LoopUnswitching could collapse the `Q[bx] < 2` guard into the
hoisted `P[bx] < 2` guard because the two `BufferLoad < const` expressions are
structurally similar. As a result, `OB` followed `P` instead of `Q`. Disabling
only LoopUnswitching restored the correct behavior, isolating the issue to this
pass.

## Root cause

The pass used structural equality directly when deciding whether an `if` matched
the hoisted condition:

```cpp
StructuralEqual()(op->condition, hoisted_condition)
```

The same structural check was also used in `CallCheckerExcludingIf` when
excluding calls inside ifs that match the candidate condition.

Structural equality is useful for proving that two expressions have the same
shape, but it does not prove that they read the same runtime values. In
particular, structurally similar `BufferLoad` expressions over `P` and `Q` can
match even though the underlying buffer data Vars are distinct.

## Fix

Use the existing `ExprDeepEqual` predicate for LoopUnswitching condition matches
instead of the looser structural equality predicate:

- update `CallCheckerExcludingIf` so calls are excluded only under truly matching
  guards;
- update `IfBranchReplacer` so only matching guards are collapsed after the
  condition is hoisted;
- keep the same predicate in both analysis and rewrite paths.

This keeps the pass aligned with other TileLang transforms such as
`MergeIfStmt`, which already uses `ExprDeepEqual` for same-condition matching,
and avoids adding a custom equality helper for this pass.

## Tested

Added regression coverage in
`testing/python/transform/test_tilelang_transform_loop_unswitching.py`:

- same-buffer rebuilt load conditions still collapse;
- same-shape guards on different buffers are not collapsed;
- same-buffer different-index guards are not collapsed;
- `CallCheckerExcludingIf` does not exclude calls under a different-buffer guard.

Also verified the CUDA repro from
https://github.com/tile-ai/tilelang/issues/2539 locally on an NVIDIA H200:

```text
expected (OB follows Q<2): [1, 0, 1, 0]
default build            : [1, 0, 1, 0]
bypass  build            : [1, 0, 1, 0]
```

Fixes https://github.com/tile-ai/tilelang/issues/2539

## Summary

Replaced LoopUnswitching's direct `StructuralEqual` condition matches with
`ExprDeepEqual`, so same-shape guards over different buffers are no longer
treated as the same predicate.

Kept the existing same-buffer optimization intact for conditions that still
compare equal under `ExprDeepEqual`.

Used the same predicate in both `CallCheckerExcludingIf` and `IfBranchReplacer`,
so analysis and rewrite decisions stay consistent.

## C++ style / lint notes

This PR touches TileLang-owned C++ transform code under `src/transform` and keeps
the change narrowly focused on behavior rather than broad style cleanup.

The implementation follows the nearby TileLang transform pattern by reusing
`ExprDeepEqual` instead of adding a new pass-local equality abstraction.

The C++ code was checked with the repository pre-commit hooks, including
`clang-format`.

## Summary

- Fixed `LoopUnswitching` predicate matching to be identity-sensitive: structurally similar guards over different buffers now remain distinct, while truly identical conditions (including same-buffer rebuilt loads) can still be merged.
- Applied the identity-sensitive condition equivalence consistently to both call exclusion (`CallCheckerExcludingIf`) and branch replacement (`IfBranchReplacer`).
- Added regression tests covering:
  - different buffers collapsing incorrectly (now prevented),
  - same-buffer rebuilt loads being safely replaced/hoisted,
  - different indices preventing collapsing,
  - call-exclusion behavior when guards differ by buffer.
- Verified CUDA output behavior via the new regression coverage.

## C++ style / lint notes

- The PR updates C++ implementation code (`src/transform/loop_unswitching.cc`) and adds Python regression tests, but does not modify documented C++ style rules in `docs/developer_guide/cpp_style.md`.
- No CI/lint configuration changes are indicated; the **C++ API Style Audit (warning only)** step should remain relevant. Any TLCPP003/TLCPP004 findings would be advisory unless this PR introduces an API/FFI/maintainability risk (none indicated by the changes).

### DIFF
--- a/src/transform/loop_unswitching.cc
+++ b/src/transform/loop_unswitching.cc
@@ -15,7 +15,6 @@
  */
 
 #include "support/check.h"
-#include <tvm/ffi/extra/structural_equal.h>
 #include <tvm/tirx/analysis.h>
 #include <tvm/tirx/builtin.h>
 #include <tvm/tirx/stmt.h>
@@ -169,7 +168,7 @@ public:
   void VisitStmt_(const IfThenElseNode *op) final {
     // Skip the interior of any if statement with matching condition
     if (excluded_condition.defined() &&
-        StructuralEqual()(op->condition, excluded_condition)) {
+        ExprDeepEqual()(op->condition, excluded_condition)) {
       return;
     }
     StmtExprVisitor::VisitStmt_(op);
@@ -415,8 +414,8 @@ public:
   }
 
   Stmt VisitStmt_(const IfThenElseNode *op) final {
-    // Replace if the condition is structurally equal to the hoisted condition
-    if (StructuralEqual()(op->condition, hoisted_condition)) {
+    // Replace if the condition matches the hoisted condition.
+    if (ExprDeepEqual()(op->condition, hoisted_condition)) {
       if (take_then) {
         return VisitStmt(op->then_case);
       } else {
--- a/testing/python/transform/test_tilelang_transform_loop_unswitching.py
+++ b/testing/python/transform/test_tilelang_transform_loop_unswitching.py
@@ -326,6 +326,99 @@ def test_multiple_identical_conditions():
     _check(before, expected)
 
 
+def test_same_buffer_rebuilt_load_conditions_are_replaced():
+    """Repeated loads from the same buffer/index are the same guard."""
+
+    @T.prim_func
+    def before(
+        A: T.Tensor((128,), T.float32),
+        B: T.Tensor((128,), T.float32),
+        C: T.Tensor((128,), T.float32),
+        P: T.Tensor((1,), T.int32),
+    ):
+        for i in range(128):
+            if P[0] < 4:
+                B[i] = A[i]
+            if P[0] < 4:
+                C[i] = A[i] * T.float32(2.0)
+
+    @T.prim_func
+    def expected(
+        A: T.Tensor((128,), T.float32),
+        B: T.Tensor((128,), T.float32),
+        C: T.Tensor((128,), T.float32),
+        P: T.Tensor((1,), T.int32),
+    ):
+        if P[0] < 4:
+            for i in range(128):
+                B[i] = A[i]
+                C[i] = A[i] * T.float32(2.0)
+        else:
+            for _i in range(128):
+                T.evaluate(0)
+                T.evaluate(0)
+
+    _check(before, expected)
+
+
+def test_same_shape_different_buffer_conditions_are_not_collapsed():
+    """Structurally similar guards on different buffers are distinct."""
+
+    @T.prim_func
+    def before(
+        A: T.Tensor((128,), T.float32),
+        B: T.Tensor((128,), T.float32),
+        C: T.Tensor((128,), T.float32),
+        P: T.Tensor((1,), T.int32),
+        Q: T.Tensor((1,), T.int32),
+    ):
+        for i in range(128):
+            if P[0] < 4:
+                B[i] = A[i]
+            if Q[0] < 4:
+                C[i] = A[i] * T.float32(2.0)
+
+    _check(before, before)
+
+
+def test_same_buffer_different_index_conditions_are_not_collapsed():
+    """Guards on the same buffer but different indices are distinct."""
+
+    @T.prim_func
+    def before(
+        A: T.Tensor((128,), T.float32),
+        B: T.Tensor((128,), T.float32),
+        C: T.Tensor((128,), T.float32),
+        P: T.Tensor((2,), T.int32),
+    ):
+        for i in range(128):
+            if P[0] < 4:
+                B[i] = A[i]
+            if P[1] < 4:
+                C[i] = A[i] * T.float32(2.0)
+
+    _check(before, before)
+
+
+def test_call_checker_does_not_exclude_different_buffer_guard():
+    """Calls under a sibling guard on another buffer still block unswitching."""
+
+    @T.prim_func
+    def before(
+        A: T.Tensor((128,), T.float32),
+        B: T.Tensor((128,), T.float32),
+        P: T.Tensor((1,), T.int32),
+        Q: T.Tensor((1,), T.int32),
+    ):
+        for i in range(128):
+            if P[0] < 4:
+                B[i] = A[i]
+            if Q[0] < 4:
+                T.evaluate(T.call_extern("handle", "generic_op"))
+
+    _check(before, before)
+
+
 def test_multiple_identical_conditions_with_else():
     """Conservative: multiple if-else statements should NOT be hoisted."""
 


LeiWang1999:
@regression-perf

zyy3077:
@muhamedfazalps Yes, this is intentional. `CallCheckerExcludingIf` should only skip calls nested under an `if` whose condition matches the hoisted condition itself. With `ExprDeepEqual`, `P[bx] < 2` and `Q[bx] < 2` are no longer treated as the same guard, so a `T.copy` under the `P` guard remains visible to the call-safety check when hoisting `Q`.

This is the conservative behavior we want: the `P`-guarded copy is not part of the region specialized by the `Q` condition, so it should not be excluded merely because the two predicates have the same shape.

PR #2611, opened 2026-07-11
[BugFix] Reject non-power-of-two AllReduce widths

### LINKED ISSUE #2542: [BUG][Fuzzer][wrong-code] `T.reduce_sum`/`T.reduce_max` silently drop lanes when the reduce-dimension thread count isn't a power of two
### Required prerequisites

0.1.11

### System information

NVIDIA H100 (sm_90) and A10G (sm_86), CUDA 13.0, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). The defect is in the JIT-included `tl::AllReduce` CUDA template, so it is target-independent.

### Problem description

A block reduction is silently wrong whenever the number of threads spanning the reduce dimension is not a power of two. The reduction is an XOR-butterfly, correct only over a power-of-two width, but the only compile-time guard is [`static_assert(threads % scale == 0)`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/reduce.h#L224) — divisibility, not the power-of-two property the butterfly needs. For a 48-thread reduce, only the first 32 lanes are combined and the other 16 are dropped, with no error: `reduce_sum` of `arange(1..48)` returns `528` (= `1+…+32`) instead of `1176`. Powers of two (32, 64, 128) are correct.

### Reproducible example code

```python
import torch
import tilelang
import tilelang.language as T

W = 48  # non-power-of-two: silently wrong. 32 / 64 / 128: correct.


@T.prim_func
def main(A: T.Tensor((1, W), "float32"), B: T.Tensor((1,), "float32")):
    with T.Kernel(1, threads=W) as _:
        a = T.alloc_fragment((1, W), "float32")
        r = T.alloc_fragment((1,), "float32")
        T.copy(A, a)
        T.reduce_sum(a, r, dim=1)
        T.copy(r, B)


k = tilelang.compile(main, out_idx=[1], target="cuda")
A = torch.arange(1, W + 1, device="cuda", dtype=torch.float32).reshape(1, W)
print(k(A)[0].item(), "expected", A.sum().item())   # 528.0 expected 1176.0
```

### Traceback

```pytb
No traceback — the headline case is silent: the kernel compiles and runs to completion and the wrong result is deterministic. (The two loud faces in the collapsed section below do raise — an `nvcc` `static_assert` at compile time, or `CUDA error: an illegal memory access` at runtime.)
```

### Expected behavior

The reduction should produce the full, correct result — or be rejected at compile time — rather than silently combining only the first power-of-two lanes.

### Additional context

**Root cause.** The reduce lowers to `tl::AllReduce<Reducer, threads, scale, 0>`, an [XOR-butterfly](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/reduce.h#L246-L266): each step combines lane `t` with `t ^ (threads/2)`, then recurses on `threads/2` until `offset == scale`. This is a correct all-reduce only for a power-of-two width — for `W=48` the halving sequence `24,12,6,3,1` never folds in lanes 32–47. The sole guard [`static_assert(threads % scale == 0)`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/reduce.h#L224) checks divisibility; for the common `scale == 1` it holds for every width, so any non-pow2 width slips through.

**Suggested fix.** Strengthen the guard so a malformed tree can't be emitted silently:

```diff
   // src/tl_templates/cuda/reduce.h (line 224)
   static_assert(threads % scale == 0);
+  static_assert(((threads / scale) & (threads / scale - 1)) == 0,
+                "AllReduce reduce width (threads / scale) must be a power of two");
```

A more complete fix reduces the tail lanes naively (or pads the launch to the next power of two with identity lanes) so non-pow2 widths compute correctly rather than being rejected. Note the pad-to-pow2 route changes `reducing_threads`, which is also the batched workspace stride, so it would need to be coordinated with the `batch > 1` workspace sizing in #2524; the `static_assert` route above is independent.

<details>
<summary>Same root cause, two loud faces (compile abort / runtime crash)</summary>

Which face appears depends on where the non-pow2 halving lands:

- **Compile abort** (`dim=0` reduce, `scale ≥ 2`, even-non-pow2 rows): `R=6` descends `6 → 3`, and `3 % 2 != 0` trips the `static_assert` at instantiation. `R=8` compiles and runs correctly.
- **Runtime crash** (non-pow2 `W ∈ (64,128)`, e.g. `W=96`): the first step `offset=48 ≥ 32` takes the shared-memory branch `red_buf[tid ^ 48]`, which for `tid ∈ [64,96)` reaches out of the workspace → `CUDA error: an illegal memory access`. `W=64`/`128` are correct.

All three vanish once the width is a power of two, and all three are prevented by the guard above.
</details>

**Provenance.** The divisibility-only guard has been present since the initial public migration of the codebase ([`57ab687`](https://github.com/tile-ai/tilelang/commit/57ab687cf7c97f040386c4d473d325b4a1731914), #10, 2025-01-11) and survived the later batched-`AllReduce` rework.

**Not a duplicate.** Distinct from #2524 (batch>1 workspace aliasing), #2294/#2305 (reduce_absmax *intra-thread* layout), and the partial-warp `tl.ThreadSync` barrier race (the `W=48` kernel has no `__syncthreads()` and is deterministic). No existing tracker report covers the non-pow2 width defect.

zyy3077:
## Issue

https://github.com/tile-ai/tilelang/issues/2542

## Summary

XOR-butterfly AllReduce assumed that the logical reduction width was a power of
two. For non-power-of-two widths such as `W=48`, lowering could still emit
`tl::AllReduce<..., 48, 1>`, which silently produced the wrong result at runtime.

## Example

```python
W = 48

@T.prim_func
def main(A: T.Tensor((1, W), "float32"), B: T.Tensor((1,), "float32")):
    with T.Kernel(1, threads=W):
        a = T.alloc_fragment((1, W), "float32")
        r = T.alloc_fragment((1,), "float32")
        T.copy(A, a)
        T.reduce_sum(a, r, dim=1)
        T.copy(r, B)
```

Before this fix, the kernel could compile and return `528.0` for
`torch.arange(1, 49).sum()`, whose expected value is `1176.0`.

## Root cause

The CUDA runtime template only checked that the physical AllReduce thread span
was divisible by the thread stride:

```cpp
static_assert(threads % scale == 0);
```

This proves that `threads / scale` is integral, but it does not prove that the
logical reduction width is valid for a recursive XOR-butterfly reduction.

The butterfly recursion halves the logical participant set at each step, so a
non-power-of-two logical width can leave participants paired incorrectly and
produce silent wrong results.

## Fix

Strengthen the AllReduce guards around the existing XOR-butterfly
implementation:

- define the logical width as `threads / scale` at the lowering call sites;
- reject non-positive or non-power-of-two logical widths before emitting the
  `tl::AllReduce<...>` template call;
- keep the original `threads % scale == 0` invariant;
- add a CUDA template `static_assert` defense-in-depth matching the issue's
  suggested fix.

This PR intentionally does not add arbitrary-width padding. Supporting that
correctly would need a broader implementation covering reducer identities,
workspace stride, batch reduction, and shuffle masks.

## Tested

Added regression coverage in
`testing/python/language/test_tilelang_language_reduce.py`:

- `48` and `96` logical widths are rejected for `reduce_sum` and `reduce_max`;
- `32`, `64`, and `128` logical widths still run correctly for `reduce_sum`;
- `32`, `64`, and `128` logical widths still run correctly for `reduce_max`;
- `scale > 1` valid combinations still run correctly;
- `scale > 1` with logical width `48` is rejected before runtime.

Fixes https://github.com/tile-ai/tilelang/issues/2542

## Summary

Added early lowering diagnostics for malformed XOR-butterfly AllReduce widths
and a CUDA template `static_assert` fallback so non-power-of-two logical reduce
widths are rejected instead of silently miscomputed.

Kept valid power-of-two paths intact, including `scale > 1` cases where the
logical reduction width is `threads / scale`.

## C++ style / lint notes

This PR touches TileLang-owned common GPU lowering code and a CUDA runtime
template. The change is intentionally local to the AllReduce emission sites and
uses existing `ICHECK` diagnostics plus TVM's existing
`tirx::is_const_power_of_two_integer` helper.

No new common helper/header was added; the checks stay next to the code that
emits the templated AllReduce call.

## Summary

- Fixes CUDA AllReduce correctness by rejecting configurations where the **logical reduction width** (`threads / scale`) is non-positive or not a power of two, while still enforcing `threads % scale == 0`.
- Centralizes validation in lowering via a new `reduce::CheckAllReduceWidth` helper (`src/backend/common/op/reduce.h`) and wires it into `FinalizeReducerLowerer::Lower` and `ReduceLowerer::Lower` call sites to fail fast instead of silently producing incorrect results.
- Adds/strengthens CUDA-side template `static_assert` checks in `tl::AllReduce` (including `scale > 1` support) so invalid `threads / scale` configurations are caught at compile time.
- Adds regression tests ensuring invalid logical widths (e.g., `48`, `96`) are rejected, valid power-of-two widths (`32`, `64`, `128`) match PyTorch runtime results for `reduce_sum`/`reduce_max`, and `scale > 1` paths are correct or rejected when the logical width is not a power of two.

## C++ style / lint notes

- The PR touches C++ implementation code but does not modify the rules documented in `docs/developer_guide/cpp_style.md`.
- No new or specifically addressed warning-only style/lint issues are introduced; the **C++ API Style Audit (warning only)** CI step remains advisory with no correctness/build impact indicated by this change.


SiriusNEO (approved):

SiriusNEO:
@regression-perf

PR #2636, opened 2026-07-12
[BugFix] Respect safe value in tile copy OOB

### LINKED ISSUE #2543: [BUG][Fuzzer][wrong-code] Whole-tile `T.copy` of an out-of-bounds tile silently ignores `T.annotate_safe_value` and zero-fills the out-of-bounds region instead of using the annotated value
### Required prerequisites

0.1.11

### System information

NVIDIA H100 (sm_90), CUDA 13.0, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). The defect is in the tile-copy SIMT lowering (`src/op/copy.cc`), which is target-independent; reproduced on sm_90.

### Problem description

A whole-tile `T.copy` of an out-of-bounds global tile into shared memory silently ignores `T.annotate_safe_value` and always zero-fills the out-of-bounds region. Copying a `(60, 60)` `int32` tensor into a `(64, 64)` tile annotated with safe value `-99` produces `0` (not `-99`) in the out-of-bounds rows/cols `[60..63]`; the string `-99` never appears in the generated CUDA. The in-bounds region `[0:60, 0:60]` is correct, the kernel compiles and runs with no error, and the wrong fill is deterministic. Any code relying on the padding (e.g. a masked softmax over a padded tile) then consumes `0` where it asked for `-99`.

The value *is* a real, honored feature on the elementwise path: a byte-identical per-element `T.Parallel` load of the same tile with the same annotation returns `-99`. Only the whole-tile `T.copy` lowering drops it.

### Reproducible example code

```python
import tilelang
tilelang.disable_cache()
import tilelang.language as T
import torch

M = N = 60      # global 60x60
BM = BN = 64    # tile 64x64 -> rows/cols [60..63] are out of bounds


@T.prim_func
def bulk(A: T.Tensor((M, N), "int32"), Out: T.Tensor((BM, BN), "int32")):
    with T.Kernel(1, threads=128) as _:
        As = T.alloc_shared((BM, BN), "int32")
        T.annotate_safe_value({A: -99})          # request -99 as the OOB fill
        T.copy(A[0:BM, 0:BN], As)                # whole-tile copy
        for i, j in T.Parallel(BM, BN):
            Out[i, j] = As[i, j]


k = tilelang.compile(bulk, out_idx=[1], target="cuda")
A = torch.arange(M * N, dtype=torch.int32, device="cuda").reshape(M, N)
Out = k(A)
print("'-99' in CUDA:", "-99" in k.get_kernel_source())   # False
print("OOB row 60 [0:8]:", Out[60, :8].cpu().tolist())     # [0,0,0,0,0,0,0,0]  (want -99)
print("inbounds correct:", torch.equal(Out[:60, :60].cpu(), A.cpu()))  # True
```

Replacing the `T.copy` with a per-element `for i, j in T.Parallel: As[i,j] = A[i,j]` load — same annotation — instead prints `'-99' in CUDA: True` and `[-99, -99, …]`, which is the control isolating the defect to the whole-tile lowering.

### Traceback

```pytb
No traceback — the kernel compiles and runs to completion; the OOB fill is silently `0` instead of the annotated `-99`.
```

### Expected behavior

A whole-tile `T.copy` of an OOB tile should fill the out-of-bounds region with the value passed to `T.annotate_safe_value` (produce `-99`), matching the elementwise load of the same tile with the same annotation. The two paths honoring the same annotation differently is an internal inconsistency, not a design choice.

### Additional context

**Root cause.** The OOB fill of a tile copy is emitted by `CopyNode::MakeSIMTLoop` ([`src/op/copy.cc#L452`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/copy.cc#L452)), which builds the guarded load with a hardcoded zero on the false branch ([`#L485`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/copy.cc#L485)):

```cpp
value = if_then_else(src_predicate, value, make_zero(dst->dtype));
```

This runs inside `LowerTileOp`, *before* `LegalizeSafeMemoryAccess`. So the load arrives at the legalization pass already guarded with `make_zero`, and the pass's safe-value machinery — which reads `attr::kSafeValueMap` and applies `GetSafeValue` to still-unguarded loads ([`legalize_safe_memory_access.cc#L631`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/legalize_safe_memory_access.cc#L631), [`#L356`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/legalize_safe_memory_access.cc#L356)) — never sees this fill and never consults the annotated value. The elementwise path stays unguarded through `LowerTileOp`, so the legalization pass reaches it and honors the annotation — exactly the asymmetry the control shows.

**Suggested fix.** Thread the annotated safe value into `MakeSIMTLoop`'s false branch instead of hardcoding `make_zero`: look up the source buffer's safe value (defaulting to `make_zero`, mirroring `GetSafeValue`) where the copy is lowered, and use it at [`copy.cc#L485`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/copy.cc#L485). This requires making `attr::kSafeValueMap` available to the copy lowering (it is on the block `LowerTileOp` processes). A local patch of this shape flips the repro to `-99` in the OOB region with the in-bounds region and the no-annotation zero-fill default both unchanged.

**Provenance.** The safe-value mechanism was added to `LegalizeSafeMemoryAccess` in [`72111642a`](https://github.com/tile-ai/tilelang/commit/72111642a) (#1050, 2025-10-17); the whole-tile copy path's `make_zero` fill in `MakeSIMTLoop` predates and was never wired to it.

**Not a duplicate.** Distinct from the TMA-copy *crash* reports (`cudaErrorIllegalInstruction` / launch abort / compile abort from TMA alignment/stride/swizzle constraints) — this is a silent wrong result on a clean-compiling kernel. The wrong fill comes from the SIMT copy lowering's `make_zero`, not from the TMA descriptor's `oob_fill` (flipping that descriptor field is inert for this OOB copy, since an OOB tile suppresses bulk/TMA selection). No open or closed tracker issue covers the whole-tile copy dropping `annotate_safe_value`.

zyy3077:
## Issue

https://github.com/tile-ai/tilelang/issues/2543

## Summary

Whole-tile `T.copy` lowered source out-of-bounds lanes with a hard-coded zero
fallback. This ignored `T.annotate_safe_value`, even though ordinary memory
access legalization already honors the same annotation.

## Example

```python
M = N = 60
BM = BN = 64

@T.prim_func
def bulk(A: T.Tensor((M, N), "int32"), Out: T.Tensor((BM, BN), "int32")):
        As = T.alloc_shared((BM, BN), "int32")
        T.annotate_safe_value({A: -99})
        T.copy(A[0:BM, 0:BN], As)
        for i, j in T.Parallel(BM, BN):
            Out[i, j] = As[i, j]
```

Before this fix, rows and columns `[60..63]` were filled with zero. After this
fix, they are filled with `-99`, while the in-bounds region remains unchanged.

## Root cause

`CopyNode::MakeSIMTLoop` generated the source predicate false branch directly:

```cpp
if_then_else(src_predicate, value, make_zero(dst->dtype))
```

This happens during `LowerTileOp`, before `LegalizeSafeMemoryAccess` can rewrite
ordinary OOB loads. As a result, the later pass never had a chance to apply the
annotated safe value for whole-tile copies.

## Fix

Resolve the source OOB safe value while parsing the tile op from its enclosing
`SBlock` annotations:

- `LowerTileOpPass` preserves the current lexical `SBlock` annotations and
  overlays the scoped `attr::kSafeValueMap` keyed by buffer data `Var`;
- `ParseOperator(op, block_annotations)` keeps call-local annotations separate
  and applies registered block-annotation handlers after the tile op is built;
- `copy`, `async_copy`, and `tma_copy` register the same handler, which resolves
  the source buffer safe value and stores it on `CopyNode`;
- `CopyNode::MakeSIMTLoop` uses the resolved value for the `src_predicate` false
  branch, falls back to source-dtype zero when no annotation exists, and casts
  the fallback to the destination dtype.

This keeps block annotation handling generic while keeping Copy-specific safe
value resolution in the Copy operator implementation.

## Tested

Added regression coverage in
`testing/python/language/test_tilelang_language_copy.py`:

- annotated source safe value `-99` is used for whole-tile copy OOB lanes;
- no annotation keeps the zero-fill default;
- row, column, and corner OOB regions are checked;
- source safe value is cast to the destination dtype;
- elementwise `T.copy` safe-value behavior remains unchanged.

Fixes https://github.com/tile-ai/tilelang/issues/2543

## C++ style / lint notes

This PR touches TileLang-owned C++ lowering and tile-op parsing code under
`src/op` and `src/transform`, plus focused Python regression tests.

The implementation keeps call-local tile-op annotations separate from enclosing
`SBlock` annotations and uses a small registered handler instead of adding
Copy-specific logic to the generic parser.

The C++ code was checked with the repository pre-commit hooks, including
`clang-format`.

## Summary

- Fixes whole-tile `T.copy` out-of-bounds handling to respect `T.annotate_safe_value` by propagating the per-block safe-value map into copy lowering.
- Uses the annotated source out-of-bounds safe value when available; otherwise falls back to a source-dtype zero and casts to the destination dtype as needed.
- Threads block annotations through tile-operator parsing and lowering with scoped restoration for nested blocks, updating parsing/lowering wiring so copy ops can consume `attr::kSafeValueMap`.
- Adds regression coverage for annotated/unannotated whole-tile OOB copies (row/column/ccorner regions), destination dtype casting, lexical scoping behavior, and unchanged in-bounds elementwise copy behavior.

## C++ style / lint notes

- The PR modifies C++ core code (`src/op/*` and `src/transform/lower_tile_op.cc`) but does not touch `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step may emit advisory TLCPP003/TLCPP004-style findings due to expanded public operator/block-annotation parsing interfaces and the new `CopyNode` field; treat any such warnings as non-blocking unless they indicate a concrete API/FFI/maintainability risk beyond style.


LeiWang1999:
Introducing a callback to get the safe memory value is too heavy, I think we can just add a map member to the lower args.

LeiWang1999 (approved):

PR #2639, opened 2026-07-12
[BugFix] Fix metal stream bridge

penguin-wwy:
## Summary

- Fixed the Metal/MPS stream bridge by disabling TVM’s Torch C DLPack integration on ROCm and MPS.
- Added MPS-specific handling to skip the DLPack C exchange API due to PyTorch stream-handle incompatibility.
- Updated lazy library loading to use the new helper.

### DIFF
--- a/tilelang/__init__.py
+++ b/tilelang/__init__.py
@@ -106,11 +106,17 @@ if not env.is_light_import():
 del _init_logger
 
 
-def _disable_rocm_tvm_ffi_torch_c_dlpack(torch_module):
-    if getattr(torch_module.version, "hip", None) is None:
+def _disable_tvm_ffi_torch_c_dlpack(torch_module):
+    is_rocm = getattr(torch_module.version, "hip", None) is not None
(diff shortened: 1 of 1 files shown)


penguin-wwy:
@LeiWang1999 fix ci metal

oraluben:
Would disabling tvm dlpack extension have performance issue?

oraluben (approved):
The current workaround works, but it disables the C DLPack exchange path globally and uses the slower Python fallback. In testing, this added roughly 3x more FFI call overhead and also disabled EnvTensorAllocator.

A better solution is to explicitly force-load tvm-ffi’s optional Torch C DLPack extension on MPS via  `_optional_torch_c_dlpack.load_torch_c_dlpack_extension()`, bypassing Torch 2.13’s broken native MPS exchange API. This preserves the fast path, supports JIT compilation when no prebuilt extension is installed, and should fall back to the current workaround only if extension loading fails.

But we can land this first to fix ci. @LeiWang1999

PR #2646, opened 2026-07-13
[BugFix] Fix 1D bulk TMA transfer alignment check

### LINKED ISSUE #2525: [BUG][Fuzzer][wrong-code] 1-D bulk TMA store of a shared→global `T.copy` whose byte size isn't a multiple of 16 silently drops the unaligned tail instead of copying the whole buffer
### Required prerequisites

0.1.11

### System information

- TileLang `0.1.11`; source pinned to tag `v0.1.11` (commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`).
- Python 3.12.3, linux; PyTorch 2.12.1+cu130, CUDA 13.0.
- **Reproduced at runtime on an NVIDIA H100 80GB HBM3 (compute capability 9.0, sm_90).** This is a Hopper-specific defect: it lives in the **1-D bulk TMA store** (`cp.async.bulk.global.shared::cta`) lowering, which is only selected on sm_90.

### Problem description

A contiguous 1-D `shared→global` `T.copy` whose total size in bytes is not a multiple of 16 **silently drops its last `total_bytes mod 16` bytes** — the kernel compiles and runs with no error and returns wrong numbers, deterministically. For a length-`N` float32 buffer that is the last `N mod 4` elements (e.g. `N=63` returns 60 correct elements and 3 zeros).

The cause is in instruction selection, not in the copy semantics: this copy is admitted to the 1-D bulk-TMA path and lowered to a `cp.async.bulk.global.shared::cta` whose size operand is the raw byte count with **no 16-byte rounding**, so the hardware transfers only `floor(size/16)*16` bytes and drops the rest. The eligibility gate `CheckBulkStore1D` ([`copy_analysis.cc#L322-L330`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L322-L330)) is missing the 16-byte-multiple check that would keep an unaligned-size copy off the bulk-TMA path — the multi-dim path already has that check and, as the 2-D control below shows, correctly falls back to the normal copy (which handles any size). So the copy is well-formed and supported; it is simply mis-routed to an instruction that cannot express its size.

<details>
<summary>Discriminating control — pack-aligned <code>N</code>, identical op/dtype/path, is exactly correct (H100, TESTED)</summary>

The single difference between a wrong run and a correct run is whether the total transfer size is a whole multiple of 16 bytes. Same bare `T.copy` round-trip, same float32 dtype, same lowering path:

| `N` (float32) | total bytes `N*4` | `% 16` | result |
|---------------|-------------------|--------|--------|
| 60 | 240 | 0 | **correct** |
| 63 | 252 | 12 | drops 3 tail elements |
| 64 | 256 | 0 | **correct** |
| 65 | 260 | 4 | drops 1 tail element |
| 66 | 264 | 8 | drops 2 tail elements |

The pack-aligned cases are exact, which isolates the defect to the unaligned-tail handling of the 1-D bulk TMA store — not to any particular op and not to the test harness.

**2-D control (proves it's a routing bug, not a size limitation):** the same unaligned last dimension in a 2-D copy `T.copy` of shape `(4, N)` takes the **non-TMA** path (no `tma_store` emitted) and is **exactly correct** for `N = 63` and `N = 127` (H100, TESTED). TileLang handles the unaligned size fine on the normal path; only the 1-D bulk-TMA gate mis-routes it.
</details>

**Root cause:** the eligibility gate `CheckBulkStore1D` calls `CheckBulkStore(..., /*check_last_dim=*/false, ...)`, which disables the only 16-byte-multiple check the multi-dim path has, so an unaligned-size copy is admitted to the 1-D bulk path and `LowerBulk1D` emits the raw byte count as a truncating `cp.async.bulk` size.

<details>
<summary>Full root-cause derivation (the eligibility gate, not the copy itself, admits the truncating store)</summary>

The corruption is entirely in the `shared→global` copy, which is lowered to a 1-D bulk TMA store. In `LowerBulk1D` ([`copy.cc#L2038-L2049`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L2038-L2049)) the size operand is `total_bytes = TMATransactionBytesFromElements(elements, dtype)` ([`copy.cc#L86-L88`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L86-L88) — a raw byte count, no alignment), emitted verbatim as the `tma_store` size with no rounding or check. `tma_store` is `cp.async.bulk.global.shared::cta` ([`copy_sm90.h#L411-L419`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/copy_sm90.h#L411-L419)), whose size operand must be a multiple of 16; the hardware uses only `floor(size/16)*16` bytes and silently drops the rest.

The root cause is the **eligibility gate**. `CheckBulkStore1D` ([`copy_analysis.cc#L322-L330`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L322-L330)) calls `CheckBulkStore(op, target, analyzer, /*check_last_dim=*/false, …)`, explicitly **disabling** the only 16-byte-multiple check — the `check_last_dim`-gated `TMAGlobalBitsFromElements(last_dim) % 128 == 0` guard at [`copy_analysis.cc#L230-L266`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L230-L266) (lines 240-255). It then calls `CheckBulkCopy1D` ([`copy_analysis.cc#L268-L310`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L268-L310)), which verifies only (a) shared/global contiguity and (b) `shared_elements == global_elements`, and **never** checks that `shared_elements * dtype.bytes` is a multiple of 16. So for `N=63` fp32 (252 bytes) `CheckBulkStore1D` returns `true`, `SelectTmaInst` ([`copy_analysis.cc#L505-L522`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L505-L522)) prefers `kBulkStore1D`, and `LowerBulk1D` emits a truncating `cp.async.bulk`.

**Why `check_last_dim=false` here is reasonable — and what's actually missing.** Disabling that guard for the 1-D path is not simply careless. The `check_last_dim` guard tests the **last-dim extent** (`op.dst_range[last]->extent * bits % 128 == 0`) — it is written for the **multi-dim, descriptor-based** TMA, where the innermost dimension is the contiguous row and the 16-byte constraint applies specifically to *that* dimension. The 1-D path uses a **different instruction** (`cp.async.bulk.global.shared::cta`, non-descriptor), whose constraint is on the **total transfer size**, not on a "last-dim extent". So the multi-dim last-dim test is genuinely the wrong check to run for 1-D, and turning it off is defensible. The bug is that, having turned off the inapplicable multi-dim check, the 1-D path never adds the check it *does* need — `total_bytes % 16 == 0` in `CheckBulkCopy1D`. It's "correctly skipped the wrong guard, but forgot to add the right one," not "removed a guard that applied."

The load side has the symmetric defect: `CheckBulkLoad1D` ([`copy_analysis.cc#L312-L320`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L312-L320)) also passes `check_last_dim=false` and routes through the same `CheckBulkCopy1D`, so a global→shared 1-D bulk load can under-fill shared memory by the same unaligned tail.
</details>

### Reproducible example code

Run on an sm_90 (Hopper) GPU so the 1-D bulk TMA store path is selected.

A `global→shared→global` `T.copy` round-trip. `float32` (4 B, pack width 4), so `N mod 4` elements are dropped.

```python
import tilelang
import tilelang.language as T
import torch


def build(N, dtype="float32"):
    @T.prim_func
    def kernel(A: T.Tensor((N,), dtype), B: T.Tensor((N,), dtype)):
            sh = T.alloc_shared((N,), dtype)
            T.copy(A, sh)            # global -> shared
            T.copy(sh, B)            # shared -> global  -> 1-D bulk TMA store (the defect)
    return kernel


def run(N, dtype="float32"):
    jit = tilelang.compile(build(N, dtype), target="cuda")
    a = torch.arange(1, N + 1, dtype=torch.float32, device="cuda")
    b = torch.empty_like(a)
    jit(a, b)
    torch.cuda.synchronize()
    got, ref = b.cpu().tolist(), a.cpu().tolist()
    tma = next((l.strip() for l in jit.get_kernel_source().splitlines() if "tma_store" in l), None)
    return got == ref, got[-3:], ref[-3:], tma


# Defect: total byte size not a multiple of 16 drops the tail (fp32 pack width = 4).
for N in [63]:
    ok, gt, rt, tma = run(N)
    print(f"N={N} MATCH={ok}  got_tail={gt}  ref_tail={rt}")
    print(f"   emit: {tma}")
# -> N=63 MATCH=False  got_tail=[0.0, 0.0, 0.0]  ref_tail=[61.0, 62.0, 63.0]
# ->    emit: tl::tma_store((&(B[0])), (&(sh[0])), 252);

# Control: N a clean multiple of the pack width transfers everything (same op/dtype/path).
for N in [60, 64]:
    ok, gt, rt, tma = run(N)
    print(f"N={N} MATCH={ok}  (control: pack-aligned)  emit: {tma}")
# -> N=60 MATCH=True   emit: tl::tma_store((&(B[0])), (&(sh[0])), 240);
# -> N=64 MATCH=True   emit: tl::tma_store((&(B[0])), (&(sh[0])), 256);
```

For `N=63` fp32, `252 = 63 * 4` bytes is passed verbatim to `cp.async.bulk`, which truncates to `floor(252/16)*16 = 240` bytes = 60 floats, dropping the last 12 bytes = 3 floats (indices 60, 61, 62) — exactly `63 mod 4 = 3` elements. The pack-aligned controls (`N=60→240`, `N=64→256`) transfer everything, isolating the defect to the unaligned total size.

### Traceback

```pytb
```

### Expected behavior

A contiguous shared→global (or global→shared) `T.copy` should transfer **all** elements, regardless of whether the total byte size is a multiple of 16 — i.e. it should produce the full `N`-element result. Grounded in TileLang's own behavior:

- The **pack-aligned** cases of the identical program (`N` a multiple of `16/sizeof(T)`) are exact (see the control), so the 1-D bulk path is meant to be a faithful copy; the unaligned tail should be handled by falling back to the multi-dim bulk path or the normal vectorized/scalar copy, which already handle unaligned tails.
- **A 2-D copy with the same unaligned last dimension already does the right thing:** it takes the non-TMA path (no `tma_store` emitted) and transfers all elements correctly (see the 2-D control below). So TileLang fully supports unaligned-size copies — the fix is not to reject them but to route the 1-D unaligned case to that same working non-TMA path.
- The 16-byte-multiple guard **already exists** in the project — `CheckBulkStore` enforces exactly this `% 128 bits == 0` test ([`copy_analysis.cc#L230-L266`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L230-L266)) for the multi-dim path; it was simply disabled (`check_last_dim=false`) for the 1-D path, so the 1-D path is the one place that neither aligns the transfer nor falls back.

### Additional context

**Suggested fix (proposed, not built or tested).** In `CheckBulkCopy1D` ([`copy_analysis.cc#L268-L310`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L268-L310)), after computing `shared_elements`, also require the total transfer size to be a multiple of 16 bytes before returning `true` — e.g. `analyzer->CanProveEqual(FloorMod(shared_elements * dtype.bits, 128), 0)` (equivalent to `total_bytes % 16 == 0`). Returning `false` otherwise lets `SelectTmaInst` ([`copy_analysis.cc#L505-L522`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L505-L522)) fall through to the multi-dim bulk path or the normal vectorized/scalar copy, which handle the unaligned tail correctly. Because both `CheckBulkStore1D` and `CheckBulkLoad1D` route through `CheckBulkCopy1D`, this also closes the symmetric BulkLoad1D under-fill. As a defensive secondary measure, `LowerBulk1D` ([`copy.cc#L2038-L2049`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L2038-L2049)) could `ICHECK` a statically-known `total_bytes % 16 == 0`, so any future caller passing an unaligned size gets a hard compile-time error instead of silent data loss.

**Provenance.** The 1-D bulk TMA path (`kBulkStore1D` / `CheckBulkCopy1D`, with the last-dim 16-byte check disabled via `check_last_dim=false`) was introduced in [#788](https://github.com/tile-ai/tilelang/pull/788) *"[TMA] Automatically lower 1d tma in appropriate cases"* (@LeiWang1999, merged 2025-09-06); the gate was later relocated by refactors [#2138](https://github.com/tile-ai/tilelang/pull/2138) and [#2297](https://github.com/tile-ai/tilelang/pull/2297) (namespace move) but the missing size guard rode along unchanged. Present on `0.1.11` (the version reproduced).

**Dedup.** I searched the open and closed `tile-ai/tilelang` tracker and found no existing report of the 1-D bulk TMA store silently dropping an unaligned tail of a contiguous shared↔global copy. The closest report is [#2506](https://github.com/tile-ai/tilelang/issues/2506) (open) *"`T.copy_cluster` emits a `cp.async.bulk` with a non-16B transfer size … hangs the kernel on H100"* — same *family* (an unguarded non-16B `cp.async.bulk` size), but a **different path, intrinsic, and symptom**: #2506 is the **cluster** copy `T.copy_cluster` → `tma_store_cluster` (`cp.async.bulk…cluster.mbarrier`) sized in `copy.cc`, and its UB manifests as a **device hang**; this bug is the **non-cluster 1-D bulk store** `T.copy` → `tma_store` (`cp.async.bulk.global.shared::cta`) gated in `copy_analysis.cc` (`CheckBulkStore1D`/`CheckBulkCopy1D`) and lowered in `LowerBulk1D`, and it produces a **silent tail-drop (wrong data, no hang)**.

UnsettingGalaxy:
## Summary

Fix the 1-D bulk TMA copy eligibility check to require the total transfer size to be a multiple of 16 bytes.

Unaligned transfers now fall back to the non-TMA copy path, while aligned transfers continue to use bulk TMA.

A defensive `ICHECK` was also added in `LowerBulk1D` to prevent future callers from emitting an invalid transfer size.

Fixes #2525

## Testing

* `N=63`, `float32`: falls back to the non-TMA path and produces the correct result
* `N=64`, `float32`: continues to use TMA and produces the correct result

## Summary
- Tightened 1-D global↔shared bulk TMA eligibility so it now additionally requires the transferred payload size to be provably 16-byte aligned when this can be determined.
- Added a defensive runtime check during 1-D bulk lowering to fatally reject invalid or unprovably aligned `total_bytes`, including the computed size in the error.

## C++ style / lint notes
- No documentation or public API changes; existing C++ style/lint audit steps remain applicable.

### DIFF
--- a/src/cuda/op/copy.cc
+++ b/src/cuda/op/copy.cc
@@ -2349,6 +2349,18 @@ Stmt Copy::LowerBulk1D(const CopyNode &op, const LowerArgs &lower_args,
   Stmt tma_copy;
   PrimExpr total_bytes =
       TMATransactionBytesFromElements(elements, shared_tensor->dtype);
+
+  // Safety net: instruction selection (CheckBulkCopy1D) must not route a
+  // provably misaligned transfer here. Symbolic sizes cannot be proven
+  // either way and are allowed through, matching the selection logic.
+  ICHECK(!analyzer->CanProve(
+      FloorMod(total_bytes, make_const(total_bytes.dtype(), 16)) !=
+          make_const(total_bytes.dtype(), 0),
+      arith::ProofStrength::kSymbolicBound))
+      << "BulkCopy1D requires total_bytes to be 16-byte aligned, but got "
+         "invalid size.\n"
+      << "  Total_bytes = " << total_bytes << "\n";
+
   if (is_load) {
     PrimExpr mbar_arg = barrier_base_id >= 0 ? mbar_handle : PrimExpr(0);
     tma_copy = Evaluate(Call(DataType::Handle(), tma_load(),
--- a/src/cuda/op/copy_analysis.cc
+++ b/src/cuda/op/copy_analysis.cc
@@ -300,13 +300,25 @@ bool CheckBulkCopy1D(const Buffer &global_tensor, const Buffer &shared_tensor,
   for (size_t i = 0; i < shared_range.size(); i++) {
     shared_elements *= shared_range[i]->extent;
   }
+
   PrimExpr global_elements = 1;
   for (size_t i = 0; i < global_range.size(); i++) {
     global_elements *= global_range[i]->extent;
   }
+
   bool element_match =
       analyzer->CanProveEqual(shared_elements, global_elements);
-  return shared_is_contiguous && global_is_contiguous && element_match;
+
+  PrimExpr total_bits = shared_elements * shared_tensor->dtype.bits();
+
+  // Reject only when the total transfer size is provably not 16-byte
+  // aligned (e.g. constant shapes like 63 x fp32). Symbolic shapes stay on
+  // the 1D TMA path, consistent with the last-dim check in CheckBulkLoad.
+  bool total_16b_aligned = !analyzer->CanProve(
+      FloorMod(total_bits, 128) != 0, arith::ProofStrength::kSymbolicBound);
+
+  return shared_is_contiguous && global_is_contiguous && element_match &&
+         total_16b_aligned;
 }
 
 bool CheckBulkLoad1D(const CopyNode &op, Target target,


LeiWang1999 (commented):
just wonder why is the alignment check of BulkCopy1D not carried out in LowerBulk1D

LeiWang1999:
@regression-perf

PR #2245, opened 2026-05-22
[Pipeline] Refactor software pipeline transforms

LeiWang1999:
## Summary

- Split shared software pipeline analysis, barrier, helper, and rewrite code into focused `src/transform/pipeline/` modules.
- Keep the public pipeline passes in `pipeline_planning.cc` and `inject_pipeline.cc`, while removing legacy raw PTX async planning paths.
- Move `IfStmtBinding` before pipeline planning so the pipeline passes operate on canonical high-level TileOP bodies.

## Changes

- Removed explicit `ptx_cp_async` / commit / wait scheduling support from pipeline planning and access analysis; pipeline planning now reasons over TileOP copies and generated async annotations.
- Consolidated `InjectSoftwarePipeline` implementation back into `inject_pipeline.cc` and deleted the temporary injector shim.
- Added reusable pipeline helpers for access analysis, body analysis, barrier handling, stage analysis, and rewriting.
- Tightened pipeline body invariants so inject expects the `SeqStmt` body produced by pipeline planning.
- Filtered loop-local buffer declarations out of pipeline scheduling annotations while still registering them for injection.
- Updated tests for high-level TMA / WGMMA / TCGEN05 planning cases and async producer annotations.

## Validation

- `pre-commit run --all-files`
- `git diff --check`
- `cmake --build build -j$(nproc)`
- `PYTHONPATH=$(pwd):$PYTHONPATH python -m pytest testing/python/transform/test_tilelang_transform_Inject_software_pipeline.py testing/python/transform/test_tilelang_transform_pipeline_planning.py testing/python/language/test_tilelang_language_tma_copy.py -q`

## Notes

- Local uncommitted changes in `examples/quickstart.py` and `docs/compiler_internals/async_mbarrier_dependency_analysis.md` were intentionally left out of this PR.

## Summary by CodeRabbit

* **New Features**
  * Full software-pipeline rewriter: buffer-versioning, replayed scalar-bind replay, and improved TMA copy synchronization.

* **Improvements**
  * Multi-version barrier expansion with parity-aware waits and TMA-aware wiring.
  * Richer per-stage analysis (reads/writes/scalar deps), replayability checks, and body normalization.
  * Configurable inlining of replayable binds in conditional binding pass.

* **Tests**
  * Expanded pipeline/TMA tests and new IfStmtBinding coverage; pipeline-planning tests now target-aware.

[...]

PR #2651, opened 2026-07-13
[BugFix][Transform] Never classify side-effecting binds as replayable (atomics re-executed at every use site since v0.1.11)

### LINKED ISSUE #2650: [Bug][Correctness] atomic_add(return_prev=True) bound to a scalar is re-executed at every use site since v0.1.11 — silent data corruption
Since v0.1.11 (through v0.1.12 and current main), a scalar bind whose value is a side-effecting call is dissolved by `tl.IfStmtBinding` and re-evaluated at every use site:

```python
pos = T.atomic_add(counter[0], 1, return_prev=True)
if pos < cap:          # use 1
    out[pos] = i       # use 2
```

lowers to CUDA that calls `AtomicAddRet` once per use site (4× total after `LegalizeSafeMemoryAccess` copies the inlined index into its bounds guards). Every kernel that uses the returned atomic value more than once is silently corrupted: the counter over-counts, interleaved slots are never written, and consumers read uninitialized memory. v0.1.8–v0.1.10 lower the same source correctly (single materialization). This is the usage shown in `atomic_add`'s docstring; nothing warns against it.

- **Repro:** ~20-line compaction kernel (above). With `n = cap = 4096` and all-positive input, the counter finishes at ~10–12k instead of 4096 and half the output slots stay unwritten. Grep the lowered source for `AtomicAddRet`: 4 occurrences instead of 1.
- **Real-world impact:** in a radix top-k kernel using this idiom (derived from the `deepseek_v32/topk_selector.py` example, which itself avoids the bug only because it routes atomic returns through `T.alloc_var`), per-row top-1024 recall drops from 1024/1024 to min 911 / mean ≈1002, selections differ across identical reruns, and garbage indices crash downstream gathers. The example's own test only prints recall and asserts nothing, so CI cannot catch this class of miscompile.
- **Root cause:** `IsReplayableScalarBind()` (`src/transform/common/bind_utils.h`, introduced in #2245) checks dtype and buffer read/write disjointness but never `SideEffect(bind->value)`. The atomic intrinsic is registered `kOpaque`, but its pointer argument enters via `tvm_access_ptr` and is recorded as a read, so the disjointness check passes and the bind is declared replayable. The same predicate also gates bind replay in `PipelinePlanning`/`InjectSoftwarePipeline`.
- **Suggested fix:** reject binds with `SideEffect(value) > kReadState` in `IsReplayableScalarBind` (mirrors the effect guard `StmtSimplifier` applies to let inlining). PR: <link>.
- **Workarounds:** pin `<=0.1.10`; pass config `tl.if_stmt_binding_inline_replayable_binds=False`; or bind through `T.alloc_var`.

---

zkyue:
### Summary

Since v0.1.11, a scalar bind whose value carries a side effect is dissolved by `tl.IfStmtBinding` and re-evaluated ("replayed") at **every use site**. The canonical victim is the documented usage of `atomic_add(..., return_prev=True)`:

```python
pos = T.atomic_add(counter[0], 1, return_prev=True)
if pos < cap:             # use 1: capacity guard
    out[pos] = i          # use 2: store index
```

The atomic executes once per use site instead of exactly once — silent data corruption for any compaction / histogram / top-k style kernel that uses the returned value more than once. `v0.1.8`–`v0.1.10` compile the same source correctly; the regression is present in `v0.1.11`, `v0.1.12` (PyPI latest) and current `main`.

### Reproduction

Minimal kernel (added as a regression test in this PR):

```python
@tilelang.jit
def compact_positive(n, cap, block, dtype="float32"):
    @T.prim_func
    def compact_kernel(
        x: T.Tensor[(n,), dtype],
        out: T.Tensor[(cap,), "int32"],
        counter: T.Tensor[(1,), "int32"],
    ):
        with T.Kernel(T.ceildiv(n, block), threads=block) as bx:
            tx = T.get_thread_binding()
            i = bx * block + tx
            if i < n and x[i] > 0:
                pos = T.atomic_add(counter[0], 1, return_prev=True)
                if pos < cap:
                    out[pos] = i
    return compact_kernel
```

v0.1.12 / main lower this to CUDA with the atomic inlined into **four** sites (the user's capacity guard, the store index, plus the two out-of-bounds guards that `tl.LegalizeSafeMemoryAccess` later builds by copying the already-inlined index expression):

```cuda
if (AtomicAddRet((&(counter[0])), 1) < 4096) {
  if (0 <= AtomicAddRet((&(counter[0])), 1)) {
    if (AtomicAddRet((&(counter[0])), 1) < 4096) {
      out[AtomicAddRet((&(counter[(int64_t)0])), (int64_t)1)] = i;
```

Functionally (n = cap = 4096, all-positive input; correct result is `counter == 4096` and `out` a permutation of `0..4095`):

| version | `AtomicAddRet` in lowered CUDA | functional |
|---|---|---|
| 0.1.8 | 1 (materialized once) | PASS |
| 0.1.9 | 1 | PASS |
| 0.1.10 | 1 | PASS |
| 0.1.11 | 4 (inlined per use site) | FAIL — counter ≈ 2.5–3×N, half the slots never written |
| 0.1.12 (PyPI latest) | 4 | FAIL |
| main @ 2047357e | 4 | FAIL |

In a larger radix-top-k kernel using this idiom (2048 rows × 12288 candidates, top-1024) the corruption is subtle enough to ship: per-row recall vs `torch.topk` drops from 1024/1024 to min 911 / mean ≈1002, the selected set changes across identical reruns (uninitialized shared memory is consumed), and out-of-range garbage indices appear that crash downstream gathers. We hit this in production code derived from the `deepseek_v32/topk_selector.py` example.

Note that the shipped `examples/deepseek_v32/topk_selector.py` itself escapes the bug only *accidentally*: it routes every atomic return value through `T.alloc_var` (a buffer store rather than a scalar bind — required anyway because the eager builder forbids rebinding). Nothing tests or documents that this indirection is load-bearing for correctness; the plain binding shown in `atomic_add`'s own docstring (`old_value = atomic_add(counter, 5, return_prev=True)`) is the form that miscompiles.

### Root cause

#2245 introduced `IsReplayableScalarBind()` (`src/transform/common/bind_utils.h`), used to decide that a scalar bind can be dissolved and its value re-evaluated at each use:

- `tl.IfStmtBinding` inlines "replayable" binds while pushing the `if` guard into each statement of the sequence (`BindIfStmtWithReplayableBindInlining`, config `tl.if_stmt_binding_inline_replayable_binds`, default true);
- `PipelinePlanning` / `InjectSoftwarePipeline` use the same predicate to pull "replayable" binds out of pipeline stages and replay them.

The predicate checks (a) that the bind is a non-pointer scalar (pointer exclusion added in #2278), and (b) that the value reads no buffer written in the same scope — but it **never consults the value's effect kind**. `tl.atomic_add_ret_elem_op` is correctly registered `kOpaque` in `src/op/builtin.cc`; however its buffer access enters the expression through `tvm_access_ptr`, which the access collector records as a *read*, so check (b) passes and the side-effecting bind is declared replayable.

`StmtSimplifier` already guards its own let-inlining correctly (`SideEffect(op->value) <= CallEffectKind::kPure`, `src/transform/simplify.cc`); the new predicate simply lacks the equivalent guard.

### Fix

One check in the shared predicate: a bind whose value has `SideEffect(...) > CallEffectKind::kReadState` (i.e. `kUpdateState`/`kOpaque`) is never replayable. This covers all three consumers (`IfStmtBinding`, `PipelinePlanning`, `InjectSoftwarePipeline`) at once. Read-only values remain replayable — replay-read consistency is exactly what the existing write-set disjointness check enforces.

### Tests

- `testing/python/transform/test_tilelang_transform_if_stmt_binding.py::test_if_stmt_binding_keeps_side_effecting_bind` — pass-level: `IfStmtBinding` must keep a bind whose value is an atomic RMW returning the previous value (structural equality).
- `testing/python/language/test_tilelang_language_atomic.py::test_atomic_add_return_prev_materialized_once` — end-to-end CUDA: exactly one `AtomicAddRet` in the generated source, `counter == N`, output is a permutation.
- `examples/deepseek_v32/topk_selector.py` — `test_topk_selector` computed recall but only printed it, so it could never fail on miscompilation; it now asserts `recall >= 0.999` (tie-tolerant; the plain-bind miscompile measures ≈0.98). This also protects the example's currently-accidental alloc_var immunity against future refactors.

Verification (B200 / sm_100a, CUDA 13.3):

- Both new tests fail on unpatched main and pass with the fix.
- Example test on the patched build: recall 1.0 on all 64 rows; kernel time unchanged (0.029 ms vs torch.topk 0.095 ms).
- Focused subset (if_stmt_binding, let_inline, simplify, Inject_software_pipeline, pipeline_planning, pipeline_barrier_ownership, language atomic/let/let_layout): unpatched baseline **99 passed / 3 skipped** → patched **101 passed / 3 skipped** (+2 = the new regression tests), no new failures.
- Full `testing/python/transform` + `testing/python/language`: **13 failed / 953 passed / 19 skipped**; the 13 failures are pre-existing, not caused by this change (the identical failure set reproduces on an unpatched build of the same base commit on the same machine: tcgen05 codegen x2, TMA gather/scatter x4, cooperative grid_sync, cutedsl PDL, reduce tolerance x3, clear matmul, hoist_broadcast dtype3). No new failures.
- Lint: `ruff check` / `ruff format --check` / `clang-format --dry-run -Werror` all clean on the changed files.

### Workarounds for users pinned to v0.1.11/v0.1.12

- pin `tilelang<=0.1.10`; or
- set pass config `tl.if_stmt_binding_inline_replayable_binds=False` (covers the `IfStmtBinding` path only); or
- route the returned value through a mutable var: `pos = T.alloc_var("int32"); pos[0] = T.atomic_add(..., return_prev=True)` (what the deepseek_v32 example happens to do).

---

Fixes #2650

## Summary

- Fixes the v0.1.11 regression by tightening `IsReplayableScalarBind()` so scalar binds are **not replayable** when their value has effects **stronger than `kReadState`**. This prevents incorrect re-execution of side-effecting operations (e.g., `T.atomic_add(..., return_prev=True)`) that could corrupt compaction, histogram, and top-k kernels.
- Extends the analysis infrastructure to correctly classify buffer accesses for pointer-based ops, including more precise handling of `tl::access_ptr`/`tvm_access_ptr` read/write regions (used by both `IfStmtBinding` and pipeline planning/injection).
- Updates `InjectSoftwarePipeline` and related pipeline planning logic to enforce the semantic replayability rules consistently, rejecting unsupported/unsafe replayability annotations during injection.

### Tests and example coverage

- Adds a CUDA-only end-to-end regression test ensuring **atomic return-value materialization happens exactly once**:
  - Verifies generated CUDA contains exactly one `AtomicAddRet`
  - Verifies counter correctness and that output indices are fully written/permuted.
- Adds structural regressions for `IfStmtBinding`:
  - Confirms side-effecting bound values aren’t duplicated across multiple uses/branches.
  - Includes coverage for binds that read atomic targets via pointer/access handling.
- Adds pipeline planning coverage:
  - Ensures binds that read atomic targets are preserved and do **not** result in `software_pipeline_replayable_scalar_binds` annotations.
- Adds an injection regression:
  - Ensures `InjectSoftwarePipeline()` fails when given a replay mask for a side-effecting scalar bind.
- Strengthens `examples/deepseek_v32/topk_selector.py` verification by enforcing **per-row `recall >= 0.999`** with improved diagnostics.

### Documentation / workarounds

- Documents the affected behavior and provides workarounds for impacted versions.

## C++ style / lint notes

- Touches C++ implementation logic only in transformation/analysis codepaths (`src/transform/common/bind_utils.h`, `src/transform/if_stmt_binding.cc`, `src/transform/pipeline_planning.cc`, `src/transform/inject_pipeline.cc`) and a small helper in `src/op/operator.h`; does **not** modify `docs/developer_guide/cpp_style.md`.
- No changes to CI/lint tooling are included. Any **`C++ API Style Audit (warning only)`** (e.g., TLCPP003/TLCPP004) findings should be treated as advisory unless the PR introduces a concrete API/FFI/maintainability risk; correctness/build implications are not expected from style-only results.

### DIFF
--- a/examples/deepseek_v32/topk_selector.py
+++ b/examples/deepseek_v32/topk_selector.py
@@ -212,6 +212,10 @@ def test_topk_selector(batch=64, seq_len=32 * 1024, topk=2048):
         set_trt = set(trt_np)
         intersection = set_ref & set_trt
         print("selected/all:", len(intersection), "/", len(set_ref), "=", len(intersection) / len(set_ref))
+        recall = len(intersection) / len(set_ref)
+        # Exact top-k selection modulo float ties at the threshold boundary.
+        # Miscompiled builds (duplicated atomic return value) drop to ~0.98.
+        assert recall >= 0.999, f"row {i}: top-{topk} recall too low: {recall:.4f}"
 
     # Performance test with CUDA events
 
--- a/examples/deepseek_v32/topk_selector.py
+++ b/examples/deepseek_v32/topk_selector.py
@@ -212,10 +212,6 @@ def test_topk_selector(batch=64, seq_len=32 * 1024, topk=2048):
         set_trt = set(trt_np)
         intersection = set_ref & set_trt
         print("selected/all:", len(intersection), "/", len(set_ref), "=", len(intersection) / len(set_ref))
-        recall = len(intersection) / len(set_ref)
-        # Exact top-k selection modulo float ties at the threshold boundary.
-        # Miscompiled builds (duplicated atomic return value) drop to ~0.98.
-        assert recall >= 0.999, f"row {i}: top-{topk} recall too low: {recall:.4f}"
 
     # Performance test with CUDA events
 
--- a/src/op/operator.h
+++ b/src/op/operator.h
@@ -48,6 +48,15 @@ enum AccessMask : int {
   kAccessReadWrite = kAccessRead | kAccessWrite,
 };
 
+// Preserve known read/write bits and conservatively assume both when a mask
+// cannot be resolved at compile time.
+inline int GetConservativeAccessMask(const PrimExpr &rw_mask) {
+  if (const auto *mask = rw_mask.as<IntImmNode>()) {
+    return static_cast<int>(mask->value) & kAccessReadWrite;
+  }
+  return kAccessReadWrite;
+}
+
 struct AccessRegion {
   tirx::BufferRegion region;
   int access_mask{kAccessReadWrite};
--- a/src/transform/common/bind_utils.h
+++ b/src/transform/common/bind_utils.h
@@ -6,6 +6,7 @@
 #define TVM_TL_TRANSFORM_COMMON_BIND_UTILS_H_
 
 #include <tvm/ir/type.h>
+#include <tvm/tirx/analysis.h>
 #include <tvm/tirx/stmt.h>
 
 #include <unordered_set>
@@ -30,6 +31,14 @@ inline bool IsReplayableScalarBind(const Stmt &stmt,
     return false;
   }
   if (!bind->value.dtype().is_scalar() || bind->value.dtype().is_handle()) {
+    return false;
+  }
+  // A bind value that carries a side effect (e.g. an atomic RMW returning the
+  // previous value, or any opaque call) must be materialized exactly once:
+  // replaying it at every use site re-executes the side effect and silently
+  // corrupts results. Pure and read-only values remain replayable; read
+  // consistency is guaranteed by the read/write disjointness check below.
+  if (SideEffect(bind->value) > CallEffectKind::kReadState) {
     return false;
   }
   for (const BufferRegion &read : reads) {
--- a/src/transform/if_stmt_binding.cc
+++ b/src/transform/if_stmt_binding.cc
@@ -59,6 +59,15 @@ private:
       writes_.push_back(BufferRegion::FullRegion(buffer));
     }
 
+    void AddAccess(const Buffer &buffer, int access_mask) {
+      if (access_mask & kAccessRead) {
+        AddRead(buffer);
+      }
+      if (access_mask & kAccessWrite) {
+        AddWrite(buffer);
+      }
+    }
+
     void VisitStmt_(const BufferStoreNode *op) final {
       AddWrite(op->buffer);
       StmtExprVisitor::VisitStmt_(op);
@@ -91,12 +100,29 @@ private:
             AddRead((*it).second);
           }
         }
+      } else if (op->op.same_as(tl::access_ptr())) {
+        ICHECK_EQ(op->args.size(), 3U);
+        const auto *load = op->args[0].as<BufferLoadNode>();
+        ICHECK(load) << "tl.access_ptr base must be a BufferLoad";
+        AddAccess(load->buffer, GetConservativeAccessMask(op->args[2]));
+        for (const PrimExpr &index : load->indices) {
+          this->VisitExpr(index);
+        }
+        if (load->predicate.defined()) {
+          this->VisitExpr(load->predicate.value());
+        }
+        this->VisitExpr(op->args[1]);
+        this->VisitExpr(op->args[2]);
+        return;
       } else if (op->op.same_as(builtin::tvm_access_ptr())) {
         if (op->args.size() > 1) {
           if (const auto *var_node = op->args[1].as<VarNode>()) {
             auto it = buffer_data_to_buffer_.find(GetRef<Var>(var_node));
             if (it != buffer_data_to_buffer_.end()) {
-              AddRead((*it).second);
+              const int access_mask =
+                  op->args.size() == 5U ? GetConservativeAccessMask(op->args[4])
+                                        : kAccessReadWrite;
+              AddAccess((*it).second, access_mask);
             }
           }
         }
--- a/src/transform/inject_pipeline.cc
+++ b/src/transform/inject_pipeline.cc
@@ -3412,20 +3412,30 @@ private:
         }
       }
     }
-    BufferSet pipeline_write_buffers;
-    if (!replayable_bind_mask.defined()) {
-      pipeline_write_buffers = CollectPipelineWriteBuffers(original_order);
-    }
+    BufferSet pipeline_write_buffers =
+        CollectPipelineWriteBuffers(original_order);
     Array<SBlock> scalar_binding_blocks;
     Array<SBlock> scheduled_order;
     std::vector<char> is_replayable_bind;
     is_replayable_bind.reserve(original_order.size());
     for (size_t i = 0; i < original_order.size(); ++i) {
       const SBlock &block = original_order[i];
-      bool replayable =
-          replayable_bind_mask.defined()
-              ? !is_zero(replayable_bind_mask.value()[i])
-              : IsReplayableScalarBindBlock(block, pipeline_write_buffers);
+      const bool semantically_replayable =
+          IsReplayableScalarBindBlock(block, pipeline_write_buffers);
+      bool replayable = semantically_replayable;
+      if (replayable_bind_mask.defined()) {
+        replayable = !is_zero(replayable_bind_mask.value()[i]);
+        if (replayable && !semantically_replayable) {
+          const auto *bind = block->body.as<BindNode>();
+          ICHECK(bind != nullptr);
+          LOG(FATAL) << "PrimFunc " << global_symbol_ << " marks scalar Bind '"
+                     << bind->var << "' as replayable via "
+                     << kPipelineReplayableScalarBinds
+                     << ", but the Bind has an unsupported type, has side "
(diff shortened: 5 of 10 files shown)


LeiWang1999:
@regression-perf

zkyue:
The `example_topk` 0.105 row is a stale-merge-base artifact of the regression run, not an effect of this PR: the generated kernel for that example is byte-identical with and without this change. The delta is host-side dispatch overhead that #2657 removed from `main` after this PR's merge ref was last computed.

What happened:

1. The regression job builds "Current" from `refs/pull/2651/merge`. GitHub had last computed that merge ref against base `1ac5a01a` (#2655, merged 12:24 +08 on 07-14). "Original" is built from current `main` (`70548a17`), which already contains #2657 (merged 14:11 +08). So the run effectively compared `main` **with** #2657 against this PR **without** #2657.
2. Before #2657, `AutoTuneImpl.__call__` re-ran `self.jit_impl.get_tir(*args, **kwargs)` — a full TIR elaboration costing tens of ms of host time — on **every** dispatch, only to stash `_prim_func_for_validation`. #2657 deleted exactly that.
3. `example_topk` is the only entry in the regression suite whose timed callable is the `@tilelang.autotune`-wrapped function itself (`do_bench(lambda: tl_topk(...))`); all other autotuned examples `.compile()` once and bench the resulting `JITKernel`. That is why exactly one row moved while everything else sat at 0.99+.

Reproduction (B200, CUDA 13.3, single GPU, A/B interleaved, `TILELANG_DISABLE_CACHE=1`, `PYTHONDEVMODE=1`, measuring `run_regression_perf()` exactly as the bot does):

| build | regression metric (ms) | kernel-only do_bench (ms) | kernel source sha256 |
|---|---|---|---|
| `main@70548a17` (= bot "Original") | 1.27 – 1.36 | 0.00947 | `219312b3…` |
| `refs/pull/2651/merge` (= bot "Current", fix on stale base `1ac5a01a`) | 28.8 – 30.6 | 0.00947 | `219312b3…` |
| merge ref + only #2657's `tuner.py` | 1.41 | 0.00941 | `219312b3…` |
| `main@70548a17` minus only #2657's `tuner.py` | 29.1 | 0.00952 | `219312b3…` |

- The lowered CUDA source is byte-identical across all four builds (single sha256). Expected: this PR only stops replay of binds whose RHS carries a side effect (`SideEffect(value) > kReadState`), and `example_topk` contains no such bind — no atomics at all.
- The real kernel time is ~9.5 µs on both sides; both reported numbers are two to three orders of magnitude above it, i.e. pure host dispatch overhead. (Absolute overhead is machine-dependent — 2.75 ms on the CI runner vs 1.3 ms here — but the mechanism is the same.)
- Moving #2657's `tuner.py` change alone across the two builds flips the result in both directions (30.6 → 1.41 ms when added to the PR build; 1.29 → 29.1 ms when removed from `main`), so that commit fully accounts for the gap.
- The example's own correctness assertion (`torch.testing.assert_close` vs `torch.topk`) passes on both sides, and `topk_selector` — the example this PR actually modifies — reads 0.999 in the same report.

The branch has been updated onto current `main` (pure rebase, no content change), so the merge ref now includes #2657; a fresh regression run should show `example_topk` at parity as well.

zkyue:
Rebased onto latest main to resolve conflicts.

LeiWang1999:
Thanks for catching these issues and made a fix! I improved some corner cases in commit `0955752ab`:

  1. `IfStmtBinding` and `PipelinePlanning` now honor the read/write masks of both `tl.access_ptr` and `tvm_access_ptr`. Non-constant masks are conservatively treated as read-write.
  2. `InjectSoftwarePipeline` now revalidates precomputed replay masks against the semantic replay-safety checks and rejects unsafe annotations with a clear diagnostic.

LeiWang1999:
@regression-perf

LeiWang1999 (dismissed):

LeiWang1999 (approved):

PR #2652, opened 2026-07-13
[BugFix] Reject invalid dtypes in T.dp4a

### LINKED ISSUE #2600: [BUG][Fuzzer][accepts-invalid] `T.dp4a` silently bit-reinterprets non-`int8`/`int32` operands instead of rejecting the dtype
### Required prerequisites

0.1.11 (built from source at cd37ed5f).

### System information

NVIDIA H100 80GB HBM3 (sm_90), Python 3.12, torch 2.12.1+cu130 (CUDA 13.0), tilelang 0.1.11 (tag `v0.1.11` = `cd37ed5f`). `T.dp4a` and the `DP4A` device template are architecture-independent (used on the SIMT path across CUDA targets); the reinterpret is not GPU-specific — reproduced on sm_90 this session.

### Problem description

`T.dp4a(A, B, C)` performs no dtype check on `A`/`B`/`C`, and the device `DP4A` template blindly reads each operand's first 4 bytes as `int` before calling `__dp4a`. If the buffers are anything other than the intended packed-`int8` inputs and an `int32` accumulator, the bytes are silently reinterpreted: the kernel compiles and runs to completion, no error or warning, and the result is garbage — neither a correct value nor a clean rejection.

Two silent cases (deterministic, reproduced twice), against the correct control:

| `A` / `B` / `C` dtype        | result | expected (int8-dot) |
|------------------------------|--------|---------------------|
| `int8` / `int8` / `int32` (control) | `80`   | `80` (correct)      |
| `float16` / `float16` / `int32`     | `8630` | should reject / not silently miscompute |
| `int8` / `int8` / `float32`         | `1.09e9` | should reject / not silently miscompute |

The `fp16` operands are read as packed `int8` lanes; the `float32` accumulator's bits are read via `*(int*)c` and an `int` result stored back into a `float`. The intended `int8`/`int32` contract is real but is enforced only externally by callers (the shipped SIMT GEMM path gates `use_dp4a` on `in_dtype == int8 and accum_dtype == int32`), never by the primitive itself.

Present since `T.dp4a` first shipped (the initial codebase-migration commit `57ab687c`); not a regression.

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import torch

def build(a_dt, b_dt, c_dt):
    @T.prim_func
    def main(A: T.Tensor((4,), a_dt), B: T.Tensor((4,), b_dt),
             C: T.Tensor((1,), c_dt), Out: T.Tensor((1,), c_dt)):
        with T.Kernel(1, threads=1):
            As = T.alloc_shared((4,), a_dt); Bs = T.alloc_shared((4,), b_dt)
            Cs = T.alloc_shared((1,), c_dt)
            T.copy(A, As); T.copy(B, Bs); T.copy(C, Cs)
            T.dp4a(As, Bs, Cs)
            for i in T.Parallel(1):
                Out[i] = Cs[i]
    return main

def run(a_dt, b_dt, c_dt):
    TT = {"int8": torch.int8, "int32": torch.int32,
          "float16": torch.float16, "float32": torch.float32}
    k = tilelang.compile(build(a_dt, b_dt, c_dt), target="cuda")
    A = torch.tensor([1, 2, 3, 4], dtype=TT[a_dt], device="cuda")
    B = torch.tensor([5, 6, 7, 8], dtype=TT[b_dt], device="cuda")
    C = torch.tensor([10], dtype=TT[c_dt], device="cuda")
    Out = torch.zeros(1, dtype=TT[c_dt], device="cuda")
    k(A, B, C, Out); torch.cuda.synchronize()
    return Out[0].item()

# 1*5 + 2*6 + 3*7 + 4*8 + 10 = 80
print("int8/int8/int32  :", run("int8", "int8", "int32"))      # -> 80    (correct)
print("fp16/fp16/int32  :", run("float16", "float16", "int32"))# -> 8630  (silent garbage)
print("int8/int8/float32:", run("int8", "int8", "float32"))    # -> 1.09e9(silent garbage)
```

### Traceback

```pytb
No traceback — every case compiles and runs to completion; the wrong cases are silently wrong and deterministic (reproduced twice, identical).
```

### Expected behavior

`T.dp4a` should reject non-`int8` inputs / non-`int32` accumulator at the frontend (raise), rather than emit code that reinterprets the operands' bytes. `__dp4a` is defined only for four packed signed 8-bit lanes accumulated into `int32`; any other dtype has no meaning there. This matches the reject-hardening the tracker already applies to other unguarded-dtype primitives, e.g. `T.atomic_add` on a 16-bit destination (#2572).

### Additional context

**Root cause.** `dp4a` issues `call_extern("DP4A", access_ptr(A,"r"), access_ptr(B,"r"), access_ptr(C,"rw"))` with no dtype check ([`customize.py#L11-L28`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/customize.py#L11-L28)). The device template reinterprets all three operands as `int*` regardless of the actual buffer dtype ([`common.h#L295-L320`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/common.h#L295-L320)):

```cpp
DP4A(InDatatype *a, InDatatype *b, OutDatatype *c) {
  const int a_int = *((int *)a);
  const int b_int = *((int *)b);
  const int c_int = *((int *)c);
  *c = __dp4a(a_int, b_int, c_int);
}
```

So `fp16` inputs have their raw bytes taken as four `int8` lanes, and a `float32` accumulator's bits are read as an `int` and the `int` result stored back into the `float` — both silent.

**Suggested fix.** A frontend dtype assert in `dp4a` (e.g. `A`/`B` element size 8 bits, `C` `int32`) would let an unsupported dtype raise instead of compiling; a `static_assert` in the template could serve as a backstop. (Proposed, not built.)

**Provenance.** `T.dp4a` and the `DP4A` template have been present since the initial codebase-migration commit [`57ab687c`](https://github.com/tile-ai/tilelang/commit/57ab687c) ("Migration of Codebase from Dev Branch into Main") with no dtype guard; this is not a regression. (The commit's trailing "(#10)" is a stale dev-branch PR number and is unrelated to the public repo's PR #10, which only edits the README.)

**Related.** Same missing-dtype-guard family as the recently-filed reject-hardening issues (e.g. #2572), but a distinct surface (the `DP4A` SIMT intrinsic) and a distinct **silent** face — the sibling issues abort loudly (`nvcc` error / internal assert), whereas `T.dp4a` silently reinterprets and returns garbage. I searched the open and closed tracker and found no existing `dp4a` bug report.

Chennesxu:
## Summary

`T.dp4a` accepted arbitrary input and accumulator dtypes. Unsupported dtypes could be silently reinterpreted by the backend DP4A implementation, producing incorrect results instead of raising an error.

## Root cause

The frontend emitted the DP4A external call without validating operand dtypes. The backend expects four packed signed `int8` lanes from each input and an `int32` accumulator.

## Fix

Validate the underlying buffer dtypes before emitting the DP4A call:

* Require both inputs to be `int8`.
* Require the accumulator to be `int32`.
* Raise `ValueError` for unsupported combinations.

The validation handles both Buffer and BufferLoad arguments used by existing SIMT kernels.

## Tested

Added regression coverage for valid Buffer and BufferLoad forms, invalid input dtypes, unsigned inputs, and invalid accumulator dtypes. Verified the existing int8 SIMT dp4a path locally.

Fixes #2600

## Summary

- Updated `T.dp4a` dtype validation to enforce the DP4A contract:
  - Operand inputs `A` and `B` must be `int8`.
  - Accumulator input `C` must be `int32`.
  - Any unsupported dtype combinations now raise `ValueError` (documented in the `dp4a` docstring).
- Validation supports both `Buffer` and `BufferLoad`-style inputs (including slice-based usage).
- Added regression coverage for:
  - Valid `int8/int8/int32` and `BufferLoad`/slice forms.
  - Invalid operand dtypes (including unsigned) and invalid accumulator dtypes.

## C++ style / lint notes

- This PR does not touch C++, CI/lint tooling, or `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step is not relevant for this change.

### DIFF
--- a/tilelang/language/customize.py
+++ b/tilelang/language/customize.py
@@ -1,25 +1,37 @@
 """Some customized operations frequently used in tensor programming, exposed on the TileLang language surface."""
 
 from __future__ import annotations
-from tilelang._typing import ShapeType, DType
+from tilelang._typing import ShapeType, DType, BufferLikeType
 import tilelang.language as T
 from tvm import arith
 from tvm.tirx import PrimExpr, Buffer, op
-from tilelang.utils.language import bits_product, prim_expr_equal
+from tilelang.utils.language import bits_product, prim_expr_equal, retrieve_buffer_and_offset
 from .atomic import atomic_max, atomic_min, atomic_add, atomic_addx2, atomic_addx4, atomic_load, atomic_or, atomic_store  # noqa: F401
 
 
-def dp4a(A: Buffer, B: Buffer, C: Buffer) -> PrimExpr:
-    """Perform a 4-element dot product with accumulation (DP4A).
+def dp4a(A: BufferLikeType, B: BufferLikeType, C: BufferLikeType) -> PrimExpr:
+    """Perform a four-element signed int8 dot product accumulated into int32.
 
     Args:
-        A (Buffer): First input buffer
-        B (Buffer): Second input buffer
-        C (Buffer): Accumulation buffer
+        A: First int8 input buffer.
+        B: Second int8 input buffer.
+        C: Int32 accumulator buffer.
 
     Returns:
-        PrimExpr: Handle to the DP4A operation
+        Handle to the DP4A operation.
+
+    Raises:
+        ValueError: If A or B is not int8, or C is not int32.
     """
+    a_dtype = T.dtype(retrieve_buffer_and_offset(A)[0].dtype)
+    b_dtype = T.dtype(retrieve_buffer_and_offset(B)[0].dtype)
+    c_dtype = T.dtype(retrieve_buffer_and_offset(C)[0].dtype)
+    if a_dtype != T.int8:
+        raise ValueError(f"dp4a requires int8 inputs, got A.dtype='{a_dtype}'")
+    if b_dtype != T.int8:
+        raise ValueError(f"dp4a requires int8 inputs, got B.dtype='{b_dtype}'")
+    if c_dtype != T.int32:
+        raise ValueError(f"dp4a requires an int32 accumulator, got C.dtype='{c_dtype}'")
     return T.call_extern(
         "handle",
         "DP4A",
--- a/testing/python/issue/test_tilelang_issue_2600.py
+++ b/testing/python/issue/test_tilelang_issue_2600.py
@@ -1,0 +1,54 @@
+"""Regression tests for T.dp4a dtype validation (issue #2600)."""
+
+import pytest
+
+import tilelang.language as T
+import tilelang.testing
+
+
+def _build_dp4a(a_dtype: str, b_dtype: str, c_dtype: str):
+    @T.prim_func
+    def kernel(
+        A: T.Tensor((4,), a_dtype),
+        B: T.Tensor((4,), b_dtype),
+        C: T.Tensor((1,), c_dtype),
+    ):
+        T.dp4a(A, B, C)
+
+    return kernel
+
+
+def test_dp4a_accepts_int8_int32():
+    kernel = _build_dp4a("int8", "int8", "int32")
+    assert kernel is not None
+
+
+def test_dp4a_accepts_buffer_load_slice():
+    # Existing SIMT kernels call dp4a with BufferLoad arguments.
+    @T.prim_func
+    def kernel(
+        A: T.Tensor((4,), "int8"),
+        B: T.Tensor((4,), "int8"),
+        C: T.Tensor((1,), "int32"),
+    ):
+        T.dp4a(A[0], B[0], C[0])
+
+    assert kernel is not None
+
+
+@pytest.mark.parametrize(
+    "a_dtype,b_dtype,c_dtype",
+    [
+        ("float16", "float16", "int32"),  # wrong input dtype
+        ("int8", "int8", "float32"),  # wrong accumulator dtype
+        ("uint8", "uint8", "int32"),  # unsigned inputs (signed-only intrinsic)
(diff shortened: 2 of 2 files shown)


LeiWang1999 (approved):

tilelang/language/customize.py
```diff
@@ -1,25 +1,37 @@
...
+        C: Int32 accumulator buffer.
 
     Returns:
-        PrimExpr: Handle to the DP4A operation
+        Handle to the DP4A operation.
+
+    Raises:
+        ValueError: If A or B is not int8, or C is not int32.
     """
+    a_dtype = str(retrieve_buffer_and_offset(A)[0].dtype)
```
LeiWang1999: do not use `str`, use `T.dtype` instead

LeiWang1999 (changes requested):
some comments about dtype, we prefer T.dtype instead str, as T.int8 is much more generao and can be both equal to "int8" and "torch.int8".

tilelang/language/customize.py
Chennesxu: Got it, thanks! I’ll update this to use T.dtype instead.

LeiWang1999 (approved):

PR #2655, opened 2026-07-13
[TIR][Runtime] Enforce host-evaluable assumptions at runtime

LeiWang1999:
## Summary

- Validate user-authored `T.assume` conditions in the host launcher before a device kernel is launched.
- Keep compiler-generated buffer shape and stride assumptions optimizer-only so nullable buffer arguments retain their existing behavior.

## Changes

- Mark user-authored assumptions with `tl.assume_requires_runtime_check` during `InjectAssumes` while preserving the paired `tl.assume` optimizer fact.
- Lift only unconditional, host-evaluable runtime checks in `SplitHostDevice`, and remove runtime-check markers from device functions.
- Extract marked checks in `MakePackedAPI` and emit `AssertStmt` nodes outside optimizer assumptions so later simplification cannot erase validation.
- Preserve device-only behavior for assumptions involving thread/block variables, buffers, calls, local bindings, or conditional control flow.
- Add transform and CUDA runtime coverage for marker propagation, runtime failures, optimizer-only assumptions, and nullable buffers.

## Validation

- `./format.sh`
- `cmake --build build -j`
- `TILELANG_DISABLE_CACHE=1 TVM_TEST_TARGETS='cuda;llvm' python -m pytest testing/python/transform/test_tilelang_transform_make_packed_api.py testing/python/transform/test_tilelang_transform_split_host_device.py testing/python/language/test_tilelang_language_assume.py testing/python/transform/test_nullable_buffer_params.py -q` (`21 passed, 4 skipped`)

## Risk

- Runtime lifting is intentionally conservative. Checks that cannot be evaluated safely and unconditionally by the launcher remain optimizer-only assumptions in device code.

### DIFF
--- a/src/transform/common/attr.h
+++ b/src/transform/common/attr.h
@@ -35,7 +35,8 @@ constexpr const char *coproc_scope = "coproc_scope";
 constexpr const char *pipeline_exec_scope = "pipeline_exec_scope";
 // Marks user-authored assumptions that require a host runtime check. The
 // corresponding tl.assume remains in the IR as an optimizer fact.
-constexpr const char *kAssumeRuntimeCheck = "tl.assume_runtime_check";
+constexpr const char *kAssumeRequiresRuntimeCheck =
+    "tl.assume_requires_runtime_check";
 
 // Attributes to implement SourceCodeBlock
 constexpr const char *kCodeBlockSource = "code_block_source";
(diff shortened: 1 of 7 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2656, opened 2026-07-13
[BugFix] Fix FP4 dequant symbolic exponent clamp

### LINKED ISSUE #2615: [BUG][Fuzzer][ice-on-valid] Exported MXFP4 dequant `_tir_u8_to_f4_to_bf16` crashes at trace time on a Python `min()` over symbolic `PrimExpr` operands (`min` should be `T.min`)
### Required prerequisites

0.1.11 (source pinned to tag `v0.1.11`, commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). Also present on `origin/main` (verified this session).

### System information

NVIDIA L40S (`sm_89`), Python 3.13, tilelang 0.1.11 (tag `v0.1.11` = `cd37ed5f`). The crash is at Python trace time, before any device codegen, so it is target-independent (also reproduced on `sm_90`/H100 this session).

### Problem description

`tilelang.quantize._tir_u8_to_f4_to_bf16` (a re-exported public helper, [`tilelang/quantize/__init__.py:8`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/__init__.py#L8)) — the per-element MXFP4-to-bfloat16 dequant used by the FP4 GEMM examples — **cannot be traced at all**. Its exponent-clamp line uses the Python builtin `min()` on two symbolic `PrimExpr` operands:

```python
# tilelang/quantize/quantization.py:68
e_bf16 = min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
```

`min(a, b)` evaluates `a < b` for truthiness, which invokes `PrimExpr.__bool__` on a symbolic comparison and raises

```
ValueError: Cannot use and / or / not operator to Expr, hint: use tvm.tirx.all / tvm.tirx.any instead
```

The left operand `e_bf16` is always a symbolic expression (it is derived from the input `val`), so the crash fires **for every call** — including the shipped `scale=0` (a Python `int`) convention, since `e_bf16 + 0` is still symbolic. The `min` was meant to be `T.min` (a lazy PrimExpr `Min` node); the builtin `min` eagerly compares. This is a clear typo-class defect: the neighbouring comment even calls it "the max function to limit the exponential part to 8 bits", i.e. a lazy clamp was intended.

The raised message is also misleading — it points the user at `tvm.tirx.all`/`tvm.tirx.any`, which are unrelated to the real `min`→`T.min` cause.

### Reproducible example code

```python
# Minimal real kernel calling the public exported helper exactly as the FP4 examples do.
import tilelang
import tilelang.language as T
from tilelang.quantize import _tir_u8_to_f4_to_bf16

N = 256

def build(scale):  # scale=0 int (bf16_fp4_hopper convention) or a symbolic uint16
    @T.prim_func
    def main(B: T.Tensor((N,), "uint8"), Out: T.Tensor((N * 2,), "bfloat16")):
            Bs = T.alloc_shared((N,), "uint8")
            Os = T.alloc_shared((N * 2,), "bfloat16")
            T.copy(B, Bs)
            for i in T.Parallel(N * 2):
                Os[i] = _tir_u8_to_f4_to_bf16(4, Bs[i // 2], i % 2, scale, dtype="bfloat16")
            T.copy(Os, Out)
    return main

# scale=0 (Python int) — the convention the shipped example passes:
tilelang.compile(build(0), out_idx=[-1], target="cuda")
# -> ValueError: Cannot use and / or / not operator to Expr ...  (at quantization.py:68)

# A symbolic per-element scale (the documented use) crashes identically:
tilelang.compile(build(T.cast(3, "uint16")), out_idx=[-1], target="cuda")
# -> same ValueError
```

### Traceback

```
File ".../tilelang/quantize/quantization.py", line 68, in _tir_u8_to_f4_to_bf16
    e_bf16 = min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
File ".../tvm/tirx/expr.py", line 181, in __nonzero__
    raise ValueError(
ValueError: Cannot use and / or / not operator to Expr, hint: use tvm.tirx.all / tvm.tirx.any instead
```

### Expected behavior

`_tir_u8_to_f4_to_bf16` should trace like the other exported dequant helpers in the same module (e.g. `_tir_u32_to_f4_to_f32`, `_tir_packed_to_fp4_to_f16`), all of which use lazy `tirx.Select`/`tirx.Min` and compile cleanly. The exponent clamp should build a lazy `Min` PrimExpr so the function is usable on the documented per-element `scale` path.

### Additional context

**Root cause.** The clamp on [`quantization.py:68`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/quantization.py#L68) is the Python builtin `min`, not `T.min`. `min(x, y)` forces `bool(x < y)`; with `x = e_bf16 + scale` a symbolic `PrimExpr`, `PrimExpr.__bool__` raises rather than folding, so the whole function is untraceable. Every other clamp in this file uses `tirx.Min`/`tirx.Select`.

**Verified fix (before → after, this session).** Replacing the one token with `T.min` (a lazy `Min`):

```python
e_bf16 = T.min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
```

turns both repro variants above from the `ValueError` into `COMPILE OK`. No other change needed.

**Reachability.** The buggy function is the **public re-export** [`tilelang.quantize._tir_u8_to_f4_to_bf16`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/quantize/__init__.py#L8) — a user calling this public API directly (as the minimal repro above does) hits the crash for any `scale`. Note the shipped FP4 examples do **not** exercise it: each defines its **own private** `_tir_u8_to_f4_to_bf16` (`example_dequant_gemm_bf16_fp4_hopper.py:217`, `..._mxfp4_hopper.py:10`, `..._mxfp4_cdna4.py:26`), so Python name resolution binds their call sites to the local copy, not the library helper (`grep` confirms nothing in the repo does `from tilelang.quantize import _tir_u8_to_f4_to_bf16`). Those private copies have in fact diverged — `mxfp4_hopper`'s has the clamp line commented out, and `bf16_fp4_hopper`'s already uses `T.min` (i.e. the fixed form). So the library helper is a public API that crashes on its own `scale` path while the examples silently forked away from it.

**Provenance.** The `min(e_bf16 + scale, ...)` line was introduced with the helper in [#725](https://github.com/tile-ai/tilelang/pull/725) ("[Feature] Low-bit twiddling dequantization and FP4 GEMM", Zhengju Tang, merged 2025-08-19); the later tirx migration [#2216](https://github.com/tile-ai/tilelang/pull/2216) renamed `tir`→`tirx` but kept the builtin `min`. Present since the helper first landed — not a regression. The library helper has no in-tree unit test and every example ships its own private copy (see Reachability), so CI never traces line 68.

**Disposition.** ice-on-valid: a valid, documented public API crashes at trace time (rather than compiling or rejecting the input). The value-oracle is not needed here — the function never reaches codegen.

Distinct from #2600 (`T.dp4a` dtype reinterpret), #2604 (packed `T.add2` mixed-dtype), and the `T.gemm_sp` sparse-MMA issues (#2603/#2605): those are silent-wrong data transforms in different components; this is a trace-time crash from a builtin-vs-`T.min` typo in the quantize helper. I searched the tracker for `u8_to_f4_to_bf16`, `Cannot use and or not operator`, and `quantize min bfloat16` and found no existing report.

mygitljf:
## Summary

Fix the trace-time crash in `_tir_u8_to_f4_to_bf16` caused by applying Python's built-in `min` to symbolic `PrimExpr` operands.

## Changes

- Replace the eager Python `min` call with TileLang's lazy `T.min`.
- Remove the misleading comment that described the upper clamp as a max operation.
- Add trace regression coverage for:
  - Python integer `scale=0`.
  - A symbolic `uint16` scale loaded from a buffer.
- Verify that tracing produces exactly one `Min<uint16>` node.

## Validation

- Reproduced the original `PrimExpr.__bool__` failure on:
  - TileLang v0.1.11.
  - TileLang v0.1.12.
  - Latest `origin/main`.
- Verified the candidate fix produces `TRACE_OK` for both scale paths.
- Verified the existing builds with `ninja -j64`.
- Local experiments were not rerun after the final commit, as requested; remote CI is pending.

## C++ style / lint notes

- Python-only change; no C++ files were modified.
- `git diff --check` passed.
- C++ formatting and lint checks are not applicable.

Fixes #2615

## Summary

- Fixed a trace-time crash in `_tir_u8_to_f4_to_bf16` by replacing Python’s eager `min` with TileLang’s lazy `T.min`, avoiding symbolic `PrimExpr` truthiness (`PrimExpr.__bool__`).
- Removed the misleading clamp-related comment.
- Added regression coverage in `testing/python/issue/test_tilelang_issue_2615.py` for both `scale=0` and symbolic per-element `uint16` scales, asserting tracing produces exactly one `Min<uint16>` node.

## Validation

- Existing builds passed.
- `git diff --check` passed.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_2615.py
+++ b/testing/python/issue/test_tilelang_issue_2615.py
@@ -1,5 +1,6 @@
 import pytest
 
+import tilelang.testing
 import tilelang.language as T
 from tilelang import tvm
 from tilelang.quantize import _tir_u8_to_f4_to_bf16
@@ -39,4 +40,4 @@ def test_u8_to_f4_to_bf16_traces_symbolic_exponent_clamp(use_buffer_scale):
 
 
 if __name__ == "__main__":
-    pytest.main([__file__])
+    tilelang.testing.main()
--- a/testing/python/issue/test_tilelang_issue_2615.py
+++ b/testing/python/issue/test_tilelang_issue_2615.py
@@ -1,0 +1,42 @@
+import pytest
+
+import tilelang.language as T
+from tilelang import tvm
+from tilelang.quantize import _tir_u8_to_f4_to_bf16
+
+
+@pytest.mark.parametrize(
+    "use_buffer_scale",
+    [False, True],
+    ids=["python-int-zero", "uint16-buffer-load"],
+)
+def test_u8_to_f4_to_bf16_traces_symbolic_exponent_clamp(use_buffer_scale):
+    @T.prim_func
+    def main(
+        packed: T.Tensor((1,), "uint8"),
+        scale: T.Tensor((1,), "uint16"),
+        output: T.Tensor((1,), "bfloat16"),
+    ):
+        with T.Kernel(1, threads=1):
+            for i in T.Parallel(1):
+                exponent_scale = scale[0] if use_buffer_scale else 0
+                output[i] = _tir_u8_to_f4_to_bf16(
+                    4,
+                    packed[i],
+                    i,
+                    exponent_scale,
+                    dtype="bfloat16",
+                )
+
+    min_nodes = []
+    tvm.tirx.stmt_functor.post_order_visit(
+        main.body,
+        lambda node: min_nodes.append(node) if isinstance(node, tvm.tirx.Min) else None,
+    )
+
+    assert len(min_nodes) == 1
+    assert min_nodes[0].dtype == T.uint16
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
--- a/tilelang/quantize/quantization.py
+++ b/tilelang/quantize/quantization.py
@@ -65,7 +65,7 @@ def _tir_u8_to_f4_to_bf16(nbit: int, val: tirx.PrimExpr, pos: tirx.PrimExpr, sca
     e_bf16 = e_f4 + tirx.const(126, T.uint16)
     # Scale is the exponential part, within the representation of uint8
     # Clamp the scaled exponent to the maximum value representable in 8 bits.
-    e_bf16 = T.min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
+    e_bf16 = T.min(e_bf16 + T.cast(scale, T.uint16), tirx.const((1 << 8) - 1, T.uint16))
     m_f4 = f4 & tirx.const(1, T.uint16)
     val_bf16 = tirx.reinterpret(T.bfloat16,
                                ((((s << tirx.const(8, T.uint16)) | e_bf16) << tirx.const(7, T.uint16))
--- a/tilelang/quantize/quantization.py
+++ b/tilelang/quantize/quantization.py
@@ -64,8 +64,8 @@ def _tir_u8_to_f4_to_bf16(nbit: int, val: tirx.PrimExpr, pos: tirx.PrimExpr, sca
     # Exponential bias between f4 and bf16 is 2^(8-1) - 2^(2-1) = 126
     e_bf16 = e_f4 + tirx.const(126, T.uint16)
     # Scale is the exponential part, within the representation of uint8
-    # To handle the overflow, we use the max function to limit the exponential part to 8 bits
-    e_bf16 = min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
+    # Clamp the scaled exponent to the maximum value representable in 8 bits.
+    e_bf16 = T.min(e_bf16 + scale, tirx.const((1 << 8) - 1, T.uint16))
     m_f4 = f4 & tirx.const(1, T.uint16)
     val_bf16 = tirx.reinterpret(T.bfloat16,
                                ((((s << tirx.const(8, T.uint16)) | e_bf16) << tirx.const(7, T.uint16))


mygitljf:
cc @LeiWang1999
Could take a moment to look at this? Thanks.

LeiWang1999 (dismissed):

testing/python/issue/test_tilelang_issue_2615.py
```diff
@@ -0,0 +1,42 @@
...
+        main.body,
+        lambda node: min_nodes.append(node) if isinstance(node, tvm.tirx.Min) else None,
+    )
+
+    assert len(min_nodes) == 1
+    assert min_nodes[0].dtype == T.uint16
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
LeiWang1999: please use `tilelang.testing.main()`

mygitljf:
also used T.min and T.cast because scale can be a symbolic TIR expression. The T.* APIs emit operations into the generated TIR instead of evaluating them in Python, and they keep the Python-int and symbolic paths consistently typed as uint16. Thanks!

testing/python/issue/test_tilelang_issue_2615.py
mygitljf: ok

LeiWang1999 (approved):

PR #2661, opened 2026-07-14
[BugFix] Respect runtime pointer alignment when vectorizing tensor views

### LINKED ISSUE #2526: [BUG][Fuzzer][ice-on-valid-code] `T.copy` over a `T.make_tensor` view emits an unconditional 16-byte `int4` load/store that crashes with `CUDA error: misaligned address` instead of honoring (or rejecting on) the view's runtime base alignment
### Required prerequisites

0.1.11

### System information

- Python 3.12.3, linux
- GPU: **NVIDIA H100 80GB HBM3** (compute capability 9.0, sm_90). The defect is in the architecture-independent loop-vectorization planner (`src/transform/loop_vectorize.cc`); it was reproduced on sm_90 with warp specialization **disabled** (`tl.disable_warp_specialized=True`), so it is not a Hopper-TMA or warp-specialization issue.

```python
import sys, tilelang, torch
print(sys.version.split()[0], sys.platform)        # 3.12.3 linux
print(tilelang.__version__)                         # 0.1.11
print(torch.__version__, torch.version.cuda)        # 2.12.1+cu130 13.0
```

### Problem description

A `T.copy` from a `T.make_tensor(ptr, shape, dtype)` view crashes at kernel launch with `AcceleratorError: CUDA error: misaligned address` (`cudaErrorMisalignedAddress`) whenever the user-supplied runtime pointer is not 16-byte aligned. The kernel compiles cleanly with no compile-time abort; it fails only at run time on a non-16B base.

The cause: `T.copy` over the view compiles to a **forced 16-byte (`int4`) vectorized** load/store, and the copy vectorizer's only alignment proof inspects the loop-relative *index base* (always `0` for a view), never the runtime pointer value — so it assumes the base is 16-byte aligned. For a **normal `T.Tensor` kernel argument** the base comes from `cudaMalloc`/PyTorch (≥ 256-byte aligned) so the identical `int4` copy is always safe; for a `make_tensor` view over an arbitrary runtime pointer nothing guarantees that, and the same `int4` load is emitted unconditionally — byte-identical code for aligned and misaligned bases.

<details>
<summary>Generated kernel (the unconditional 16-byte load over the runtime pointer)</summary>

```cpp
extern "C" __global__ void __launch_bounds__(4, 1)
main_kernel(const int64_t* __restrict__ base_ptr, int* __restrict__ out) {
  extern __shared__ __align__(1024) int sh[];
  int64_t v_ = base_ptr[0];
  void* V = (*(void**)(&(v_)));
  *(int4*)(sh + (((int)threadIdx.x) * 4)) = *(int4*)(((int*)V) + (((int)threadIdx.x) * 4));  // 16-byte load over runtime V
  __syncthreads();
  for (int i = 0; i < 4; ++i)
    for (int j = 0; j < 4; ++j)
      out[((i * 4) + j)] = sh[((i * 4) + j)];
}
```

</details>

<details>
<summary>Trigger boundary and discriminating control (all run this session on H100, sm_90, 0.1.11)</summary>

The kernel copies an `int32[4,4]` view (each row 4×`int32` = 16 bytes) from a `make_tensor` over `base_ptr[0] + off*4` bytes into shared memory; the single lever is the view's base offset in elements:

```
[off0]          off=0 -> MATCH, int4-vectorized-view-load=True   # base+0 bytes  (16B aligned)
[off4_aligned]  off=4 -> MATCH, int4-vectorized-view-load=True   # base+16 bytes (16B aligned)
[off1_misalign] off=1 -> *** CRASH: AcceleratorError: CUDA error: misaligned address   # base+4 bytes (4B aligned, not 16)
```

The single difference between a crashing run and a clean run is whether the copied buffer's runtime base is 16-byte aligned:

| buffer source | runtime base | result |
|---------------|--------------|--------|
| `make_tensor(base_ptr[0] + 0)` | 16B aligned (`+0`) | **MATCH**, `int4`-vectorized |
| `make_tensor(base_ptr[0] + 4 elems)` | 16B aligned (`+16B`) | **MATCH**, `int4`-vectorized |
| `make_tensor(base_ptr[0] + 1 elem)` | 4B aligned (`+4B`) | **CRASH: misaligned address** |
| normal `T.Tensor` `int32[4,4]`/`float32[512]` arg | `cudaMalloc` base (≥256B) | vectorizes identically, **never crashes** |

The normal-`T.Tensor` control (a `T.copy` over a `cudaMalloc`-backed argument) vectorizes to the identical `int4`/`float4` copy but **matches and never crashes**, because its base is over-aligned. This isolates the defect to the vectorizer assuming the `make_tensor` view's runtime base is 16-byte aligned when nothing guarantees it.

</details>

### Root cause

`make_tensor` builds a buffer over an opaque runtime pointer with no `data_alignment`, and the loop-vectorization planner's only alignment proof inspects the loop-relative *index* (which is `0` for a view), never that runtime pointer — so it forces the 16-byte `int4` copy unconditionally.

<details>
<summary>Step-by-step: from <code>make_tensor</code> to the unchecked <code>int4</code> load</summary>

1. **`make_tensor` carries no alignment.** `make_tensor(ptr, …)` ([`proxy.py#L329`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/proxy.py#L329)) routes to `make_tensor_from_addr`, which builds the buffer as `buffer(shape, dtype=dtype, data=pointer_var, strides=strides, scope=storage_scope)` ([`proxy.py#L326`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/proxy.py#L326)) — no `elem_offset`, no `data_alignment`. The data `Var` is the user's runtime pointer, whose alignment is unknown.

2. **The vectorizer picks `int4` purely from dtype lanes and loop extent.** `ComputeBufferVectorSize` ([`loop_vectorize.cc#L768`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L768)) builds an offset from *loop indices × strides* ([`#L781-L784`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L781-L784)) and a lower bound from dtype lane capacity (`vector_load_bits_max_ / dtype.bits`) ([`#L791`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L791)). For `int32` that capacity is 128/32 = 4 lanes = `int4`.

3. **The only divisibility check examines the index base, not the pointer.** `IndicesCanVectorize` ([`#L975`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L975)) computes `Simplify(Substitute(expr, {var -> 0}))` — the index with the loop var zeroed — and requires `FloorMod(simplified_expr, vec_size) == 0` ([`#L1001-L1006`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L1001-L1006)). For a `make_tensor` view the access starts at element `0` of the view, so this index base is literally `0`, and `FloorMod(0, vec_size) == 0` for any `vec_size`. The actual misalignment lives in the runtime pointer value `V`, which is never part of the index expression and is never examined.

4. **The planner never consults buffer base alignment.** `ComputeBufferVectorSize` and the access-ptr path ([`#L654-L690`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L654-L690)) reason only about loop-relative offsets and dtype lanes; there is no read of `buffer->data_alignment` anywhere in `loop_vectorize.cc` (grep: zero occurrences). `data_alignment` *is* carried through `lower_tile_op.cc` ([`#L86`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/lower_tile_op.cc#L86), [`#L1123`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/lower_tile_op.cc#L1123)), but the vectorizer never reads it.

So for a normal kernel-argument `T.Tensor` the `cudaMalloc`/PyTorch base is over-aligned and the `int4` copy is always safe; for a `make_tensor` view the same unconditional `int4` is emitted with no fact relating it to the runtime base alignment — and crashes the moment that base is not 16-byte aligned.

</details>

### Reproducible example code

```python
import torch
import tilelang as tl
import tilelang.language as T


# T.copy over a make_tensor view: does it honor the view's runtime base alignment?
def copy_view(R, C, srow, scol, dtype=T.int32):
    @T.prim_func
    def main(base_ptr: T.Tensor([1], T.ptr), out: T.Tensor([R, C], dtype)):
        with T.Kernel(1, threads=R) as _:
            V = T.make_tensor(base_ptr[0], (R, C), dtype, strides=(srow, scol))
            sh = T.alloc_shared((R, C), dtype)
            T.copy(V, sh)                      # <-- forced int4 (16B) vectorized load over V
            for i in T.serial(R):
                for j in T.serial(C):
                    out[i, j] = sh[i, j]       # writeback so the aligned control can be checked
    return main


def run(off_elems, label):
    kernel = tl.compile(copy_view(4, 4, 4, 1), execution_backend="cython",
                        pass_configs={"tl.disable_warp_specialized": True})
    vec = "int4*)(((int*)V)" in kernel.get_kernel_source()
    base = torch.arange(256, device="cuda", dtype=torch.int32)
    base_ptrs = torch.tensor([base.data_ptr() + off_elems * 4], device="cuda", dtype=torch.int64)
    out = torch.empty((4, 4), device="cuda", dtype=torch.int32)
    try:
        kernel(base_ptrs, out)
        torch.cuda.synchronize()
        ref = base.as_strided((4, 4), (4, 1), storage_offset=off_elems)
        print(f"[{label}] off={off_elems} -> {'MATCH' if torch.equal(out, ref) else '*** MISMATCH ***'}, int4-view-load={vec}")
    except Exception as e:
        print(f"[{label}] off={off_elems} -> *** CRASH: {type(e).__name__}: {str(e)[:40]}, int4-view-load={vec}")


if __name__ == "__main__":
    run(0, "off0")            # -> MATCH,   16B-aligned base
    run(4, "off4_aligned")    # -> MATCH,   16B-aligned base (+16B)
    run(1, "off1_misalign")   # -> CRASH: misaligned address, 4B-aligned base
```

Output (captured this session on H100, sm_90, tilelang 0.1.11):

```
[off0]          off=0 -> MATCH, int4-view-load=True
[off4_aligned]  off=4 -> MATCH, int4-view-load=True
[off1_misalign] off=1 -> *** CRASH: AcceleratorError: CUDA error: misaligned address, int4-view-load=True
```

### Traceback

```pytb
The `off=1` view triggers a **device-side** illegal access (`cudaErrorMisalignedAddress`) at kernel launch; there is no compile-time abort. The kernel compiles cleanly and the copy fails at run time:


AcceleratorError: CUDA error: misaligned address
Search for `cudaErrorMisalignedAddress' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1


(Because the illegal access poisons the CUDA context, a *subsequent* compile's `cudaFuncSetAttribute` may then fail with `Failed to set the allowed dynamic shared memory size … misaligned address` — a downstream symptom of the same crash, not a separate failure.)
```

### Expected behavior

`T.copy` over a `T.make_tensor` view must not assume the view's runtime base pointer is 16-byte aligned. It should either copy correctly for any legal runtime pointer (e.g. by capping the vector width to the proven base alignment) or reject the program at compile time, rather than emit an unconditional `int4` that crashes with `cudaErrorMisalignedAddress` on a non-16-byte-aligned base.

Grounded in TileLang's own behavior:

- The **identical** copy over a normal `T.Tensor` argument (`cudaMalloc`-aligned base) and over an aligned view (`off=0`/`off=4`) is correct and `int4`-vectorized, so the vectorization itself is sound — only the *unverified base-alignment assumption* for `make_tensor` is the defect.
- `T.make_tensor` is an **exported public** primitive (`tilelang/language/proxy.py`); a public primitive should not produce a kernel that issues a misaligned 16-byte access for an ordinary (4-byte-aligned) runtime pointer.
- The project already tracks `buffer->data_alignment` and threads it through `lower_tile_op.cc` ([`#L86`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/lower_tile_op.cc#L86)…[`#L1123`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/lower_tile_op.cc#L1123)); the vectorizer simply fails to consult it, so the alignment fact needed to bound the vector width already exists in the IR.

### Additional context

<details>
<summary>Suggested fix (proposed, not built or tested)</summary>

The vectorizer should not assume the runtime base pointer of a `make_tensor` (externally-pointed) buffer is 16-byte aligned. In rough order of preference:

1. **Cap the vector size by the buffer's proven base alignment.** In `ComputeBufferVectorSize` ([`loop_vectorize.cc#L768`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L768)) / the access-ptr path ([`#L654-L690`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/loop_vectorize.cc#L654-L690)), when a buffer's data `Var` is an externally-supplied pointer carrying no proven alignment, read `buffer->data_alignment` and `GCD` the candidate lane count with `data_alignment * 8 / dtype.bits`, so the alignment proof has a real base fact instead of the implicit "aligned" assumption.
2. **Have `make_tensor` record a conservative `data_alignment` and expose an opt-in hint.** In `make_tensor_from_addr` ([`proxy.py#L326`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/proxy.py#L326)), default `data_alignment` to the dtype element size, and add an explicit hint (e.g. `T.make_tensor(..., align=16)`) so a user who *can* prove a 16-byte base opts into the `int4` path; the vectorizer then only emits `int4` when `data_alignment >= 16`.

The preferred combination is (1)+(2): propagate `data_alignment` from `make_tensor` and consult it in the vectorizer. I have not built or verified any of these changes.

</details>

**Provenance.** The `make_tensor`/`make_tensor_from_addr` path that builds a buffer with no `data_alignment` was introduced by PR [#1923](https://github.com/tile-ai/tilelang/pull/1923) ("Support ptr-table grouped GEMM kernels", merged 2026-03-12; `git log -S'buffer(shape, dtype=dtype, data=pointer_var' -- tilelang/language/proxy.py`). The zeroed-index divisibility check in `IndicesCanVectorize` predates that and has never consulted base alignment; its origin before subsequent NFC vectorizer refactors is unverified. Reproduced fresh and deterministically this session on an H100 (sm_90, tilelang 0.1.11); no fix built.

**Distinct from previously reported issues.** I searched the open and closed tracker (including `make_tensor`, `misaligned`, `vectorize`, `int4`, `alignment`). The nearest existing report is [#218](https://github.com/tile-ai/tilelang/issues/218) — a `T.copy` `misaligned address` crash for `int8` with a small block — but there the misalignment comes from the *tile's element offset within a normal buffer* (loop-relative, index-visible), which the vectorizer can in principle reason about; here the index base is `0` and the misalignment lives entirely in the **opaque runtime pointer** of a `make_tensor` view that the alignment proof never inspects. Different input primitive and different root cause. I found no report of the copy vectorizer forcing an `int4` load/store over a `make_tensor` view without checking the view's runtime base alignment.

UnsettingGalaxy:
## Summary

Fixes #2526.

When a tensor view is created from a runtime address, the underlying pointer alignment may be weaker than the default alignment assumed by the compiler. This can cause the vectorizer to generate vectorized memory accesses (e.g. int4 loads) for pointers that are not properly aligned, resulting in CUDA misaligned address errors.

This PR adds explicit alignment propagation for tensors created from runtime addresses and uses the alignment information to constrain vectorization.

## Root Cause

The vectorization decision depends on `Buffer.data_alignment`.

For tensors created from runtime addresses, the alignment information was not explicitly specified. As a result, the generated Buffer could contain an inaccurate alignment assumption.

A pointer with insufficient alignment could therefore be treated as vectorizable:

## Summary

- Propagate explicit or inferred alignment through `make_tensor` and `make_tensor_from_addr`.
- Constrain global-buffer vectorization using `buffer->data_alignment`, preventing unsafe wide accesses from insufficiently aligned runtime pointers.
- Preserve vectorization when alignment is proven, while reducing vector width for weaker alignment.

## C++ style / lint notes

- The C++ change does not modify rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) may report advisory findings; these are not correctness or build blockers.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -805,6 +805,23 @@ private:
 
     int buffer_vec_size = loop_extent_vector_size_;
 
+    // A vector access of `vec_size` lanes issues a `vec_size * elem_bits`-wide
+    // load/store whose address is only guaranteed to be a multiple of the
+    // buffer's base alignment plus an elem_offset the divisibility checks
+    // below reason about. The base alignment itself is recorded in
+    // buffer->data_alignment: ordinary allocations and kernel arguments carry
+    // the default kAllocAlignment (64B, never restrictive here), but a buffer
+    // viewing an opaque runtime pointer (e.g. T.make_tensor over an address)
+    // may only be element-aligned, so cap the vector width by it.
+    if (IsGlobalBuffer(buffer)) {
+      int elem_bits = buffer->dtype.bits() * buffer->dtype.lanes();
+      if (buffer->data_alignment > 0 && elem_bits > 0) {
+        int align_lanes =
+            std::max(buffer->data_alignment * 8 / elem_bits, 1);
+        buffer_vec_size = arith::ZeroAwareGCD(buffer_vec_size, align_lanes);
+      }
+    }
+
     // Transform indices using layout_map if present
     auto transformed_indices = TransformIndices(indices, buffer);
 
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -816,8 +816,7 @@ private:
     if (IsGlobalBuffer(buffer)) {
       int elem_bits = buffer->dtype.bits() * buffer->dtype.lanes();
       if (buffer->data_alignment > 0 && elem_bits > 0) {
-        int align_lanes =
-            std::max(buffer->data_alignment * 8 / elem_bits, 1);
+        int align_lanes = std::max(buffer->data_alignment * 8 / elem_bits, 1);
         buffer_vec_size = arith::ZeroAwareGCD(buffer_vec_size, align_lanes);
       }
     }
--- a/tilelang/language/proxy.py
+++ b/tilelang/language/proxy.py
@@ -325,7 +325,6 @@ def make_tensor_from_addr(
 
     dtype = _normalize_tensor_dtype(dtype)
     if align is None:
-        
         dtype_info = DataType(dtype)
         align = max(dtype_info.bits * dtype_info.lanes // 8, 1)
     pointer_var = _materialize_pointer_from_addr(addr, dtype, storage_scope)
--- a/tilelang/language/proxy.py
+++ b/tilelang/language/proxy.py
@@ -313,7 +313,9 @@ def make_tensor_from_addr(
     dtype: DType = "float32",
     strides: tuple[PrimExpr, ...] | None = None,
     storage_scope: str = "global",
+    align: int | None = None,
 ) -> tirx.Buffer:
+
     from tilelang.language.eager.builder import Builder
 
     if Builder.current() is None:
@@ -322,12 +324,20 @@ def make_tensor_from_addr(
         )
 
     dtype = _normalize_tensor_dtype(dtype)
+    if align is None:
+        
+        dtype_info = DataType(dtype)
+        align = max(dtype_info.bits * dtype_info.lanes // 8, 1)
     pointer_var = _materialize_pointer_from_addr(addr, dtype, storage_scope)
-    return buffer(shape, dtype=dtype, data=pointer_var, strides=strides, scope=storage_scope)
+    return buffer(shape, dtype=dtype, data=pointer_var, strides=strides, scope=storage_scope, align=align)
 
 
 def make_tensor(
-    ptr: Var | PrimExpr, shape: ShapeType, dtype: DType = "float32", strides: tuple[PrimExpr, ...] | None = None
+    ptr: Var | PrimExpr,
+    shape: ShapeType,
+    dtype: DType = "float32",
+    strides: tuple[PrimExpr, ...] | None = None,
+    align: int | None = None,
 ) -> tirx.Buffer:
     from tilelang.language.eager.builder import Builder
 
@@ -340,4 +350,4 @@ def make_tensor(
 
     ptr_type = _get_pointer_type_annotation(ptr)
     storage_scope = ptr_type.storage_scope if ptr_type is not None and ptr_type.storage_scope else "global"
-    return make_tensor_from_addr(ptr, shape, dtype=dtype, strides=strides, storage_scope=storage_scope)
+    return make_tensor_from_addr(ptr, shape, dtype=dtype, strides=strides, storage_scope=storage_scope, align=align)


PKU-wcx:
I would like to clarify whether the test case in the issue is expected to produce the results described there. Specifically, should int4-view-load be true for the first two runs? My understanding is that the compiler should not be able to determine the alignment of base.data_ptr() at compile time, since the actual pointer value is only known at runtime.

PR #2663, opened 2026-07-14
[BugFix] Fix WGMMA C-store layout for multiple warpgroups along M

### LINKED ISSUE #2522: [BUG][Fuzzer][wrong-code] WGMMA `T.gemm` with `FullRow` and >1 warpgroup along M (e.g. `block_M=256`) silently swaps 64-row output sub-blocks instead of storing each row to its correct place
### Required prerequisites

0.1.11

### System information

- Python 3.13.13, linux
- GPU: **NVIDIA H100 80GB HBM3** (compute capability 9.0, sm_90). This is a Hopper-specific defect: it lives in the **WGMMA** code path (`tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py`), which is only used on sm_90.

```python
import sys, tilelang, torch
print(sys.version.split()[0], sys.platform)        # 3.13.13 linux
print(tilelang.__version__)                         # 0.1.11
print(torch.__version__, torch.version.cuda)        # 2.11.0+cu130 13.0
```

### Problem description

**Silent miscompile:** a WGMMA GEMM with `GemmWarpPolicy.FullRow` and more than one warpgroup along M (`num_warp_m >= 8`, e.g. `block_M=256, threads=256`) compiles and runs with no error but **swaps 64-row sub-blocks of the output across the warpgroup boundary**. With `block_M=256`, output rows `[64:128)` and `[128:192)` are exchanged — `ref[64]==out[128]`, 128 of 256 rows wrong, `max|out-ref| = 41.0`, deterministic. The hardware computes the right values; they are **stored to the wrong rows**.

**Discriminating control (isolates the defect to the WGMMA store path):** the *identical* program with the WGMMA path disabled (`pass_configs={"tl.disable_wgmma": True}`, same `FullRow` policy) is bit-exact (`max_err = 0.0`). So it is not the program, the policy, or numerics — only the WGMMA accumulator→C store, and only when M spans >1 warpgroup.

**Root cause (one line):** the C-store layout `make_mma_store_layout` decomposes the M axis differently from the WGMMA instruction-issue side — it flattens all warps onto a 16-row thread axis and treats the per-warpgroup 64-row-atom count as a stride-128 *outer index*, inverting the "which warpgroup" vs "which 64-row atom" roles. The two decompositions coincide only for a single warpgroup (`num_warp_m == 4`), and diverge for `num_warp_m >= 8`.

<details>
<summary>Captured H100 output — <code>block_M=256, block_N=128, K=64, threads=256, FullRow</code></summary>

```
WGMMA fired: True
disable_wgmma path uses wgmma: False
max|wgmma - ref|         = 41.0
max|disable_wgmma - ref| = 0.0
max|wgmma - disable_wgmma (cross-config) = 41.0
wgmma deterministic (run1==run2): True
rows wrong (of 256): 128
row 0   correct: True
ref[64]==wgmma[128]: True
ref[128]==wgmma[64]: True
ref[64,:6]    = [58.0, 47.0, 60.0, 41.0, 65.0, 49.0]
wgmma[64,:6]  = [67.0, 60.0, 71.0, 55.0, 83.0, 67.0]   # == ref[128]
ref[128,:6]   = [67.0, 60.0, 71.0, 55.0, 83.0, 67.0]
wgmma[128,:6] = [58.0, 47.0, 60.0, 41.0, 65.0, 49.0]   # == ref[64]
```
</details>

<details>
<summary>Failing boundary — trigger is <code>num_warp_m >= 8</code>, independent of block_N, no column analogue</summary>

Swept on H100. Row-swap fires **iff `num_warp_m >= 8`** (>1 warpgroup along M):

| policy | warps along | `num_warp_m` | result |
|--------|----------------|--------------|--------|
| `FullRow` | **M** | **8** (two warpgroups) | **row-swap** `[64:128)`↔`[128:192)` (`max_err` 39–48) |
| `FullCol` | **N** | 4 (single warpgroup) | correct |
| `Square` | **N** | 4 (single warpgroup) | correct |

- **Independent of `block_N`:** identical swap at `block_N = 16, 64, 128, 256`.
- **`block_M <= 128` always correct:** `num_warp_m <= 4`, `warp_rows = 1`, the index-repeat is a no-op, store and issue coincide.
- **`disable_wgmma` bit-exact `0.0` in every cell**, confirming the defect is in the WGMMA store path.
- No column-swap analogue: N is laid out by `warp_cols` consistently on both sides. `FullRow` is the policy that most directly reaches `m_warp >= 8` ([`src/cuda/op/gemm.cc#L204-L211`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/gemm.cc#L204-L211)); the true trigger is the resulting `warp_rows >= 2`.

Sibling (distinct failure mode, not this issue): `FullCol`/`Square` at `block_N=16, threads=256` deadlock at launch (RC=124) rather than row-swap — the 8-warp warp-specialization deadlock family (cf. #359). Observed in an earlier run, not re-run here; the row-swap itself was reproduced fresh on H100.
</details>

<details>
<summary>Root cause — full derivation (issue-side vs store-side M-axis decomposition)</summary>

**Instruction-issue side (hardware truth).** `warp_i` at [`wgmma_macro_generator.py#L677`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L677) = `(warp_m // 4) * num_inst_m + inst_m_idx`, with the A row offset `warp_i * 64` ([`#L680-L686`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L680-L686)) and `C_offset = inst_m_idx * warp_cols * local_size_out` ([`#L671`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L671)). This lays out M as: warpgroup `g = warp_m//4` owns `[g*128, g*128+128)` (outer); within it two 64-row atoms; within each atom, 4 warps own 16-row strips. So **row `r` ⇒ warp `(r//128)*4 + (r%64)//16`** — matching the PTX ISA wgmma D-fragment layout (warp `w` owns band `16*w` within a 64-row warpgroup).

**Store side** ([`wgmma_macro_generator.py#L867-L875`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L867-L875)):
```python
block_layout  = warp_n_layout.repeat([block_row_warps, block_col_warps], True, False)  # repeat_on_thread=True
warp_m_layout = block_layout.repeat([warp_rows, 1], False, False)                       # repeat_on_thread=False (index)
```
For `FullRow` `block_M=256`: `block_row_warps=8`, `warp_rows=2`. Per `FragmentNode::Repeat` ([`layout.cc#L510-L554`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/layout/layout.cc#L510-L554)) the `block_row_warps=8` **thread**-repeat flattens all 8 warps to `warp = (r%128)//16`, and the `warp_rows=2` **index**-repeat becomes the stride-128 outer dim `atom = r//128`. So the store believes **row `r` ⇒ warp `(r%128)//16`, atom `r//128`**.

**Mismatch at `r=128`:** hardware says warpgroup 1 / warp 4; store says warp 0 / atom 1 — and warp 0 of warpgroup 0 holds the data for row **64**. So `[128:192)` and `[64:128)` are exchanged. The C++ reference `makeGemmFragmentCHopper` ([`gemm_layouts.cc#L190-L201`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/layout/gemm_layouts.cc#L190-L201)) distributes warp**groups** on thread and 16-row strips on index (correct); the Python store layout's comment claims to reproduce it but does not for the multi-warpgroup case.
</details>

### Reproducible example code

```python
import tilelang
tilelang.disable_cache()
import tilelang.language as T
from tilelang.language import GemmWarpPolicy
import torch

# WGMMA GEMM, FullRow policy, block_M=256 -> 8 warps along M (num_warp_m=8, two warpgroups).
bM, bN, bK, threads = 256, 128, 32, 256
M, N, K = bM, bN, bK * 2


def make(disable_wgmma):
    @T.prim_func
    def main(A: T.Tensor((M, K), "float16"),
             C: T.Tensor((M, N), "float16")):
        with T.Kernel(T.ceildiv(N, bN), T.ceildiv(M, bM), threads=threads) as (bx, by):
            As = T.alloc_shared((bM, bK), "float16")
            Bs = T.alloc_shared((bK, bN), "float16")
            Cf = T.alloc_fragment((bM, bN), "float32")
            T.clear(Cf)
            for k in T.Pipelined(T.ceildiv(K, bK), num_stages=2):
                T.copy(A[by * bM, k * bK], As)
                T.copy(B[k * bK, bx * bN], Bs)
                T.gemm(As, Bs, Cf, policy=GemmWarpPolicy.FullRow)
            T.copy(Cf, C[by * bM, bx * bN])
    return tilelang.compile(main, out_idx=-1,
                            pass_configs={"tl.disable_wgmma": disable_wgmma})


torch.manual_seed(7)
A = torch.randint(0, 3, (M, K), dtype=torch.float16, device="cuda")
B = torch.randint(0, 3, (K, N), dtype=torch.float16, device="cuda")
ref = (A.float() @ B.float())

kw = make(False)            # WGMMA path (suspect)
Cw = kw(A, B).float()
km = make(True)             # same FullRow policy, WGMMA disabled (trusted reference)
Cm = km(A, B).float()

print("WGMMA fired:", ("wgmma" in kw.get_kernel_source()) and ("warpgroup" in kw.get_kernel_source()))
print("max|wgmma - ref|         =", (Cw - ref).abs().max().item())   # -> 41.0
print("max|disable_wgmma - ref| =", (Cm - ref).abs().max().item())   # -> 0.0
print("cross-config max|wgmma - disable_wgmma| =", (Cw - Cm).abs().max().item())  # -> 41.0
print("rows wrong (of 256):", ((Cw - ref).abs().max(dim=1).values > 0.5).sum().item())  # -> 128
print("ref[64]==wgmma[128]:", torch.equal(ref[64], Cw[128]))   # -> True (the swap)
print("ref[128]==wgmma[64]:", torch.equal(ref[128], Cw[64]))   # -> True
```

The cross-config control (`disable_wgmma=True`, same `FullRow` policy) is bit-exact (`max_err = 0.0`), isolating the defect to the WGMMA store path.

### Traceback

```pytb
There is **no traceback**: the kernel compiles and runs to completion, and the result is silently wrong and deterministic.
```

### Expected behavior

A WGMMA GEMM with `GemmWarpPolicy.FullRow` and any legal warp count (including the `num_warp_m >= 8` / multi-warpgroup case that `block_M=256, threads=256` produces) should write each accumulator row to its correct global row — i.e. produce `allclose` output without any manual row permutation.

This is grounded in TileLang's own behavior: the **non-WGMMA** lowering of the identical program with the identical `FullRow` policy is **bit-exact** (`max_err = 0.0`, see the control). The WGMMA path is meant to be a faster, numerically-equivalent lowering of the same GEMM, and the project already has a C++ reference layout, `makeGemmFragmentCHopper` ([`src/layout/gemm_layouts.cc#L190-L201`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/layout/gemm_layouts.cc#L190-L201)), that the Python `make_mma_store_layout` comments that it is reproducing ([`wgmma_macro_generator.py#L867-L875`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L867-L875)). The project's own intent is therefore that the store layout match the hardware accumulator layout for all warp counts; for `num_warp_m >= 8` it does not.

The hardware layout is fixed by the PTX ISA: the `wgmma.mma_async` D-fragment for a 64-row warpgroup gives warp `w` (of the 4 warps in the warpgroup) the 16-row band starting at `16*w` (PTX ISA, *Asynchronous Warpgroup Level Matrix Multiply-Accumulate — Register Fragment Layout*), so globally `row r` is owned by warp `(r // 128) * 4 + (r % 64) // 16`. The instruction-issue side (`warp_i` at [`wgmma_macro_generator.py#L677`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L677)) follows this; the store side flattens all warps to `(r % 128) // 16`, which agrees only for a single warpgroup. So the fix should make the store-side thread→row map equal the PTX warpgroup-major layout the issue side already uses.

### Additional context

I verified *where* the bug is (the M-axis decomposition mismatch between the instruction-issue side at [`wgmma_macro_generator.py#L677`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L677) / [`#L680-L686`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L680-L686) and the store layout at [`#L867-L875`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L867-L875), via `FragmentNode::Repeat` at [`src/layout/layout.cc#L510-L554`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/layout/layout.cc#L510-L554)) and reproduced it **fresh and deterministically** on an H100, but I did **not** build a fix.

**Suggested fix direction (proposed, not built or tested):** make `make_mma_store_layout` ([`wgmma_macro_generator.py#L867-L875`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py#L867-L875)) decompose the M axis to match the instruction-issue mapping in `wgmma_ss_atom`, i.e.

`[warpgroup (4 warps, 128 rows) outer] × [inst_m 64-row atoms] × [4 in-group warps × 16-row strips] × [base 16×8 lane fragment]`.

Concretely the row axis must split as: warpgroup index `g = warp_m // 4` selecting a 128-row block on **thread** (outermost), then within each warpgroup the two 64-row atoms on **index**, then the 4 in-group warps (`warp_m % 4`) as 16-row strips on **thread**, then the base lane fragment. Equivalently: build a single-warpgroup store fragment (4 warps × `num_inst_m` 64-row atoms, matching exactly how `C_offset = inst_m_idx*warp_cols*local_size_out` and `warp_i = (warp_m//4)*num_inst_m + inst_m_idx` lay out the accumulator registers), then thread-repeat that whole warpgroup tile by `block_row_warps // 4` along M as the **outermost** dimension. The current `.repeat([block_row_warps, ...], repeat_on_thread=True)` followed by `.repeat([warp_rows, 1], repeat_on_thread=False)` should be restructured so the `warp_rows`/atom expansion lives **inside** a warpgroup (between the in-group warp thread-repeat and the inter-warpgroup thread-repeat), not as a stride-128 outer index.

A targeted check: after the fix, `store_layout(r)` must yield `warp = (r // 128)*4 + (r % 64)//16` and atom-local index `= (r % 128)//64`, identical to the issue side. Validate with the differential repro above (`FullRow`, `block_M=256`, `threads=256`), which must then satisfy `allclose` without the manual 64↔128 swap. I have not built or verified this change.

**Provenance.** The Python WGMMA macro generator that carries `make_mma_store_layout` was most recently reshaped by the backend-refactor PR [#2165](https://github.com/tile-ai/tilelang/pull/2165) (*"Move backend-specific GEMM implementations and transforms into backend directories"*) and the TIR-x update [#2216](https://github.com/tile-ai/tilelang/pull/2216); the divergent M-axis decomposition is present at `v0.1.11` (`cd37ed5f`). I did not bisect the exact introducing commit, so this is "present no later than `v0.1.11`", not an attribution.

**Dedup.** I searched the open and closed `tile-ai/tilelang` tracker and found no report of this defect (the WGMMA `FullRow` multi-warpgroup C-store M-axis row-block swap). Two potentially-confusable neighbors, both distinct: **#359** (8-warps-on-Hopper, but a launch **deadlock**, not a silent row-swap — this is the sibling noted above) and **#2381** (also a `T.gemm` silent miscompile, but on the **K axis** — non-multiple `block_K`; here `block_K` is a clean multiple and the corruption is on the **M axis**).

ColmaLiu:
### Summary

Fixes #2522: a WGMMA `T.gemm` with more than one warpgroup along M (`block_row_warps > 4`, e.g. `FullRow` with `block_M=256, threads=256`) compiled and ran without error but silently swapped 64-row output sub-blocks across the warpgroup boundary (for `block_M=256`, rows `[64:128)` and `[128:192)` were exchanged — 128 of 256 rows wrong, deterministic).

The WGMMA instruction-issue side (`wgmma_ss_atom`) decomposes M warpgroup-major: warp `w` owns the 16-row strip at `(w // 4) * 128 * num_inst_m + inst_m * 64 + (w % 4) * 16`, matching the PTX wgmma D-fragment layout, with the per-warpgroup 64-row atoms stored as consecutive local accumulator indices (`C_offset = inst_m_idx * warp_cols * local_size_out`). `make_mma_store_layout` instead flattened all `block_row_warps` warps onto the thread axis and pushed the atom count to a
stride-`block_row_warps * 16` outer index — inverting the "which warpgroup" vs "which 64-row atom" roles. The two decompositions coincide only for a single warpgroup (`block_row_warps == 4`), so the bug was invisible in all `block_M <= 128`-per-warpgroup configs.

### Changes

- `tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py` (`make_mma_store_layout`): decompose the M axis warpgroup-major to match the instruction-issue side, built in four `repeat` steps:
  1. 4 in-group warps along M on the **thread** axis (one warpgroup, 128 rows);
  2. per-warpgroup 64-row WGMMA atoms (`warp_rows`) on the local  **index** axis, matching `C_offset` in `wgmma_ss_atom`;
  3. warpgroups along M on the **thread** axis;
  4. `block_col_warps` as the **outermost** thread repeat, preserving the physical warp order `warp_id = warp_m + block_row_warps * warp_n` from `extract_thread_binding`.

  For `block_row_warps == 4` this is algebraically identical to the previous layout (step 3 is the identity and steps 1+4 fuse into the old `[block_row_warps, block_col_warps]` thread repeat), so single-warpgroup configs are bit-for-bit unchanged. A `block_row_warps % 4 == 0` assert guards the invariant already enforced by `ComputeWgmmaWarpPartition`.

- `testing/python/issue/test_tilelang_issue_2522.py`: new regression tests (sm_90) covering each layer of the M decomposition:
  - `FullRow, block_M=256, threads=256` — two warpgroups along M (the reported config);
  - `FullRow, block_M=128, threads=128` — single warpgroup baseline;
  - `FullRow, block_M=128, threads=512` — `m_warp=8, n_warp=2`, guards the ordering of the inter-warpgroup and `warp_n` thread repeats;
  - `Square, block_M=256, block_N=16, threads=512, transpose_B` — `m_warp=8, n_warp=2, warp_rows=2`, all three M-decomposition layers active at once;

## Summary

- Fixed WGMMA `T.gemm` C-store row mapping for `GemmWarpPolicy.FullRow` when using multiple warpgroups along M, preventing silent 64-row output sub-block swapping across warpgroup boundaries (e.g., `block_M=256`).
- Reworked `make_mma_store_layout` in the WGMMA macro generators to use a warpgroup-major M decomposition (with `wg_row_warps=4`) aligned with WGMMA instruction issue order and accumulator offsets.
- Preserved the existing layout behavior for single-warpgroup configurations and added an assertion that `block_row_warps` is divisible by 4.
- Applied the same warpgroup-major fix to the sparse WGMMA (`WGSparseTensorCoreIntrinEmitter`) variant.
- Added SM90 regression tests for multi-warpgroup and single-warpgroup FullRow GEMM correctness across multiple warp arrangements and multi-atom configurations, validating compiled kernels include `"wgmma"` and match PyTorch references.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_2522.py
+++ b/testing/python/issue/test_tilelang_issue_2522.py
@@ -1,0 +1,112 @@
+"""Regression test for GitHub issue #2522."""
+
+import torch
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+from tilelang.language import GemmWarpPolicy
+
+
+def _matmul(block_M, block_N, block_K, threads, policy, pass_configs=None):
+    """Single-block GEMM: C[block_M, block_N] = A @ B over K = 2 * block_K."""
+    M, N, K = block_M, block_N, block_K * 2
+
+    @T.prim_func
+    def main(A: T.Tensor((M, K), "float16"), B: T.Tensor((K, N), "float16"), C: T.Tensor((M, N), "float16")):
+        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=threads) as (bx, by):
+            A_shared = T.alloc_shared((block_M, block_K), "float16")
+            B_shared = T.alloc_shared((block_K, block_N), "float16")
+            C_frag = T.alloc_fragment((block_M, block_N), "float32")
+            T.clear(C_frag)
+            for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=2):
+                T.copy(A[by * block_M, k * block_K], A_shared)
+                T.copy(B[k * block_K, bx * block_N], B_shared)
+                T.gemm(A_shared, B_shared, C_frag, policy=policy)
+            T.copy(C_frag, C[by * block_M, bx * block_N])
+
+    return tilelang.compile(main, out_idx=-1, pass_configs=pass_configs)
+
+
+def _check(kernel, M, N, K, seed=0):
+    torch.manual_seed(seed)
+    A = torch.randint(0, 3, (M, K), dtype=torch.float16, device="cuda")
+    B = torch.randint(0, 3, (K, N), dtype=torch.float16, device="cuda")
+    ref = A.float() @ B.float()
+    out = kernel(A, B).float()
+    torch.testing.assert_close(out, ref, atol=1e-2, rtol=1e-2)
+
+
+@tilelang.testing.requires_cuda_compute_version(9, 0)
+def test_wgmma_fullrow_two_warpgroups_along_m():
+    """FullRow, block_M=256, threads=256: m_warp=8 (two warpgroups along M).
+
+    The configuration reported in issue #2522: without the warpgroup-major
+    store layout, output rows [64:128) and [128:192) are exchanged.
+    """
+    kernel = _matmul(256, 128, 32, 256, GemmWarpPolicy.FullRow)
+    assert "wgmma" in kernel.get_kernel_source()
+    _check(kernel, 256, 128, 64)
+
+
+@tilelang.testing.requires_cuda_compute_version(9, 0)
+def test_wgmma_fullrow_single_warpgroup():
+    """FullRow, block_M=128, threads=128: single warpgroup along M."""
+    kernel = _matmul(128, 128, 32, 128, GemmWarpPolicy.FullRow)
+    assert "wgmma" in kernel.get_kernel_source()
+    _check(kernel, 128, 128, 64)
+
+
+@tilelang.testing.requires_cuda_compute_version(9, 0)
+def test_wgmma_two_warpgroups_with_col_warps():
+    """FullRow, block_M=128, threads=512: m_warp=8, n_warp=2.
+
+    Exercises block_col_warps > 1 together with multiple warpgroups along M
+    (warp_rows=1).  Guards the ordering of the inter-warpgroup and warp_n
+    thread repeats: warp_id = warp_m + block_row_warps * warp_n, so warp_n
+    must be the outermost repeat, not interleaved between the in-group warps
+    and the warpgroup index.
+    """
+    kernel = _matmul(128, 256, 32, 512, GemmWarpPolicy.FullRow)
+    assert "wgmma" in kernel.get_kernel_source()
+    _check(kernel, 128, 256, 64)
+
+
+@tilelang.testing.requires_cuda_compute_version(9, 0)
+def test_wgmma_two_warpgroups_col_warps_multi_atom():
+    """Square, block_M=256, block_N=16, threads=512: m_warp=8, n_warp=2, warp_rows=2.
+
+    All three layers of the M decomposition are active at once: multiple
+    warpgroups along M, multiple warps along N, and multiple 64-row WGMMA
+    atoms per warpgroup.  Uses K-major B (transpose_B) since block_N=16
+    N-major shared layouts are not supported.
+    """
+    M, N, K = 256, 16, 64
+    block_K, threads = 32, 512
+
+    @T.prim_func
+    def main(A: T.Tensor((M, K), "float16"), B: T.Tensor((N, K), "float16"), C: T.Tensor((M, N), "float16")):
+        with T.Kernel(1, 1, threads=threads) as (bx, by):
+            A_shared = T.alloc_shared((M, block_K), "float16")
+            B_shared = T.alloc_shared((N, block_K), "float16")
+            C_frag = T.alloc_fragment((M, N), "float32")
+            T.clear(C_frag)
+            for k in T.Pipelined(T.ceildiv(K, block_K), num_stages=2):
+                T.copy(A[0, k * block_K], A_shared)
+                T.copy(B[0, k * block_K], B_shared)
+                T.gemm(A_shared, B_shared, C_frag, transpose_B=True, policy=GemmWarpPolicy.Square)
+            T.copy(C_frag, C[0, 0])
+
+    kernel = tilelang.compile(main, out_idx=-1)
+    assert "wgmma" in kernel.get_kernel_source()
+
+    torch.manual_seed(0)
+    A = torch.randint(0, 3, (M, K), dtype=torch.float16, device="cuda")
+    B = torch.randint(0, 3, (N, K), dtype=torch.float16, device="cuda")
+    ref = A.float() @ B.float().T
+    out = kernel(A, B).float()
+    torch.testing.assert_close(out, ref, atol=1e-2, rtol=1e-2)
+
+
+if __name__ == "__main__":
+    tilelang.testing.main()
--- a/testing/python/issue/test_tilelang_issue_2522.py
+++ b/testing/python/issue/test_tilelang_issue_2522.py
@@ -37,7 +37,7 @@ def _check(kernel, M, N, K, seed=0):
     torch.testing.assert_close(out, ref, atol=1e-2, rtol=1e-2)
 
 
-@tilelang.testing.requires_cuda_compute_version(9, 0)
+@tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_wgmma_fullrow_two_warpgroups_along_m():
     """FullRow, block_M=256, threads=256: m_warp=8 (two warpgroups along M).
 
@@ -49,7 +49,7 @@ def test_wgmma_fullrow_two_warpgroups_along_m():
     _check(kernel, 256, 128, 64)
 
 
-@tilelang.testing.requires_cuda_compute_version(9, 0)
+@tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_wgmma_fullrow_single_warpgroup():
     """FullRow, block_M=128, threads=128: single warpgroup along M."""
     kernel = _matmul(128, 128, 32, 128, GemmWarpPolicy.FullRow)
@@ -57,7 +57,7 @@ def test_wgmma_fullrow_single_warpgroup():
     _check(kernel, 128, 128, 64)
 
 
-@tilelang.testing.requires_cuda_compute_version(9, 0)
+@tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_wgmma_two_warpgroups_with_col_warps():
     """FullRow, block_M=128, threads=512: m_warp=8, n_warp=2.
 
@@ -72,7 +72,7 @@ def test_wgmma_two_warpgroups_with_col_warps():
     _check(kernel, 128, 256, 64)
 
 
-@tilelang.testing.requires_cuda_compute_version(9, 0)
+@tilelang.testing.requires_cuda_compute_version_eq(9, 0)
 def test_wgmma_two_warpgroups_col_warps_multi_atom():
     """Square, block_M=256, block_N=16, threads=512: m_warp=8, n_warp=2, warp_rows=2.
 
--- a/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py
+++ b/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py
@@ -909,6 +909,20 @@ class TensorCoreIntrinEmitter(MMAIntrinEmitter):
             forward_index_fn=forward_index,
         )
         warp_n_layout = base_fragment.repeat([1, warp_cols], False, False)
-        block_layout = warp_n_layout.repeat([block_row_warps, block_col_warps], True, False)
-        warp_m_layout = block_layout.repeat([warp_rows, 1], False, False)
+        # Decompose M warpgroup-major to match the instruction-issue side:
+        # each group of 4 warps owns a contiguous 128-row tile, and the
+        # per-warpgroup 64-row WGMMA atoms are consecutive local accumulator
+        # indices (see C_offset in wgmma_ss_atom).
+        wg_row_warps = 4
+        assert block_row_warps % wg_row_warps == 0, f"block_row_warps must be a multiple of {wg_row_warps} for WGMMA, got {block_row_warps}"
+        num_warp_groups_m = block_row_warps // wg_row_warps
+        # Four M warps forming one warpgroup.
+        warpgroup_layout = warp_n_layout.repeat([wg_row_warps, 1], True, False)
+        # Per-warpgroup WGMMA M atoms are local accumulator indices.
+        warpgroup_layout = warpgroup_layout.repeat([warp_rows, 1], False, False)
+        # Repeat complete warpgroups along M.
+        warp_m_layout = warpgroup_layout.repeat([num_warp_groups_m, 1], True, False)
+        # Preserve the physical warp order:
+        # warp_id = warp_m + block_row_warps * warp_n.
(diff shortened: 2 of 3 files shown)


testing/python/issue/test_tilelang_issue_2522.py
```diff
@@ -0,0 +1,112 @@
...
+def _check(kernel, M, N, K, seed=0):
+    torch.manual_seed(seed)
+    A = torch.randint(0, 3, (M, K), dtype=torch.float16, device="cuda")
+    B = torch.randint(0, 3, (K, N), dtype=torch.float16, device="cuda")
+    ref = A.float() @ B.float()
+    out = kernel(A, B).float()
+    torch.testing.assert_close(out, ref, atol=1e-2, rtol=1e-2)
+
+
+@tilelang.testing.requires_cuda_compute_version(9, 0)
```
LeiWang1999: eq, as sm100 doesn't have wgmma.

LeiWang1999 (changes requested):
LGTM, but should also fix wgmma_sp.

ColmaLiu:
The sparse variant had the identical divergence and is fixed the same way. @LeiWang1999

testing/python/issue/test_tilelang_issue_2522.py
ColmaLiu: also fixed

LeiWang1999 (approved):

PR #2672, opened 2026-07-15
[BugFix] Return previous value for scalar atomic_min/atomic_max

### LINKED ISSUE #2667: [BUG][Fuzzer][ice-on-valid] `T.atomic_min`/`T.atomic_max` with `return_prev=True` fails to compile (`int prev = ;`) instead of returning the previous value
### Required prerequisites

0.1.11 (`cd37ed5f`)

### System information

TileLang 0.1.11 / NVIDIA L40S (sm_89) / CUDA 12.x. The defect is in host-side op selection (`tilelang/language/atomic.py`) + CUDA codegen, so it is architecture-independent.

### Problem description

A scalar `T.atomic_min(dst, value, return_prev=True)` (or `T.atomic_max`) fails to compile: codegen emits the void form `AtomicMin(&dst, value, 0);` as a *statement* and then a binding with an empty right-hand side, `int prev = ;`, which `nvcc` rejects with a syntax error. `T.atomic_add(..., return_prev=True)` on the same kernel compiles and returns the previous value correctly — so `return_prev` works for `add` but is broken for `min`/`max`.

`return_prev=True` is a documented, in-contract use of these ops: `atomic_max`'s own docstring shows `prev_value = atomic_max(counter, 50.0, return_prev=True)` and `atomic_min`'s shows `old_min = atomic_min(threshold, 3.14, return_prev=True)`. Both `int32` and `float32` reproduce.

This is not a regression — the behavior has been present since `atomic_min`/`atomic_max` were first added (see Provenance).

### Reproducible example code

```python

def build(op):                       # op in {"atomic_min", "atomic_max", "atomic_add"}
    atom = getattr(T, op)
    @tilelang.jit(out_idx=[1])
    def k():
        @T.prim_func
        def main(src: T.Tensor((256,), "int32"), acc: T.Tensor((1,), "int32")):
            with T.Kernel(1, threads=256):
                tx = T.get_thread_binding(0)
                prev = atom(acc[0], src[tx], return_prev=True)   # scalar, return_prev=True
                src[tx] = prev                                    # use the returned value
        return main
    return k()

for op in ("atomic_add", "atomic_min", "atomic_max"):
    try:
        build(op); print(f"{op}: COMPILED")
    except Exception:
        print(f"{op}: COMPILE-CRASH")

# atomic_add:  COMPILED          (control — return_prev works)
# atomic_min:  COMPILE-CRASH     (emits `int prev = ;`)
# atomic_max:  COMPILE-CRASH
```

### Traceback

```
Compilation error:
    emitted: AtomicMin((&(acc[0])), src[((int)threadIdx.x)], 0);
tvm_kernels.cu(22): error: expected an expression
    int prev = ;
               ^
```

### Expected behavior

`atomic_min`/`atomic_max` with `return_prev=True` should return the previous value, exactly as `atomic_add` already does — the scalar `return_prev` path is behaviorally identical across the three ops and the docstrings promise it.

### Additional context

**Root cause.** The scalar path of `atomic_max`/`atomic_min` always selects the *void* intrinsic regardless of `return_prev`, whereas `atomic_add` selects the return-value intrinsic when `return_prev=True`.

<details>
<summary>mechanism + line pointers</summary>

- `atomic_max` scalar path sets `return_type = dst.dtype if return_prev else "handle"` but hardcodes `op.Op.get("tl.atomic_max_elem_op")` (the void op) — [atomic.py#L66-L75](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/atomic.py#L66-L75). `atomic_min` is identical — [atomic.py#L148-L157](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/atomic.py#L148-L157).
- `atomic_add` does it correctly: `op.Op.get("tl.atomic_add_ret_elem_op") if return_prev else op.Op.get("tl.atomic_add_elem_op")` — [atomic.py#L235](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/atomic.py#L235).
- In CUDA codegen the void `atomic_max_elem_op` prints via `this->stream` as a statement (no expression returned), while `atomic_max_ret_elem_op` prints via `os` as an expression — [codegen_cuda.cc#L4387-L4428](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L4387-L4428). Because the call was typed with a non-void `return_type` but lowered to the void op, the generated binding has no RHS → `int prev = ;`.

</details>

**Suggested fix.** Mirror `atomic_add`: in the scalar path of `atomic_max`/`atomic_min`, select `tl.atomic_max_ret_elem_op` / `tl.atomic_min_ret_elem_op` when `return_prev=True`. These ops already exist and are already handled downstream — the CUDA templates define `AtomicMaxRet`/`AtomicMinRet` ([atomic.h#L283](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/atomic.h#L283)), codegen has the emitter (`os << "AtomicMaxRet(...)"`), and `LegalizeSafeMemoryAccess`/`ThreadSync` already recognize them; the Python scalar path is the only place they are never selected. (Suggested, not built.)

**Provenance.** Introduced by [#1676](https://github.com/tile-ai/tilelang/pull/1676) ("Atomic Reduction Operations and Vectorization Enhancement", merged 2026-01-15), which added `atomic_max`/`atomic_min` with the `return_prev` parameter and the `return_type = dst.dtype if return_prev` line but hardcoded the void `*_elem_op` in the scalar path — the same PR even defined `atomic_max_ret_elem_op`/`atomic_min_ret_elem_op`, but the Python path never selects them. Broken since these functions first shipped; not a regression. I searched the open and closed tracker and found no existing report of this defect.

**Reach.** Triggering requires the scalar (single-element) form of `atomic_min`/`atomic_max` with `return_prev=True`. The tile-region form already raises a clear `NotImplementedError` for `return_prev`, so only the scalar form reaches this path — and it fails for every dtype/arch, not an edge case. The shipped atomic test uses `atomic_min`/`atomic_max` only in their void form (`test_tilelang_language_atomic.py`, no `return_prev`), and the one `return_prev` test exercises only `atomic_add`, so CI never compiles a `min`/`max` + `return_prev` kernel.

Hughshine:
Scalar `atomic_min`/`atomic_max` and vectorized `atomic_addx2`/`atomic_addx4` typed their call with a non-void return under `return_prev=True` but always selected the void `*_elem_op`, so codegen emitted an empty binding (`prev = ;`) and nvcc failed with "expected an expression".

This selects the `*_ret_elem_op` variant under `return_prev`, as `atomic_add` already does:

- **min/max**: the ret ops, codegen handlers, and device templates (`AtomicMaxRet`/`AtomicMinRet`) already existed.
- **addx2/addx4, float32**: register the ret ops and add codegen handlers emitting `AtomicAddx2Ret`/`AtomicAddx4Ret` (already in the device headers) as expressions; the return is typed as the packed 2-/4-lane vector so the store matches.
- **addx2, fp16/bf16**: the ret store LHS is `uint1`, so codegen wraps the `half2`/`bf162` result with `tl::to_uint1`.
- **addx4, fp16/bf16**: add `AtomicAddx4Ret(half_t*)`/`(bfloat16_t*)` device overloads that run two `AtomicAddx2Ret` halves and return them packed as `uint2` (a fp16x4 is stored as `uint2`).

The 16-bit x4 path is **per-pair atomic** (two `half2` atomics), not whole-quad single-atomic — hardware has no vector atomic wider than `.f16x2` for 16-bit types. This matches the existing fp32-x4 scalar fallback, and the docstring only promises "quad-width atomic addition". Tests cover scalar/x2/x4 across float32/fp16/bf16.

Fixes #2667

## Summary

- Fixed scalar `atomic_min`/`atomic_max` with `return_prev=True` by selecting the correct `*_ret_elem_op` intrinsic variants, preventing invalid generated CUDA such as `prev = ;` for both integer and floating-point cases.
- Extended `atomic_addx2` and `atomic_addx4` with `return_prev=True` by registering new `atomic_addx{2,4}_ret_elem_op` intrinsics and adding CUDA codegen handlers that emit `AtomicAddx{2,4}Ret(...)` (including optional `memory_order`), returning packed vector types for the previous values.
- Updated `tilelang/language/atomic.py` to route `return_prev=True` through the new `*_ret_elem_op` intrinsics and to use packed vector return types for the previous-value result.
- Added CUDA template support for fp16/bf16 `atomic_addx4` previous-value returns (`AtomicAddx4Ret`) via two `AtomicAddx2Ret` calls and repacking into `uint2`; for `atomic_addx2` fp16/bf16, codegen additionally uses `tl::to_uint1(...)` bridging to match storage representation.
- Added/extended Python tests to validate previous-value correctness for scalar `atomic_min`/`atomic_max` (including the min/max test adjustment), and for vectorized `atomic_addx2`/`atomic_addx4` `return_prev` across float32, fp16, and bf16, including kernel codegen marker-string checks (`AtomicAddx2Ret`/`AtomicAddx4Ret`).
- Added legalization coverage for `atomic_addx2`/`atomic_addx4` `return_prev` on `access_ptr` bases, including an OOB guard test.

## C++ style / lint notes

- Touches C++ implementation files (`src/cuda/codegen/codegen_cuda.cc`, `src/op/builtin.cc/h`, `src/tl_templates/cuda/atomic.h`, `src/transform/legalize_safe_memory_access.cc`), but does **not** modify `docs/developer_guide/cpp_style.md` (no style-rule documentation changes).
- The “C++ API Style Audit (warning only)” CI step may surface advisory TLCPP003/TLCPP004 findings; any such warnings are expected to be non-blocking unless this PR introduces an actual API/FFI/maintainability risk.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -4629,6 +4629,24 @@ bool CodeGenTileLangCUDA::HandleLateIntrinsicCall(const CallNode *op,
     }
     this->stream << ");\n";
     return true;
+  } else if (op->op.same_as(tl::atomic_addx2_ret_elem_op())) {
+    need_atomic_h_ = true;
+    // fp16/bf16 x2 is stored as uint1 but AtomicAddx2Ret returns half2/bf162;
+    // bridge with to_uint1. float32 is native float2, emitted bare.
+    bool need_cast = op->dtype.is_float16() || op->dtype.is_bfloat16();
+    if (need_cast) {
+      os << "tl::to_uint1(";
+    }
+    os << "AtomicAddx2Ret(" << PrintExpr(op->args[0]) << ", "
+       << PrintExpr(op->args[1]);
+    if (op->args.size() > 2) {
+      os << ", " << PrintExpr(op->args[2]);
+    }
+    os << ")";
+    if (need_cast) {
+      os << ")";
+    }
+    return true;
   } else if (op->op.same_as(tl::atomic_addx4_elem_op())) {
     need_atomic_h_ = true;
     // atomic_addx4_elem_op(dst_ptr, src_ptr[, memory_order])
@@ -4640,6 +4658,16 @@ bool CodeGenTileLangCUDA::HandleLateIntrinsicCall(const CallNode *op,
       this->stream << ", " << PrintExpr(op->args[2]);
     }
     this->stream << ");\n";
+    return true;
+  } else if (op->op.same_as(tl::atomic_addx4_ret_elem_op())) {
+    need_atomic_h_ = true;
+    // AtomicAddx4Ret already returns the store type (float4 / uint2), so bare.
+    os << "AtomicAddx4Ret(" << PrintExpr(op->args[0]) << ", "
+       << PrintExpr(op->args[1]);
+    if (op->args.size() > 2) {
+      os << ", " << PrintExpr(op->args[2]);
+    }
+    os << ")";
     return true;
   } else if (op->op.same_as(tl::atomic_load_elem_op())) {
     need_atomic_h_ = true;
--- a/src/op/builtin.cc
+++ b/src/op/builtin.cc
@@ -615,7 +615,17 @@ TIR_DEFINE_TL_BUILTIN(atomic_addx2_elem_op)
     .set_attr<TCallEffectKind>("TCallEffectKind",
                                Integer(CallEffectKind::kOpaque));
 
+TIR_DEFINE_TL_BUILTIN(atomic_addx2_ret_elem_op)
+    .set_num_inputs(3)
+    .set_attr<TCallEffectKind>("TCallEffectKind",
+                               Integer(CallEffectKind::kOpaque));
+
 TIR_DEFINE_TL_BUILTIN(atomic_addx4_elem_op)
+    .set_num_inputs(3)
+    .set_attr<TCallEffectKind>("TCallEffectKind",
+                               Integer(CallEffectKind::kOpaque));
+
+TIR_DEFINE_TL_BUILTIN(atomic_addx4_ret_elem_op)
     .set_num_inputs(3)
     .set_attr<TCallEffectKind>("TCallEffectKind",
                                Integer(CallEffectKind::kOpaque));
--- a/src/op/builtin.h
+++ b/src/op/builtin.h
@@ -1117,12 +1117,30 @@ TVM_DLL const Op &atomic_add_ret_elem_op();
 TVM_DLL const Op &atomic_addx2_elem_op();
 
 /*!
+ * \brief tilelang intrinsic for vectorized (x2) atomic addition with return
+ * value.
+ *
+ *  This op is used to represent a vectorized atomic add operation (2 elements)
+ * in tilelang that returns the previous packed value.
+ */
+TVM_DLL const Op &atomic_addx2_ret_elem_op();
+
+/*!
  * \brief tilelang intrinsic for vectorized (x4) atomic addition.
  *
  *  This op is used to represent a vectorized atomic add operation (4 elements)
  * in tilelang.
  */
 TVM_DLL const Op &atomic_addx4_elem_op();
+
+/*!
+ * \brief tilelang intrinsic for vectorized (x4) atomic addition with return
+ * value.
+ *
+ *  This op is used to represent a vectorized atomic add operation (4 elements)
+ * in tilelang that returns the previous packed value.
+ */
+TVM_DLL const Op &atomic_addx4_ret_elem_op();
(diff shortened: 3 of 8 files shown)


Hughshine:
Good catch — adopted. The min case now uses `val=1.0` so `min(3, 1)=1` differs from the initial `3.0`, making the dst-update assertion meaningful (verified: dst goes 3→1).

PR #2674, opened 2026-07-15
[BugFix][WS] Fix pipeline replacement under persistent T.serial

### LINKED ISSUE #2548: [BUG][Fuzzer][ice-on-valid-code] Regression (0.1.10): the Hopper warp-specialization pass hard-aborts on a `T.Pipelined` loop inside a `while` with an internal `ICHECK`, instead of declining WS and falling back to the standard pipeline as it did on 0.1.9
### Required prerequisites

0.1.11

### System information

- TileLang `0.1.11` (source pinned to tag `v0.1.11`, commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`).
- Python 3.12.3, linux; PyTorch 2.12.1+cu130 (CUDA 13.0).
- GPU: NVIDIA H100 80GB HBM3 (compute capability 9.0, sm_90). The defect is on the Hopper warp-specialization (WS) lowering path — it triggers only when WS is active (sm_90) and disappears when WS is disabled (control below).

### Problem description

On Hopper, a `num_stages` `T.Pipelined` loop nested inside a `while` loop aborts compilation with an internal invariant assertion:

```
tvm.error.InternalError: Check failed: (replaced.defined()) is false: ProducerConsumerWS: failed to replace pipeline loop
```

This is a **regression in the failure mode**: on `0.1.9` this kernel compiles, because WS quietly declines the `while`-enclosed loop and lets the standard `cp.async` pipeline lower it; on `0.1.10`/`0.1.11` the same situation hard-aborts on an internal `ICHECK`. (To be precise: WS *itself* was never applied under a `while` on any version — `0.1.9` silently skipped it and fell back; `0.1.10` began trying to handle the `while` and crashes partway. So the ask is to restore the graceful skip, not to newly support WS-under-`while`.) The `while`-enclosing-`T.Pipelined` shape is the core of a stream-K GEMM, and the shipped `gemm_streamk` example is itself affected: running [`examples/gemm_streamk/example_tilelang_gemm_streamk.py`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/gemm_streamk/example_tilelang_gemm_streamk.py) verbatim on sm_90 aborts with this exact `ICHECK` (its test is gated `@requires_cuda_compute_version_le(8, 9)`, so CI never exercises it on Hopper). The minimal reproducer below is a trimmed version of that kernel.

The kernel is well-formed — with WS disabled it compiles and runs correct, and the identical pipeline loop warp-specializes fine when it is *not* wrapped in a `while` (both controls below). `ICHECK(replaced.defined())` is an internal-consistency guard and should not be reachable from a user kernel: when WS can't rewrite this shape it should **decline WS and fall back to the standard `cp.async` pipeline** (as it already does for an `if`-wrapped loop, #2184, and as `0.1.9` did here), or emit a clear diagnostic.

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import torch

M = N = 128; K = 256
block_M = block_N = 64; block_K = 64
dtype = "float16"; accum = "float32"
iters = K // block_K


def build(num_stages, pass_configs):
    @T.prim_func
    def main(A: T.Tensor((M, K), dtype), B: T.Tensor((N, K), dtype), C: T.Tensor((M, N), "float32")):
        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128) as (bx, by):
            B_shared = T.alloc_shared((block_N, block_K), dtype)
            C_local = T.alloc_fragment((block_M, block_N), accum)
            start_iter = T.alloc_fragment((1,), "int32", "local")
            end_iter = T.alloc_fragment((1,), "int32", "local")
            start_iter[0] = 0
            while start_iter[0] < iters:  # stream-K core: while advances K-tiles
                end_iter[0] = T.min(start_iter[0] + 1, iters)
                T.clear(C_local)
                for k in T.Pipelined(end_iter[0] - start_iter[0], num_stages=num_stages):
                    kk = k + start_iter[0]
                    T.copy(A[by * block_M, kk * block_K], A_shared)
                    T.copy(B[bx * block_N, kk * block_K], B_shared)
                    T.gemm(A_shared, B_shared, C_local, transpose_B=True)
                    T.atomic_add(C[by * block_M + i, bx * block_N + j], C_local[i, j])
                start_iter[0] = end_iter[0]
    return tilelang.compile(main, out_idx=[], target="cuda", pass_configs=pass_configs)


# Hopper, WS on (default): aborts at compile time.
build(2, {})
# -> InternalError: Check failed: (replaced.defined()) is false:
#    ProducerConsumerWS: failed to replace pipeline loop

# Control: same kernel, WS disabled -> compiles and runs correct (maxdiff 0.0).
kernel = build(2, {"tl.disable_warp_specialized": True})
a = torch.randint(0, 3, (M, K), device="cuda").half()
b = torch.randint(0, 3, (N, K), device="cuda").half()
c = torch.zeros(M, N, device="cuda", dtype=torch.float32)
kernel(a, b, c); torch.cuda.synchronize()
print((c - (a.float() @ b.float().T)).abs().max().item())  # -> 0.0
```

<details>
<summary>Second control: the same pipeline loop with no enclosing <code>while</code> warp-specializes correctly</summary>

Replacing the `while` with the pipeline loop directly under `T.Kernel` (same shapes, WS on) compiles and the WS+TMA path activates (`feats={'wgmma': True, 'mbarrier': True}`), `maxdiff=0.0`. So it is specifically the `while` enclosure, not the pipeline loop, that defeats the pass.
</details>

### Traceback

```pytb
tvm.error.InternalError: Check failed: (replaced.defined()) is false: ProducerConsumerWS: failed to replace pipeline loop


raised at [`producer_consumer_ws.cc#L1940-L1944`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L1940-L1944).
```

### Expected behavior

A well-formed kernel should not reach an internal `ICHECK`. When `ProducerConsumerWS` cannot rewrite this shape it should decline WS and fall back to the standard `cp.async` pipeline (the behavior TileLang already shows for the `if`-wrapped case #2184, and the behavior of `0.1.9` before this regressed), or emit a clear diagnostic.

### Additional context

**Root cause (one line):** in `ProducerConsumerWSRewriter`, the loop **finder** descends into `While` bodies but the loop **matcher/replacer** does not — so the pass commits to a WS rewrite of a loop it then cannot re-locate, and asserts.

<details>
<summary>Root cause detail (finder vs. matcher/replacer traversal of <code>While</code>)</summary>

`PipelineLoopFinder` derives from `StmtVisitor`, whose default recursion descends into `While` bodies, so it locates the `num_stages` `For` even inside a `While` and commits the pass to a rewrite ([`#L2089-L2111`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L2089-L2111)). The pass then calls `ReplacePipelineLoopInStmt` and asserts the result is defined ([`#L1940-L1944`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L1940-L1944)). But the two helpers that do the replacement — `PipelineLoopContainmentChecker` ([`#L2269-L2319`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L2269-L2319)) and `PipelineLoopInStmtReplacer` ([`#L2321-L2429`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L2321-L2429)) — are `StmtFunctor`s with explicit `VisitStmt_` handlers only for `SeqStmt`/`BlockRealize`/`Block`/`AttrStmt`/`IfThenElse` and **no `WhileNode` handler**. The pipeline `For` is matched by pointer identity, so reaching it requires descending through every enclosing node; a `While` hits `VisitStmtDefault_`, returns the not-found sentinel, and the `SeqStmt` `loop_idx < 0` bail ([`#L2342-L2353`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L2342-L2353)) yields the undefined result the `ICHECK` treats as fatal.
</details>

<details>
<summary>Regression provenance — bisected to #2231 (0.1.9 → 0.1.10), tag-ancestry verified</summary>

| version | `while` + WS on (default) |
|---------|---------------------------|
| `0.1.9`  | **compiles** (WS silently declines) |
| `0.1.10` / `0.1.11` | **CRASH** `ICHECK(replaced.defined())` |

`0.1.9`'s hand-rolled `FindPipelineLoop` dispatched on `For`/`SeqStmt`/`LetStmt`/`BlockRealize`/`Block`/`AttrStmt` and returned `nullptr` otherwise; finder and replacer were **symmetric** (both blind to `While`), so a `while`-enclosed loop fell through to the non-WS path. [#2231](https://github.com/tile-ai/tilelang/pull/2231) "Refactor TIR statement traversal helpers" (merged 2026-05-20, first shipped `0.1.10`) replaced the finder with `PipelineLoopFinder : StmtVisitor` (which recurses into `While`) but introduced the matcher/replacer as `StmtFunctor`s with no `While` handler — the asymmetry that fires the `ICHECK`. (`git log -S'class PipelineLoopContainmentChecker'` locates these classes at that commit; `grep WhileNode producer_consumer_ws.cc` = 0 at `cd37ed5f`.) #2297 later relocated the file without changing the traversal.
</details>

**Suggested fix — decline WS and fall back.** At the call site ([`#L1940-L1944`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L1940-L1944)), when `ReplacePipelineLoopInStmt` returns `!defined()`, skip the WS rewrite and keep the original block so a later pass lowers it via the standard `cp.async` pipeline — the decline-and-fall-back accepted for #2184, and exactly what `0.1.9` did here. (Equivalently, gate WS off when the pipeline `For` is `While`-enclosed.) This restores the graceful skip rather than trying to warp-specialize under a `while` — which is out of scope here: I verified that simply teaching the matcher/replacer to descend into `While` compiles but then fails at **launch** (`unspecified launch failure`, i.e. #972), because WS partitioning doesn't handle the loop-carried `int32` fragments mutated across iterations. So the correct fix is to decline, not to support.

**Not a duplicate.** No tracker report of a `while`-enclosed pipeline loop aborting `ProducerConsumerWS` at `ICHECK(replaced.defined())`. Nearest neighbors are distinct: **#2184** (closed) is the `if`-wrapped case (matcher/replacer *do* handle `if`) that silently declines WS rather than aborting — cited as the fall-back precedent; **#972** (closed) is a *runtime* failure in the same example (`unspecified launch failure`), not this compile-time `ICHECK`.

RuneFang:
## Summary

Fix `ProducerConsumerWarpSpecialized` (WS) aborting with `ProducerConsumerWS: failed to replace pipeline loop` whenever a `T.Pipelined` K-loop sits underneath  outer `ForNode` — most commonly the two idiomatic persistent-kernel patterns in TileLang.
image

## Motivation

TileLang has two standard ways to write a persistent kernel, and **both
currently break the WS pass**.

### Pattern A — `T.Persistent` primitive 
```python
with T.Kernel(sm_num, threads=threads) as (block_id,):
    for by, bx in T.Persistent(
            [T.ceildiv(M, block_M), T.ceildiv(N, block_N)],
            sm_num, block_id):
        T.clear(C_local)
        for ko in T.Pipelined(T.ceildiv(K, block_K), num_stages=num_stages):
            T.copy(A[by * block_M, ko * block_K], A_shared)
            T.copy(B[ko * block_K, bx * block_N], B_shared)
        T.copy(C_local, C[by * block_M, bx * block_N])
```

### Pattern B — hand-written `T.serial` tile scheduler 

```python
with T.Kernel(sm_num, threads=threads) as (block_id,):
    for w in T.serial(waves):
        tile_id = sm_num * w + block_id
        bx = (tile_id // group_size) % m_blocks
        by = (tile_id %  group_size) + (tile_id // group_size) // m_blocks * group_size
        if bx * block_M < M and by * block_N < N:
            T.clear(C_local)
            for ko in T.Pipelined(T.ceildiv(K, block_K), num_stages=num_stages):
                ...
```

## Root Cause

`ProducerConsumerWSRewriter` locates and rewrites the pipeline loop in two steps:

1. **Discovery** — `PipelineLoopFinder::Find(orig_block->body)` (a`StmtVisitor`) finds the inner `T.Pipelined` `For` marked with `num_stages`. It correctly recurses through arbitrary `ForNode`s because `StmtVisitor` provides a default `VisitStmt_` for every node.
2. **Replacement** — `ReplacePipelineLoopInStmt(...)` swaps that loop with a `dummy_ws` placeholder using two `StmtFunctor<Ret(Stmt)>`
   subclasses:
   - `PipelineLoopContainmentChecker` (returns `bool`)
   - `PipelineLoopInStmtReplacer` (returns `Optional<Stmt>`)

Both step-2 helpers override `VisitStmt_` for `SeqStmtNode`,`SBlockRealizeNode`, `SBlockNode`, `AttrStmtNode`, and
`IfThenElseNode`, but **neither overrides `VisitStmt_(const ForNode*)`**.Unlike `StmtVisitor`, `StmtFunctor` has no per-node default: unknown nodes hit `VisitStmtDefault_`, which returns `false` / empty `Optional`.

Consequence: as soon as the target pipeline loop sits inside **any** outer `For` (whether synthesized by `T.Persistent` or written by hand via `T.serial`), the containment checker reports "no pipeline loop here", the replacer returns an empty `Optional`, and the outer `ICHECK(replaced.defined())` fires.

This is a genuine **coverage gap** — `PipelineLoopFinder` already treats outer `For`s as first-class, so the downstream helpers were expected to be symmetric.

## Fix

Add the missing `VisitStmt_(const ForNode*)` override to both helpers, mirroring the shape of the existing `AttrStmtNode` handlers.

`PipelineLoopContainmentChecker` — just recurse into the body; the
`same_as` check at the entry point catches the target loop:

```cpp
bool VisitStmt_(const ForNode *op) final { return VisitStmt(op->body); }
```

`PipelineLoopInStmtReplacer` — recurse into the body, and if it was rewritten, clone the enclosing `For` via `CopyOnWrite()` and re-attach the new body:

```cpp
Optional<Stmt> VisitStmt_(const ForNode *op) final {
  Optional<Stmt> body = VisitStmt(op->body);
  if (!body.defined()) {
    return Optional<Stmt>();
  }
  For new_for = GetRef<For>(op);
  new_for.CopyOnWrite()->body = body.value();
  return new_for;
}
```

## Tests

Added
[`testing/python/transform/test_tilelang_transform_producer_consumer_ws_persistent.py`](../../testing/python/transform/test_tilelang_transform_producer_consumer_ws_persistent.py)
with five regression tests:

## Summary

- Fixed `ProducerConsumerWarpSpecialized` aborts when a `T.Pipelined` K-loop is nested inside `ForNode` constructs, including loops produced by `T.Persistent` and hand-written `T.serial` schedulers.
- Updated pipeline-loop containment checking and in-statement replacement to explicitly recurse into `ForNode` bodies and correctly rebuild enclosing `For` nodes when replacement occurs inside them.
- Improved persistent warp specialization by hoisting phase-counter initialization out of the per-pipeline loop when the pipelined `For` is under an outer persistent/serial scheduler, disabling the prior per-pipeline `WrapLoopWithAlloc` path in that nested case.
- Reworked stage/parity expression rewriting to avoid rewriting unrelated user expressions: replaced the older loop-var-based stage expression replacer with `MVBStageIndexReplacer`, backed by a new internal intrinsic `tl.mvb_stage_index` (`mvb_stage_index` in `src/op/builtin.{h,cc}`), plus a final cleanup sweep to remove leftover `mvb_stage_index` wrappers.
- Updated `MultiVersionBufferRewriter` pipeline `ForNode` stage/version index handling to route version indices through the new `mvb_stage_index` flow.
- Added five regression tests covering persistent-kernel patterns (including `T.Persistent`, persistent serial variants with symbolic outer bounds, misaligned K, guarded inner K-loop variants, `num_stages==1`, and a deep guarded pipeline case) with CUDA end-to-end correctness checks and WS-vs-non-WS comparisons for the WS-specific regressions.

## C++ style / lint notes

- Touches C++ implementation code (`src/cuda/transform/producer_consumer_ws.cc`, `src/cuda/transform/multi_version_buffer_rewriter.cc`, and the internal builtin definitions in `src/op/builtin.{h,cc}`) but does not change the documented rules in `docs/developer_guide/cpp_style.md`.
- No CI/lint tooling or style-guide documents are modified.
- The “C++ API Style Audit (warning only)” CI step should remain applicable; this PR is correctness-focused and does not introduce any clear API/FFI/maintainability red flags that would be expected to generate new warning-only findings.

Close https://github.com/tile-ai/tilelang/issues/2548


RuneFang:
May this PR could close https://github.com/tile-ai/tilelang/issues/2548

reoLantern:
I ran into exactly the problem this PR targets — I wanted to write a GEMM that combines an **outer `T.Persistent` loop + inner `T.Pipelined` K-loop + automatic warp specialization** on Hopper (sm_90), and on the released versions it aborts with:

```
ProducerConsumerWS: failed to replace pipeline loop
```

I found this PR, cherry-picked the fix commit onto current `main`, built from source, and tested on an H100. The good news: **this PR does fix the crash and does apply WS** (mbarrier + producer/consumer split + `warpgroup_reg_alloc/dealloc` all show up). But I believe **the fix is not yet correct** — for the general persistent case it produces wrong results, and in some configs a hard `cudaErrorLaunchFailure`.

## Summary

The WS lowering now descends into the outer persistent `For`, but it does **not reconcile the mbarrier phase across outer-loop iterations**. The kernel is correct **only** when either:
- the grid covers all tiles in a single wave (each block runs the persistent body once), **or**
- the K-loop trip count is a multiple of the pipeline phase period `2 * num_stages`.

Otherwise every tile computed in the **2nd or later persistent iteration (wave)** is wrong, or the kernel crashes.

## Reproduction

Environment: H100 PCIe (sm_90), CUDA 12.4, built from this branch cherry-picked onto `main` (`tilelang 0.1.12+cuda.gitaff4479d`), `torch 2.13.0`. Config: `BLOCK_M=128, BLOCK_N=256, BLOCK_K=64, threads=256, num_stages=3` (auto-WS on by default).

```python
import torch, tilelang
import tilelang.language as T
from tilelang.carver.arch import driver
sm_num = driver.get_num_sms()

@tilelang.jit(out_idx=[2])
def matmul(M, N, K, bm, bn, bk, threads, ns, dt="float16", ac="float"):
    @T.prim_func
    def main(A: T.Tensor((M, K), dt), B: T.Tensor((K, N), dt), C: T.Tensor((M, N), dt)):
        with T.Kernel(sm_num, threads=threads) as bid:
            As = T.alloc_shared((bm, bk), dt); Bs = T.alloc_shared((bk, bn), dt)
            Cl = T.alloc_fragment((bm, bn), ac); Cs = T.alloc_shared((bm, bn), dt)
            for bx, by in T.Persistent([T.ceildiv(M, bm), T.ceildiv(N, bn)], sm_num, bid):
                T.clear(Cl)
                for kk in T.Pipelined(T.ceildiv(K, bk), num_stages=ns):
                    T.copy(A[bx*bm, kk*bk], As); T.copy(B[kk*bk, by*bn], Bs)
                    T.gemm(As, Bs, Cl)
                T.copy(Cl, Cs); T.copy(Cs, C[bx*bm, by*bn])
    return main

def check(M, N, K):
    k = matmul(M, N, K, 128, 256, 64, 256, 3)
    a = torch.randn(M, K, dtype=torch.float16, device="cuda")
    b = torch.randn(K, N, dtype=torch.float16, device="cuda")
    md = (k(a, b).float() - (a @ b).float()).abs().max().item()
    trip = (K + 63) // 64
    tiles = ((M+127)//128) * ((N+255)//256)
    waves = (tiles + sm_num - 1) // sm_num
    print(f"M={M} N={N} K={K}  trip={trip} trip%6={trip%6}  waves={waves}  max_diff={md:8.3f}")

check(1024, 1024, 2048)  # 32 tiles, waves=1              -> 0.000   OK
check(2048, 2048, 1152)  # trip=18, 18%6=0, waves=2       -> 0.000   OK
check(2048, 2048, 1536)  # trip=24, 24%6=0, waves=2       -> 0.000   OK
check(2048, 2048, 1024)  # trip=16, 16%6=4, waves=2       -> ~81     WRONG
check(4096, 4096, 4096)  # trip=64, 64%6=4, waves=5       -> ~83     WRONG
check(2048, 2048, 1280)  # trip=20, 20%6=2, waves=2       -> cudaErrorLaunchFailure
```

Observed on my machine (`num_sms=114`):

| M,N,K | K-trip | trip % (2·num_stages) | waves | result |
|---|---|---|---|---|
| 1024,1024,2048 | 32 | — | 1 | ✅ max_diff 0.000 |
| 2048,2048,1152 | 18 | 0 | 2 | ✅ max_diff 0.000 |
| 2048,2048,1536 | 24 | 0 | 2 | ✅ max_diff 0.000 |
| 2048,2048,1024 | 16 | 4 | 2 | ❌ max_diff 81.4 |
| 4096,4096,4096 | 64 | 4 | 5 | ❌ max_diff 82.9 (228/512 tiles) |
| 2048,2048,1280 | 20 | 2 | 2 | 💥 `cudaErrorLaunchFailure` |

For the 2-wave failing case the wrong tiles are **exactly** the `tiles - sm_num` tiles that fall into the second wave; the first wave is always correct. Disabling WS (`tl.disable_warp_specialized=True`) makes all of the above correct.

## Root cause

In the generated CUDA, the mbarriers are initialized **once, before** the persistent loop and are never re-initialized per wave, while the producer/consumer phase parity is derived **only from the inner K index**:

```cpp
    mbarrier[0].init(1); mbarrier[1].init(1); mbarrier[2].init(1);
    mbarrier[3].init(256); mbarrier[4].init(256); mbarrier[5].init(256);
}
tl::fence_barrier_init();
__syncthreads();
for (int w = 0; w < 5; ++w) {                              // persistent loop: no barrier reset here
    ...
    mbarrier[((kk % 3) + 3)].wait((((kk % 6) / 3) ^ 1));   // producer phase = f(kk) only
    ...
    mbarrier[(kk_1 % 3)].wait(((kk_1 % 6) / 3));           // consumer phase = f(kk_1) only
}
```

The phase argument `(kk % 6) / 3` (period `2 * num_stages = 6` for `num_stages=3`) assumes each K-loop starts from a fresh barrier phase. But the **hardware mbarrier phase persists across the outer persistent loop**. After a wave runs `trip` K-iterations, the barriers are left phase-advanced by `trip mod (2*num_stages)`. Unless that is `0`, the next wave's `wait(f(kk))` uses the wrong parity, so producers/consumers desync — giving wrong results, or (when the mismatch stalls the handshake) an illegal-state launch failure. When `trip % (2*num_stages) == 0` the phase happens to realign at the wave boundary, which is why those cases pass.

## Suggested direction

The producer/consumer split under an outer loop needs the pipeline barrier phase to be consistent across iterations. Either:
- re-initialize the mbarriers (with `fence_barrier_init` + a barrier) at the top of each persistent iteration, or
- carry the phase parity across the outer loop instead of recomputing it from the inner index alone (e.g. seed the phase with the accumulated `iter * trip` count).

Lyscoria:
In addition to the issue already reported, the persistent-loop path newly enabled by this PR appears to expose two additional correctness problems involving guarded pipelines. (Both ran on H100 PCIe, `sm_90`)

### 1. The guarded phase counter is reset for every persistent tile

With `guard(k) = k < 2`, the first tile is a correctness control, while the second tile fails:

```python
import torch, tilelang
import tilelang.language as T

M = N = 64
K, BK = 128, 32

def matmul():
    @T.prim_func
    def main(A: T.Tensor((2, M, K), "float16"), B: T.Tensor((2, K, N), "float16"), C: T.Tensor((2, M, N), "float16")):
        with T.Kernel(1, threads=128) as bid:
            As = T.alloc_shared((M, BK), "float16")
            Bs = T.alloc_shared((BK, N), "float16")
            Cl = T.alloc_fragment((M, N), "float32")
            for tile, _ in T.Persistent([2, 1], 1, bid):
                T.clear(Cl)
                for k in T.Pipelined(K // BK, num_stages=2):
                    if k < 2:
                        T.copy(A[tile, 0, k*BK], As)
                        T.copy(B[tile, k*BK, 0], Bs)
                        T.gemm(As, Bs, Cl)
                T.copy(Cl, C[tile, 0, 0])
    return main

torch.manual_seed(0)
a = torch.randn(2, M, K, dtype=torch.float16, device="cuda") * 0.125
b = torch.randn(2, K, N, dtype=torch.float16, device="cuda") * 0.125
ref = a[:, :, :2*BK].float() @ b[:, :2*BK, :].float()

func = matmul()
for label, disable_ws in (("control_ws_off", True), ("repro_ws_on", False)):
    kernel = tilelang.compile(
        func,
        out_idx=[2],
        pass_configs={
            tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: disable_ws,
        },
    )
    c = kernel(a, b).float()
    print(f"{label}_tile0_diff=", (c[0] - ref[0]).abs().max().item())
    print(f"{label}_tile1_diff=", (c[1] - ref[1]).abs().max().item())
```

```text
control_ws_off_tile0_diff= 0.00012058019638061523
control_ws_off_tile1_diff= 0.00012093782424926758
repro_ws_on_tile0_diff= 0.00012058019638061523
repro_ws_on_tile1_diff= 0.654445469379425
```

**Root cause**: `PhaseCounter::WrapLoopWithAlloc` initializes the counter immediately around the inner pipeline loop. After this PR re-inserts that loop under the persistent `For`, the counter is reset to zero on every tile. The mbarriers, however, are block-scoped and initialized only once. For two stages, tile 1 should continue with parity 1 after tile 0's two transactions, but it restarts with parity 0 and can consume stale data.

### 2. The barrier stage and shared-buffer version use different clocks

With `guard(k) = (k == 0 or k == 2)`, even the first tile fails:

```python
import torch, tilelang
import tilelang.language as T

M = N = 64
K, BK = 128, 32

def matmul():
    @T.prim_func
    def main(A: T.Tensor((2, M, K), "float16"), B: T.Tensor((2, K, N), "float16"), C: T.Tensor((2, M, N), "float16")):
        with T.Kernel(1, threads=128) as bid:
            As = T.alloc_shared((M, BK), "float16"); Bs = T.alloc_shared((BK, N), "float16")
            Cl = T.alloc_fragment((M, N), "float32")
            for tile, _ in T.Persistent([2, 1], 1, bid):
                T.clear(Cl)
                for k in T.Pipelined(K // BK, num_stages=2):
                    if k == 0 or k == 2:
                        T.copy(A[tile, 0, k*BK], As); T.copy(B[tile, k*BK, 0], Bs); T.gemm(As, Bs, Cl)
                T.copy(Cl, C[tile, 0, 0])
    return main

torch.manual_seed(0)
a = torch.randn(2, M, K, dtype=torch.float16, device="cuda") * 0.125
b = torch.randn(2, K, N, dtype=torch.float16, device="cuda") * 0.125
ref = a[:, :, :BK].float() @ b[:, :BK, :].float() + a[:, :, 2*BK:3*BK].float() @ b[:, 2*BK:3*BK, :].float()

func = matmul()
for label, disable_ws in (("control_ws_off", True), ("repro_ws_on", False)):
    kernel = tilelang.compile(
        func,
        out_idx=[2],
        pass_configs={
            tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: disable_ws,
        },
    )
    c = kernel(a, b).float()
    print(f"{label}_tile0_diff=", (c[0] - ref[0]).abs().max().item())
    print(f"{label}_tile1_diff=", (c[1] - ref[1]).abs().max().item())
```

```text
control_ws_off_tile0_diff= 0.00012162327766418457
control_ws_off_tile1_diff= 0.0001195669174194336
repro_ws_on_tile0_diff= 0.4828925132751465
repro_ws_on_tile1_diff= 0.955072820186615
```

**Root cause**: MVB runs before the WS rewrite and versions shared buffers using a linearized ancestor-loop expression such as `(outer * trip_count + k) % num_stages`. For a guarded loop, WS instead selects the barrier using `phase_counter % num_stages`.

`StageExprReplacer` only recognizes `k % stages` or `(k - min) % stages`, so it does not replace the MVB expression. For executed iterations `k=0,2` with two stages, the barrier stages are `0,1`, but both shared-buffer accesses still use version `0`. The barrier therefore does not protect the buffer version actually being accessed.

### Summary

These paths were unreachable for persistent loops before this PR because loop replacement aborted at the outer `For`. Since the PR now claims support for that structure, silently generating incorrect WS code is a correctness gap.

The guarded transaction ordinal needs to live across persistent iterations and drive both mbarrier stage/parity and shared-buffer versioning. If that cannot be guaranteed, guarded pipelines under an outer persistent `For` should be rejected instead of transformed.

RuneFang:
@Lyscoria @reoLantern I’ve addressed the review feedback. Can you help brainstorm if there are any other hidden issues? 😁

chengyupku:
Thanks for addressing the issues. I tested the current head (`74b16c6e`) on an NVIDIA H200 and found another correctness regression caused by the broadened
  `StageExprReplacer::MatchLinearIdx()`.

  The problematic change is:

  ```cpp
  return tvm::tirx::UsesVar(
      expr, [this](const VarNode *v) { return v == loop_var_.get(); });
  ```

  `StageExprReplacer` traverses the entire producer/consumer body, including the original loop guard. Therefore, any user expression matching:

  ```text
  FloorMod(<expression using k>, num_stages)
  ```

  may be replaced with the phase counter, even when it is unrelated to MVB buffer versioning.

  This reproduces without `T.Persistent`:

  ```python
  @T.prim_func
  def main(
      A: T.Tensor((64, 128), "float16"),
      B: T.Tensor((128, 64), "float16"),
      C: T.Tensor((64, 64), "float16"),
  ):
          As = T.alloc_shared((64, 32), "float16")
          Bs = T.alloc_shared((32, 64), "float16")
          Cl = T.alloc_fragment((64, 64), "float32")

          T.clear(Cl)
          for k in T.Pipelined(4, num_stages=2):
              if (k + 1) % 2 == 0:
                  T.copy(A[:, k * 32], As)
                  T.copy(B[k * 32, :], Bs)
                  T.gemm(As, Bs, Cl)
          T.copy(Cl, C)
  ```

  The original condition should execute `k=1` and `k=3`. With the current PR, the generated producer CUDA contains:

  ```cpp
  producer_phase_cnt[0] = 0;
  for (int k = 0; k < 4; ++k) {
    if ((producer_phase_cnt[0] % 2) == 0) {
      ...
      producer_phase_cnt[0]++;
    }
  }
  ```

  The consumer condition is rewritten similarly. This changes the user program semantics: `k=0` executes incorrectly, the counter becomes 1, and subsequent iterations no longer enter the branch.

  Results on H200:

  ```text
  commit              74b16c6ef347d74ca84d53084a7620a0cda6db58
  ws_applied          True
  wsoff_vs_expected   0.0001226664
  wson_vs_expected    0.6124069691
  wson_vs_wsoff       0.6123046875
  ```

  The important distinction is between the logical loop coordinate and the physical pipeline transaction coordinate:

  ```text
  guard(k)                         keep k
  global_buffer[k * block_K]       keep k
  user arithmetic involving k     keep k

  MVB shared-buffer version        use txn_idx
  mbarrier stage                   use txn_idx
  mbarrier parity                  use txn_idx
  ```

  For a guarded pipeline, `txn_idx` is the number of previously executed transactions, not the original loop index:

  ```cpp
  if (original_guard(k)) {
    stage = txn_idx % num_stages;
    parity = (txn_idx / num_stages) % 2;

    // Only MVB versions and mbarrier bookkeeping use stage/parity.
    ...
    txn_idx++;
  }
  ```

  Producer and consumer should retain separate transaction counters because they progress asynchronously.

  I think the counter-hoisting change is conceptually correct. The unsafe part is identifying MVB-generated version indices by scanning arbitrary `FloorMod` expressions.

  A robust fix would be to establish explicit provenance between MVB and WS. For example, MVB could emit an internal marker:

  ```cpp
  tl.mvb_stage_index(
      FloorMod(linear_index, num_stages))
  ```

  MVB would use this marker only for its synthetic shared-buffer version index. WS could then perform:

  ```text
  producer: tl.mvb_stage_index(...) -> producer_txn_idx % num_stages
  consumer: tl.mvb_stage_index(...) -> consumer_txn_idx % num_stages
  ```

  When no transaction counter is required, the marker can lower back to its original argument. Returning explicit versioned-buffer metadata from MVB would be another possible implementation.

  Restricting the broad `UsesVar()` matching to persistent pipelines would not be sufficient, because the same user expression can appear inside a persistent loop. The replacement must be scoped by provenance/
  location, not merely by whether an expression depends on `k`.

  I suggest removing the `UsesVar()` fallback until MVB-generated indices can be identified explicitly, and adding WS-on/WS-off regression tests for both non-persistent and persistent versions of the shifted-
  modulo condition above.

chengyupku:
I reproduced the `StageExprReplacer` correctness issue on H200 and pushed a follow-up fix to this branch.

The fix preserves provenance for MVB-generated stage indices through an internal marker, so WS only replaces  synthetic shared-memory/barrier version indices. User expressions such as `(k + 1) % num_stages` remain unchanged.

reoLantern:
Follow-up: I rebuilt the current PR head (`bd7ddcfce7`) from source and re-ran everything from my earlier report on an H100 — **all of the correctness problems I hit are now fixed.** 🎉

Environment: H100 PCIe (sm_90), CUDA 12.4, `tilelang 0.1.12+cuda.gitbd7ddcfc`, `torch 2.13.0`, `num_sms=114`. Config: `BLOCK_M=128, BLOCK_N=256, BLOCK_K=64, threads=256, num_stages=3`, auto-WS on.

## Every case I previously reported as broken now passes

| case | before (my earlier comment) | now (`bd7ddcfce7`) |
|---|---|---|
| `2048×2048×1024` (trip=16, 2 waves) | max_diff ≈ 81 | ✅ max_diff 0.000 |
| `2048×2048×1280` (trip=20, 2 waves) | `cudaErrorLaunchFailure` | ✅ max_diff 0.000 |
| `4096×4096×4096` (trip=64, 5 waves) | max_diff ≈ 83 | ✅ max_diff 0.000 |
| K-trip sweep @ 2 waves (trip = 16,18,20,22,24,26) | correct **only** when `trip % (2·num_stages) == 0` | ✅ correct for **all** trip values |
| `examples/gemm/example_gemm_persistent.py` @ 4096³ | 43% elements mismatched | ✅ `assert_allclose` PASS |
| `examples/gemm/example_gemm_persistent.py` @ 8192³ (default) | `cudaErrorLaunchFailure` | ✅ `assert_allclose` PASS |

The K-trip phase dependency I described is completely gone — results are exact regardless of how the trip count lines up with the pipeline phase period.

## WS is genuinely still applied (not silently declined)

I checked the generated CUDA for the multi-wave 4096³ persistent kernel: the persistent loop (`for (int w ...)`), `__launch_bounds__(384)` (1 producer + 2 consumer warpgroups), `mbarrier`, `tma_load`, and `wgmma` are all present. So this is persistent + pipelined + real WS, all three active and correct — not correctness bought by turning WS off.

## Matches the root cause

For what it's worth, the fix lines up exactly with the phase-carryover cause. The producer/consumer waits now derive their phase from persistent counters that carry across the outer loop:

```cpp
mbarrier[... producer_phase_cnt[0] % 3 ...].wait(... (producer_phase_cnt[0] % 6) / 3 ...);
mbarrier[... consumer_phase_cnt[0] % 3 ...].wait(... (consumer_phase_cnt[0] % 6) / 3 ...);
```

instead of the old `f(kk)` phase that reset every wave. That's precisely what was needed so the hardware mbarrier phase stays in sync across persistent iterations.

---

While verifying, I also probed nearby edge cases in the persistent + WS path. Two are still broken on `bd7ddcfce7`. Both are outside the multi-wave phase fix this PR nails, so I don't think either blocks merging — flagging for awareness / follow-up.

### A. Divergent per-tile `if/else` guard under persistent + WS → wrong results

A persistent kernel whose per-tile guard has an **`else` branch that also writes output** (the pipeline living in the `then` branch) miscompiles under WS. Minimal repro (a checkerboard block mask — a valid kernel; WS-off is exact):

```python
for bx, by in T.Persistent([T.ceildiv(M, bm), T.ceildiv(N, bn)], sm_num, bid):
    if (bx + by) % 2 == 0:
        T.clear(Cl)
        for kk in T.Pipelined(T.ceildiv(K, bk), num_stages=3):
            T.copy(A[bx*bm, kk*bk], As); T.copy(B[kk*bk, by*bn], Bs); T.gemm(As, Bs, Cl)
        T.copy(Cl, Cs); T.copy(Cs, C[bx*bm, by*bn])
    else:
        T.clear(Cl); T.copy(Cl, Cs); T.copy(Cs, C[bx*bm, by*bn])   # else also writes output
```

On H100 at 2048³ this gives `max_diff ≈ 183` (and non-deterministic `nan`/`inf` across runs) with WS on, vs exact with `tl.disable_warp_specialized=True`. It needs all three conditions — dropping any one makes it correct:

| persistent | has `else` | WS | result |
|---|---|---|---|
| yes | yes | on | ❌ wrong |
| yes | yes | off | ✅ |
| yes | `then`-only guard (no `else`) | on | ✅ |
| no (plain grid) | yes | on | ✅ |

This is a different shape from the inner-body `if k < 2` guard that is now fixed — here the guard is at the tile level with a divergent, output-writing `else`.

### B. `T.Pipelined` inside a `while` (stream-K) — already tracked elsewhere

For completeness: `T.Pipelined` nested in a `while` loop still aborts with the same `ProducerConsumerWS: failed to replace pipeline loop` ICHECK on this head (the shipped `examples/gemm_streamk` reproduces it). This is expected — the PR teaches the replacer to descend through `For` but not `While`. It's **already being handled separately in #2692 (for issue #2548)**, so this is just a note that the two efforts don't overlap; nothing to do here.

Thanks @RuneFang and @chengyupku for the quick turnaround!

chengyupku:
> Divergent per-tile `if/else` guard under persistent + WS → wrong results

 Thanks for the thorough verification, I also reproduced issue A on H200.

  Since it is separate from the persistent phase issue fixed here, I suggest to merge this PR and handle the per-tile `if/else` case in a separate PR. The `while` case remains tracked by #2692.

PR #2685, opened 2026-07-16
[BugFix] Track GEMM accumulator writes in warp-specialization liveness collector

### LINKED ISSUE #2547: [BUG][Fuzzer][ice-on-valid-code] Regression: two `T.gemm` accumulating into one fragment no longer compile under default warp specialization on Hopper (worked on 0.1.8) — aborts with an internal `LayoutInference` error instead of compiling
### Required prerequisites

0.1.11

### System information

NVIDIA H100 (sm_90), CUDA 13.0, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). Hopper-specific: the crash appears only when warp specialization is enabled (the sm_90 default), which splits the kernel into producer/consumer warpgroups before layout inference.

### Problem description

Running `examples/linear_attention/example_mamba_chunk_scan.py` verbatim at its default configuration on sm_90 aborts compilation with an internal `LOG(FATAL)`. The kernel accumulates **two** `T.gemm` calls into one `acc_o` fragment (a prelude `C @ prev_states` then a pipelined-loop `cb @ x`), and under the **default** pipeline (warp specialization enabled) this aborts:

```
tvm.error.InternalError: Get different layout for acc_o
 current layout:  Fragment((64, 64) -> (32,), ..., thread_range: I.Range(128, 256))
 previous layout: Fragment((64, 64) -> (16,), ..., thread_range: I.Range(0, 256))
```

The abort fires in `LayoutInference` ([`layout_inference.cc#L260-L276`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/layout_inference.cc#L260-L276)), reached from `tilelang/cuda/pipeline.py:113`. The two layouts for the **same** `acc_o` fragment differ only in `thread_range`: one inferred over the full block `I.Range(0, 256)`, the other over the consumer warpgroup `I.Range(128, 256)`. It is deterministic and gated entirely by warp specialization: it crashes under the default and compiles + runs correctly the moment `tl.disable_warp_specialized=True` is set. It is **not** the WGMMA instruction — `tl.disable_wgmma=True` (WS still on) still crashes with the identical `thread_range` conflict on an MMA-shaped fragment. The single discriminating lever is warp specialization.

### Reproducible example code

The trigger is the shipped `examples/linear_attention/example_mamba_chunk_scan.py` pattern — a prelude `T.gemm(C_shared, prev_state_shared, acc_o, transpose_B=True)` followed by a pipelined-loop `T.gemm(cb_local, x_shared, acc_o)` into the same accumulator — compiled with the default (WS-on) pipeline at 128 threads (256 effective under WS).

<details>
<summary>Full standalone repro (default crashes; <code>disable_ws</code> compiles + runs at ~5e-4 vs torch; <code>disable_wgmma</code> still crashes)</summary>

Use a real `.py` file (TileLang's eager front end calls `inspect.getsource`). `tilelang.disable_cache()`; do not set `TILELANG_CLEAR_CACHE` (removed in 0.1.11). Run on sm_90.

```python
import tilelang
tilelang.disable_cache()
import torch
import tilelang.language as T
import sys


def chunk_scan_fwd(batch, seqlen, chunk_size, ngroups, nheads, headdim, dstate,
                   block_M, block_N, block_K, block_Dstate, num_stages, threads,
                   disable_ws=False, disable_wgmma=False):
    dtype, accum_dtype = T.float16, T.float32
    nchunks = T.ceildiv(seqlen, chunk_size)
    p = 1.44269504

    @T.prim_func
    def main(
        cb: T.Tensor((batch, nchunks, ngroups, chunk_size, chunk_size), dtype),
        x: T.Tensor((batch, seqlen, nheads, headdim), dtype),
        dt: T.Tensor((batch, nheads, nchunks, chunk_size), dtype),
        dA_cumsum: T.Tensor((batch, nheads, nchunks, chunk_size), dtype),
        C: T.Tensor((batch, seqlen, ngroups, dstate), dtype),
        prev_states: T.Tensor((batch, nchunks, nheads, headdim, dstate), dtype),
        D: T.Tensor((nheads), dtype),
        Output: T.Tensor((batch, seqlen, nheads, headdim), dtype),
    ):
        with T.Kernel(nheads, T.ceildiv(chunk_size, block_M) * T.ceildiv(headdim, block_N),
                      batch * nchunks, threads=threads) as (bz, bx, by):
            acc_o = T.alloc_fragment((block_M, block_N), accum_dtype)
            acc_o_shared = T.alloc_shared((block_M, block_N), dtype)
            cb_shared = T.alloc_shared((block_M, block_K), dtype)
            cb_local = T.alloc_fragment((block_M, block_K), dtype)
            dA_cs_k_shared = T.alloc_shared((block_K), dtype)
            dA_cs_k_local = T.alloc_fragment((block_K), accum_dtype)
            dA_cs_m_local = T.alloc_fragment((block_M), accum_dtype)
            dt_shared = T.alloc_shared((block_K), dtype)
            dt_local = T.alloc_fragment((block_K), accum_dtype)
            x_shared = T.alloc_shared((block_K, block_N), dtype, scope="shared.dyn")
            dA_cs_m_shared = T.alloc_shared((block_M), dtype)
            scale_m_local = T.alloc_fragment((block_M), accum_dtype)
            C_shared = T.alloc_shared((block_M, block_Dstate), dtype)
            prev_state_shared = T.alloc_shared((block_N, block_Dstate), dtype)
            D_local = T.alloc_fragment((1), accum_dtype)
            x_residual_shared = T.alloc_shared((block_M, block_N), dtype)
            x_residual_local = T.alloc_fragment((block_M, block_N), accum_dtype)
            batch_idx, chunk_idx = by % batch, by // batch
            m_idx, n_idx = bx // T.ceildiv(headdim, block_N), bx % T.ceildiv(headdim, block_N)
            T.annotate_layout({
                cb_shared: tilelang.layout.make_swizzled_layout(cb_shared),
                x_residual_shared: tilelang.layout.make_swizzled_layout(x_residual_shared),
            })
            T.no_set_max_nreg()
            T.copy(dA_cumsum[batch_idx, bz, chunk_idx, m_idx * block_M:(m_idx + 1) * block_M], dA_cs_m_shared)
            T.copy(dA_cs_m_shared, dA_cs_m_local)
            T.clear(acc_o)
            for i in T.Parallel(block_M):
                scale_m_local[i] = T.exp2(dA_cs_m_local[i] * p)
            T.copy(C[batch_idx, chunk_idx * chunk_size + m_idx * block_M:chunk_idx * chunk_size + (m_idx + 1) * block_M,
                     bz // (nheads // ngroups), 0:block_Dstate], C_shared)
            T.copy(prev_states[batch_idx, chunk_idx, bz, n_idx * block_N:(n_idx + 1) * block_N, 0:block_Dstate], prev_state_shared)
            T.gemm(C_shared, prev_state_shared, acc_o, transpose_B=True)   # prelude GEMM -> acc_o
                acc_o[i, j] *= scale_m_local[i]
            loop_range = T.ceildiv((m_idx + 1) * block_M, block_K)
            for k in T.Pipelined(loop_range, num_stages=num_stages):
                T.copy(cb[batch_idx, chunk_idx, bz // (nheads // ngroups),
                          m_idx * block_M:(m_idx + 1) * block_M, k * block_K:(k + 1) * block_K], cb_shared)
                T.copy(cb_shared, cb_local)
                T.copy(dA_cumsum[batch_idx, bz, chunk_idx, k * block_K:(k + 1) * block_K], dA_cs_k_shared)
                T.copy(dA_cs_k_shared, dA_cs_k_local)
                for i, j in T.Parallel(block_M, block_K):
                    cb_local[i, j] = cb_local[i, j] * T.exp2(dA_cs_m_local[i] * p - dA_cs_k_local[j] * p)
                T.copy(dt[batch_idx, bz, chunk_idx, k * block_K:(k + 1) * block_K], dt_shared)
                T.copy(dt_shared, dt_local)
                for i, j in T.Parallel(block_M, block_K):
                    cb_local[i, j] *= dt_local[j]
                for i, j in T.Parallel(block_M, block_K):
                    cb_local[i, j] = T.if_then_else(m_idx * block_M + i >= k * block_K + j, cb_local[i, j], 0)
                T.copy(x[batch_idx, chunk_idx * chunk_size + k * block_K:chunk_idx * chunk_size + (k + 1) * block_K,
                         bz, n_idx * block_N:(n_idx + 1) * block_N], x_shared)
                T.gemm(cb_local, x_shared, acc_o)                          # loop GEMM -> same acc_o
            D_local[0] = D[bz]
            T.copy(x[batch_idx, chunk_idx * chunk_size + m_idx * block_M:chunk_idx * chunk_size + (m_idx + 1) * block_M,
                     bz, n_idx * block_N:(n_idx + 1) * block_N], x_residual_shared)
            T.copy(x_residual_shared, x_residual_local)
                acc_o[i, j] += x_residual_local[i, j] * D_local[0]
            T.copy(acc_o, acc_o_shared)
            T.copy(acc_o_shared, Output[batch_idx,
                   chunk_idx * chunk_size + m_idx * block_M:chunk_idx * chunk_size + (m_idx + 1) * block_M,
                   bz, n_idx * block_N:(n_idx + 1) * block_N])

    pass_configs = {tilelang.PassConfigKey.TL_ENABLE_FAST_MATH: True}
    if disable_ws:
        pass_configs[tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED] = True
    if disable_wgmma:
        pass_configs[tilelang.PassConfigKey.TL_DISABLE_WGMMA] = True
    return tilelang.compile(main, pass_configs=pass_configs)


mode = sys.argv[1] if len(sys.argv) > 1 else "default"
chunk_scan_fwd(1, 512, 256, 1, 8, 64, 128, 64, 64, 64, 128, 2, 128,
               disable_ws=(mode == "disable_ws"), disable_wgmma=(mode == "disable_wgmma"))
print("COMPILE OK")
```

```
$ python repro.py            -> InternalError: Get different layout for acc_o  (I.Range(128,256) vs I.Range(0,256))
$ python repro.py disable_ws -> COMPILE OK   (and RUN OK, max abs err = 0.00048828125 vs torch)
$ python repro.py disable_wgmma -> InternalError: Get different layout for acc_o  (same conflict, MMA-shaped)
```
</details>

### Traceback

```pytb
File ".../tilelang/cuda/pipeline.py", line 113, in CUDAPassPipelineBodyPrologue
  ... BufferUseDefCollector::Run -> GemmNode::InferLayout -> LogFatal
tvm.error.InternalError: Get different layout for acc_o
```

### Expected behavior

The dual-GEMM-into-one-accumulator kernel should compile under the default sm_90 pipeline — the byte-identical program compiles *and runs correctly* with `tl.disable_warp_specialized=True` (max abs err ≈ 5e-4 vs torch), so it is well-formed and the WS path is the regression. Since disabling WGMMA does not fix it, this is not an "accumulator unsupported" guard; it is a thread-range bookkeeping defect. The WS pass should assign `acc_o` a single thread range (compile), not abort on the generic layout-consistency invariant.

### Additional context

**Root cause.** Warp specialization runs before `LayoutInference` and splits the kernel into a producer warpgroup `[0,128)` and consumer `[128,256)`. Pre-loop ("prelude") statements are kept in the shared prelude (full `[0,256)`) or extracted into a branch, driven by the liveness sets `LocalAccessCollector` builds. But `LocalAccessCollector::VisitExpr_(CallNode)` ([`producer_consumer_ws.cc#L453-L525`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L453-L525)) special-cases only `CopyNode` and `FillNode` — there is **no `GemmNode` case**, so a `T.gemm`'s accumulator `acc_o` is never inserted into `read_buffers`/`write_buffers`. The prelude GEMM therefore has empty tracked defs, `ClassifyPreludeStmt` returns `kKeepSharedPrelude` ([`#L553-L555`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L553-L555)), and it stays in the full `[0,256)` prelude while the consumer-loop GEMM runs in `[128,256)`. At `LayoutInference` the two GEMMs bake incompatible `thread_range`s into the one `acc_o` fragment and the `LOG(FATAL)` fires. The WS pass's own comment ([`#L1947-L1951`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L1947-L1951)) states the very invariant being violated ("a single fragment layout cannot represent both thread ranges"), but its remedy covers only producer-private buffers, not a fragment shared between the shared prelude and the consumer branch. *(Line numbers above are the pinned `v0.1.11` SHA; on current main they have drifted a few lines — the fatal to ~273-276, the collector to ~454-527 — but the mechanism is unchanged and the permalinks resolve.)*

**Suggested fix.** Teach `LocalAccessCollector::VisitExpr_(CallNode)` about `GemmNode`, mirroring the existing `Copy`/`Fill` cases: insert the operands `a_`/`b_` into `read_buffers` and the accumulator `c_` into `write_buffers` (plus `read_buffers` unless `clearAccum`), matching `GemmNode::GetAccessRegions`. The load-bearing line is `write_buffers.insert(gemm->c_)`: today the accumulator only reaches the collector through the generic `BufferLoad` fall-through, i.e. recorded read-only, so the WS liveness (`HasTrackedDefs`/`NeedsAnyDef`, which key off `write_buffers`) never sees the GEMM as *defining* `acc_o`. Once it does, the prelude GEMM is classified `kConsumerOnly`, extracted + `threadIdx`-rewritten into the consumer branch, and both GEMMs infer `acc_o` over the same range. This closes the gap without touching pass order — preferable to reverting `WarpSpecialized`/`LayoutInference` order, since WS was moved ahead of layout inference deliberately (tiled-WS / `use_2cta` annotation visibility) and a blanket revert risks regressing that. (A more general fix would derive read/write sets from each tile-op's own `GetAccessRegions()` rather than per-op cases — the same read-only-capture gap exists for any other accumulating tile-op, e.g. `Reduce`.)

Verified end-to-end on a from-source build at this SHA (sm_90): pristine build aborts with `Get different layout for acc_o`; after adding the `GemmNode` case above (recompile + relink `libtilelang.so`, no other change) the same example compiles and passes numerics (`All checks pass.`).

**Provenance.** `mamba_chunk_scan` is a shipped example; the dual-GEMM-into-one-accumulator pattern is at [`example_mamba_chunk_scan.py`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/linear_attention/example_mamba_chunk_scan.py#L161) (prelude GEMM) + [`#L198`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/linear_attention/example_mamba_chunk_scan.py#L198) (loop GEMM). This is a regression: the example compiles and runs correctly on `tilelang==0.1.7`/`0.1.8`, and first crashes on `0.1.9` — the release that first runs `ProducerConsumerWarpSpecialized` *before* `LayoutInference`. The order is the trigger: in the crashing `0.1.9` build, moving the `LayoutInference()` call back before the `WarpSpecialized()` call in `phase.py` (a pure-Python pass-order swap, nothing else changed) makes the example compile and pass numerics. Before that reorder `LayoutInference` laid `acc_o` out once over the full block, so the producer/consumer split never produced two ranges; running WS first exposes the latent `LocalAccessCollector` gap below. (The example has no CI test, so this went unnoticed.)

**Not a duplicate.** #1117 (open) shares the `LayoutInference` `LOG(FATAL)` message family but is an attention layout-inference disagreement, not the WS prelude-vs-consumer thread-range split of a dual-GEMM accumulator — the signal here is that the layouts differ *only* in `thread_range` and the crash is gated entirely by `tl.disable_warp_specialized`. Distinct from #1954/#1263 (both the `MakePackedAPI` undefined-var abort — a different pass and message) and from our sibling grouped-GEMM draft (that is the `MakePackedAPI` `ICHECK` for a dropped `make_tensor` TMA descriptor; no `acc_o`/layout involvement).

zyy3077:
## Issue

https://github.com/tile-ai/tilelang/issues/2547

## Summary

On Hopper (sm_90) with the default warp-specialized (WS) pipeline, a kernel
with a prelude `T.gemm` and a pipelined-loop `T.gemm` accumulating into the
same fragment aborts compilation in `LayoutInference` with:

```
tvm.error.InternalError: Get different layout for acc_o
 current layout:  thread_range: I.Range(128, 256)
 previous layout: thread_range: I.Range(0, 256)
```

This is a regression hit by the shipped `examples/linear_attention/example_mamba_chunk_scan.py`,
which compiles and runs correctly with `tl.disable_warp_specialized=True`.

## Root cause

`ProducerConsumerWarpSpecialized` runs before `LayoutInference` and splits the
kernel into a producer warpgroup `[0, threads)` and a consumer warpgroup
`[threads, 2*threads)`. Pre-loop ("prelude") statements are classified as
shared/producer-only/consumer-only by `ClassifyPreludeStmt`, based on read/write
buffer sets built by `LocalAccessCollector::VisitExpr_(CallNode*)`.

That collector special-cased `CopyNode` and `FillNode` only. `GemmNode` had no
case, so a prelude `T.gemm`'s accumulator write fell through to the generic
`BufferLoad` visitor and was recorded as a read only — never as a write. With
no tracked write, `ClassifyPreludeStmt` short-circuited to `kKeepSharedPrelude`
before even checking live sets, so the prelude GEMM stayed in the shared
`[0, 2*threads)` range while the loop-body GEMM (already consumer-only) ran in
`[threads, 2*threads)`. Both GEMMs then baked incompatible `thread_range`s into
the same accumulator fragment, and `LayoutInference` aborted.

## Fix

Add a `GemmNode` case to `LocalAccessCollector::VisitExpr_(CallNode*)`,
mirroring `GemmNode::GetAccessRegions()`: record `a_`/`b_` as reads and `c_` as
a write (plus a read unless `clearAccum_` is set), and revisit
`offsetA_`/`offsetB_`/`cCoords_` so index-variable dependencies are still
captured after the early return (matching how the existing `Copy`/`Fill` cases
revisit their own ranges/value).

With the accumulator write now tracked, the prelude GEMM is correctly
classified `kConsumerOnly` and extracted into the consumer branch alongside the
loop-body GEMM, so both share one `thread_range`.

## Tests

Added two regression tests in
`testing/python/transform/test_tilelang_transform_producer_consumer_ws.py`,
using a minimal prelude-GEMM + pipelined-loop-GEMM-into-the-same-fragment
kernel:

- `test_tiled_ws_dual_gemm_shared_accumulator_correctness`: compiles and runs
  under the default (WS-on) pipeline, checked against a PyTorch reference.
- `test_tiled_ws_dual_gemm_shared_accumulator_disable_wgmma`: same pattern with
  `TL_DISABLE_WGMMA=True`, covering the issue's point that this is a
  thread-range bookkeeping defect in the WS pass, not specific to WGMMA.

Verified both tests reproduce the original `Get different layout for acc`
failure when the `producer_consumer_ws.cc` fix is reverted, and pass with it
applied. Also re-ran `examples/linear_attention/example_mamba_chunk_scan.py`
(default WS-on config) end-to-end against its PyTorch reference — passes.

## C++ style / lint notes

This PR touches TileLang-owned C++ transform code under
`src/cuda/transform/producer_consumer_ws.cc` and keeps the change narrowly
scoped to the missing `GemmNode` case, following the existing `Copy`/`Fill`
case pattern in the same visitor rather than introducing a new abstraction.
Checked with the repository's `clang-format` and `ruff-format` pre-commit
hooks.

Fixes https://github.com/tile-ai/tilelang/issues/2547

## Summary
- **Bug Fixes**
  - Improved local access/liveness tracking for GEMM tile-ops, including accumulator writes, accumulator reuse reads, and related scale-factor region reads.
- **New Features / Testing**
  - Added regression coverage for a dual-GEMM shared-accumulator pattern, validating both default warp-specialization behavior and a configuration with WGMMA disabled.

## C++ style / lint notes
- No exported/public entity changes. Any C++ API style audit findings should be treated as advisory unless they impact public surfaces.

### DIFF
--- a/src/cuda/transform/producer_consumer_ws.cc
+++ b/src/cuda/transform/producer_consumer_ws.cc
@@ -481,6 +481,26 @@ private:
         }
         return;
       }
+      if (const auto *gemm = tile_op.as<GemmNode>()) {
+        if (IsBranchPrivateBuffer(gemm->a_)) {
+          summary_.read_buffers.insert(gemm->a_);
+        }
+        if (IsBranchPrivateBuffer(gemm->b_)) {
+          summary_.read_buffers.insert(gemm->b_);
+        }
+        if (IsBranchPrivateBuffer(gemm->c_)) {
+          summary_.write_buffers.insert(gemm->c_);
+          if (!is_one(gemm->clearAccum_)) {
+            summary_.read_buffers.insert(gemm->c_);
+          }
+        }
+        VisitExpr(gemm->offsetA_);
+        VisitExpr(gemm->offsetB_);
+        for (const auto &coord : gemm->cCoords_) {
+          VisitExpr(coord);
+        }
+        return;
+      }
     }
 
     if (op->op.same_as(tl::access_ptr())) {
--- a/src/cuda/transform/producer_consumer_ws.cc
+++ b/src/cuda/transform/producer_consumer_ws.cc
@@ -482,22 +482,57 @@ private:
         return;
       }
       if (const auto *gemm = tile_op.as<GemmNode>()) {
-        if (IsBranchPrivateBuffer(gemm->a_)) {
-          summary_.read_buffers.insert(gemm->a_);
+        // Mirrors GemmNode::GetAccessRegions(): a_/b_ (and the optional MX
+        // scale-factor regions) are reads, c_ is always a write and also a
+        // read unless clearAccum_ clears it first. Region mins/extents are
+        // revisited (rather than just offsetA_/offsetB_/cCoords_) so any
+        // loop-variable dependency in a sliced operand or accumulator
+        // sub-block is still captured after the early return below.
+        if (IsBranchPrivateBuffer(gemm->aRegion_->buffer)) {
+          summary_.read_buffers.insert(gemm->aRegion_->buffer);
         }
-        if (IsBranchPrivateBuffer(gemm->b_)) {
-          summary_.read_buffers.insert(gemm->b_);
+        for (const auto &range : gemm->aRegion_->region) {
+          VisitExpr(range->min);
+          VisitExpr(range->extent);
         }
-        if (IsBranchPrivateBuffer(gemm->c_)) {
-          summary_.write_buffers.insert(gemm->c_);
+        if (IsBranchPrivateBuffer(gemm->bRegion_->buffer)) {
+          summary_.read_buffers.insert(gemm->bRegion_->buffer);
+        }
+        for (const auto &range : gemm->bRegion_->region) {
+          VisitExpr(range->min);
+          VisitExpr(range->extent);
+        }
+        if (IsBranchPrivateBuffer(gemm->cRegion_->buffer)) {
+          summary_.write_buffers.insert(gemm->cRegion_->buffer);
           if (!is_one(gemm->clearAccum_)) {
-            summary_.read_buffers.insert(gemm->c_);
+            summary_.read_buffers.insert(gemm->cRegion_->buffer);
           }
         }
-        VisitExpr(gemm->offsetA_);
-        VisitExpr(gemm->offsetB_);
-        for (const auto &coord : gemm->cCoords_) {
-          VisitExpr(coord);
+        for (const auto &range : gemm->cRegion_->region) {
+          VisitExpr(range->min);
+          VisitExpr(range->extent);
+        }
+        if (gemm->sfaRegion_.defined()) {
+          if (IsBranchPrivateBuffer(gemm->sfaRegion_->buffer)) {
+            summary_.read_buffers.insert(gemm->sfaRegion_->buffer);
+          }
+          for (const auto &range : gemm->sfaRegion_->region) {
+            VisitExpr(range->min);
+            VisitExpr(range->extent);
+          }
+        }
+        if (gemm->sfbRegion_.defined()) {
+          if (IsBranchPrivateBuffer(gemm->sfbRegion_->buffer)) {
+            summary_.read_buffers.insert(gemm->sfbRegion_->buffer);
+          }
+          for (const auto &range : gemm->sfbRegion_->region) {
+            VisitExpr(range->min);
+            VisitExpr(range->extent);
+          }
+        }
+        VisitExpr(gemm->clearAccum_);
+        if (gemm->sfKStart_.defined()) {
+          VisitExpr(gemm->sfKStart_);
         }
         return;
       }
--- a/testing/python/transform/test_tilelang_transform_producer_consumer_ws.py
+++ b/testing/python/transform/test_tilelang_transform_producer_consumer_ws.py
@@ -37,6 +37,61 @@ def matmul_pipelined(M, N, K, block_M, block_K, block_N, num_stages, dtype="floa
     return main
 
 
+def dual_gemm_shared_accumulator(
+    M,
+    N,
+    K0,
+    K1,
+    block_M,
+    block_N,
+    block_K,
+    num_stages,
+    dtype="float16",
+    threads=128,
+):
+    """A prelude GEMM and a pipelined-loop GEMM accumulating into the same
+    fragment, mirroring the #2547 `example_mamba_chunk_scan.py` pattern:
+    `T.gemm(A0, B0, acc)` runs once before the loop, then
+    `T.gemm(A1_k, B1_k, acc)` accumulates into the same `acc` inside a
+    `T.Pipelined` loop. Under warp specialization, the prelude GEMM must be
+    classified consumer-only so both GEMMs share one thread_range for `acc`.
+    """
+
+    @T.prim_func
+    def main(
+        A0: T.Tensor((M, K0), dtype),
+        B0: T.Tensor((K0, N), dtype),
+        A1: T.Tensor((M, K1), dtype),
+        B1: T.Tensor((K1, N), dtype),
+        C: T.Tensor((M, N), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=threads) as (
+            bx,
+            by,
+        ):
+            acc = T.alloc_fragment((block_M, block_N), "float32")
+            A0_shared = T.alloc_shared((block_M, K0), dtype)
+            B0_shared = T.alloc_shared((K0, block_N), dtype)
+            A1_shared = T.alloc_shared((block_M, block_K), dtype)
+            B1_shared = T.alloc_shared((block_K, block_N), dtype)
+            C_local = T.alloc_fragment((block_M, block_N), dtype)
+
+            T.clear(acc)
+            T.copy(A0[by * block_M, 0], A0_shared)
+            T.copy(B0[0, bx * block_N], B0_shared)
+            T.gemm(A0_shared, B0_shared, acc)  # prelude GEMM -> acc
+
+            for ko in T.Pipelined(T.ceildiv(K1, block_K), num_stages=num_stages):
+                T.copy(A1[by * block_M, ko * block_K], A1_shared)
+                T.copy(B1[ko * block_K, bx * block_N], B1_shared)
+                T.gemm(A1_shared, B1_shared, acc)  # loop GEMM -> same acc
+
+            T.copy(acc, C_local)
+            T.copy(C_local, C[by * block_M, bx * block_N])
+
+    return main
+
+
 def matmul_windowed_pipelined(
     M,
     N,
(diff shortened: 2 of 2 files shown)


LeiWang1999 (approved):
LGTM, but we should refactor this function in future to classify tile operators in a general approach.

PR #2687, opened 2026-07-17
[BugFix] Skip TMA promotion for copies whose base pointer is bound inside the kernel body

### LINKED ISSUE #2549: [BUG][Fuzzer][ice-on-valid-code] Ptr-backed grouped GEMM (`T.make_tensor`) aborts the default sm_90 compile with an internal `MakePackedAPI` `ICHECK` instead of compiling or diagnosing
### Required prerequisites

0.1.11

### System information

NVIDIA H100 (sm_90), CUDA 13.0, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). Hopper-specific: the crash appears only when warp specialization is enabled (the sm_90 default), which rewrites the `T.copy` into a TMA copy.

### Problem description

The ptr-backed grouped GEMM — the pattern shipped in `examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py`, where each group's `A`/`B` is reconstructed inside the kernel from a pointer table via the public `T.make_tensor(ptr, shape, dtype)` primitive — aborts compilation under the **default** sm_90 pipeline with an internal `ICHECK`:

```
tvm.error.InternalError: Check failed: undefined.size() == 0 (2 vs. 0) :
In PrimFunc kernel variables (A, B) are used, but are not passed in as API arguments
```

The abort fires in `MakePackedAPI` ([`make_packed_api.cc#L577`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/make_packed_api.cc#L577)), reached from `tilelang/cuda/pipeline.py:248`. It is deterministic and gated entirely by warp specialization: it crashes under the default (and under an explicit `tl.disable_warp_specialized=False`), and compiles the moment `tl.disable_warp_specialized=True` is set. `T.make_tensor` is an exported public primitive, and the failing `ICHECK` is a generic TVM invariant ("all used `Var`s must be bound/passed"), not a deliberate "unsupported" `LOG(FATAL)`.

The shipped `examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py` runs its `tl.compile` with `pass_configs={"tl.disable_warp_specialized": True}` hardcoded — i.e. that example only compiles because it ships with this path pre-disabled. Confirmed on sm_90: the example runs verbatim, but flipping that one line to `False` (what a user gets under default configs) reproduces the exact `ICHECK` above.

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import math


def grouped_gemm_ptr(batch_sizes_list, K, N, block_M, block_N, block_K, threads=128, dtype=T.float16):
    batch_count = len(batch_sizes_list)
    max_M = max(batch_sizes_list)
    batch_tile_offsets = [0]
    for size in batch_sizes_list[:-1]:
        batch_tile_offsets.append(batch_tile_offsets[-1] + math.ceil(size / block_M))
    total_m_blocks = sum(math.ceil(size / block_M) for size in batch_sizes_list)

    @T.prim_func
    def kernel(
        A_ptrs: T.Tensor([batch_count], T.ptr),
        B_ptrs: T.Tensor([batch_count], T.ptr),
        C_ptrs: T.Tensor([batch_count], T.ptr),
        batch_tile_offsets: T.Tensor([batch_count], T.int32),
    ):
        with T.Kernel(total_m_blocks, T.ceildiv(N, block_N), threads=threads) as (bx, by):
            B_shared = T.alloc_shared((block_K, block_N), dtype)
            C_local = T.alloc_fragment((block_M, block_N), T.float32)
            cur_batch_idx = T.alloc_var(dtype=T.int32)
            cur_tile_offset = T.alloc_var(dtype=T.int32)
            cur_batch_idx = 0
            cur_tile_offset = 0
            for i in range(batch_count):
                in_cur = bx >= batch_tile_offsets[i]
                cur_batch_idx = T.if_then_else(in_cur, i, cur_batch_idx)
                cur_tile_offset = T.if_then_else(in_cur, batch_tile_offsets[i], cur_tile_offset)
            m_start = (bx - cur_tile_offset) * block_M
            A = T.make_tensor(A_ptrs[cur_batch_idx], (max_M, K), dtype)   # ptr-backed base
            B = T.make_tensor(B_ptrs[cur_batch_idx], (K, N), dtype)       # ptr-backed base
            C = T.make_tensor(C_ptrs[cur_batch_idx], (max_M, N), dtype)
            T.clear(C_local)
            for ko in T.Pipelined(T.ceildiv(K, block_K), num_stages=1):
                T.copy(A[m_start, ko * block_K], A_shared)
                T.copy(B[ko * block_K, by * block_N], B_shared)
            T.copy(C_local, C[m_start, by * block_N])
    return kernel


def try_compile(cfg, label):
    prog = grouped_gemm_ptr([16, 33, 64], 128, 96, 32, 32, 32)
    try:
        tl.compile(prog, target="cuda", pass_configs=cfg)
        print(f"[{label}] cfg={cfg} -> COMPILE OK")
    except Exception as e:
        print(f"[{label}] cfg={cfg} -> {type(e).__name__}: {str(e)[:120]}")


try_compile({}, "default")                                    # ICHECK abort at make_packed_api.cc:577
try_compile({"tl.disable_warp_specialized": True}, "ws-off")  # COMPILE OK
```

Output:

```
[default] cfg={} -> InternalError: Check failed: undefined.size() == 0 (2 vs. 0) : In PrimFunc kernel variables (A, B) are used, but are not passed in as API arguments
[ws-off]  cfg={'tl.disable_warp_specialized': True} -> COMPILE OK
```

### Traceback

```pytb
File ".../tilelang/cuda/pipeline.py", line 248, in CUDAPassPipelineBody
    mod = tilelang.transform.MakePackedAPI()(mod)
File "/project/src/transform/make_packed_api.cc", line 577, in tvm::tl::MakePackedAPI(tvm::tirx::PrimFunc)
tvm.error.InternalError: Check failed: undefined.size() == 0 (2 vs. 0) : In PrimFunc kernel variables (A, B) are used, but are not passed in as API arguments
```

### Expected behavior

The `make_tensor` grouped-GEMM pattern should compile under the default sm_90 pipeline — the byte-identical program compiles *and runs correctly* with `tl.disable_warp_specialized=True`, so it is well-formed and the WS path is the regression. Since the pattern is computable (the WS-off path proves it), the WS path should fall back to a plain copy for a ptr-backed source rather than aborting; failing that, it should emit a clear diagnostic — not an internal invariant `ICHECK`.

### Additional context

**Root cause (one line):** a TMA tensor-map descriptor is host-side machinery, but a `make_tensor` base pointer is computed *inside the device kernel body* (indexing the pointer table by a runtime batch index) — so when WS promotes the copy to TMA, the descriptor init lands in the host prologue referencing a device-body-local `Var`, which is then undefined.

<details>
<summary>Root cause detail (where the descriptor init lands)</summary>

With WS on, `ProducerConsumerWS` rewrites the `T.copy` into a TMA copy, and `LowerHopperIntrin` records a tensor-map `DescInit` keyed on the copy source's base pointer — here the `make_tensor` base `Var` `A`/`B` ([`lower_hopper_intrin.cc#L181-L206`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/lower_hopper_intrin.cc#L181-L206)). A `DescInit` not emitted after an `AllocBufferNode` ([`#L138-L150`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/lower_hopper_intrin.cc#L138-L150)) — and `A`/`B` are never an `AllocBuffer`'s `data`, they come from a `make_tensor` `let` — is instead pushed into the **host prologue** ([`#L38-L40`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/lower_hopper_intrin.cc#L38-L40)), where the tensor-map-create call references `A`/`B`. But `A`/`B` are bound only inside the device kernel body, so from the host prologue they are undefined, and `MakePackedAPI`'s `ICHECK_EQ(undefined.size(), 0)` ([`make_packed_api.cc#L575-L579`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/make_packed_api.cc#L575-L579)) aborts.
</details>

**Suggested fix.** Because the descriptor is host-side and the base pointer is device-body-local, this source cannot be TMA-promoted at all. The minimal safe fix is a clear diagnostic: in `LowerHopperIntrin::Substitute`, before pushing an unemitted `DescInit` into the host prologue, build the set of argument-backed data `Var`s from `f->buffer_map` and, if the `DescInit`'s `base_var` is not among them, `LOG(FATAL)` with an actionable message instead of letting the generic undefined-`Var` `ICHECK` fire later. Verified on a from-source build at this SHA: the confusing `make_packed_api.cc:577` ICHECK is replaced by a diagnostic that names the offending base pointer and points at the workaround (`"Cannot TMA-promote a copy whose source base pointer `A` is computed inside the device kernel body … Compile with pass_configs={\"tl.disable_warp_specialized\": True}"`), and the already-compiling WS-off path is unchanged. A more complete fix would instead make `RewriteCopyToTmaCopy` ([`producer_consumer_ws.cc#L786`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/producer_consumer_ws.cc#L786)) keep the plain `T.copy` for a `let`/`make_tensor`-defined source, so the default path compiles (the WS-off behavior) rather than requiring the flag.

**Provenance.** `T.make_tensor` is an exported public primitive ([`tilelang/language/__init__.py#L17`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/__init__.py#L17)) and the basis of the `examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py` example, whose own `tl.compile` hardcodes `tl.disable_warp_specialized=True` — the default path ships pre-disabled. This is not a recent regression: the ptr-table pattern has crashed under default WS since the first released version that can express it (`0.1.9`); the preceding release (`0.1.8`) cannot even construct the ptr-table tensor, so no released version ever compiled this pattern under the default pipeline.

**Not a duplicate.** #1954 (closed, v0.1.8, *"ProducerConsumerWarpSpecialized only collects LetStmt on main body"*) surfaces the same `MakePackedAPI` message, but its undefined var is a pipeline-`phase` **scalar** LetStmt moved out of scope by WS restructuring; the fix collected branch-body `LetStmt`s. Here the undefined vars are the ptr-backed `make_tensor` **base pointers**, dropped because their TMA `DescInit` is host-side and references a device-body var — a different mechanism, still live on 0.1.11. Distinct from #1263 (generic `T.Pipelined` regression, compiles on 0.1.11) and #1117 (layout-inference disagreement).

Lyscoria:
## Summary

Fix #2549.

The ptr-backed grouped GEMM pattern — reconstructing each group's `A`/`B` inside the kernel from a pointer table via `T.make_tensor(ptrs[i], shape, dtype)`, as shipped in `examples/grouped_gemm/example_grouped_gemm_fwd_ptr.py` — aborted at compile time under the default `sm_90` pipeline:

```
tvm.error.InternalError: Check failed: undefined.size() == 0 (2 vs. 0) :
```

It only compiled with `pass_configs={"tl.disable_warp_specialized": True}` (which the shipped example hardcodes).

Root cause: copy instruction selection classified every TMA-shaped global↔shared copy as a TMA candidate, including copies whose buffer comes from `T.make_tensor`. Such a base pointer is a `Bind`-bound handle var computed inside the kernel body (indexed from a device-side pointer table, per-block), while a TMA descriptor (`CUtensorMap`) must be encoded on the host before launch. `LowerHopperIntrin` therefore pushed the descriptor init into the host prologue referencing the device-local var, and `MakePackedAPI` aborted on the undefined-var check.

## Changes

The constraint ("a copy endpoint whose base pointer is bound inside the device body can never be TMA'd") is enforced once in the CUDA copy analysis layer, so every TMA selection path inherits it. Callers provide the context (the set of body-bound handle vars) that a bare `CopyNode` cannot carry.

- `src/cuda/op/copy.h`: added a `VarSet` alias and a `device_bound_ptr_vars` field on `CopyAnalysisContext`; `ClassifyWarpSpecializedProducerCopy` takes the set as an optional parameter.
- `src/cuda/op/copy_analysis.cc`:
  - `AnalyzeCopyFacts` clears all `can_bulk_*` facts for an endpoint whose base pointer is in the set, so `kBulkLoad`/`kBulkStore`/`kBulkLoad1D`/`kBulkStore1D` are all rejected in one place and selection falls through to a plain sync copy;
  - `MakeTmaUnavailableReason` gained a dedicated branch: an explicit `T.tma_copy` / `prefer_instruction="tma"` on such a copy now fails with an actionable diagnostic naming the offending base pointer, instead of the undefined-var ICHECK in `MakePackedAPI`.
- `src/cuda/op/copy.cc`: `Copy::SelectInst` builds the set from `bind_var_to_expr` (already available on `LayoutInferArgs`/`LowerArgs`) and passes it through `CopyAnalysisContext`. This also covers the non-WS auto bulk-store path, which had the same latent issue.
- `src/cuda/transform/producer_consumer_ws.cc`: added `CollectBodyBoundHandleVars` (handle-typed vars bound by `Bind` in the function body) and threaded the set into `ClassifyStmt`/`ClassifyCopy`/`IsSyncGlobalToSharedCopyLikeStmt` at both decision points (`TiledWSCandidate` and `ProducerConsumerWSRewriter`). The WS pass no longer second-guesses the classification result — the copy analysis returns the correct (non-TMA) classification directly.

## Testing

`testing/python/issue/test_tilelang_issue_2549.py`: 3 tests:

- `test_make_tensor_grouped_gemm_compiles_under_default_ws`: the issue repro (ragged ptr-table grouped GEMM) compiles under default configs, takes the non-WS path, and matches PyTorch per group;
- `test_mixed_param_and_make_tensor_copies_keep_ws_for_param_copy`: a mixed kernel keeps TMA + warp specialization for the param-backed copy, emits no descriptor for the make_tensor-backed copy, and matches PyTorch;
- `test_explicit_tma_copy_on_make_tensor_reports_actionable_error`: an explicit `T.tma_copy` store into a make_tensor destination fails with the new diagnostic (`"bound inside the device kernel body"`) instead of the `MakePackedAPI` ICHECK.

Ran on H100 (sm_90):
```
python3 -m pytest testing/python/issue/test_tilelang_issue_2549.py -q
python3 -m pytest testing/python/issue/test_tilelang_issue_tma_no_ws.py -q
```
all tests passed. `pre-commit run --all-files` was also run before committing.

## Summary

- Prevents TMA and bulk-copy promotion for pointers bound inside device kernel bodies, preserving synchronous CUDA copies.
- Propagates device-local pointer tracking through copy analysis and warp-specialized and non-warp-specialized classification paths.
- Adds actionable diagnostics for explicit TMA requests using device-local pointers.
- Adds regression coverage for ptr-backed grouped GEMM, mixed parameter- and `T.make_tensor`-backed copies, compilation, correctness, and diagnostics.

## C++ style / lint notes

- C++ implementation and API declarations were updated; no documented `cpp_style.md` rules are intentionally changed.
- The C++ API Style Audit remains warning-only. Any TLCPP003/TLCPP004 findings should be treated as advisory unless they indicate a concrete API or maintainability risk.
- No correctness or build issues are indicated; tests passed on H100 and pre-commit checks were run.


src/cuda/transform/producer_consumer_ws.cc
```diff
@@ -758,15 +703,36 @@ static bool IsProducerMovableLoopPrefixStmt(const Stmt &stmt, Target target) {
...
   if (copy == nullptr) {
     return TileStmtKind::kConsumer;
   }
 
   cuda::CopyInstSelection result = ClassifyWarpSpecializedCopy(copy, target);
   if (cuda::CopyInstIsTMA(result.inst)) {
+    // TMA descriptors are built on the host; a copy whose source base
+    // pointer is bound inside the kernel body cannot be TMA-promoted.
+    // Keep it as a plain consumer copy instead.
+    if (device_local && IsGlobalBuffer(copy->src) &&
```
LeiWang1999: this copy never ends up being lowered to TMA, so we need to improve CopyInstIsTMA in copy_analysis to prevent it from incorrectly considering this op as lowerable to TMA.

Lyscoria: Thank you! I have updated copy_analysis, and now this kind of copy is no longer considered lowerable to TMA.

src/cuda/op/copy_analysis.cc
```diff
@@ -622,6 +636,20 @@ CopyFacts AnalyzeCopyFacts(const CopyNode &op, const CopyAnalysisContext &ctx) {
   facts.can_stsm = CheckSTSMCopy(op, ctx.target);
   facts.can_tmem_load = CheckTMemLoad(op, ctx.target);
   facts.can_tmem_store = CheckTMemStore(op, ctx.target);
+  // TMA descriptors are encoded on the host; a copy endpoint whose base
+  // pointer is bound inside the device kernel body can never be TMA.
+  if (ctx.device_bound_ptr_vars) {
+    if (ctx.device_bound_ptr_vars->count(op.src->data)) {
+      facts.can_bulk_load = false;
+      facts.can_bulk_load_1d = false;
```
SiriusNEO, line 644: LowerBulk1D passes the device-computed global address directly without creating a CUtensorMap, so body-bound pointers are valid and the Bulk1D lowering path should remain enabled.

src/cuda/op/copy.cc
```diff
@@ -692,22 +693,31 @@ LayoutMap Copy::InferBulkLayout(const CopyNode &op,
 
 CopyInst Copy::SelectInst(const CopyNode &op, Target target,
                           const LayoutMap &layout_map,
-                          arith::Analyzer *analyzer, bool buffer_oob) {
+                          arith::Analyzer *analyzer, bool buffer_oob,
+                          const Map<Var, PrimExpr> &bind_var_to_expr) {
+  cuda::VarSet device_bound_local_vars;
+  for (const auto &[var, expr] : bind_var_to_expr) {
```
SiriusNEO, line 699: In this PR, I found three places where `device_bound_local_vars` (or something similar) is constructed or populated, but they use different definitions of a device-bound pointer: `ProducerConsumerWS` scans every handle-typed Bind, `LayoutInference` records every Bind, while `LowerTileOp` records only bindings whose RHS is a direct `BufferLoad`. Consequently, the `reinterpret(handle, BufferLoad(...))` binding produced by `T.make_tensor(ptrs[i], ...)` is absent from the final lowering context, allowing an automatic shared-to-global copy to be re-promoted to descriptor-based TMA. Could we use one dedicated provenance representation and propagate it consistently through these stages instead of deriving it from pass-specific `bind_var_to_expr_` maps?

SiriusNEO (changes requested):
The current change fixes the main warp-specialized repro, but I suggest refactoring it around a single, explicit notion of TMA descriptor eligibility instead of reconstructing device-bound pointer sets from multiple pass-local binding maps

SiriusNEO:
Close since we have discussed a better implementation for this issue: https://github.com/tile-ai/tilelang/pull/2803. Also thanks for your time and effort on this!

Issue #2184, opened 2026-05-11
[BUG] ProducerConsumerWarpSpecialized does not trigger when `T.Pipelined` is wrapped by an outer runtime `if`


dcaox:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.9

### System information

3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] linux
0.1.9
2.8.0a0+5228986c39.nv25.05

PyTorch version: 2.8.0a0+5228986c39.nv25.05
CUDA used to build PyTorch: 12.9
OS: Ubuntu 24.04 x86_64
Python version: 3.12.3
Is CUDA available: True
GPU: NVIDIA H200
Nvidia driver version: 570.86.15
CUDA runtime version: 12.9.41
cuDNN version: 9.x

CPU: x86_64, Intel Xeon, 2 sockets, 192 logical CPUs

Relevant packages:
numpy==1.26.4
torch==2.8.0a0+5228986c39.nv25.5
torchvision==0.22.0a0
triton==3.1.0
pytorch-triton==3.3.0+<internal-build>
torch_tensorrt==2.8.0a0

### Problem description

After upgrading from TileLang 0.1.7 to TileLang 0.1.9, a varlen FlashAttention forward kernel stopped generating the automatic producer/consumer warp-specialized TMA path. The generated CUDA falls back to ordinary global loads, `__launch_bounds__(256, 1)`, and no `CUtensorMap` / `tma_load` / `mbarrier`.

I reduced the issue to a minimal GEMM repro: when `T.Pipelined` is directly visible in the kernel body, TileLang generates automatic WS/TMA. If the same `T.Pipelined` loop is wrapped by an outer runtime `if`, automatic WS/TMA is not generated.

### Reproducible example code

The Python snippets:

```python

import argparse
import re

import torch
import tilelang
import tilelang.language as T


M = 256
N = 128
K = 256
BLOCK_M = 128
BLOCK_N = 128
BLOCK_K = 64
THREADS = 256
NUM_STAGES = 2


def summarize_source(src):
    markers = [
        "CUtensorMap",
        "tma_load",
        "mbarrier",
        "warpgroup_reg",
        "__launch_bounds__(384",
        "__launch_bounds__(256",
[...]

Issue #972, opened 2025-10-10
[bug] failed to run gemm_streamk example on sm90


LingYeAI:
cuda:12.6
arch: sm90 (H20)
file: example_tilelang_gemm_streamk.py

**cmd output:**
CUDA error: unspecified launch failure

**compute-santinizer ouput:**
```
========= Illegal barrier arrive operation
=========     at cutlass::arch::ClusterBarrier::arrive(const unsigned long *)+0x6560 in /workspace/tilelang/3rdparty/cutlass/include/cutlass/arch/barrier.h:511
=========     by thread (96,0,0) in block (17,0,0)
=========     Device Frame:cutlass::arch::ClusterBarrier::arrive() const+0x6560 in /workspace/tilelang/3rdparty/cutlass/include/cutlass/arch/barrier.h:377
=========     Device Frame:main_kernel+0x6560 in /tmp/tmpbpv8o6_n.cu:49
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x381257]
=========                in /usr/lib64/libcuda.so.1
=========     Host Frame:libcudart_static_4d8b33a106dceb3c07a56e26de61f2d53bb62a68 [0x12532]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:cudaLaunchKernel [0x7619d]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:cudaError cudaLaunchKernel<char>(char*, dim3, dim3, void**, unsigned long, CUstream_st*) [0xcb80]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:__device_stub__Z11main_kernel14CUtensorMap_stS_PN7cutlass6half_tE(CUtensorMap_st const&, CUtensorMap_st const&, cutlass::half_t*) [0xc840]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:main_kernel [0xc869]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:call [0xc609]
=========                in /tmp/tmpbpv8o6_n.so
=========     Host Frame:ffi_call_unix64 [0xa051]
=========                in /opt/conda310/lib/python3.10/lib-dynload/../../libffi.so.8
=========     Host Frame:ffi_call_int [0x88cc]
=========                in /opt/conda310/lib/python3.10/lib-dynload/../../libffi.so.8
[...]

PR #2692, opened 2026-07-17
[BugFix] Skip warp specialization for pipeline loops inside while loops

### LINKED ISSUE #2548: [BUG][Fuzzer][ice-on-valid-code] Regression (0.1.10): the Hopper warp-specialization pass hard-aborts on a `T.Pipelined` loop inside a `while` with an internal `ICHECK`, instead of declining WS and falling back to the standard pipeline as it did on 0.1.9

zyy3077:
## Issue

https://github.com/tile-ai/tilelang/issues/2548

## Summary

On Hopper (sm_90) with the default warp-specialized (WS) pipeline, a `T.Pipelined` loop nested inside a `while` loop — the core structure of stream-K GEMM, including the shipped `examples/gemm_streamk` — aborts compilation with:

```
tvm.error.InternalError: Check failed: (replaced.defined()) is false:
ProducerConsumerWS: failed to replace pipeline loop
```

The same kernel compiles and runs correctly with `tl.disable_warp_specialized=True`, and the same pipeline loop without the enclosing `while` warp-specializes fine. This is a 0.1.9 → 0.1.10 regression: 0.1.9 silently skipped this shape and fell back to the `cp.async` pipeline.

## Root cause

`ProducerConsumerWSRewriter` locates and rewrites the pipeline loop asymmetrically:

- `PipelineLoopFinder` derives from `StmtVisitor`, whose default visitor recurses into `While` bodies, so it finds the pipeline `For` inside a `while` and commits to the WS rewrite.
- The replacement helpers (`PipelineLoopContainmentChecker` and `PipelineLoopInStmtReplacer`) are `StmtFunctor`s with explicit handlers only for `SeqStmt`/`SBlockRealize`/`SBlock`/`AttrStmt`/`IfThenElse`; a `While` falls through to `VisitStmtDefault_` and reports "not found", so `ReplacePipelineLoopInStmt` returns an undefined result and the `ICHECK(replaced.defined())` fires.

Teaching the replacer to descend into `While` is not the right fix: WS partitioning cannot handle the loop-carried mutable state a `while` carries across iterations (the issue reporter verified this variant compiles but fails at launch, echoing https://github.com/tile-ai/tilelang/issues/972).

## Fix

Make the finder symmetric with the replacer instead: add an empty `VisitStmt_(const WhileNode *)` override to `PipelineLoopFinder` so it does not descend into `While` bodies. While-wrapped pipeline loops then take the existing non-WS `cp.async` fallback — the same accepted behavior as the `if`-wrapped case (https://github.com/tile-ai/tilelang/issues/2184) and as 0.1.9.

Orthogonal known gap, not addressed here: pipeline loops under an outer `For` (`T.Persistent` / `T.serial` persistent kernels) hit the same `ICHECK` via the missing `ForNode` handler in the replacement helpers, and is being fixed independently in https://github.com/tile-ai/tilelang/pull/2674 by supporting the rewrite. Because the trailing `T.serial`-wrapped pipeline loop in `examples/gemm_streamk` still trips that path, this PR keeps the example's sm90 test gate unchanged; it can be lifted once https://github.com/tile-ai/tilelang/pull/2674 lands.

## Tests

Added two regression tests in `testing/python/transform/test_tilelang_transform_producer_consumer_ws.py`, using a minimal stream-K style while-wrapped pipelined GEMM:

- `test_tiled_ws_declines_pipeline_loop_inside_while`: runs `ProducerConsumerWarpSpecialized` standalone on sm_90 and asserts the pass declines (no `tl_tiled_ws_applied` / `kWarpSpecializationScope` in the output IR) instead of crashing.
- `test_tiled_ws_while_pipelined_correctness`: end-to-end compile and run on sm_90 via the non-WS fallback, checked against a PyTorch reference.

Both tests reproduce the original `ICHECK` failure without the one-line fix and pass with it. The full `test_tilelang_transform_producer_consumer_ws.py` file passes on H200.

## C++ style / lint notes

This PR touches TileLang-owned C++ transform code under `src/cuda/transform/producer_consumer_ws.cc`; the change is a single empty visitor override on the pass-local `PipelineLoopFinder` with a comment explaining the invariant, following the surrounding visitor style. Checked with the repository's `clang-format` and `ruff-format` pre-commit hooks.

Fixes https://github.com/tile-ai/tilelang/issues/2548

## Summary

- Updated `ProducerConsumerWSRewriter`’s `PipelineLoopFinder` to stop traversal at `while` nodes, preventing Hopper warp-specialization rewriting from triggering on `T.Pipelined` loops nested inside `while` bodies.
- With traversal disabled for `while`, these nested pipeline loops decline warp specialization and fall back to the standard `cp.async`/non-WS pipeline behavior instead of hitting Hopper internal assertions.
- Added Hopper regression coverage by introducing a `matmul_while_pipelined` kernel plus:
  - a test (`test_tiled_ws_declines_pipeline_loop_inside_while`) asserting warp-specialization markers are absent in the compiled output when the pipeline loop is inside a `while`, and
  - a correctness test (`test_tiled_ws_while_pipelined_correctness`) compiling/running the non-WS fallback on `sm_90` and validating results vs `A @ B.T`.
- Left the existing `gemm_streamk` `sm90` test gate unchanged, since pipeline loops under outer `For` constructs are addressed by separate logic/tests.

## C++ style / lint notes

- This PR changes C++ implementation in `src/cuda/transform/producer_consumer_ws.cc` but does not modify `docs/developer_guide/cpp_style.md` (no C++ style-doc updates).
- No CI/lint tooling or configuration files were touched.
- The “C++ API Style Audit (warning only)” CI step should still run as usual; this change does not introduce public API/FFI/maintainability risk and is not expected to add new TLCPP003/TLCPP004-style warnings.


src/cuda/transform/producer_consumer_ws.cc
```diff
@@ -2177,6 +2177,12 @@ class ProducerConsumerWSRewriter : public StmtExprMutator {
       StmtVisitor::VisitStmt_(op);
     }
 
+    // WS partitioning cannot handle pipeline loops whose enclosing `while`
+    // carries mutable state across iterations, and the replacement helpers
+    // below do not traverse While nodes. Skip them so such loops fall back
+    // to the non-WS software pipeline.
+    void VisitStmt_(const WhileNode *op) final {}
```
LeiWang1999: would be better to introduce a warning.

LeiWang1999 (changes requested):

PR #2694, opened 2026-07-18
[BugFix] Align multi-version shared buffer strides for TMA

### LINKED ISSUE #2528: [BUG][Fuzzer][ice-on-valid-code] Pipelined `T.copy` into a 1-D `block_K=32` fp16 shared buffer crashes with `CUDA error: misaligned address` instead of running
### Required prerequisites

0.1.11

### System information

- TileLang `0.1.11`; source pinned to commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`.
- Python 3.12, linux; torch 2.12.1+cu130, CUDA 13.0.
- GPU: **NVIDIA H100 80GB HBM3** (compute capability 9.0, sm_90). **Reproduced at runtime on this H100.** This is a Hopper-path defect: the `cp.async.bulk.tensor` (TMA) lowering of these 1-D copies is selected on sm_90, and only the TMA shared destination carries a 128-byte-alignment requirement.

### Problem description

A pipelined `T.copy` from global into a **1-D** `T.alloc_shared((block_K,), float16)` buffer hard-crashes with **`CUDA error: misaligned address`** whenever `block_K=32` is paired with `num_stages>=2`. The kernel compiles and launches, then faults asynchronously at the next sync. `block_K=64` and `num_stages=1` both run and produce correct results.

The cause is that with `num_stages>=2` the pipeline double-buffers the shared buffer by prepending a version dimension with **no alignment padding**, giving a 64-byte per-stage stride at `block_K=32` fp16 (`32*2` bytes). On sm_90 the copy lowers to `cp.async.bulk.tensor.1d`, whose shared destination must be 128-byte aligned; the odd buffer version lands at `64 mod 128` and the hardware rejects it. This surfaced in the shipped mamba `chunk_state` example (`examples/linear_attention/example_mamba_chunk_state.py`), whose autotune grid includes `block_K=32, num_stages>=2` and therefore materializes the crashing configs — but the defect is in the generic pipelined-1-D-TMA-copy path, not the example.

<details>
<summary>Discriminating control — same kernel, only <code>block_K</code>/<code>num_stages</code> differ (H100, TESTED)</summary>

The only difference between a crashing and a clean run is whether the per-stage stride of the 1-D shared TMA destination is a multiple of 128 bytes. Same kernel, same fp16 dtype, same TMA path:

| `block_K` | `num_stages` | per-stage stride of `dt_shared` | result |
|-----------|-------------|----------------------------------|--------|
| 64 | 2 | `(k&1)*64` halfs = 128 B (128-aligned) | **OK, `match=True`** |
| 32 | 1 | no version dim (single buffer) | **OK, `match=True`** |
| 32 | 2 | `(k&1)*32` halfs = 64 B (`64 mod 128`) | **`CUDA error: misaligned address`** |

Both passing configs produce the correct result over the same inputs, isolating the defect to the 64-byte per-stage shared-destination stride — not the math, the inputs, or the harness.
</details>

<details>
<summary>Generated CUDA for the crashing vs the passing config</summary>

The base of the 1-D shared buffer is declared `__align__(1024)` (well over 128-byte aligned); the sole misalignment is the odd-version element offset. `block_K=32, num_stages=2` (crashing):

```cpp
extern __shared__ __align__(1024) half_t dt_shared[];   // base is 128-aligned
...
tl::tma_load(dt_desc, mbarrier[(k & 1)], (&(dt_shared[((k & 1) * 32)])), (k * 32));
//          odd-version offset = 32 halfs = 64 bytes -> absolute addr % 128 == 64  (MISALIGNED)
```

`block_K=64, num_stages=2` (the passing control) — per-stage offset is now 128 bytes, aligned:

```cpp
tl::tma_load(dt_desc, mbarrier[(k & 1)], (&(dt_shared[((k & 1) * 64)])), (k * 64));
//          (k&1)*64 halfs = 128 B -> ALIGNED
```

In the full mamba example the same pattern holds for both `dt_shared` and `dA_cumsum_shared`; the 2-D shared buffers in that loop are immune because their per-stage strides are large multiples of 128 (`x_shared` steps 4096 B, `B_shared` steps 8192 B).
</details>

<details>
<summary>Root cause — three cooperating facts (source, pinned)</summary>

**(1) The pipeline multi-version expansion prepends an unpadded version dimension.** When `num_stages>1`, `RewriteAllocBuffer` double-buffers each pipelined shared buffer ([`multi_version_buffer_rewriter.cc#L410-L426`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/multi_version_buffer_rewriter.cc#L410-L426)). For a non-barrier shared buffer it simply prepends `num_versions` as a new leading dimension — `new_buffer->shape.insert(new_buffer->shape.begin(), PrimExpr(num_versions))` ([`#L417-L418`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/multi_version_buffer_rewriter.cc#L417-L418)) — with **no alignment padding** of the innermost extent. So `(block_K,) = (32,)` becomes `(num_versions, 32)`, and version `v` starts at element offset `v*32` = `v*64` bytes for fp16, never rounded up to a 128-byte boundary. For `block_K=64` the per-version stride is `64*2 = 128` bytes (aligned by coincidence); the 2-D buffers are aligned because their per-version strides are large multiples of 128. Only the small 1-D buffers land at `64 mod 128`. The `if (num_stages <= 1)` filter at [`#L494-L519`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/multi_version_buffer_rewriter.cc#L494-L519) returns only barrier buffers, so `num_stages=1` never gets a version dimension and never crashes.

**(2) The TMA-eligibility analysis never checks the shared destination's alignment, and runs before multi-versioning.** The 1-D copy is decided TMA-eligible via `CheckBulkLoad1D` ([`copy_analysis.cc#L312-L320`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L312-L320)) → `CheckBulkLoad` + `CheckBulkCopy1D`. `CheckBulkLoad` ([`#L192-L228`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L192-L228)) validates scope, the global last-dim being a multiple of 16 bytes, the dtype pair, and global strides — never the shared destination's byte alignment. `CheckBulkCopy1D` ([`#L268-L310`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L268-L310)) checks shared-side contiguity and element-count match — again no per-stage shared-dest alignment. This decision is made on the **pre-multi-versioned** buffer, so it cannot see the `(num_versions, block_K)` layout that fact (1) later produces.

**(3) The emitted instruction requires 128-byte alignment.** The 1-D `tma_load` ([`copy_sm90.h#L212-L240`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/tl_templates/cuda/copy_sm90.h#L212-L240)) emits `cp.async.bulk.tensor.1d.shared::cta.global` with the shared pointer `&dt_shared[(k&1)*32]`. That instruction family requires a 128-byte-aligned shared destination; the 64-byte odd-version offset violates it and the hardware raises `misaligned address`.
</details>

### Reproducible example code

Run on an sm_90 (Hopper) GPU so the TMA path is selected.

```python
# Pipelined T.copy into a 1-D fp16 shared buffer: block_K=32 with num_stages>=2
# faults with "CUDA error: misaligned address". The pipeline double-buffers
# dt_shared with a 64-byte per-stage stride, so the odd version's shared
# destination is at 64 mod 128 -- and cp.async.bulk.tensor.1d (sm_90) requires
# a 128-byte-aligned shared destination. block_K=64 (128-byte stride) and
# num_stages=1 (no double-buffer) both run.
import tilelang
import torch
import tilelang.language as T

def make_kernel(chunk_size, block_K, num_stages, threads=128):
    dtype = T.float16; accum = T.float32
    @T.prim_func
    def f(
        dt: T.Tensor((chunk_size,), dtype),
        Out: T.Tensor((chunk_size,), accum),
    ):
        with T.Kernel(1, threads=threads) as bx:
            dt_shared = T.alloc_shared((block_K,), dtype)      # 1-D pipelined TMA dest
            dt_local  = T.alloc_fragment((block_K,), accum)
            for k in T.Pipelined(T.ceildiv(chunk_size, block_K), num_stages=num_stages):
                T.copy(dt[k*block_K:(k+1)*block_K], dt_shared)  # -> cp.async.bulk.tensor.1d on sm_90
                T.copy(dt_shared, dt_local)
                for i in T.Parallel(block_K):
                    Out[k*block_K+i] = dt_local[i] * 2.0
    return tilelang.compile(f)

def trial(block_K, num_stages):
    chunk_size = 256
    torch.manual_seed(0)
    dt = torch.randint(0, 3, (chunk_size,), device="cuda").half()
    out = torch.empty((chunk_size,), device="cuda", dtype=torch.float32)
    make_kernel(chunk_size, block_K, num_stages)(dt, out)
    torch.cuda.synchronize()
    print(f"  block_K={block_K} num_stages={num_stages}: OK  match={torch.allclose(out, dt.float()*2.0)}")

if __name__ == "__main__":
    print("block_K=64 num_stages=2 (expected OK):");    trial(64, 2)  # -> OK  match=True
    print("block_K=32 num_stages=1 (expected OK):");    trial(32, 1)  # -> OK  match=True
    print("block_K=32 num_stages=2 (expected CRASH):"); trial(32, 2)  # -> CUDA error: misaligned address
```

Output (captured on H100 this session):

```
block_K=64 num_stages=2 (expected OK):
  block_K=64 num_stages=2: OK  match=True
block_K=32 num_stages=1 (expected OK):
  block_K=32 num_stages=1: OK  match=True
block_K=32 num_stages=2 (expected CRASH):
torch.AcceleratorError: CUDA error: misaligned address
```

The shipped mamba `chunk_state` example hits the same crash directly: its autotune grid is `block_K=[32, 64], num_stages=[1, 2, 3, 4, 5]` ([`example_mamba_chunk_state.py#L48-L50`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/linear_attention/example_mamba_chunk_state.py#L48-L50)), and it declares `dt_shared`/`dA_cumsum_shared` as 1-D `T.alloc_shared((block_K))` ([`#L73-L74`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/linear_attention/example_mamba_chunk_state.py#L73-L74)) filled by pipelined `T.copy` ([`#L103-L104`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/examples/linear_attention/example_mamba_chunk_state.py#L103-L104)), so the autotuner benchmarks the crashing `block_K=32, num_stages>=2` configs.

### Traceback

```pytb
  File "repro.py", line 40, in <module>
    trial(32, 2)
  File "repro.py", line 34, in trial
    torch.cuda.synchronize()
  File ".../torch/cuda/__init__.py", line 1181, in synchronize
    return torch._C._cuda_synchronize()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: misaligned address


The error surfaces asynchronously at `torch.cuda.synchronize()`; the kernel launch itself returns. The faulting instruction is the `cp.async.bulk.tensor.1d` TMA load into the odd version of the 1-D shared buffer.
```

### Expected behavior

A well-formed pipelined `T.copy` into a 1-D shared buffer should run for every `block_K`/`num_stages` combination — including `block_K=32, num_stages>=2` — and produce correct results, rather than faulting with `CUDA error: misaligned address`. (If a config is genuinely unsupportable, it should be rejected at compile time with a clear diagnostic rather than emitting an instruction the hardware rejects at runtime.)

Grounded in TileLang's own behavior:

- The **byte-identical** kernel runs correctly at `block_K=64, num_stages=2` and at `block_K=32, num_stages=1`, so the program is well-formed and TMA-promotion of these 1-D loads is intended; the `block_K=32, num_stages>=2` path is the regression.
- The 2-D pipelined shared buffers are already aligned because their per-version strides are large multiples of 128 — the layout machinery already produces aligned multi-version buffers in the common case; it simply fails to pad the small 1-D case.

### Additional context

**Provenance.** The unpadded version-dimension insert in `RewriteAllocBuffer` ([`#L417-L418`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/multi_version_buffer_rewriter.cc#L417-L418)) was introduced by [#1909](https://github.com/tile-ai/tilelang/pull/1909) (*Add Producer-Consumer Warp Specialization and `T.tma_copy()` API*, merged 2026-03-18) — `git blame -L417,418` on the pinned commit points at commit `ded6a99`, which is #1909. [#2297](https://github.com/tile-ai/tilelang/pull/2297) (merged 2026-06-02) only relocated the file into the `cuda/` transform namespace and did not change the logic.

<details>
<summary>Suggested fix (proposed, not built or tested)</summary>

Two complementary options; (1) keeps the fast TMA path, (2) is a safety net.

1. **Pad the per-version stride of TMA-destination shared buffers to 128 bytes** in `RewriteAllocBuffer` ([`#L410-L426`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/multi_version_buffer_rewriter.cc#L410-L426)). For the non-barrier shared case, after inserting the version dimension, synthesize the version stride as `round_up(prod(trailing_extents) * elem_bytes, 128) / elem_bytes` (or pad the innermost shape dim to the 128-byte multiple), leaving the logical innermost extent unchanged so indexing stays correct. This rounds the 64-byte `block_K=32` region up to 128 bytes, matching how the 2-D buffers already happen to be aligned.

2. **Reject or fall back when the per-stage shared destination cannot be 128-byte aligned.** Since `CheckBulkLoad1D` currently runs before multi-versioning, propagate `num_stages` into the eligibility decision (or re-check where pipelining is known) and fall back to a non-TMA copy (`cp.async` / vectorized, which has no 128-byte requirement) when `num_stages>1` and the innermost shared extent in bytes is not a multiple of 128 (for fp16, `block_K % 64 != 0`).

**Validation:** with either fix, the `block_K=32, num_stages=2` trial must complete without the misaligned-address crash and match the reference. I have not built or verified any of these changes.
</details>

I searched the open and closed tracker and found no existing report of this defect. The closest live report is [#2506](https://github.com/tile-ai/tilelang/issues/2506) (open) — a `T.copy_cluster` **transfer-size** (non-16B) hang on the descriptor-less cluster path; this is a distinct failure (a **shared-destination** 128-byte misalignment on the descriptor-based 1-D load path, faulting rather than hanging).

UnsettingGalaxy:
## Summary

This PR fixes a shared-memory alignment issue in the multi-version buffer rewriter for pipelined 1-D copies on Hopper GPUs.

When a compact 1-D shared buffer is used by a multi-stage pipeline, TileLang prepends a version dimension to create one buffer region for each pipeline stage. Previously, the original buffer extent was used directly as the per-version stride.

For an `fp16[32]` buffer, each version therefore occupied only 64 bytes. Although the shared-memory allocation itself was aligned, the second pipeline version started at a 64-byte offset and produced a misaligned destination for the TMA load.

The fix makes the rewritten physical layout consistent with the TMA alignment requirement:

* Pad the extent of compact 1-D versioned buffers so that each version occupies a multiple of 128 bytes.
* Compute flattened per-version pointer offsets from the rewritten buffer shape, ensuring that `tvm_access_ptr` and regular buffer indexing use the same padded layout.

For example, an `fp16[32]` buffer with two versions is rewritten from:

```text
[2, 32]
```

to:

```text
[2, 64]
```

The logical buffer still contains 32 elements per version, but the physical distance between versions becomes 128 bytes instead of 64 bytes.

Fixes #2528.

## Testing

Added the reproduction from issue #2528 as a regression test in testing/python/issue/test_tilelang_issue_2528.py.

The test covers:

* `block_K=32, num_stages=2`, which previously failed with `CUDA error: misaligned address`;
* `block_K=64, num_stages=2`, where the original per-version stride is already 128-byte aligned;
* `block_K=32, num_stages=1`, where no multi-version buffer is introduced.

All configurations are checked against the reference result.

## Summary

- Fixed Hopper `sm_90` 1-D TMA multi-version shared-memory alignment by rewriting compact buffers to insert a pipeline-version dimension and pad each version’s innermost extent so every version’s physical span remains 128-byte aligned (`kTmaSharedMemAlignBytes`).
- Kept pointer math consistent by updating `tvm_access_ptr` rewriting to compute per-version flattened offsets from the rewritten (padded) buffer shape, aligning the version spacing used for regular buffer indexing with the versioned pointer offset (`ComputeVersionOffset`).
- Updated allocation/access rewriting logic for shared-memory multi-version buffers: `RewriteAllocBuffer` now pads for alignment when necessary and maintains a contiguous rewritten layout (without synthesizing leading strides from the original buffer’s `strides`), and `RewriteBufferAccess` always uses the new helper for version offsets.
- Added regression coverage in `testing/python/issue/test_tilelang_issue_2528.py` for pipelined 1-D FP16 shared-memory staging:
  - `block_K=64, num_stages=2` (expected OK)
  - `block_K=32, num_stages=1` (expected OK)
  - `block_K=32, num_stages=2` (expected “misaligned address” CUDA error reproducer)

## C++ style / lint notes

- The C++ change is confined to `src/cuda/transform/multi_version_buffer_rewriter.cc` and does not touch rules documented in `docs/developer_guide/cpp_style.md`.
- No C++ public API/FFI changes were introduced; any “C++ API Style Audit (warning only)” findings (e.g., TLCPP003/TLCPP004) should be treated as advisory rather than correctness/build blockers unless a concrete API/FFI/maintainability risk is introduced by this change.

### DIFF
--- a/src/cuda/transform/multi_version_buffer_rewriter.cc
+++ b/src/cuda/transform/multi_version_buffer_rewriter.cc
@@ -75,7 +75,7 @@ static void PadShapeForVersionAlignment(BufferNode *new_buffer,
   const int64_t dtype_bytes =
       orig_buffer->dtype.bytes() * orig_buffer->dtype.lanes();
   if ((elems * dtype_bytes) % align_bytes == 0)
-    return; 
+    return;
 
   const auto *last_imm = orig_buffer->shape.back().as<IntImmNode>();
   const int64_t last = last_imm->value;
--- a/src/cuda/transform/multi_version_buffer_rewriter.cc
+++ b/src/cuda/transform/multi_version_buffer_rewriter.cc
@@ -14,6 +14,7 @@
 #include <tvm/tirx/stmt_functor.h>
 
 #include <functional>
+#include <numeric>
 #include <unordered_set>
 #include <utility>
 #include <vector>
@@ -34,6 +35,66 @@ using namespace ffi;
 using tirx::GetSBlockReadWriteRegion;
 
 namespace {
+
+// TMA (cp.async.bulk.tensor.*) requires 128-byte aligned shared-memory
+// addresses. Every version of a multi-versioned shared buffer is a
+// potential TMA destination base, so the stride between versions must
+// preserve this alignment. Whether a copy really lowers to TMA is only
+// decided later (in copy lowering), so all versioned shared buffers are
+// padded; the cost is at most 127 bytes per version.
+constexpr int64_t kTmaSharedMemAlignBytes = 128;
+
+// Compact (row-major) strides for `shape`: [prod(shape[1:]), ..., 1].
+Array<PrimExpr> MakeCompactStrides(const Array<PrimExpr> &shape) {
+  ICHECK(!shape.empty());
+  std::vector<PrimExpr> strides(shape.size());
+  PrimExpr stride = make_const(shape[0].dtype(), 1);
+  for (size_t i = shape.size(); i > 0; --i) {
+    strides[i - 1] = stride;
+    stride = stride * shape[i - 1];
+  }
+  return Array<PrimExpr>(strides.begin(), strides.end());
+}
+
+// Stride in elements between consecutive versions of `buffer`, where
+// `version_strides` is the physical layout of one version (its explicit
+// strides, or synthesized compact strides). The physical span of one
+// version is version_strides[0] * shape[0] -- the same rule
+// Buffer::GetFlattenedBuffer and GetBufferAllocationShape use -- rounded
+// up to the TMA shared-memory alignment.
+//
+// The buffer stride-nesting invariant (strides[i-1] % strides[i] == 0,
+// enforced by GetBufferAllocationShape) must keep holding for the pair
+// (version_stride, version_strides[0]), so the rounding unit is
+// lcm(alignment_in_elements, version_strides[0]).
+//
+// Falls back to the unpadded span when the alignment is not expressible:
+// an element bit width that does not divide 128 bytes, or a symbolic
+// version_strides[0] (the nesting invariant could not be guaranteed).
+PrimExpr AlignedVersionStride(const Buffer &buffer,
+                              const Array<PrimExpr> &version_strides) {
+  ICHECK_EQ(version_strides.size(), buffer->shape.size());
+  PrimExpr span = version_strides[0] * buffer->shape[0];
+  int64_t elem_bits =
+      static_cast<int64_t>(buffer->dtype.bits()) * buffer->dtype.lanes();
+  constexpr int64_t align_bits = kTmaSharedMemAlignBytes * 8;
+  if (elem_bits <= 0 || align_bits % elem_bits != 0) {
+    return span;
+  }
+  int64_t align_elems = align_bits / elem_bits;
+  const int64_t *inner_stride = as_const_int(version_strides[0]);
+  if (inner_stride == nullptr || *inner_stride <= 0) {
+    return span;
+  }
+  int64_t unit = std::lcm(align_elems, *inner_stride);
+  if (const int64_t *span_imm = as_const_int(span)) {
+    return make_const(span.dtype(), (*span_imm + unit - 1) / unit * unit);
+  }
+  // Symbolic leading extent: round up symbolically. `unit` is a multiple
+  // of version_strides[0], so the nesting invariant still holds.
+  PrimExpr unit_expr = make_const(span.dtype(), unit);
+  return floordiv(span + (unit_expr - 1), unit_expr) * unit_expr;
+}
 
 bool ShapesEqual(const Array<PrimExpr> &lhs, const Array<PrimExpr> &rhs,
                  arith::Analyzer *analyzer) {
@@ -497,30 +558,21 @@ private:
       // Barrier buffers: expand first dimension to keep 1D shape.
       // (1,) -> (num_versions,) so lower_shared_barrier.cc still works.
       new_buffer->shape.Set(0, PrimExpr(num_versions) * new_buffer->shape[0]);
-    } else {
-      if (new_buffer->shape.size() == 1 && new_buffer->strides.empty()) {
-        constexpr int64_t kTmaSmemAlignBits = 128 * 8;
-        const int64_t *extent = as_const_int(new_buffer->shape[0]);
-        int64_t elem_bits = static_cast<int64_t>(new_buffer->dtype.bits()) *
-                            new_buffer->dtype.lanes();
-        if (extent != nullptr && elem_bits > 0 &&
-            kTmaSmemAlignBits % elem_bits == 0) {
-          int64_t total_bits = *extent * elem_bits;
-          if (total_bits % kTmaSmemAlignBits != 0) {
-            int64_t padded_bits = (total_bits + kTmaSmemAlignBits - 1) /
-                                  kTmaSmemAlignBits * kTmaSmemAlignBits;
-            new_buffer->shape.Set(0, IntImm(new_buffer->shape[0].dtype(),
-                                            padded_bits / elem_bits));
-          }
-        }
-      }
-      new_buffer->shape.insert(new_buffer->shape.begin(),
-                               PrimExpr(num_versions));
-      if (!new_buffer->strides.empty()) {
-        ICHECK(new_buffer->strides.size() + 1 == new_buffer->shape.size());
-        PrimExpr stride_0 = new_buffer->strides[0] * new_buffer->shape[1];
-        new_buffer->strides.insert(new_buffer->strides.begin(), stride_0);
-      }
+      return Buffer(new_buffer);
+    }
+    // Versioned layout: logical shape [num_versions] + original_shape,
+    // physical strides [aligned_version_stride] + original/compact strides.
+    // The original logical shape is preserved; inter-version padding is
+    // expressed only through the leading stride, which every access path
+    // reads back as the canonical per-version offset.
+    new_buffer->shape.insert(new_buffer->shape.begin(), PrimExpr(num_versions));
+    if (!buffer->shape.empty()) {
+      Array<PrimExpr> strides = buffer->strides.empty()
+                                    ? MakeCompactStrides(buffer->shape)
+                                    : buffer->strides;
+      PrimExpr version_stride = AlignedVersionStride(buffer, strides);
+      strides.insert(strides.begin(), version_stride);
+      new_buffer->strides = std::move(strides);
     }
     return Buffer(new_buffer);
   }
@@ -881,13 +933,6 @@ private:
 
   PrimExpr RewriteBufferAccess(const Call &call,
                                const std::vector<int> &arg_indices) {
-    auto product = [](const Array<PrimExpr> &input) {
-      return foldl(
-          [](PrimExpr a, PrimExpr b, Span span) {
-            return mul(std::move(a), std::move(b), std::move(span));
-          },
-          make_const(DataType::Int(32), 1), input);
-    };
     Array<PrimExpr> new_args = call->args;
     for (int i : arg_indices) {
       auto buffer_var = Downcast<Var>(call->args[i]);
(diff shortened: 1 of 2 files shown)


src/cuda/transform/multi_version_buffer_rewriter.cc
```diff
@@ -498,6 +498,22 @@ class MultiVersionBufferRewriter : public StmtExprMutator {
       // (1,) -> (num_versions,) so lower_shared_barrier.cc still works.
       new_buffer->shape.Set(0, PrimExpr(num_versions) * new_buffer->shape[0]);
     } else {
+      if (new_buffer->shape.size() == 1 && new_buffer->strides.empty()) {
```
LeiWang1999: It's inelegant, strictly 1D, and visually unappealing."

LeiWang1999 (changes requested):

LeiWang1999:
made another fix for this issue, checkout #2737

PR #2695, opened 2026-07-18
[BugFix] Correct the spelling of fragment

penguin-wwy:
## Summary

- Fixed the buffer allocation scope spelling in `src/transform/layout_inference.cc` by correcting `local.framgent` to `local.fragment`, ensuring layout inference attaches inferred layouts only for buffers allocated in the intended fragment scope.
- Updated `examples/gdn/example_chunk_o_bwd.py` so `dk_fragment_2` is allocated only inside the `else` branch (`use_g == False`), matching where the fragment is actually used.

## C++ style / lint notes

- Does not touch rules in `docs/developer_guide/cpp_style.md`.
- Does not modify CI/lint tooling (including `maint/scripts/audit_cpp_api_style.py`); the “C++ API Style Audit (warning only)” step remains advisory.
- No new TLCPP003/TLCPP004-style correctness/build concerns are expected, since this is a scope-string/layout-inference fix with no public API/FFI/header surface changes.

### DIFF
--- a/src/transform/layout_inference.cc
+++ b/src/transform/layout_inference.cc
@@ -1215,11 +1215,11 @@ private:
    * @brief Visit and mutate a Block node to attach inferred layout information.
    *
    * Converts the visited Block via the base visitor, asserts that every buffer
-   * allocated with scope "local.framgent" has an inferred layout in
+   * allocated with scope "local.fragment" has an inferred layout in
    * result_.layout_map, and attaches result_.layout_map to the Block's
    * annotations under attr::kLayoutMap.
    *
-   * If any "local.framgent" buffer lacks an entry in result_.layout_map an
+   * If any "local.fragment" buffer lacks an entry in result_.layout_map an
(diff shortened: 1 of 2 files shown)


src/transform/layout_inference.cc
```diff
@@ -1229,7 +1229,7 @@ class LayoutInferencer : public IRMutatorWithAnalyzer {
     SBlock block = Downcast<SBlock>(IRMutatorWithAnalyzer::VisitStmt_(op));
 
     for (auto buffer : block->alloc_buffers) {
-      if (buffer.scope() == "local.framgent") {
```
LeiWang1999, line 1232: would be better to use `IsFragmentBuffer`

LeiWang1999 (commented):
LGTM, but left a simple suggestion :)

src/transform/layout_inference.cc
penguin-wwy: done

PR #2696, opened 2026-07-18
[Fix] `T.__exp` must compute e**x, not 2**x (docstring + CuTeDSL codegen)

Hughshine:
`T.__exp` is the fast **exponential** (`e**x`): the CUDA backend lowers it to `__expf` (base-e), and the fast-math test asserts exactly that (`testing/python/fastmath/test_mathops_fastmath.py` checks `torch.exp`). A stale `2**x` was followed in two places; this corrects both so the whole `__exp` path is base-e:

- **Docstring** said `Calculate 2**x` in both copies (`language/math_intrinsics.py`, `language/fastmath.py`) → `e**x`.
- **CuTeDSL codegen** lowered `tl.__exp` to `tl.exp2` (base-2) at `src/cuda/codegen/codegen_cutedsl.cc`, diverging from the CUDA backend and the frontend contract for the same op — identical source computes `e**x` on CUDA and `2**x` on CuTeDSL. Mapped to `tl.exp` (base-e), matching every sibling (`__log`→`tl.log`, `__log2`→`tl.log2`, `__exp10`→`tl.exp10`, …), which already map to their matching base. The base-2 exponential remains available as the separate `exp2` op.

The CuTeDSL fix is confirmed by source (frontend contract + CUDA backend + the base-e value test + cutlass `exp2`=base-2 semantics); I have not run the CuTeDSL backend on silicon to observe the numeric change (requires `nvidia-cutlass-dsl`). Reach is small — CuTeDSL is an opt-in backend and no shipped kernel calls `__exp` today — so this is primarily a correctness cleanup of a latent divergence.

### DIFF
--- a/src/cuda/codegen/codegen_cutedsl.cc
+++ b/src/cuda/codegen/codegen_cutedsl.cc
@@ -1182,7 +1182,7 @@ void CodeGenTileLangCuTeDSL::VisitExpr_(const CallNode *op,
     os << "tl.increase_descriptor_offset(" << descriptor << ", " << offset
        << ")";
   } else if (op->op.same_as(tl::__exp())) {
-    os << "tl.exp2(" << PrintExpr_(op->args[0]) << ", fastmath=True)";
+    os << "tl.exp(" << PrintExpr_(op->args[0]) << ", fastmath=True)";
   } else if (op->op.same_as(tl::__exp10())) {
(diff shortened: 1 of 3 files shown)


Hughshine:
@coderabbitai review

LeiWang1999 (approved):

PR #2716, opened 2026-07-20
[BugFix] Prevent partial 1-D TMA stores from bypassing bounds checks

### LINKED ISSUE #2529: [BUG][Fuzzer][ice-on-valid-code] A partial-tile 1-D shared→global `T.copy` crashes the compiler with `IndexError` in `InferBulkLayout` instead of lowering or rejecting cleanly
### Required prerequisites

0.1.11

### System information

- TileLang `0.1.11`; source pinned to tag `v0.1.11` (commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`), byte-identical to the installed wheel.
- Python 3.12.3, linux; torch 2.12.1+cu130, CUDA 13.0.
- GPU: **NVIDIA H100 80GB HBM3** (compute capability 9.0, sm_90). Hopper-specific: the 1-D TMA bulk-store path (`src/cuda/op/copy.cc`, `src/cuda/op/copy_analysis.cc`) is only selected on sm_90.

### Problem description

A 1-D `T.copy` that stores a shared buffer back to a contiguous global slice crashes the compiler with `IndexError: Index -1 out of bounds 1` during layout inference whenever the copy is a **partial tile** — i.e. the tensor length is not a multiple of the block, so the last block's slice does not cover the buffer. The crash is independent of dtype and byte-alignment: it fires even for a 16-byte-aligned block (`block=256` fp16 → 512-byte chunk). The identical program with the TMA path disabled (`pass_configs={"tl.disable_tma_lower": True}`) compiles and runs bit-exact, so the program is well-formed; only the 1-D-bulk-store lowering falls over.

The culprit is the `shared→global` store; the `global→shared` load in the repro is not required — a bare shared→global store on a partial tile crashes the same way. Root cause in one line: `Copy::InferBulkLayout` reads `shared_tensor->shape[dim-2]`, which is out of bounds for a genuinely 1-D buffer (`dim==1` → `shape[-1]`), on the code path a partial-tile store falls into (details below).

<details>
<summary>Trigger boundary (H100, fp16, N=1000) — partial tile crashes at any alignment; full tile is fine (TESTED)</summary>

| `block` | `N % block` | chunk bytes | `% 16` | selected inst | result |
|---------|-------------|-------------|--------|---------------|--------|
| 256 | ≠ 0 | 512 | 0 | `kBulkStore` (multi-dim) | **COMPILE CRASH** `IndexError: Index -1 out of bounds 1` |
| 248 | ≠ 0 | 496 | 0 | `kBulkStore` (multi-dim) | **COMPILE CRASH** `IndexError: Index -1 out of bounds 1` |
| 200 | 0 | 400 | 0 | `tma_store` (1-D) | OK, `maxdiff = 0.0` |

`block=256` gives a 512-byte (16-aligned) chunk yet still crashes, which isolates the trigger to the **partial tile** (`N % block != 0`), not to any size/alignment condition. Control — same kernel with `tl.disable_tma_lower`: the partial-tile config compiles and runs bit-exact (`maxdiff = 0.0`). A full-tile TMA copy (`N=1000`, `block=200`) selects the 1-D `tma_store` and is bit-exact, so the 1-D store path works on a full tile.
</details>

<details>
<summary>Root cause — <code>InferBulkLayout</code> confuses copy-range rank with buffer-shape rank (pinned to <code>cd37ed5f</code>)</summary>

`Copy::InferBulkLayout` ([`copy.cc#L594`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L594)) has two early-outs for the 1-D-TMA case, both keyed off the *chosen instruction* / copy range: [`copy.cc#L607`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L607) (`is_tma_1d && shared_range.size() == 1`) and [`copy.cc#L628`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L628) (`if (is_tma_1d) return;`). But the store swizzle code at [`copy.cc#L638-L639`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L638-L639) indexes the *buffer shape*:

```cpp
int dim = shared_tensor->shape.size();                              // L638
const int64_t mat_stride = *as_const_int(shared_tensor->shape[dim - 2]);  // L639 -> shape[-1] when dim==1
```

When the selected instruction is `kBulkStore` (so `is_tma_1d` is false) *and* the shared buffer is genuinely 1-D, neither guard fires and `shape[dim-2] = shape[-1]` is out of bounds. There is no shape-rank guard before this access; the load branch at [`copy.cc#L655`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L655) just builds a linear layout and is fine.

A genuinely-1-D store reaches `kBulkStore` (instead of `kBulkStore1D`) because a partial tile sets `buffer_oob`, which clears `layout_dependent_tma_available = ... && !ctx.buffer_oob` ([`copy_analysis.cc#L568`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L568)), so `can_bulk_store_1d` stays false. `SelectTmaInst` then skips the 1-D store arm ([`copy_analysis.cc#L510`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L510)) and falls through to the multi-dim arm, returning `kBulkStore` on a 1-D buffer ([`copy_analysis.cc#L517`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy_analysis.cc#L517)).
</details>

### Reproducible example code

Run on an sm_90 (Hopper) GPU so the TMA path is selected. The compile crash needs no launch.

```python
import tilelang
import tilelang.language as T
import torch, traceback

dt = "float16"

def make(N, block, disable_tma=False):
    @T.prim_func
    def main(A: T.Tensor((N,), dt), B: T.Tensor((N,), dt)):
        with T.Kernel(T.ceildiv(N, block), threads=128) as bx:
            As = T.alloc_shared((block,), dt)              # genuinely 1-D shared buffer
            T.copy(A[bx * block:(bx + 1) * block], As)     # global -> shared
            T.copy(As, B[bx * block:(bx + 1) * block])     # shared -> global : the 1-D bulk store
    cfg = {"tl.disable_tma_lower": True} if disable_tma else None
    return tilelang.compile(main, out_idx=[1], target="cuda", pass_configs=cfg)

# DEFECT: partial tile (N % block != 0). block=256 -> 512-byte (16-aligned) chunk, still crashes.
try:
    make(1000, 256)
    print("compiled (unexpected)")
except Exception:
    traceback.print_exc()                                  # -> IndexError: Index -1 out of bounds 1

# CONTROL: identical kernel, TMA path disabled -> compiles and runs bit-exact.
kc = make(1000, 256, disable_tma=True)
A = (torch.arange(1000, device="cuda") % 100).to(torch.float16)
out = kc(A); torch.cuda.synchronize()
print("control tma:", "tma_store" in kc.get_kernel_source(),
      "maxdiff:", (out.float() - A.float()).abs().max().item())   # -> tma False maxdiff 0.0
```

Captured on the H100 (sm_90):

```
IndexError: Index -1 out of bounds 1
control tma: False maxdiff: 0.0
```

### Traceback

```pytb
File ".../tilelang/cuda/pipeline.py", line 113, in CUDAPassPipelineBodyPrologue
  ...
  tvm::tl::CopyNode::InferLayout(...)
  tvm::tl::cuda::Copy::InferLayout(...)
  tvm::tl::cuda::Copy::InferBulkLayout(...)
  File ".../tvm/ffi/container/seq_base.h", line 86, in tvm::ffi::SeqBaseObj::operator[](int64_t)
IndexError: Index -1 out of bounds 1                        # <- shape[dim-2] on a 1-D buffer
```

### Expected behavior

A 1-D `T.copy` between a 1-D shared buffer and a contiguous global slice should either lower to a correct instruction — falling back to a normal copy when the 1-D TMA preconditions (here, a full tile) are not met — or be rejected at compile time with a clean diagnostic. It should not throw an internal `IndexError` during layout inference. Grounded in TileLang's own behavior: the disable-TMA lowering of the identical program compiles and runs **bit-exact**, and the full-tile TMA case is bit-exact too, so TileLang intends 1-D copies of this shape to lower correctly — the partial tile is what it mishandles.

### Additional context

**Suggested fix (proposed, not built).** Guard the swizzle path by buffer rank in `Copy::InferBulkLayout`. Before computing `mat_stride = shared_tensor->shape[dim - 2]` ([`copy.cc#L638-L639`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L638-L639)), add a `shared_tensor->shape.size() < 2` early-out that falls through to the linear layout the load branch already uses at [`copy.cc#L655`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/copy.cc#L655) (`result_map.Set(shared_tensor, ComputeLinearLayout(shared_tensor))`) — a 1-D shared buffer cannot be swizzled, so a linear layout is the correct result. (One could also address the mis-routing itself so a genuinely-1-D partial-tile store does not reach the multi-dim descriptor arm.)

**Provenance.** The `shape[dim-2]` store-swizzle logic (with no rank guard) was introduced in [#1509](https://github.com/tile-ai/tilelang/pull/1509) (`[Refactor] Support auto swizzling for tma store …`, merged 2025-12-24) and later relocated into the per-backend `src/cuda/op/` files by the NFC refactors [#2138](https://github.com/tile-ai/tilelang/pull/2138) (`[Refactor][Backend] Split tl.copy lowering by backend`) and [#2297](https://github.com/tile-ai/tilelang/pull/2297) (`Move backend-specific transforms into separate namespaces`) — those moved the code, they did not add the missing guard. (Traced via `git log -S 'shared_tensor->shape[dim - 2]'` / `gh pr view` on the clone at `cd37ed5f`.)

**Distinct from existing reports.** I searched the open and closed tracker (including `misaligned`, `1d tma`, `InferBulkLayout`, `tma_store`) and found no report of this compile crash. The nearest neighbor, #2525, is in the same 1-D-bulk-store subsystem (same file, same `CheckBulkStore1D`) but is a different defect: a *full-tile* shared→global store whose byte size isn't a multiple of 16 is admitted to the 1-D-TMA arm and *silently drops* the tail. That is a missing size check on the accepted 1-D path; this is a missing shape-rank guard on the *multi-dim* path that a partial tile falls back to — orthogonal triggers (partial tile vs unaligned size) and orthogonal symptoms (compile crash vs silent wrong result). (A related runtime `misaligned address` crash also exists when a partial-tile store lands on a nonzero, non-16B-aligned base offset, but that only arises together with the non-16B size #2525 already covers, so it is not filed separately here.)

UnsettingGalaxy:
## Motivation

Fixes #2529.

A partial-tile 1-D shared-to-global `T.copy` could fall through to the descriptor-based multidimensional `kBulkStore` path.

The multidimensional BulkStore layout inference assumes that the shared tensor has at least two dimensions and accesses `shape[dim - 2]`. When the shared source is rank 1, this results in an internal `IndexError` during compilation.

## Changes

* Only consider the multidimensional `kBulkStore` path when the shared source buffer has rank 2 or higher.
* Add a regression test based on the original issue reproducer.
* Verify the output against the original disable-TMA control case.

## Testing

Tested on an NVIDIA H100 PCIe with compute capability 9.0.

```bash
cmake --build build -j
python3 -m pytest -q testing/python/issue/test_tilelang_issue_2529.py
```

## Summary

- Fixes a compiler crash in partial-tile 1-D shared-to-global `T.copy` operations by preventing `Copy::InferBulkLayout` from taking the multidimensional `kBulkStore` descriptor-based path for rank-1 shared source buffers (avoids `IndexError` in `InferBulkLayout` / `InferBulkLayout` layout inference).
- Ensures rank-1 partial-tile copies use the existing linear/fallback layout path instead of the swizzled bulk-store path.
- Updates copy analysis to derive TMA/bulk-store availability from `arith::Analyzer`-proven in-bounds regions (adds `CanProveRegionInBounds` / `CanProveCopyInBounds`) and removes the previous `buffer_oob` plumbing by adjusting `InferLayout`/`SelectInst` argument lists across CUDA/CPU/Metal/ROCm/WebGPU lowering paths.
- Adds a CUDA regression test for issue `#2529` covering partial tiles (including aligned cases like `fp16 block=256`), verifying both lowered-source selection behavior and runtime correctness against the TMA-disabled control.
- Verified on NVIDIA H100 PCIe (compute capability 9.0).

## C++ style / lint notes

- Does not modify rules documented in `docs/developer_guide/cpp_style.md` or any C++ style/lint tooling configuration.
- Touches C++ operator/layout inference and context struct internals (e.g., removes `buffer_oob` from layout inference/copy analysis plumbing), but does not introduce any change intended to address TLCPP003/TLCPP004-type API style warnings.
- The “C++ API Style Audit (warning only)” CI step remains advisory; no correctness/build impact is indicated by these changes beyond the intended compiler/crash fix.


LeiWang1999:
@regression-perf

PR #2718, opened 2026-07-20
[Transform] Replace CPU fallback thread placeholder with a constant-zero logical thread index

### LINKED ISSUE #2226: Track CPU fallback thread placeholder cleanup
## Context

CPU `c` lowering currently uses a synthetic fallback thread variable (`v_thread`) in `LayoutInference` and `LowerTileOp` so CPU fragment/tile lowering can reuse thread-oriented helper logic even when there is no real `threadIdx.x`.

PR #2203 needs a late canonicalization step after `LowerTileOp` to rewrite that synthetic CPU fallback to `0`, otherwise it can survive into later passes and be treated as a free variable / kernel ABI parameter.

## Follow-up

Replace the ad hoc fallback placeholder with a cleaner representation so CPU lowering does not materialize a naked thread-like `Var` that can escape. Possible directions:

- represent the CPU fallback thread index as constant `0` once helper code has consumed it;
- share an explicit synthetic fallback object/metadata between `LayoutInference` and `LowerTileOp` instead of relying on a name hint;
- update helper APIs to accept a `PrimExpr` thread index where CPU can pass `0` and GPU can pass the real thread var.

## Goal

After this is addressed, `LowerTileOp` should not need a compatibility cleanup for the CPU fallback thread placeholder, and later passes such as `SplitHostDevice` / `MakePackedAPI` should never see it.

penguin-wwy:
Fix #2226 

## Summary

- Replaced the synthetic CPU `v_thread` with a logical thread index expression, using constant `0` when no real thread binding exists.
- Propagated `thread_index` through loop partitioning, predicates, layout inference, and backend/GEMM lowerings.
- Removed CPU placeholder cleanup and normalized thread indices relative to thread bounds.
- Added regression tests for CPU lowering, clean host/device ABIs, real GPU bindings, and legitimate `v_thread` parameters.

## C++ style / lint notes

- No C++ style documentation or CI/lint files are changed.
- The C++ API Style Audit remains warning-only; any advisory findings should be considered separately from correctness, build, and test issues.

### DIFF
--- a/src/backend/common/op/atomic_reduce.h
+++ b/src/backend/common/op/atomic_reduce.h
@@ -183,9 +183,9 @@ struct AtomicReduce {
                           level);
     }
     auto loop_layout = par_op->GetLoopLayout();
-    return LowerParallelLoop(fused_loop, loop_layout, lower_args.thread_var,
+    return LowerParallelLoop(fused_loop, loop_layout, lower_args.thread_index,
                              analyzer, lower_args.layout_map,
-                             par_op->GetPredicate(lower_args.thread_var),
+                             par_op->GetPredicate(lower_args.thread_index),
                              /*parallel_loop=*/true, /*should_vectorize=*/true,
                              par_op->LoopLayoutRequiresPaddingGuard());
   }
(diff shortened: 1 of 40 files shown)


LeiWang1999:
@regression-perf

LeiWang1999 (approved):

PR #2719, opened 2026-07-21
[BugFix] Reject contracting shared-buffer layouts in T.annotate_layout

### LINKED ISSUE #2632: [BUG][Fuzzer][ice-on-invalid-code] Non-bijective `T.annotate_layout` aborts with an unrelated `2 vs. 1` ICHECK instead of rejecting
### Required prerequisites

0.1.11 (tag `cd37ed5f`)

### System information

The crash is in host-side lowering, before any target codegen, so it is arch-independent; it reproduces identically on:

- L40S (sm_89), pip/dev wheel, byte-identical to source `cd37ed5f`; Python 3.13; torch 2.11.
- H100 (sm_90), isolated `tl_venv` 0.1.11.

```python
import sys, tilelang, torch
print(sys.version, sys.platform)   # 3.13.x linux
print(tilelang.__version__)        # 0.1.11
print(torch.__version__)           # 2.11.x
```

### Problem description

`T.annotate_layout({shared_buf: layout})` accepts an arbitrary user forward-map. If the map is
**many-to-one** (non-injective — two distinct logical indices land in the same physical slot), the
compiler does not reject it. Instead `makeBufferWithLayout` mis-classifies the shrunken codomain as a
*replication factor*, rewrites the 1-D shared buffer into a 2-D `[replicate_extent, out]` buffer, and
then the still-1-D `BufferStore` (`s[i] = ...`) hits an unrelated internal check and aborts with
`Check failed: buffer->shape.size() == indices.size() (2 vs. 1)`.

The error names neither `annotate_layout` nor "layout is not a bijection" — it is an opaque
dimensionality ICHECK deep in the TVM `BufferStore` constructor. A user who writes a wrong layout map
gets no actionable diagnostic.

**Trigger boundary (the discriminator — proves it is the non-bijectivity, not codomain rank).**
`OutputShape()` values below are the observed C++ `layout->OutputShape()` (matches the Python
`Layout.get_output_shape()`):

| forward map | `OutputShape()` | injective? | outcome |
|---|---|---|---|
| `lambda i: i` (identity) | `[128]` | yes (bijection) | **compiles + runs exact** |
| `lambda i: i ^ 1` (pair swap) | `[128]` | yes (bijection) | **compiles + runs exact** |
| `lambda i: i * 2` | `[255]` (grows) | yes (injective) | **compiles + runs exact** |
| `lambda i: i + 10` | `[138]` (grows) | yes (injective) | **compiles + runs exact** |
| `lambda i: i % 64` | `[64]` (shrinks) | **no (2-to-1)** | **CRASH `2 vs. 1`** |
| `lambda i: i * 0` | `[1]` (shrinks) | **no (N-to-1)** | **CRASH `2 vs. 1`** |

Injective maps whose output shape is *larger* than the input (`i*2` → out `[255]`, `i+10` → out `[138]`)
compile and run correctly (exact round-trip verified on L40S), so the crash is **not** an
"output shape ≠ input shape" artifact — it is specifically the *contracting* (many-to-one) case that the
compiler cannot represent and does not reject.

**Root cause.** The tile-op lowering pass mis-analyzes a contracting (many-to-one) shared-buffer
layout as a *replication* layout and rewrites the buffer to an extra dimension, so the still-1-D store
against it fails an internal rank ICHECK — an opaque crash instead of a clean layout reject.

Mechanism, in [`src/transform/lower_tile_op.cc`, `makeBufferWithLayout` (L63–83)](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/transform/lower_tile_op.cc#L63-L83):

```cpp
Array<PrimExpr> layout_shape = layout->OutputShape();   // [64] for i%64          (L63)
Array<PrimExpr> output_shape = layout_shape;
if (IsSharedBuffer(buffer)) {
  ...
  replicate_extent = buffer_extent / layout_extent;     // 128 / 64 = 2           (L80)
  if (replicate_extent > 1) {                            //                        (L81)
    output_shape.insert(output_shape.begin(), replicate_extent);  // -> [2, 64]  (2-D!) (L82)
  }
}
return Buffer(new_var, buffer->dtype, output_shape, ...);
```

The `replicate_extent = buffer_extent / layout_extent > 1` branch assumes that a codomain smaller than
the domain means the layout is a **replicated** layout (the same data physically stored `replicate_extent`
times — valid for thread-replicated *fragments*). For a **shared** buffer a contracting map is not a
replication; it is a genuine non-bijection where distinct elements alias one slot. The pass silently
reinterprets it as a 2-D `[2, 64]` buffer, but the BufferStore index that reads/writes `s[i]` stays 1-D,
so `tvm::tirx::BufferStore::BufferStore` (`3rdparty/tvm/src/tirx/ir/stmt.cc:349`) fails
`buffer->shape.size() == indices.size()` (2 vs 1).

There is an injectivity checker in the codebase ([`FragmentNode::DetectInjective`,
`src/layout/layout.cc:859`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/layout/layout.cc#L859), with `IterMapLevel::Bijective`), but it is only invoked on the
*fragment* thread-mapping path — the shared-buffer `annotate_layout` path never runs it.

### Reproducible example code

Single-stage, no pipelining (default pipeline):

```python
N = 128

@T.prim_func
def k(A: T.Tensor((N,), "int32"), Out: T.Tensor((N,), "int32")):
    with T.Kernel(1, threads=N) as bx:
        s = T.alloc_shared((N,), "int32")
        T.annotate_layout({s: Layout((N,), lambda i: i % 64)})   # 2-to-1: MALFORMED (not a bijection)
        for i in T.Parallel(N):
            s[i] = A[i]
        for i in T.Parallel(N):
            Out[i] = s[i]

tilelang.compile(k, out_idx=[1], target="cuda")   # -> InternalError "2 vs. 1"
```

### Traceback

```pytb
  File ".../tilelang/transform/__init__.py", line ..., in LowerTileOpPass::VisitStmt_(BufferStoreNode)
  File "/project/3rdparty/tvm/src/tirx/ir/stmt.cc", line 349, in tvm::tirx::BufferStore::BufferStore(...)
tvm.error.InternalError: Check failed: buffer->shape.size() == indices.size() (2 vs. 1) : Buffer s is 2-dimensional, cannot be indexed with the 1-dimensional indices provided.
```

(The full C++ stack passes through `tvm::tl::LowerTileOpPass::VisitStmt_(BufferStoreNode)` →
`tvm::tirx::BufferStore::BufferStore` at `stmt.cc:349`.)

### Expected behavior

A clean frontend reject, e.g.
`"annotate_layout on shared buffer 's': forward map must be injective (bijective onto its codomain); it
maps 128 input elements into 64 slots"`, rather than an opaque internal `2 vs. 1` dimensionality ICHECK
that names neither `annotate_layout` nor the non-bijectivity.

### Additional context

**Fix direction.** In `makeBufferWithLayout`, when `IsSharedBuffer(buffer)` and
`buffer_extent % layout_extent != 0` **or** the layout is not provably injective over its input domain,
emit a clear frontend error. The cheap guard `buffer_extent < layout_extent` is insufficient (`i%64` has
`layout_extent==64 < 128`); use the existing `Layout::InverseWithLevel` / an injectivity probe (as
`DetectInjective` does for fragments) before deciding the `replicate_extent` interpretation.

**Disposition.** `ice-on-invalid-code` (severity ④). The input is *malformed* — a many-to-one
`annotate_layout` map is not a valid (bijective) layout — so the compiler *should* cleanly reject it at
the frontend with an actionable message; instead it aborts with an unrelated internal `2 vs. 1` ICHECK
(an opaque compiler crash on invalid input). This is not `ice-on-valid-code`: the input is not a
program the compiler is expected to accept.

**Dedup.**
- **DISTINCT from #2606** (`make_cutlass_metadata_layout` non-bijective → *silent* miscompute in sparse
  GEMM): different entry point (an internal sm8x helper vs the generic public `T.annotate_layout`),
  different mechanism (`gemm_sp.py` row-permutation vs `lower_tile_op.cc` `replicate_extent`), different
  symptom (silent wrong result vs opaque crash).
- **DISTINCT from #2309** (LayoutInference crash for sibling `T.Pipelined` loops sharing `alloc_shared`
  with different `num_stages`): that is a **pipelined** (`num_stages>=2`) stage-versioning rank clash at
  `layout.cc` `CanProveEqual` (`layout.cc:678`; InputShape `(4096,)` vs `(3,128,32)`). This crash is
  **single-stage, no pipelining**, at `stmt.cc:349` via `makeBufferWithLayout` `replicate_extent`, on a
  plainly-contracting map.
- **DISTINCT from #2609** (`T.view` cross-dtype reshape of a swizzled buffer, silent): different op
  (`T.view`/`Reshape` vs `annotate_layout`), silent vs crash.
- Not the fragment `DetectInjective` path (`layout.cc:859`).

**Provenance.** The `replicate_extent` rewrite was introduced by PR #447 (`fce16b00`, "Separate warp
specialize rewriter and tma barrier injector pass", 2025-05-03), which replaced the prior
`return Buffer(..., layout->OutputShape(), ...)` (verified in `fce16b00^`) with the
`buffer_extent / layout_extent` replication branch that produces the 2-D buffer. Before that commit a
contracting shared layout did not trip the `2 vs. 1` store ICHECK, so this crash is a **regression**
introduced at #447. Two later commits only relocated/annotated the same code and are not the origin:
#1632 (`6e439530`) added the constant-integer `ICHECK` inside the loop; #1909 (`ded6a992`) swapped the
`storage_scope == "shared"` string test for `IsSharedBuffer(buffer)`. (Provenance git-verified in the
`cd37ed5f` clone via `git log -L` / `git blame` / `git show fce16b00^`; PR numbers are the ones in the
commit subjects.)

**Cross-arch / version.**
- L40S sm_89 (wheel 0.1.11, byte-identical to source `cd37ed5f`): CRASH.
- H100 sm_90 (tl_venv 0.1.11): identical CRASH — arch-independent (host-side lowering, before any target codegen).
- Live on the pinned dev wheel; `makeBufferWithLayout` unchanged on `main`.

Chennesxu:
## Summary

`T.annotate_layout` did not reject a contracting (many-to-one) forward map on a shared buffer. When the layout output extent was smaller than its input extent, `makeBufferWithLayout` misread the smaller output extent as a replication factor, rewrote the 1-D buffer into a 2-D buffer, and eventually aborted in the TVM `BufferStore` constructor with an unrelated dimensionality ICHECK (`2 vs. 1`) that named neither `annotate_layout` nor the non-injective map.

## Fix

Add a pigeonhole check in `makeBufferWithLayout` before the replication branch. For a shared buffer whose layout has static shapes, if the layout output extent is smaller than its input extent, the map cannot be injective, so raise a `ValueError` that names the buffer and reports both extents. Identity, permutation, and output-growing maps are unchanged.

The check intentionally targets contracting maps; general injectivity analysis for maps with a larger output bounding box, such as `(i % 64) * 4`, is outside the scope of this change.

## Tests

Added `testing/python/issue/test_tilelang_issue_2632.py`:

* Rejects the contracting maps from the issue (`i % 64` and `i * 0`).
* Confirms that identity, `i ^ 1`, `i * 2`, and `i + 10` still compile and round-trip exactly.

Also ran the related `lower_tile_op` and `layout_inference` transform suites, as well as `pre-commit`.

Fixes #2632

## Summary

- Rejects non-injective shared-buffer `T.annotate_layout` mappings with clear `ValueError` diagnostics instead of triggering lowering ICHECKs.
- Adds symbolic and exact static-domain injectivity checks for layouts and fragments, including constant-expression evaluation support.
- Validates layouts after reshaping while preserving valid identity, permutation, padded, and output-growing mappings.
- Adds CUDA regression tests for rejecting `i % 64`/`i * 0` and for successful injective mappings.

## C++ style / lint notes

- The PR changes C++ layout APIs and implementation but does not modify `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) remains relevant; no blocking style or correctness issues are reported.

### DIFF
--- a/src/transform/lower_tile_op.cc
+++ b/src/transform/lower_tile_op.cc
@@ -78,6 +78,30 @@ static Buffer makeBufferWithLayout(const Buffer &buffer, const Layout &layout,
                     << layout_shape[i];
       layout_extent *= shape->value;
     }
+    // A forward map whose codomain is smaller than its domain is provably
+    // non-injective (pigeonhole): distinct logical elements would alias the
+    // same physical slot. Reject it here; otherwise the replication branch
+    // below would misread the shrunken codomain as a replicate extent and
+    // rewrite the buffer to a higher rank, crashing later on an unrelated
+    // BufferStore rank ICHECK.
+    int64_t layout_input_extent = 1;
+    bool layout_input_is_static = true;
+    for (const auto &dim : layout->InputShape()) {
+      const auto *imm = dim.as<IntImmNode>();
+      if (!imm) {
+        layout_input_is_static = false;
+        break;
+      }
+      layout_input_extent *= imm->value;
+    }
+    if (layout_input_is_static && layout_extent < layout_input_extent) {
+      TVM_FFI_THROW(ValueError)
+          << "Invalid layout for shared buffer `" << buffer->name
+          << "`: the forward map must be injective, but it maps "
+          << layout_input_extent << " logical elements into only "
+          << layout_extent
+          << " physical slots. Check the layout passed to T.annotate_layout.";
+    }
     replicate_extent = buffer_extent / layout_extent;
     if (replicate_extent > 1) {
       output_shape.insert(output_shape.begin(), replicate_extent);
--- a/src/transform/lower_tile_op.cc
+++ b/src/transform/lower_tile_op.cc
@@ -78,30 +78,6 @@ static Buffer makeBufferWithLayout(const Buffer &buffer, const Layout &layout,
                     << layout_shape[i];
       layout_extent *= shape->value;
     }
-    // A forward map whose codomain is smaller than its domain is provably
-    // non-injective (pigeonhole): distinct logical elements would alias the
-    // same physical slot. Reject it here; otherwise the replication branch
-    // below would misread the shrunken codomain as a replicate extent and
-    // rewrite the buffer to a higher rank, crashing later on an unrelated
-    // BufferStore rank ICHECK.
-    int64_t layout_input_extent = 1;
-    bool layout_input_is_static = true;
-    for (const auto &dim : layout->InputShape()) {
-      const auto *imm = dim.as<IntImmNode>();
-      if (!imm) {
-        layout_input_is_static = false;
-        break;
-      }
-      layout_input_extent *= imm->value;
-    }
-    if (layout_input_is_static && layout_extent < layout_input_extent) {
-      TVM_FFI_THROW(ValueError)
-          << "Invalid layout for shared buffer `" << buffer->name
-          << "`: the forward map must be injective, but it maps "
-          << layout_input_extent << " logical elements into only "
-          << layout_extent
-          << " physical slots. Check the layout passed to T.annotate_layout.";
-    }
     replicate_extent = buffer_extent / layout_extent;
     if (replicate_extent > 1) {
       output_shape.insert(output_shape.begin(), replicate_extent);
--- a/src/layout/cute_layout.cc
+++ b/src/layout/cute_layout.cc
@@ -5,6 +5,7 @@
 
 #include "cute_layout.h"
 #include "layout.h"
+#include "utils.h"
 
 #include "support/check.h"
 #include <algorithm>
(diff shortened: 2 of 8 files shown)


src/transform/lower_tile_op.cc
```diff
@@ -78,6 +78,30 @@ static Buffer makeBufferWithLayout(const Buffer &buffer, const Layout &layout,
...
+    bool layout_input_is_static = true;
+    for (const auto &dim : layout->InputShape()) {
+      const auto *imm = dim.as<IntImmNode>();
+      if (!imm) {
+        layout_input_is_static = false;
+        break;
+      }
+      layout_input_extent *= imm->value;
+    }
+    if (layout_input_is_static && layout_extent < layout_input_extent) {
```
LeiWang1999: **this does not actually validate injectivity.**

  `OutputShape()` describes the physical bounding box/footprint, not the cardinality of the forward map’s image. The pigeonhole check is only one-way: `layout_extent < layout_input_extent` proves that
  the map is non-injective, but `layout_extent >= layout_input_extent` does not prove that it is injective. Padding and holes make this distinction important. For example, `i + 10` is injective with
  output extent 138, while `(i % 64) * 4` is non-injective with output extent 253 but only 64 distinct destinations. The current patch accepts the latter and still allows shared-memory aliasing.

  Therefore, this fixes the specific contracting-layout rank ICHECK, but it does not enforce the injectivity invariant claimed by the diagnostic or fully address #2632.

  I think the proper fix is to add a real `LayoutNode::DetectInjective` API and call it when applying `T.annotate_layout` to a shared buffer. `FragmentNode` already has a similarly named implementation;
  the common forward-index check should live on `LayoutNode`, while `FragmentNode` can extend it with thread and replication coordinates.

  Note that the existing `FragmentNode::DetectInjective` uses `DetectIterMap(..., Bijective)`, which is stricter than injectivity and is not sufficient as-is: it may reject valid sparse/padded mappings
  such as `i * 2`, and it cannot normalize every supported expression such as `i ^ 1`. The base implementation therefore needs true “injective onto its image” semantics, potentially using iter-map
  analysis as a fast path with an exact fallback for static domains.

  Please also add `(i % 64) * 4` as a rejection regression while preserving the existing `i ^ 1`, `i * 2`, and `i + 10` controls.

LeiWang1999 (changes requested):
Thanks but I left a comment.

LeiWang1999 (approved):

LeiWang1999:
@regression-perf

LeiWang1999 (approved):

Issue #2603, opened 2026-07-10
[BUG][Fuzzer][wrong-code] Sparse WGMMA `T.gemm_sp` with `block_N ≡ 8 (mod 16)` silently miscomputes the whole output tile on sm_90 (and hangs under warp specialization)


Hughshine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.11

### System information

- Installed: pip (dev build reporting `0.1.11`)
- GPU: NVIDIA H100 (sm_90a)
- Python 3.12, `torch 2.12.1+cu130`, CUDA 13

### Problem description

On sm_90, a 2:4-structured-sparse WGMMA `T.gemm_sp` whose N tile is congruent to 8 mod 16 (`block_N ∈ {24, 40, 56, 72, …}`) **compiles clean, runs with no error or warning, and returns a silently wrong result** — about 96–97% of the output-tile elements are incorrect. There is no exception and no diagnostic, and the result is deterministic (bit-identical across two runs at the same seed, and reproduced at independent seeds), so this is a compile/layout miscompute, not a race.

The trigger is `block_N ≡ 8 (mod 16)` **with more than one N column-atom** on the sparse-WGMMA path with warp specialization disabled (`pass_configs={"tl.disable_warp_specialized": True}`). `block_N` a multiple of 16 is correct; a single `block_N = 8` (one `n=8` sparse atom, `num_inst_n = 1`) is also correct.

The same `block_N ≡ 8 (mod 16)` config additionally **hangs at launch** under the default (warp-specialized) pipeline — the kernel launches, the GPU pins at 100%, and it never returns (see the WS-on row below).

[...]

Issue #2593, opened 2026-07-10
[BUG][Fuzzer][wrong-code] WGMMA `T.gemm` with `block_N ≡ 8 (mod 16)` silently miscomputes the output tile on sm_90 (no error)


Hughshine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.11

### System information

- Installed: pip (dev build reporting `0.1.11`)
- GPU: NVIDIA H100 (sm_90)
- Python 3.12, `torch 2.12.1+cu130`, CUDA 13

```python
import sys, tilelang, torch
print(sys.version, sys.platform)   # 3.12 ... linux
print(tilelang.__version__)        # 0.1.11
print(torch.__version__)           # 2.12.1+cu130
```

### Problem description

On sm_90 (Hopper), a WGMMA `T.gemm` whose N tile is congruent to 8 mod 16 (`block_N ∈ {24, 40, 56, 72, …}`) **compiles clean, runs with no error or warning, and returns a silently wrong result** — roughly 97% of the output-tile elements are incorrect. There is no exception, no diagnostic, and the result is deterministic (bit-identical across runs with the same seed, so this is not a race — it is a compile/layout miscompute).

The trigger is `block_N ≡ 8 (mod 16)` **with more than one N column-atom** (i.e. `warp_cols > 1`), on the WGMMA path with warp specialization disabled (`pass_configs={"tl.disable_warp_specialized": True}`). N as a multiple of 16 is correct; a single `block_N = 8` (one n=8 atom, `warp_cols = 1`) is also correct.

[...]

Issue #1554, opened 2025-12-29
[Question] I got num_warps mismatch when compiling the kernel


iclementine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Questions

The error stack is 
```txt
  File "/usr/local/lib/python3.12/dist-packages/tilelang/jit/__init__.py", line 205, in wrapper
    kernel_result = compile(
                    ^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/jit/__init__.py", line 70, in compile
    return cached(
           ^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/cache/__init__.py", line 29, in cached
    return _kernel_cache_instance.cached(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/cache/kernel_cache.py", line 185, in cached
    kernel = JITKernel(
             ^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/jit/kernel.py", line 121, in __init__
    adapter = self._compile_and_create_adapter(func, out_idx)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/jit/kernel.py", line 219, in _compile_and_create_adapter
    artifact = tilelang.lower(
               ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/engine/lower.py", line 227, in lower
    mod = LowerAndLegalize(mod, target)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/engine/phase.py", line 106, in LowerAndLegalize
    mod = tilelang.transform.LayoutInference()(mod)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tilelang/3rdparty/tvm/python/tvm/ir/transform.py", line 167, in __call__
    return _ffi_transform_api.RunPass(self, mod)
[...]

PR #2724, opened 2026-07-21
[BugFix] Reject uncovered warp partitions in T.gemm instead of producing silently wrong results

### LINKED ISSUE #2537: [BUG][Fuzzer][wrong-code] `T.gemm` silently leaves output columns unwritten (wrong result) instead of computing or rejecting them when the warp count doesn't divide N
### Required prerequisites

0.1.11

### System information

NVIDIA H100 80GB HBM3 (sm_90), TileLang 0.1.11, CUDA 13. The `src/cuda/op/gemm.cc` warp-partition logic is unchanged on current `main`.

### Problem description

`T.gemm` on `M=16, N=104, threads=384` compiles, runs without error, and **silently leaves output columns 96–103 unwritten** — a `16×104` GEMM returns a `16×96` result with stale data in the rest. Deterministic.

`N=104` is a valid input (`T.gemm`'s only N constraint is `N % 8 == 0` at [gemm.cc:123-126](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/gemm.cc#L123-L126), and `104 % 8 == 0`). The compiler picks `n_warp = 12` warps along N, gives each `104 // 12 = 8` columns (floor), and never checks that `12 * 8 = 96` covers all 104 — so the last 8 columns get no warp. The same program at `N=96` (which 12 divides) is bit-exact; the only difference is whether the warp count divides N.

### Reproducible example code

```python
from tilelang.tileop.base import GemmWarpPolicy

def run(tag, M, N, K, threads, policy):
    @T.prim_func
    def g(A: T.Tensor((M, K), "float16"), B: T.Tensor((K, N), "float16"),
        with T.Kernel(1, 1, threads=threads):
            As = T.alloc_shared((M, K), "float16"); Bs = T.alloc_shared((K, N), "float16")
            Cl = T.alloc_fragment((M, N), "float32"); T.clear(Cl)
            T.copy(A, As); T.copy(B, Bs)
            T.gemm(As, Bs, Cl, policy=policy)
            T.copy(Cl, C)
    k = tilelang.compile(g, out_idx=[2])
    torch.manual_seed(0)
    A = torch.randn(M, K, device="cuda", dtype=torch.float16)
    B = torch.randn(K, N, device="cuda", dtype=torch.float16)
    diff = (k(A, B).float() - A.float() @ B.float()).abs()
    cols_wrong = int((diff.amax(0) > 1e-2).sum())
    print(f"[{tag}] correct={cols_wrong==0} cols_wrong={cols_wrong}/{N}")

run("N=104 (12 warps, 12 does not divide 104)", 16, 104, 32, 384, GemmWarpPolicy.Square)   # -> cols_wrong=8/104
run("N=96  (12 warps, 12 divides 96)          ", 16,  96, 32, 384, GemmWarpPolicy.Square)   # -> cols_wrong=0/96  (control)
```

Observed on H100: the `N=104` case reports `cols_wrong=8/104` (columns 96–103 wrong), the `N=96` control reports `cols_wrong=0/96`.

### Traceback

```pytb
No traceback — the kernel compiles and runs to completion; the result is silently wrong (trailing N columns unwritten) and deterministic.
```

### Expected behavior

The full `16×104` result should be computed — the shape is coverable (e.g. `threads=416` / 13 warps computes `N=104` correctly, since `13 * 8 = 104`), so the partition search can choose a covering warp count instead of `n_warp=12`. If for some reason it cannot, it must **reject at compile time** with a clear error. Either is fine; silently dropping columns is not. TileLang already errors rather than miscompiles on the neighboring `M % 16` / `N % 8` misalignment ([gemm.cc:123-126](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/gemm.cc#L123-L126)), and #2390 added a `block_K % micro_size_k == 0` assert to these backends "to prevent silent miscompilation" — this is the same guarantee on the N/warp axis.

### Additional context

**Root cause.** The warp-partition search (`ComputeDefaultWarpPartition` / `ComputeWgmmaWarpPartition`, `src/cuda/op/gemm.cc`) picks the N divisor by "each warp gets ≥1 tile" and checks the product `m_warp * n_warp == num_warps`, but never checks that the warps *cover* N (`n_warp * (N // n_warp) == N`). The emitter's `assert warp_col_tiles % 8 == 0` only checks the per-warp tile is legal, not that the tiles span N — and `104//12 = 8` happens to pass it.

<details>
<summary>Suggested fix + the compile-crash sibling faces</summary>

**Fix.** In the partition search, accept a candidate only if it covers the axis (`n_warp * (N // n_warp) == N`, symmetrically for M); if none does, raise an early front-end error naming `threads`/`block_N`. Equivalently, assert `n_warp * warp_col_tiles == N` before building the emitter.

**Same root, louder faces (a separate reject-hardening ask).** The same missing coverage/divisibility check also makes the search pick a non-factoring partition on other legal inputs, which then hits an internal `ICHECK(m_warp*n_warp==num_warps)` ([gemm.cc:179-181](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/gemm.cc#L179-L181)) or a `warp_col_tiles % 8` `AssertionError` — e.g. FullRow `M=48`/8 warps, Square `M=N=16`/7 warps, WGMMA FullRow `M=128`/12 warps. These fail loudly (an internal assert where a clean diagnostic belongs), unlike the silent drop above. The coverage fix resolves both.
</details>

**Provenance.** The FullRow floor-div + terminal `ICHECK` were introduced in [#517](https://github.com/tile-ai/tilelang/pull/517) (merged 2025-05-24); `ComputeWgmmaWarpPartition` repeated the pattern in the backend split [#2153](https://github.com/tile-ai/tilelang/pull/2153) (2026-05-06). The same product `Check failed` was reported in [#1554](https://github.com/tile-ai/tilelang/issues/1554) and closed as a question (user changed policy); no fix landed and it did not cover the silent-column-drop.

li-ruinan:
# Title

`[BugFix] Reject uncovered warp partitions in T.gemm instead of producing silently wrong results`

# Summary

Fixes #2537.

`T.gemm` with `M=16, N=104, threads=384` (12 warps) produces **silently wrong results**: output columns 96–103 are never written and hold stale values. The shape passes every declared front-end check (`M % 16 == 0`, `N % 8 == 0`), the kernel compiles and runs cleanly — only the numbers are wrong. The same root cause also surfaces as internal `ICHECK` failures and emitter `AssertionError`s on other shape/policy combinations (previously reported as a "question" in #1554 and closed without a fix).

This PR turns the warp-partition computation in `src/cuda/op/gemm.cc` into a real search over *valid* partitions, and raises a clear, actionable front-end error when no valid partition exists.

# Root Cause

`ComputeDefaultWarpPartition` / `ComputeWgmmaWarpPartition` (`src/cuda/op/gemm.cc`) accept a candidate `(m_warp, n_warp)` without checking that the partition **covers** the M/N extents. A fully valid partition must satisfy four predicates:

1. factoring: `m_warp * n_warp == num_warps`
2. M-coverage: `M % (m_warp * 16) == 0`
3. N-coverage: `N % (n_warp * 8) == 0`
4. (WGMMA only) warpgroup alignment: `m_warp % 4 == 0`

The old code checked different subsets of these per policy branch, so each branch had its own failure face:

|                | Default FullRow | Default FullCol | Default Square | WGMMA FullRow | WGMMA FullCol | WGMMA Square |
| -------------- | --------------- | --------------- | -------------- | ------------- | ------------- | ------------ |
| ① factoring    | ✗ → ICHECK      | ✗ → ICHECK      | ✓              | ✗ → ICHECK    | ✓             | ✓            |
| ② M-coverage   | ◎               | ✗ → silent      | ✗ → silent     | ✓             | ✗ → silent    | ✗ → silent   |
| ③ N-coverage   | ✗ → silent      | ◎               | ✗ → silent     | ✗ → assert    | ✗ → assert    | ✗ → assert   |

(✗ = predicate not checked; ◎ = satisfied by construction of that branch.)

Causal chain for the issue's repro (Square, N=104, 12 warps): the search picks `n_warp = 12`; `warp_col_tiles = 104 // 12 = 8` floors, so the 12 warps cover only 96 columns. In `make_mma_store_layout`, elements with `j >= 96` map to `thread_id >= blockDim.x` — a thread that does not exist — so their clear/mma/store all silently vanish. Nothing ever writes columns 96–103.

The uncovered-partition logic dates back to #517 and was carried into `src/cuda/op/gemm.cc` unchanged by the backend split in #2153.

# Changes

**`src/cuda/op/gemm.cc`** — both partition functions now share the same structure:

- A single `is_valid(m, n)` predicate encodes all coverage requirements (the WGMMA variant adds `m % 4 == 0`).
- Each policy keeps its original preference, applied over the **valid** candidate set only:
  - `FullRow`: largest valid `m_warp` (was: hard fallback `m = M/16`, which crashes an `ICHECK` when it doesn't divide `num_warps`);
  - `FullCol`: largest valid `n_warp` (symmetric);
  - `Square`: the balance-scoring formula is **unchanged**; only the candidate set is restricted. Any shape that previously received a covering partition receives exactly the same partition after this PR.
- If the valid set is empty, `LOG(FATAL)` with an actionable message naming M, N, `num_warps`, the policy, and the per-warp granularity constraints — instead of wrong results or an internal-invariant crash.

Behavior changes are exactly these three classes, and nothing else:

1. **Silent wrong-code → clean error.** E.g. the issue's repro, and its M-axis siblings (`M=288`, 17 warps, rows 272–287 stale). Strictly a tightening; a "result" that was wrong has no compatibility claim.
2. **Internal ICHECK/assert crash → either a clean error, or a working kernel.** Shapes like FullCol `M=192, N=104`/12 warps (now `(12,1)`) or WGMMA FullRow `M=128, N=192`/12 warps (now `(4,3)`) previously passed every *declared* operator check and then crashed on internal invariants; they now compile and produce correct results. Since these code paths were never generated before, each such case carries a numeric ground-truth regression test, not just a "does not crash" check.
3. **Previously working shapes: no change.** Verified by diffing the selected `(m_warp, n_warp)` over a grid of covering shape/policy/warp-count combinations on both paths before and after the fix — zero diff (mechanically guaranteed for Square by the untouched scoring formula).

**`testing/python/issue/test_tilelang_issue_2537.py`** (new) — 14 regression tests: 4 uncovered-partition rejections + 4 numeric ground-truth cases + 1 no-partition clear-error case on the MMA path; 3 rejections + 2 numeric ground-truth cases on the WGMMA path (gated `requires_cuda_compute_version_eq(9, 0)`). The two default-search tests whose shapes would silently reroute to the WGMMA search on sm_90 (`M >= 64`, warps divisible by 4) pin their backend via `tl.disable_wgmma`, so they exercise the intended search on every arch.

## Notes for reviewers / related issues

- **WGMMA descriptor expressibility is a separate, pre-existing constraint layer** (related territory: #2593 / #2603). For a 16-bit MN-major B whose row width fits no canonical GMMA swizzle atom (e.g. N=104 → 208 bytes), `compute_gmma_descriptor` raises `Not a canonical GMMA_MN layout` at lowering. This depends only on the whole B tile's layout, not on the warp partition: at `threads=128` the old and the fixed search select the identical partition `(4, 1)` and hit the identical assert (verified), so this PR neither introduces nor exposes it — such shapes never compiled on the WGMMA route. The MMA-intent regression tests pin their backend via `tl.disable_wgmma` so they keep testing the default search on Hopper CI; the WGMMA numeric tests use canonical-width N. A follow-up note on #2593 (same shape family) will record the current-main status and propose surfacing this as an actionable front-end error at the WGMMA lowering entry (which knows N, dtype and majorness) — encoding it into the partition search instead would mis-attribute a shape constraint to `threads` and require plumbing operand majorness/dtype into `ComputeWarpPartition`.
- `src/cuda/op/gemm_sp.cc` contains independent copies of both partition functions with the same missing coverage checks; not touched here to keep the diff focused (and to avoid conflicting with #2603's candidate fixes). Happy to follow up.

# Tested

- Issue repro (`M=16, N=104, threads=384`, Square): wrong columns 96–103 before; clear front-end error after. Counterfactual `threads=416` (13 warps, covering) computes correctly in both.
- Full defect-face matrix (every ✗ cell above, both paths, all three policies) verified before/after: every silent face and every internal-crash face now yields either the clean error or a numerically correct kernel.
- Partition-selection diff over a grid of previously-working combinations on both MMA and WGMMA paths: identical `(m_warp, n_warp)` before and after.
- New regression tests pass on sm_89 (WGMMA cases skip; they are compile-verified via an sm_90 target and need Hopper CI for the numeric assertions).

## Summary

- Fixed `T.gemm` warp-partition selection for MMA and WGMMA paths.
- Candidates now validate warp factoring, complete M/N coverage, and WGMMA warpgroup alignment.
- Added clear front-end errors when no valid partition exists, preventing silent miscompilation and internal assertion failures.
- Preserved existing policy preferences among valid partitions.
- Added 14 regression tests covering rejected uncovered partitions, numeric correctness, and no-partition diagnostics.

## C++ style / lint notes

- The PR changes C++ implementation code but does not modify the documented rules in `docs/developer_guide/cpp_style.md`.
- The warning-only C++ API Style Audit CI step remains relevant; this PR does not introduce any new expected correctness/build failures related to linting.
- Any advisory style findings should not block merging absent a clear API, FFI, or maintainability risk.

### DIFF
--- a/src/cuda/op/gemm.cc
+++ b/src/cuda/op/gemm.cc
@@ -125,55 +125,65 @@ ComputeDefaultWarpPartition(const GemmWarpPolicyNode &policy, int M, int N,
   ICHECK(N % k_n_per_warp == 0)
       << "N must be divisible by " << k_n_per_warp << ", but got " << N;
 
+  auto is_valid = [&](int m, int n) {
+    return m * n == num_warps && M % (m * kMPerWarp) == 0 &&
+           N % (n * k_n_per_warp) == 0;
+  };
+
+  bool found = false;
   if (policy.IsFullRow()) {
-    m_warp = num_warps;
-    n_warp = 1;
-    if (M % (m_warp * kMPerWarp) != 0) {
-      int max_m_warps = M / kMPerWarp;
-      m_warp = max_m_warps;
-      n_warp = num_warps / m_warp;
-      if (n_warp == 0)
-        n_warp = 1;
+    for (int m = num_warps; m >= 1; m--) {
+      if (num_warps % m != 0 || !is_valid(m, num_warps / m))
+        continue;
+      m_warp = m;
+      n_warp = num_warps / m;
+      found = true;
+      break;
     }
   } else if (policy.IsFullCol()) {
-    m_warp = 1;
-    n_warp = num_warps;
-    if (N % (n_warp * k_n_per_warp) != 0) {
-      int max_n_warps = N / k_n_per_warp;
-      n_warp = max_n_warps;
-      m_warp = num_warps / n_warp;
-      if (m_warp == 0)
-        m_warp = 1;
+    for (int n = num_warps; n >= 1; n--) {
+      if (num_warps % n != 0 || !is_valid(num_warps / n, n))
+        continue;
+      n_warp = n;
+      m_warp = num_warps / n;
+      found = true;
+      break;
     }
   } else if (policy.IsSquare()) {
-    int max_m_warps = M / kMPerWarp;
     float ideal_ratio = N > 0 ? static_cast<float>(M) / N : 1.0f;
 
-    int best_m = 1;
-    int best_n = 1;
     float best_balance = std::numeric_limits<float>::max();
-    for (int m = 1; m <= max_m_warps && m <= num_warps; m++) {
+    for (int m = 1; m <= num_warps; m++) {
+      if (num_warps % m != 0)
+        continue;
       int n = num_warps / m;
+      if (!is_valid(m, n))
+        continue;
 
       float m_per_warp = static_cast<float>(M) / (m * kMPerWarp);
       float n_per_warp = static_cast<float>(N) / (n * k_n_per_warp);
-      if (m_per_warp < 1 || n_per_warp < 1)
-        continue;
-      if (m * n != num_warps)
-        continue;
-
       float balance = std::abs(m_per_warp / n_per_warp - ideal_ratio);
       if (balance < best_balance) {
         best_balance = balance;
-        best_m = m;
-        best_n = n;
+        m_warp = m;
+        n_warp = n;
+        found = true;
       }
     }
-
-    m_warp = best_m;
-    n_warp = best_n;
   } else {
     ICHECK(0) << "Unknown GemmWarpPolicy";
+  }
+
+  if (!found) {
+    LOG(FATAL) << "No valid warp partition for T.gemm: M=" << M << ", N=" << N
+               << " cannot be evenly covered by " << num_warps
+               << " warps (policy="
+               << (policy.IsFullRow()   ? "FullRow"
+                   : policy.IsFullCol() ? "FullCol"
+                                        : "Square")
+               << "). Each warp must own a multiple of " << kMPerWarp
+               << " rows and " << k_n_per_warp
+               << " columns; adjust `threads` or the block tile shape.";
   }
 
   ICHECK(m_warp * n_warp == num_warps)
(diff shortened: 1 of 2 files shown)


LeiWang1999:
@regression-perf

PR #2742, opened 2026-07-23
[BugFix] Round up sub-byte TVM FFI output storage

cla7aye15I4nd:
## Summary

- round sub-byte TVM FFI output storage up to the next complete backing element
- add a CPU-only boundary regression test for int4 outputs backed by int8 storage

## Intended behavior

A logical sub-byte tensor is passed through PyTorch using a wider storage dtype. The backing tensor must contain enough complete storage elements for every logical bit.

For a logical output with three int4 elements:

```text
logical size:  3 × 4 bits = 12 bits
storage dtype: int8 = 8 bits
required size: ceil(12 / 8) = 2 int8 elements
```

Returning one int8 element provides only 8 bits and cannot hold the 12-bit logical output.

## Root cause and impact

The TVM FFI adapter previously converted the logical final dimension with integer floor division. Whenever the logical bit count was not exactly divisible by the backing dtype width, it discarded the partial final storage element.

The resulting output tensor could therefore be too small for a valid sub-byte output, potentially truncating data or allowing the compiled kernel to write past the allocated backing storage. The fix performs integer ceiling division while retaining dtype lane counts in both the logical and storage widths.

## Regression test

The parameterized test exercises the byte boundary directly:

```text
2 int4 elements =  8 bits -> 1 int8 storage element
3 int4 elements = 12 bits -> 2 int8 storage elements
4 int4 elements = 16 bits -> 2 int8 storage elements
```

The executable is intentionally a no-op: the behavior under test is the output tensor shape allocated and returned by the host-side TVM FFI wrapper before kernel execution.

## Validation

- verified the three-element case returns `torch.Size([1])` with the original floor division
- verified the fixed boundary cases return `(1,)`, `(2,)`, and `(2,)`
- `pytest -q testing/python/jit/test_tilelang_jit_tvm_ffi_executable.py` — 6 passed
- `pre-commit run --files tilelang/jit/adapter/tvm_ffi.py testing/python/jit/test_tilelang_jit_tvm_ffi_executable.py`

## Summary

- Updated TVM FFI storage sizing for sub-byte outputs to use ceiling division, ensuring sufficient backing elements.
- Added CPU-only regression coverage for int4 outputs at 2-, 3-, and 4-element boundaries.

## Testing

- Added parametrized pytest coverage verifying int8 storage dtype and rounded-up output sizes.

### DIFF
--- a/testing/python/jit/test_tilelang_jit_tvm_ffi_executable.py
+++ b/testing/python/jit/test_tilelang_jit_tvm_ffi_executable.py
@@ -1,6 +1,7 @@
 import threading
 from types import SimpleNamespace
 
+import pytest
 import torch
 
 from tilelang.jit.adapter.tvm_ffi import TVMFFIKernelAdapter
@@ -13,6 +14,17 @@ class _FakeKernelParam:
     @staticmethod
     def torch_dtype():
         return torch.float32
+
+
+class _FakeInt4KernelParam:
+    dtype = SimpleNamespace(bits=4, lanes=1)
+
+    def __init__(self, logical_elements):
+        self.shape = [logical_elements]
+
+    @staticmethod
+    def torch_dtype():
+        return torch.int8
 
 
 class _TestAdapter(TVMFFIKernelAdapter):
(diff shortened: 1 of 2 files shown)


LeiWang1999:
Thanks for working on this. The ceil-division itself is mathematically correct, but I believe we should first clarify the packed ABI contract. In my opinion, rounding (or padding) is risky behavior in this context, so we should raise a clear error instead of silently rounding.

cla7aye15I4nd:
Thanks @LeiWang1999

I agree. Silently rounding the storage size is the wrong behavior here; the right move is to pin down the packed sub-byte ABI contract and raise a clear error rather than paper over a non-byte-aligned request.

Digging into this, though, I realized the fix probably shouldn't live here at all, the sub-byte output storage-shape logic is duplicated across every execution backend, and they don't even agree with each othe.

Maybe it should be enforced in one shared place instead of being re-implemented, differently, per backend.

That's a design decision bigger than this PR and I don't have a settled answer yet, so I'll close this for now and maybe follow up with an issueto nail down the contract and a shared sub-byte handling layer.

PR #2745, opened 2026-07-23
[BugFix] Preserve if condition value when splitting statements

cla7aye15I4nd:
## What changed

Snapshot an `if` condition once before `IfStmtBinding` splits a multi-statement body into separately guarded statements. Every generated guard now uses the snapshot instead of re-evaluating the original condition.

The shared helper is used by both the replayable-bind path and the legacy path. A body that produces only one guarded statement keeps the existing direct guard without an unnecessary snapshot.

## Why

A source `if` evaluates its condition exactly once before executing its body. Previously, `IfStmtBinding` could transform:

```python
if A[0] > 0:
    A[0] = 0
    A[1] = 1
```

into the equivalent of:

```python
if A[0] > 0:
    A[0] = 0
if A[0] > 0:
    A[1] = 1
```

The first generated statement changes `A[0]`, so the second generated condition becomes false. For input `[1, 7]`, the transformed function incorrectly produced `[0, 7]` instead of the source program's `[0, 1]`.

The pass now binds `A[0] > 0` once and guards both generated statements with that immutable value, preserving the original evaluation semantics.

## Validation

- Added an end-to-end behavioral regression test that transforms the function, compiles the result with the C target, executes it, and checks the output buffer.
- Confirmed the test fails without the fix (`[0, 7]`) and passes with the fix (`[0, 1]`).
- `pytest -q testing/python/transform/test_tilelang_transform_if_stmt_binding.py` (`8 passed`)
- `pre-commit run --files src/transform/if_stmt_binding.cc testing/python/transform/test_tilelang_transform_if_stmt_binding.py`

## Summary

- Fixed `IfStmtBinding` so multi-statement `if` bodies evaluate the condition once and reuse a snapshot for generated guards, preserving source semantics.
- Added shared guarding logic that supports both replayable-bind and legacy paths; single-statement bodies retain direct guards.
- Updated structural expectations and added a runtime regression test ensuring the original condition value is preserved.
- Validation includes 8 targeted tests plus pre-commit checks.

## C++ style / lint notes

- The PR touches C++ implementation code (`src/transform/if_stmt_binding.cc`) but does not modify the rules in `docs/developer_guide/cpp_style.md`, nor any CI/lint tooling or style documentation.
- The C++ API Style Audit runs as a warning-only CI step; this change does not introduce an obvious API/FFI/maintainability risk that would make new advisory findings blocking for correctness/build purposes.

### DIFF
--- a/src/transform/if_stmt_binding.cc
+++ b/src/transform/if_stmt_binding.cc
@@ -190,6 +190,50 @@ private:
     return IfThenElse(condition, stmt, Stmt(), span);
   }
 
+  bool NeedsConditionSnapshot(const PrimExpr &condition,
+                              const Array<Stmt> &stmts) const {
+    if (SideEffect(condition) > CallEffectKind::kReadState) {
+      return true;
+    }
+
+    auto [condition_reads, _] = CollectStmtAccessRegions(Evaluate(condition));
+    if (condition_reads.empty()) {
+      return false;
+    }
+
+    const BufferSet write_buffers = CollectWriteBuffers(stmts);
+    for (const BufferRegion &read : condition_reads) {
+      if (write_buffers.count(read->buffer)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  Stmt GuardStmts(const PrimExpr &condition, Array<Stmt> stmts,
+                  Span span) const {
+    ICHECK(!stmts.empty());
+    if (stmts.size() == 1) {
+      return GuardStmt(condition, stmts[0], span);
+    }
+
+    if (!NeedsConditionSnapshot(condition, stmts)) {
+      Array<Stmt> guarded_stmts;
+      guarded_stmts.reserve(stmts.size());
+      for (const Stmt &stmt : stmts) {
+        guarded_stmts.push_back(GuardStmt(condition, stmt, span));
+      }
+      return MakeSeq(std::move(guarded_stmts));
+    }
+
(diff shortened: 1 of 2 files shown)


src/transform/if_stmt_binding.cc
```diff
@@ -190,14 +190,29 @@ class IfStmtBindingRewriter : public StmtExprMutator {
...
   }
 
+  Stmt GuardStmts(const PrimExpr &condition, Array<Stmt> stmts,
+                  Span span) const {
+    ICHECK(!stmts.empty());
+    if (stmts.size() == 1) {
+      return GuardStmt(condition, stmts[0], span);
+    }
+
+    Var condition_snapshot("if_condition", condition.dtype());
```
LeiWang1999, line 229: var name `if_condition` is quite too long in my opinion, would be better to just use `"cond"`?

LeiWang1999 (commented):
Thanks, lgtm but I think we should checkout the ci

cla7aye15I4nd:
Closing this — the equivalent fix already landed in #2764, which snapshots the `if` condition before splitting a multi-statement body using the canonical `IsReplayableScalarBind` predicate. `main` already covers these cases (`test_if_stmt_binding_splits_plain_seq_stmt`, `test_if_stmt_binding_does_not_snapshot_pure_condition`, `test_if_stmt_binding_replays_condition_disjoint_from_body_writes`), so this PR is now redundant. Thanks for the equivalent fix in #2764.

PR #2746, opened 2026-07-23
[BugFix] Handle strided global buffers correctly in 1D TMA copies

cla7aye15I4nd:
## Summary

- use the global buffer's declared strides when computing 1D TMA source/destination offsets
- retain the existing contiguous-stride derivation only for buffers without explicit strides
- add a CUDA execution test that uses ordinary `T.copy(..., prefer_instruction="tma")` on a padded tensor view

## Root cause

`Copy::LowerBulk1D` reconstructed global strides solely from the buffer shape. For a legal strided buffer such as shape `(2, 32)` with strides `(40, 1)`, this produced a row stride of `32` instead of the declared `40`.

As a result, copying `A[1, 0:32]` started at physical element offset `32` instead of `40`. The access remained in bounds, but returned the wrong row data.

## Intended behavior

A buffer's explicit strides define how logical indices map to physical storage. `LowerBulk1D` should therefore compute the global address using `buffer.strides` when they are present. Shape-derived contiguous strides remain the fallback for ordinary contiguous buffers whose stride list is empty.

## Validation

- `pre-commit run --files src/cuda/op/copy.cc testing/python/language/test_tilelang_language_tma_copy.py`
- `cmake --build build -j4` (macOS non-CUDA build)
- `pytest -q testing/python/language/test_tilelang_language_tma_copy.py -k explicit_global_stride` (collected successfully; skipped on the local non-CUDA host)

The regression test creates a padded `(2, 40)` allocation, passes its first 32 columns as a normal `(2, 32)` view with stride `(40, 1)`, copies the second logical row through the 1D TMA path, and compares the output with that row.

## Summary

- Fixed 1D TMA source/destination offset calculation to honor explicitly declared global buffer strides, while keeping the contiguous row-major stride derivation as the fallback when explicit strides are not provided.
- Strengthened 1D bulk copy analysis to more accurately detect the required “contiguous global region” shape/stride packing.
- Added CUDA regression coverage for `T.copy(..., prefer_instruction="tma")` on padded/strided tensor views, including a row-stride `(40, 1)` scenario that ensures the physical offset is correct (not assumed contiguous with row stride `32`).

## Validation

- Formatting checks passed.
- Non-CUDA build passed.
- Targeted CUDA test(s) were collected but skipped on the local non-CUDA host.

## C++ style / lint notes

- The PR touches C++ implementation code (`src/cuda/op/*`) but does not change documented C++ style rules in `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step remains advisory; no correctness/build issues are indicated by the changes described here.

### DIFF
--- a/src/cuda/op/copy.cc
+++ b/src/cuda/op/copy.cc
@@ -2307,11 +2307,16 @@ Stmt Copy::LowerBulk1D(const CopyNode &op, const LowerArgs &lower_args,
     global_indices.push_back(r->min);
   }
   std::vector<PrimExpr> global_strides;
-  PrimExpr global_stride = 1;
-  for (size_t i = 0; i < global_tensor->shape.size(); i++) {
-    auto s = global_tensor->shape[global_tensor->shape.size() - i - 1];
-    global_strides.insert(global_strides.begin(), global_stride);
-    global_stride *= s;
+  if (global_tensor->strides.empty()) {
+    PrimExpr global_stride = 1;
+    for (size_t i = 0; i < global_tensor->shape.size(); i++) {
+      auto s = global_tensor->shape[global_tensor->shape.size() - i - 1];
+      global_strides.insert(global_strides.begin(), global_stride);
+      global_stride *= s;
+    }
+  } else {
+    global_strides.assign(global_tensor->strides.begin(),
+                          global_tensor->strides.end());
   }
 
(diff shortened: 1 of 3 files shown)


SiriusNEO:
@regression-perf

PR #2747, opened 2026-07-23
[Pipeline] Add pipeline copy strategy pass config (occupancy/latency/balance)

Da1L8-X:
## Pipeline Copy Strategy Pass Config

Add `tl.pipeline_copy_strategy` pass config to control how PipelinePlanning schedules global→shared copy operations. Three strategies are supported:

- **occupancy** (default): place copies by `last_use` — minimizes shared buffer residence time for better memory reuse and occupancy
- **latency**: place copies by `first_use` — starts prefetch as early as possible for maximum latency hiding
- **balance**: sort copies by `first_use`, then apply chain constraint `effective = max(own_last_use, prev_effective)` — balances prefetch priority with buffer safety

### Example

Given a kernel with two copies where `A.first_use=2, A.last_use=4` and `B.first_use=3, B.last_use=3`:

```python
for k in T.Pipelined(K // block_K, num_stages=2):
    T.copy(A[...], A_s)          # stmt 0: copy A
    T.copy(B[...], B_s)          # stmt 1: copy B
    T.copy(A_s, Out_A[...])      # stmt 2: first_use of A
    T.copy(B_s, Out_B[...])      # stmt 3: first & last use of B
    T.copy(A_s, Out_A2[...])     # stmt 4: last_use of A
```

Pipeline annotations produced by each strategy:

| Strategy | stages | orders | Copy order | Rationale |
|----------|--------|--------|------------|-----------|
| occupancy | `[0,0,2,2,2]` | `[4,2,0,1,3]` | B → A | B.last_use(3) < A.last_use(4), shorter residence first |
| latency | `[0,0,2,2,2]` | `[1,3,0,2,4]` | A → B | A.first_use(2) < B.first_use(3), earlier consumer first |
| balance | `[0,0,1,1,1]` | `[0,1,2,3,4]` | A → B | Chain constraint pushes both to end → Step 2 rotation to front, stage reduced |

### Usage

```python
mod = tilelang.compile(
    kernel,
    target="cuda",
    pass_configs={
        PassConfigKey.TL_PIPELINE_COPY_STRATEGY: "latency",
    },
)
```

### Changes

- `src/transform/pipeline_planning.cc`: add `effective_placement_index` field, three strategy branches, first_use analysis and propagation
- `src/op/builtin.h/cc`: register `kPipelineCopyStrategy` constant
- `tilelang/transform/pass_config.py`: add `TL_PIPELINE_COPY_STRATEGY` enum entry
- `testing/python/transform/test_tilelang_transform_pipeline_copy_strategy.py`: IR-level test validating all three strategies

## Summary

Adds configurable pipeline copy scheduling via `tl.pipeline_copy_strategy`, with three supported strategies:
- `occupancy` (default): order copies by `last_use` to reduce shared-buffer residency.
- `latency`: order copies by `first_use` to start prefetching earlier.
- `balance`: order by `first_use` while applying placement constraints to balance prefetch priority and buffer safety.

Implementation updates pipeline planning to compute copy-stage `first_use` and derive a strategy-specific `effective_placement_index`, then schedules copy stages and their eligibility based on that index instead of `last_use`. The new pass-config key is registered end-to-end in C++ and exposed in the Python pass configuration. IR-level tests are added to validate exact stage/order behavior for all three strategies, and a GEMM autotune benchmark script is included to compare strategy performance (including a latency-focused non-GEMM memory-bound case that shows ~2.5% improvement on H200).

## C++ style / lint notes

This PR touches C++ code in `src/op/builtin.*` and `src/transform/pipeline_planning.cc`, but does not modify rules documented in `docs/developer_guide/cpp_style.md`. The C++ API Style Audit (warning only) may inspect the updated public constructor/method signatures that are used to thread the new `pipeline_copy_strategy` parameter through planning; any TLCPP003/TLCPP004 findings would be advisory and should not block merge absent a concrete API/FFI/maintainability risk introduced by this PR.

## Testing

- Added an IR-level Python test (`test_tilelang_transform_pipeline_copy_strategy.py`) that runs `PipelinePlanning` with `occupancy`, `latency`, and `balance` and asserts expected `software_pipeline_stage` / `software_pipeline_order` sequences.
- Added a GEMM autotune benchmark script (`examples/gemm/example_gemm_autotune_pipeline_strategy.py`) to compare strategy outcomes across cases.

### DIFF
--- a/examples/gemm/example_gemm_autotune_pipeline_strategy.py
+++ b/examples/gemm/example_gemm_autotune_pipeline_strategy.py
@@ -1,0 +1,195 @@
+"""Benchmark: Pipeline Copy Strategy autotune comparison.
+
+Compares three pipeline copy strategies (occupancy/latency/balance) across
+different workloads using per-config pass_configs. Each case is designed to
+favor a specific strategy.
+
+Usage:
+    python example_gemm_autotune_pipeline_strategy.py
+"""
+
+import tilelang as tl
+import tilelang.language as T
+from tilelang.autotuner import AutoTuner
+from tilelang.transform import PassConfigKey
+
+
+def get_strategy_configs():
+    return [{"pass_configs": {PassConfigKey.TL_PIPELINE_COPY_STRATEGY: s}} for s in ("occupancy", "latency", "balance")]
+
+
+# ==================== Case 1: balance wins ====================
+
+
+def balance_favored_kernel():
+    """Small tiles + many A-only ops before B is used.
+
+    Balance's Step 2 rotation consolidates copies and reduces stage count,
+    improving occupancy on H200 (~10%+ faster than alternatives).
+    """
+    M, N, K = 4096, 4096, 4096
+    block_M, block_N, block_K = 32, 32, 32
+    num_stages, thread_num = 2, 128
+    dtype, accum_dtype = T.float16, T.float32
+
+    @T.prim_func
+    def main(
+        A: T.Tensor((M, K), dtype),
+        B: T.Tensor((N, K), dtype),
+        C: T.Tensor((M, N), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=thread_num) as (bx, by):
+            A_shared = T.alloc_shared((block_M, block_K), dtype)
+            B_shared = T.alloc_shared((block_N, block_K), dtype)
+            C_local = T.alloc_fragment((block_M, block_N), accum_dtype)
+            Acc1 = T.alloc_fragment((block_M, block_K), accum_dtype)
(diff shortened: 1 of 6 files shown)


LeiWang1999:
quite interesting and thanks for contribution on that! would you mind show some case that we can benefit from it?

Da1L8-X:
Thanks for the review! I've added a benchmark example with three cases where each strategy wins under different configurations, all verified on H200.
The latency-favored case is hard to construct a significant advantage on H200, so the example uses a non-GEMM memory-bound kernel where it shows ~2.5% improvement. On hardware with higher memory latency, the benefit would be larger.
The main idea of this patch is to expose pipeline scheduling as a tunable knob in the autotune search space. Future optimization passes can follow a similar pattern — providing pass configs that users can tune per workload rather than relying on a single fixed heuristic.

PR #2753, opened 2026-07-23
[BugFix] Support BufferRegion destinations in atomic_addx2 return_prev

### LINKED ISSUE #2750: [BUG] atomic_addx2 return_prev rejects sliced BufferRegion destinations
### Required prerequisites

- [x] I searched the issue and pull-request tracker. The directly related [#2590](https://github.com/tile-ai/tilelang/pull/2590) fixes the x4 sibling; no report covers x2 with both a sliced destination and `return_prev=True`.

Current `main` at `235077cb`.

### System information

Source checkout on Linux x86_64 with Python 3.12. This is a Python eager-builder failure and occurs before target-specific device code runs.

### Problem description

`T.atomic_addx2` computes the return vector type from `dst.dtype` whenever `return_prev=True`:

[`tilelang/language/atomic.py#L325-L329`](https://github.com/tile-ai/tilelang/blob/235077cb5835c6a3632544f1cb4ec8acc396ac28/tilelang/language/atomic.py#L325-L329)

A sliced destination such as `Dst[0:2]` is represented by `tirx.BufferRegion`. That object exposes `buffer` and `region`, but not `dtype`:

[`3rdparty/tvm/python/tvm/tirx/stmt.py#L464-L481`](https://github.com/tile-ai/tilelang/blob/235077cb5835c6a3632544f1cb4ec8acc396ac28/3rdparty/tvm/python/tvm/tirx/stmt.py#L464-L481)

Therefore the valid combination of a sliced destination and `return_prev=True` raises:

```text
AttributeError: 'BufferRegion' object has no attribute 'dtype'
```

The x4 sibling already handles this distinction by reading `dst.buffer.dtype` for a `BufferRegion`:

[`atomic.py#L365-L373`](https://github.com/tile-ai/tilelang/blob/235077cb5835c6a3632544f1cb4ec8acc396ac28/tilelang/language/atomic.py#L365-L373)

### Reproducible example code

```python
import tilelang
import tilelang.language as T


@T.prim_func
def main(
    Dst: T.Tensor((2,), T.float32),
    Val: T.Tensor((2,), T.float32),
    Prev: T.Tensor((2,), T.float32),
):
    with T.Kernel(1, threads=1):
        Prev[0:2] = T.atomic_addx2(
            Dst[0:2],
            Val[0:2],
            return_prev=True,
        )


tilelang.compile(main)
```

The existing return-value regression uses scalar element references, `Dst[0]` and `Val[0]`, so it does not enter the failing `BufferRegion` path:

[`testing/python/language/test_tilelang_language_atomic.py#L829-L850`](https://github.com/tile-ai/tilelang/blob/235077cb5835c6a3632544f1cb4ec8acc396ac28/testing/python/language/test_tilelang_language_atomic.py#L829-L850)

### Expected behavior

The sliced x2 form should derive its element dtype from the underlying buffer, compile successfully, and emit `AtomicAddx2Ret`, matching the region-aware behavior of `atomic_addx4`.

### Actual behavior

The eager builder raises `AttributeError` before lowering because it reads `dst.dtype` directly.

### Desired outcome

Make the x2 return-value path region-aware and add a focused regression using `Dst[0:2]` and `Val[0:2]`. Keep broader atomic validation and memory-order behavior outside this fix.

morluto:
## Problem

`T.atomic_addx2(..., return_prev=True)` computes its packed return type from `dst.dtype`. Sliced destinations are represented as `BufferRegion`, which exposes the element type through `dst.buffer.dtype` instead. As a result, this valid form fails during PrimFunc construction:

```python
Prev[0:2] = T.atomic_addx2(Dst[0:2], Val[0:2], return_prev=True)
```

On current `main`, the focused regression raises:

```text
AttributeError: BufferRegion object has no attribute dtype
```

For the accepted destination domain `D = Buffer | BufferRegion`, the packed return type must be total:

```text
dtype(d) = d.dtype         if d is Buffer
           d.buffer.dtype  if d is BufferRegion

return_type(d) = dtype(d) x 2
```

The previous implementation defined only the first branch.

## Change

Resolve the return element type from the underlying buffer for `BufferRegion` destinations, while preserving direct `Buffer` behavior and the non-returning `handle` path. This mirrors the x4 parity fix merged in #2590.

The existing x2 return-value program now uses sliced destination and value regions. A CPU-safe frontend regression verifies that this form constructs successfully, while the existing CUDA test continues to verify `AtomicAddx2Ret` emission and runtime results in GPU CI.

Broader atomic validation and memory-order semantics are intentionally unchanged.

## Validation

- `bash format.sh --files tilelang/language/atomic.py testing/python/language/test_tilelang_language_atomic.py`
- `python -m pytest -q testing/python/language/test_tilelang_language_atomic.py -k "test_atomic_addx2_return_prev_accepts_sliced_destination or test_atomic_addx2_return_prev"` (`1 passed, 3 skipped` on CPU)
- Repeated the focused frontend regression on an independent Ubuntu 24.04 CPU environment (`1 passed`)

Fixes #2750

## Summary

- Fixed `T.atomic_addx2(..., return_prev=True)` for sliced `BufferRegion` destinations by deriving the return element/vector `DataType` from the underlying `dst.buffer.dtype` (instead of relying on `dst` having a direct `dtype`).
- Preserved existing behavior for direct `Buffer` destinations and kept the non-returning `handle` path unchanged.
- Updated the x2 return-value tests to use sliced destination/value regions (`Dst[0:2]`, `Val[0:2]`, matching the `Prev[0:2]` writeback).
- Added a CPU regression that ensures PrimFunc construction succeeds for the sliced `return_prev` program, while the CUDA test continues to validate `AtomicAddx2Ret` emission and runtime results.

## C++ style / lint notes

Not applicable: this PR does not touch C++, CI, lint tooling, or `docs/developer_guide/cpp_style.md`, and it does not introduce/adjust any C++ API style audit warnings (including the “C++ API Style Audit (warning only)” step).

### DIFF
--- a/testing/python/language/test_tilelang_language_atomic.py
+++ b/testing/python/language/test_tilelang_language_atomic.py
@@ -830,7 +830,7 @@ def atomic_addx2_return_prev_program(dtype=T.float32):
     @T.prim_func
     def main(Dst: T.Tensor((2,), dtype), Val: T.Tensor((2,), dtype), Prev: T.Tensor((2,), dtype)):
         with T.Kernel(1, threads=1):
-            Prev[0:2] = T.atomic_addx2(Dst[0], Val[0], return_prev=True)
+            Prev[0:2] = T.atomic_addx2(Dst[0:2], Val[0:2], return_prev=True)
 
     return main
 
@@ -842,6 +842,10 @@ def atomic_addx4_return_prev_program(dtype=T.float32):
             Prev[0:4] = T.atomic_addx4(Dst[0], Val[0], return_prev=True)
 
     return main
+
+
+def test_atomic_addx2_return_prev_accepts_sliced_destination():
+    atomic_addx2_return_prev_program(T.float32)
 
 
 @tilelang.testing.requires_cuda
--- a/testing/python/language/test_tilelang_language_atomic.py
+++ b/testing/python/language/test_tilelang_language_atomic.py
@@ -835,11 +835,35 @@ def atomic_addx2_return_prev_program(dtype=T.float32):
     return main
 
 
+def atomic_addx2_return_prev_let_bound_program(dtype=T.float32):
+    @T.prim_func
+    def main(Dst: T.Tensor((2,), dtype), Val: T.Tensor((2,), dtype), Prev: T.Tensor((2,), dtype)):
+        with T.Kernel(1, threads=1):
+            dst = Dst[0:2]
+            val = Val[0:2]
+            Prev[0:2] = T.atomic_addx2(dst, val, return_prev=True)
+
+    return main
+
+
+def atomic_addx2_return_prev_ramp_program(dtype=T.float32):
+    @T.prim_func
+    def main(Dst: T.Tensor((2,), dtype), Val: T.Tensor((2,), dtype), Prev: T.Tensor((2,), dtype)):
+        with T.Kernel(1, threads=1):
+            Prev[0:2] = T.atomic_addx2(
+                Dst[T.Ramp(0, 1, 2)],
+                Val[T.Ramp(0, 1, 2)],
+                return_prev=True,
+            )
+
+    return main
+
+
 def atomic_addx4_return_prev_program(dtype=T.float32):
     @T.prim_func
     def main(Dst: T.Tensor((4,), dtype), Val: T.Tensor((4,), dtype), Prev: T.Tensor((4,), dtype)):
         with T.Kernel(1, threads=1):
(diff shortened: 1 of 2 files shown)


LeiWang1999 (approved):
I replaced the string-based dtype construction with a shared structured helper that:
  - Normalizes `Buffer`, `BufferRegion`, and `BufferLoad` to the underlying buffer.
  - Constructs the vector return dtype using `buffer.dtype.with_lanes(lanes)`.
  - Is shared by both `atomic_addx2` and `atomic_addx4`.

PR #2757, opened 2026-07-23
[BugFix] Make T.transpose swap only the final two axes

### LINKED ISSUE #2626: [BUG][Fuzzer][wrong-code] `T.transpose` on 3+ non-trivial dims silently reverses all axes instead of swapping the last two
### Required prerequisites

0.1.11 (commit `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`, tag `v0.1.11`)

### System information

- GPU: NVIDIA L40S (`sm_89`)
- CUDA 13.0, PyTorch 2.11.0+cu130, TileLang 0.1.11
- OS: Linux x86_64

### Problem description

`T.transpose(src, dst)` is documented to transpose the last two dimensions of a
`(..., M, N)` buffer into `(..., N, M)`, preserving the leading (batch) dimensions.
For a plain 2D buffer this works. For any buffer with **3 or more non-trivial
dimensions** it silently **reverses every axis** instead, producing a wrong result
with no error. Because the reversed index range often exceeds the destination
extents, the out-of-bounds writes are masked to zero, so the output is a mix of
misrouted values and zeros.

The compiler computed `dst[k, j, i] = src[i, j, k]` (full 3-axis reversal), then
masked the out-of-range writes to 0, instead of the documented
`dst[a, n, m] = src[a, m, n]` (batch `a` preserved, last two swapped).

- Reproduces identically on re-run (deterministic).
- Independent of dtype (int32 / fp16) and rank (also wrong for 4D `(2,2,3,4)->(2,2,4,3)`).
- 2D `(M,N)->(N,M)` is **correct**; `(1,M,N)->(1,N,M)` (leading dim trivial) is also
  **correct** — the trigger is strictly **>= 3 non-trivial dims**.

### Reproducible example code

```python

A_, M, N = 2, 3, 4  # a batched transpose: (2,3,4) -> (2,4,3)

@T.prim_func
def main(A: T.Tensor((A_, M, N), "int32"), B: T.Tensor((A_, N, M), "int32")):
        As = T.alloc_shared((A_, M, N), "int32")
        Bs = T.alloc_shared((A_, N, M), "int32")
        T.copy(A, As)
        T.transpose(As, Bs)
        T.copy(Bs, B)

k = tilelang.compile(main, out_idx=[1])
A = torch.arange(A_ * M * N, dtype=torch.int32, device="cuda").reshape(A_, M, N)
B = k(A)
ref = A.transpose(1, 2).contiguous()   # documented batched transpose
print("match:", torch.equal(B.cpu(), ref.cpu()))   # -> False
print(B[0].cpu())
```

Observed `B[0]`:
```
[[ 0, 12,  0],
 [ 4, 16,  0],
 [ 8, 20,  0],
 [ 0,  0,  0]]
```
Expected `ref[0]` (`A[0].T`):
```
[[ 0,  4,  8],
 [ 1,  5,  9],
 [ 2,  6, 10],
 [ 3,  7, 11]]
```

<details>
<summary>Mechanism proof: cube shape isolates full-reversal vs swap (no OOB masking)</summary>

On a cube `(3,3,3)->(3,3,3)` the full-reversal permutation is dimension-legal, so no
boundary masking confounds the comparison. The compiler output equals
`np.transpose(A, (2,1,0))` (full reversal) **exactly**, and does **not** equal
`np.transpose(A, (0,2,1))` (documented swap of the last two axes):

```python
import numpy as np

D = 3

@T.prim_func
def main(A: T.Tensor((D, D, D), "int32"), B: T.Tensor((D, D, D), "int32")):
        As = T.alloc_shared((D, D, D), "int32")
        Bs = T.alloc_shared((D, D, D), "int32")
        T.copy(A, As)
        T.transpose(As, Bs)
        T.copy(Bs, B)

k = tilelang.compile(main, out_idx=[1])
xin = torch.arange(D * D * D, dtype=torch.int32, device="cuda").reshape(D, D, D)
B = k(xin).cpu().numpy()
xnp = xin.cpu().numpy()
print("B == swap-last-two (documented):", np.array_equal(B, np.transpose(xnp, (0, 2, 1))))
print("B == full-reversal (2,1,0)     :", np.array_equal(B, np.transpose(xnp, (2, 1, 0))))
```

```
B == swap-last-two (documented): False
B == full-reversal (2,1,0)     : True
```
</details>

### Traceback

No traceback — this is a silent miscompute (wrong output, no error or warning is
raised at compile or run time).

### Expected behavior

`T.transpose(src, dst)` on a `(..., M, N)` buffer should produce `(..., N, M)`:
the leading (batch) dimensions are preserved and only the **last two** dimensions
are swapped, i.e. `dst[a, n, m] = src[a, m, n]`. This is the standard "transpose"
semantics — the docstring's own `Args` describe `(..., M, N) -> (..., N, M)`, and it
matches Triton's `tl.trans` (defaults to "swapping the last two axes, thereby
performing an (optionally batched) 2D transpose") and torch's `x.transpose(-2, -1)`.
Full-axis reversal is a different operation (a `permute((2,1,0))`), not a transpose.
For the reproducer above, `B` should equal `A.transpose(1, 2)`; instead the compiler
fully reverses all axes.

(If a batched transpose is genuinely out of scope for this op, rejecting 3+ non-trivial
dims at the frontend with a clear diagnostic would at least be better than the silent
wrong result — but per the standard semantics above, computing the batched swap is the
natural fix.)

### Additional context

**Root cause.** `T.transpose`'s index construction reverses *all* non-trivial axes instead of swapping only the last two, so for 3+ non-trivial dims it computes a full permutation reversal rather than the documented batched transpose. Concretely, in `src/op/transpose.cc`, `TransposeNode::MakeIndices`, destination path
(`src_dst == 1`) — [lines 109-117](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/transpose.cc#L109-L117):

```cpp
size_t N = nontrivial.size();
size_t nt_idx = 0;
for (size_t i = 0; i < ranges.size(); i++) {
  if (is_one(ranges[i]->extent)) {
    indices.push_back(ranges[i]->min);
  } else {
    size_t rev = N - 1 - nt_idx;              // <-- line 113: reverses ALL non-trivial axes
    indices.push_back(ranges[i]->min + ivs[rev]->var);
    nt_idx++;
  }
}
```

`rev = N - 1 - nt_idx` reverses the loop-variable-to-axis mapping across **all** N
non-trivial dimensions. For `N == 2` this is exactly a swap (correct 2D transpose).
For `N >= 3` it is a full reversal — e.g. `N == 3` maps to `(2,1,0)` instead of the
intended `(0,2,1)` (identity on leading dims, swap only the last two). The same
reversal appears in `MakePredicate`
([line 155](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/transpose.cc#L155):
`iv_idx = do_transpose ? (num_nontrivial - 1 - idx) : idx`), so the boundary predicate
matches the wrong indexing and masks the OOB writes to 0.

**Suggested fix.** Swap only the **last two** non-trivial axes; map the leading
non-trivial axes with the identity. Concretely, for the k-th non-trivial destination
axis (0-based among N): map to loop var `k` for `k < N-2`, to `N-1` for `k == N-2`,
and to `N-2` for `k == N-1`. Apply the same mapping in `MakePredicate`.

**Architecture matrix.**

| Arch | Behavior | Status |
|------|----------|--------|
| L40S `sm_89` | 3D+ transpose silently fully-reversed (wrong output, no error) | tested |
| all archs | root cause is target-independent IR construction in `op/transpose.cc` (SIMT loop, no arch gate) | inferred |

The defect is in target-independent op lowering (`MakeIndices`/`MakePredicate`), so it
is expected on every backend; only L40S `sm_89` was run.

**Provenance.** Not a regression — the reversal was present when `T.transpose` was introduced. Added by [#1971](https://github.com/tile-ai/tilelang/pull/1971) ("Introduce `T.deallocate_tmem` and `T.transpose`", commit `1ede0bd7`, 2026-03-30): `git blame` shows line 113 `size_t rev = N - 1 - nt_idx;` was written in that commit. The shipped transpose tests are all 2D (`num_nontrivial == 2`, where full reversal happens to equal the correct swap), so the `>= 3`-dim path was never exercised. Still present on `main`.

**Dedup.** Searched the open and closed tracker (`transpose`, `T.transpose`, batched transpose, axis reversal) and FILED_ISSUES.md — no existing report of `T.transpose` mis-permuting 3+-dim buffers. The nearby transpose-related items are about the 2D layout/`transpose_A`/`transpose_B` flags in GEMM, a different mechanism.

**Reach.** `T.transpose` is a low-traffic op — no shipped example uses it and its only in-tree callers are the 2D transpose tests, so the 3+-non-trivial-dim path has no coverage (which is why the bug went unnoticed). In practice this affects a user who calls `T.transpose` on a 3D+ (batched) buffer, as its `Args` describe.

morluto:
## Summary

`T.transpose` documented a swap of the final two axes, but its destination index
construction reversed every nontrivial axis. Those operations coincide for a
2D buffer, which allowed the existing tests to pass, but rank-3 and higher
buffers produced wrong code.

For a source shaped `[B, M, N]`, the old and corrected mappings are:

```text
source index:       src[b, m, n]
old destination:   dst[n, m, b]  (full reversal)
new destination:   dst[b, n, m]  (final-two-axis swap)

destination axis:      0  1  2
mapped source axis:    0  2  1
```

This change defines that mapping once and applies it to both destination indices
and boundary predicates. It also maps through the source axes before accounting
for extent-one dimensions, preserving the semantics of cases such as
`[B, M, 1] -> [B, 1, M]`.

Closes #2626.

## Regression evidence

The CPU tests cover rank-3, rank-4, both positions of an extent-one final axis,
and a shifted destination region that requires boundary-predicate clipping.
Against `main`, the four full-region rank and singleton cases fail. With this
change, all five cases match their PyTorch references.

For the ordinary rank-3 case, generated C contains the equivalent of:

```text
dst[i, k, j] = src[i, j, k]
```

For the shifted destination region, generated C proves that the predicate uses
the source penultimate-axis iterator:

```text
if (j < 2) {
  dst[i, k, j + 1] = src[i, j, k];
}
```

## Test plan

- `pytest -q testing/python/cpu/test_tilelang_cpu_transpose.py` — 5 passed
- `bash format.sh --files src/op/transpose.cc src/op/transpose.h testing/python/cpu/test_tilelang_cpu_transpose.py` — passed
- Full source CMake build — passed

The implementation is in target-independent transpose lowering; the regression
suite executes the resulting kernels on the CPU backend.

## Summary

- Fixed `T.transpose` to swap only the final two axes while preserving leading dimensions for rank-3 and higher buffers.
- Updated the transpose implementation so the same final-two-axis mapping is used consistently for both index construction and boundary predicates, including cases where one of the transposed extents is `1` (e.g. `[B, M, 1] -> [B, 1, M]`).
- Updated C++/doc comments to reflect the corrected “final-two-axes swap” behavior.

## Validation

- Added CPU regression tests covering rank-3, rank-4, extent-one final-axis cases, and a shifted-destination boundary/predicate scenario.
- CPU regression suite, formatting checks, and full source CMake build pass.

## C++ style / lint notes

- Does not appear to touch `docs/developer_guide/cpp_style.md` or related C++ style guidance.
- The C++ API Style Audit (warning only) remains advisory; this PR does not introduce new/changed lint tooling or CI configuration, and no correctness/build impact is indicated by any warning-only style checks.

### DIFF
--- a/src/op/transpose.cc
+++ b/src/op/transpose.cc
@@ -150,40 +150,18 @@ Array<PrimExpr> TransposeNode::MakeIndices(const Array<IterVar> &ivs,
 }
 
 PrimExpr TransposeNode::MakePredicate(arith::Analyzer *analyzer,
-                                      const Array<IterVar> &ivs,
-                                      Array<PrimExpr> extents,
-                                      int src_dst) const {
-  bool do_transpose = (src_dst == 1);
-  Array<Range> ranges = src_dst == 0 ? src_range : dst_range;
-
-  std::vector<int> source_axis_to_iv;
-  if (do_transpose) {
-    ICHECK_EQ(src_range.size(), ranges.size())
-        << "Transpose source and destination ranks must match";
-    source_axis_to_iv = MakeSourceAxisToIterVarMap(src_range, ivs.size());
-  }
-
+                                      const Array<PrimExpr> &indices,
+                                      const Array<PrimExpr> &extents) const {
   Array<PrimExpr> cond_list;
-  ICHECK(extents.size() == ranges.size()) << extents << " " << ranges;
-  size_t idx = 0;
-  for (size_t i = 0; i < ranges.size(); i++) {
-    size_t iv_idx;
-    if (do_transpose) {
-      size_t src_axis = SourceAxisForDestinationAxis(i, ranges.size());
-      int mapped_iv_idx = source_axis_to_iv[src_axis];
-      if (mapped_iv_idx < 0)
-        continue;
-      iv_idx = static_cast<size_t>(mapped_iv_idx);
-    } else {
-      if (is_one(ranges[i]->extent))
-        continue;
-      iv_idx = idx++;
-    }
-    PrimExpr cond = ranges[i]->min + ivs[iv_idx]->var < extents[i];
+  ICHECK_EQ(indices.size(), extents.size())
+      << "Transpose index rank (" << indices.size()
+      << ") must match buffer rank (" << extents.size() << ")";
+  for (size_t i = 0; i < indices.size(); i++) {
+    PrimExpr cond = indices[i] < extents[i];
     if (!analyzer->CanProve(cond, arith::ProofStrength::kSymbolicBound)) {
       cond_list.push_back(cond);
     }
-    cond = ranges[i]->min + ivs[iv_idx]->var >= 0;
+    cond = indices[i] >= 0;
     if (!analyzer->CanProve(cond, arith::ProofStrength::kSymbolicBound)) {
       cond_list.push_back(cond);
     }
@@ -206,8 +184,8 @@ For TransposeNode::MakeSIMTLoop(arith::Analyzer *analyzer) const {
   Array<PrimExpr> src_indices = MakeIndices(loop_vars, 0);
   Array<PrimExpr> dst_indices = MakeIndices(loop_vars, 1);
 
-  PrimExpr src_predicate = MakePredicate(analyzer, loop_vars, src->shape, 0);
-  PrimExpr dst_predicate = MakePredicate(analyzer, loop_vars, dst->shape, 1);
+  PrimExpr src_predicate = MakePredicate(analyzer, src_indices, src->shape);
+  PrimExpr dst_predicate = MakePredicate(analyzer, dst_indices, dst->shape);
 
   PrimExpr value = BufferLoad(src, src_indices);
   if (src->dtype != dst->dtype)
--- a/src/op/transpose.cc
+++ b/src/op/transpose.cc
@@ -1,6 +1,6 @@
 /*!
  * \file tl/op/transpose.cc
- * \brief Transpose operator: dst[j, i] = src[i, j] using SIMT loops.
+ * \brief Transpose operator that swaps the final two axes using SIMT loops.
  */
 
 #include "transpose.h"
@@ -23,6 +23,30 @@ using namespace tirx;
 using namespace ffi;
 
 namespace {
+
+size_t SourceAxisForDestinationAxis(size_t dst_axis, size_t rank) {
+  ICHECK(rank >= 2) << "Transpose requires at least two dimensions";
+  if (dst_axis == rank - 2)
+    return rank - 1;
+  if (dst_axis == rank - 1)
+    return rank - 2;
+  return dst_axis;
+}
+
+std::vector<int> MakeSourceAxisToIterVarMap(Array<Range> ranges,
+                                            size_t iv_count) {
+  std::vector<int> source_axis_to_iv(ranges.size(), -1);
+  size_t iv_idx = 0;
+  for (size_t axis = 0; axis < ranges.size(); axis++) {
+    if (is_one(ranges[axis]->extent))
+      continue;
+    source_axis_to_iv[axis] = static_cast<int>(iv_idx++);
+  }
+  ICHECK_EQ(iv_idx, iv_count)
+      << "Transpose: source nontrivial dimensions (" << iv_idx
+      << ") != iterator count (" << iv_count << ")";
+  return source_axis_to_iv;
+}
 
 std::vector<TransposeImpl> &TransposeImplRegistry() {
   static std::vector<TransposeImpl> registry;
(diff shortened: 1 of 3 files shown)


morluto:
Addressed the boundary-predicate coverage request in 41fd718c. The new shifted-destination test emits `if (j < 2) { dst[i, k, j + 1] = src[i, j, k]; }`, proving the predicate uses the corrected final-two-axis mapping. The focused CPU suite passes (5 tests).

LeiWang1999 (approved):
Thanks for the original fix — the final-two-axis mapping itself is correct. I pushed a follow-up commit, `9bd9c9b96`, to address a remaining boundary-safety issue.

  The previous `MakePredicate` implementation independently reconstructed the transpose mapping from `IterVar`s. For an extent-one axis, no iterator exists, so the predicate skipped that axis even though `MakeIndices`
  still generated an actual index from `range.min`. A shifted singleton region could therefore cause an unguarded out-of-bounds load or store.

  The follow-up changes make the generated indices the single source of truth:

  - `MakeIndices` remains responsible for transpose semantics.
  - `MakePredicate` now checks the exact source/destination index expressions against the corresponding buffer shape.
  - Extent-one, negative, constant-OOB, and symbolic indices are all handled consistently.
  - Provably in-bounds checks are still removed by the analyzer, so the normal path gains no extra runtime guards.
  - Added regression tests for both out-of-bounds singleton source and destination regions.

Issue #2538, opened 2026-07-07
[BUG][Fuzzer][wrong-code] `MergeIfStmt` silently merges same-condition `if`s that the first body invalidates, giving wrong results


Hughshine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.11

### System information

NVIDIA H100 80GB HBM3 (sm_90), CUDA 13.0, Python 3.11, from-source `v0.1.11` (`cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). The defect is target-independent — it is in the `MergeIfStmt` TIR transform, applied unconditionally in the CUDA lowering pipeline ([`pipeline.py#L241`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/pipeline.py#L241)).

### Problem description

`MergeIfStmt` merges two consecutive same-condition `if` statements into one, but doesn't check whether the first body writes a variable the shared condition reads. When it does, the merge is unsound: the second guard should now be false, but the merged form evaluates the condition only once (against the stale value), so the second body runs anyway — a silently wrong result, no error.

```python
Out[0] = 0
if A[0] == 0:
    A[0] = 7      # first body writes the guard variable A[0]
if A[0] == 0:     # A[0] is now 7 -> this guard is false, body must be skipped
    Out[0] = 20
```

With `A[0] == 0` initially, the first `if` fires and sets `A[0] = 7`, so the second guard is false and `Out[0]` must stay `0`. Instead `MergeIfStmt` collapses the two into `if (A[0] == 0) { A[0] = 7; Out[0] = 20; }` — the guard is tested once against the original `A[0]`, both bodies run, and `Out[0]` becomes `20`.

The trigger is specifically *the first body writing the guard variable*, not merging itself: when the first body writes a **different** element (`A[1] = 7`), the guard is not invalidated, the merge is sound, and the result is correct (control below).

### Reproducible example code

```python
import torch
import tilelang
import tilelang.language as T


@tilelang.jit
def kernel(A: T.Tensor((4,), "int32"), Out: T.Tensor((4,), "int32")):
    with T.Kernel(1, threads=1) as bx:
        Out[0] = 0
        if A[0] == 0:
[...]

PR #2764, opened 2026-07-24
[BugFix] Preserve if condition evaluation during fan-out

SiriusNEO:
## Motivation

`IfStmtBinding` fans an `if` body out into one guarded statement per body
statement.  #2744 correctly prevents `MergeIfStmt` from merging independently
authored state-reading conditions, because those conditions must be
re-evaluated.  The generated guards from one original `if`, however, must keep
the original single condition evaluation.

Without that distinction, an earlier generated guard can mutate state read by
a later guard.  For example, resetting the innermost tensor index makes the
next generated carry guard false, so outer indices no longer advance.

This is a follow-up to #2744 and the wrong-code report in #2538.

## Changes

- collect the bodies produced by `IfStmtBinding` before wrapping them in guards
- replay pure/read-only conditions when guarded bodies cannot modify their
  inputs; otherwise bind the condition once and reuse that snapshot
- leave pure conditions and single guarded bodies unchanged
- add a CPU-only runtime regression covering `IfStmtBinding` followed by
  `MergeIfStmt`, alongside #2744's independent-if regression
- cover both default and legacy bind handling, and verify pure conditions do
  not gain a snapshot

## Testing

- `cmake -S . -B build && cmake --build build -j$(nproc)`
- `python -m pytest testing/python/issue/test_tilelang_issue_merge_if.py testing/python/transform/test_tilelang_transform_if_stmt_binding.py -q` (`12 passed`)
- the four block-sparse pipeline tests that failed the first CUDA CI run (`4 passed`)
- related pipeline-planning, lexical-scope, and allocation-location tests (`31 passed`)
- full transform suite excluding one unrelated CUDA 13.1 `half_t` conversion failure (`339 passed, 7 skipped`)
- original `normal_kernel` reproducer (`33 passed`)
- pre-commit on all three changed files

## Summary

- Preserve single evaluation of state-reading `if` conditions when `IfStmtBinding` fans out guarded statements by binding the condition once to a fresh boolean and reusing it across generated guards.
- Centralize guard generation with a new `GuardStmts` helper that snapshots only potentially invalidated conditions; replayable read-only and pure conditions continue to be reused without snapshots.
- Refactor both `BindIfStmtLegacy` and `BindIfStmtWithReplayableBindInlining` to collect `IfStmtBinding` bodies first, then apply guarding once at the end (including correct handling of grouped `bind_scope` sequences), instead of emitting per-element guarded statements.
- Add/verify CPU-only runtime regressions for default and legacy bind handling, including ensuring pure conditions are not snapshotted and exercising `MergeIfStmt` interactions.

## Testing

- Passed targeted, related, transform-suite, reproducer, build, and pre-commit checks; one unrelated CUDA 13.1 `half_t` conversion failure remains excluded from the full transform suite.
- CUDA CI failures/spot checks and transform/pipeline-planning verification are reported as passing according to the PR’s verification notes.

## C++ style / lint notes

- This PR touches C++ implementation and related transform/tests, but does not update rules documented in `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” remains advisory; no new blocking correctness/build issues or new TLCPP003/TLCPP004-style findings are indicated by the PR changes.

### DIFF
--- a/src/transform/if_stmt_binding.cc
+++ b/src/transform/if_stmt_binding.cc
@@ -199,11 +199,15 @@ private:
 
     Array<Stmt> guarded_stmts;
     PrimExpr guard_condition = condition;
-    // Fan-out must preserve the original if's single condition evaluation.
-    // Otherwise an earlier guarded body may mutate state read by a later guard.
-    if (SideEffect(condition) > CallEffectKind::kPure) {
-      Var condition_snapshot("if_condition", condition.dtype(), span);
-      guarded_stmts.push_back(Bind(condition_snapshot, condition, span));
+    Var condition_snapshot("if_condition", condition.dtype(), span);
+    Stmt condition_bind = Bind(condition_snapshot, condition, span);
+    auto [condition_reads, _] = CollectStmtAccessRegions(condition_bind);
+    const BufferSet write_buffers = CollectWriteBuffers(stmts);
+    // Replay a read-only condition when no guarded body can change its inputs.
+    // Otherwise preserve the original if's single condition evaluation.
+    if (!IsReplayableScalarBind(condition_bind, condition_reads,
+                                write_buffers)) {
+      guarded_stmts.push_back(condition_bind);
       guard_condition = condition_snapshot;
     }
     for (const Stmt &stmt : stmts) {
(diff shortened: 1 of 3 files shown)


SiriusNEO:
Fixed the first CUDA CI failure

PR #2765, opened 2026-07-25
[BugFix] Add arithmetic operators to vec_type in common.h for CPU codegen

GY-Bai:
## Problem

When the `c` target generates vectorized code via the AutoVectorize pass, it produces expressions like:

```cpp
*(float4*)(C + (i * 4)) = (*(float4*)(A + (i * 4)) - *(float4*)(B + (i * 4)));
```

PR #1901 introduced the `vec_type` template and its type aliases (`float4`, `int32_t4`, etc.), which resolved the `'float4' was not declared in this scope` error. However, the arithmetic operators (`+`, `-`, `*`, `/`) for `vec_type` were not implemented, causing the compiler to fail with:

```
error: no match for operator- (operand types are float4 {aka vec_type<float, 4>} and float4 {aka vec_type<float, 4>})
```

This error occurs on any platform where the C codegen uses the `vec_type` template instead of GCC native vector extensions (e.g., ARM64 macOS/Linux).

## Solution

This commit adds the four arithmetic operators (`+`, `-`, `*`, `/`) to the `vec_type` struct in `src/tl_templates/cpp/common.h`.

These are element-wise operations that simply delegate to the underlying scalar operator for each element. No compound assignment operators are needed since the codegen only emits binary arithmetic expressions.

## Verification

- A minimal reproducer (`reproducer_vec_type.py`) is included
- Verified on ARM64 macOS (M2) and ARM64 Linux (Neoverse-N1)
- All existing CPU `c` target tests pass after the fix

## Related

- PR #1901 introduced the initial vec_type definitions
- Fixes the remaining compilation issue on ARM64 platforms where GCC native vector extensions are not available

## Summary

- Add element-wise `+`, `-`, `*`, and `/` operators to `vec_type`, enabling arithmetic on vector aliases such as `float4` and `int32_t4` when native GCC vector extensions are unavailable.
- Add a minimal TileLang C-backend reproducer that compiles and validates vector addition.

## C++ style / lint notes

- This PR updates C++ code under `src/` (`src/tl_templates/cpp/common.h`), so it should be reviewed against `docs/developer_guide/cpp_style.md`.
- The new `vec_type` arithmetic operator overloads take the RHS as `const vec_type&`, matching the style guidance to pass non-trivial inputs by `const&` unless the function consumes the value.
- The “C++ API Style Audit (warning only)” CI step runs `maint/scripts/audit_cpp_api_style.py --github-warnings --limit 20`; while it may report advisory TLCPP003/TLCPP004-style findings, nothing in this change suggests a correctness/build/test problem or an API/FFI/maintainability risk that would warrant blocking merge on warning-only output.

### DIFF
--- a/reproducer_vec_type.py
+++ b/reproducer_vec_type.py
@@ -1,8 +1,11 @@
 """
 Minimal reproducer for the missing vec_type arithmetic operators in common.h.
 
-Without the fix, compiling a CPU kernel with `c` target fails:
-  error: no match for 'operator-' (operand types are 'float4' and 'float4')
+Without the fix, compiling a CPU kernel with `c` target fails when the
+AutoVectorize pass generates vector operations like +, -, *, / on
+vec_type (float4, int32_t4, etc.) because these operators are not defined.
+
+The fix adds operator+, operator-, operator*, and operator/ to vec_type.
 
 Requires: tilelang, torch
 """
--- a/reproducer_vec_type.py
+++ b/reproducer_vec_type.py
@@ -1,0 +1,37 @@
+"""
+Minimal reproducer for the missing vec_type arithmetic operators in common.h.
+
+Without the fix, compiling a CPU kernel with `c` target fails:
+  error: no match for 'operator-' (operand types are 'float4' and 'float4')
+
+Requires: tilelang, torch
+"""
+
+import tilelang
+import tilelang.language as T
+import torch
+
+
+@T.prim_func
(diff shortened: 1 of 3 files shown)


GY-Bai:
Thanks for the reminder! Running pre-commit now.

GY-Bai:
pre-commit formatting applied (clang-format). Docstring updated to reference all four operators.

PR #2767, opened 2026-07-25
[BugFix] Add threadgroup address space qualifier in Metal codegen for shared memory pointer arithmetic

GY-Bai:
## Problem

The Metal codegen (`codegen_metal.cc`) inherits the CUDA-oriented codegen patterns
that emit unqualified pointer types like `void*` and untyped casts like `((void*)((char*)...))`.
In CUDA these work without address space qualifiers, but **Metal Shading Language (MSL)
requires explicit qualifiers on all pointers**.

Without this fix, any kernel using `T.alloc_shared` with `target="metal"` generates
invalid MSL code that fails at compilation:

```
error: pointer type must have explicit address space qualifier
  void* A_shared = ((void*)((char*)buf_dyn_shmem + 0));
```

## Solution

Add string post-processing after codegen to inject `threadgroup` qualifiers:

| Pattern | Before | After |
|---------|--------|-------|
| Variable declarations | `void* A_shared = ...` | `threadgroup void* A_shared = ...` |
| Pointer arithmetic | `((void*)((char*)...)` | `((threadgroup void*)((threadgroup char*)...)` |
| Typed casts | `((float2*)A_shared)` | `((threadgroup float2*)A_shared)` |
| General pattern | `((TYPE*)_shared)` | `((threadgroup TYPE*)_shared)` |

## Verification

Captured the generated Metal source via monkey-patching `torch.mps.compile_shader`,
manually applied the same `threadgroup` transformations, and confirmed compilation
succeeds with `torch.mps.compile_shader()` on macOS M2.

## Scope

Only `codegen_metal.cc` is modified. Other backends (CUDA, HIP, CPU) are unaffected.

## Related

- PR #799 introduced the initial Metal backend support
- PR #2765 fixes a separate CPU `c` target compilation issue in common.h

## Summary

- Updated Metal code generation (`src/metal/codegen/codegen_metal.cc`) to add explicit `threadgroup` address-space qualifiers for shared-memory pointers, including shared-memory `void*` declarations, pointer arithmetic, and typed casts (via targeted rewrites and generalized cast handling).
- Added a reproducer for the Metal address space issue (`reproducer_metal_address_space.py`) and an additional CPU vectorization reproducer (`reproducer_vec_type.py`).
- Extended C++ `vec_type<T, N>` with element-wise `+`, `-`, `*`, and `/` operators (`src/tl_templates/cpp/common.h`).
- Updated `3rdparty/cutlass` and `3rdparty/tvm` submodule revisions.

## Validation

[...]

PR #2768, opened 2026-07-25
[BugFix] Add arithmetic operators to vec_type in common.h for CPU codegen

GY-Bai:
## Problem

When the `c` target generates vectorized code via the AutoVectorize pass, it produces expressions like:

```cpp
*(float4*)(C + (i * 4)) = (*(float4*)(A + (i * 4)) - *(float4*)(B + (i * 4)));
```

PR #1901 introduced the `vec_type` template and its type aliases (`float4`, `int32_t4`, etc.), which resolved the `float4 was not declared in this scope` error. However, the arithmetic operators (`+`, `-`, `*`, `/`) for `vec_type` were not implemented, causing g++ to fail with:

```
error: no match for operator- (operand types are float4 {aka vec_type<float, 4>} and float4 {aka vec_type<float, 4>})
```

This error occurs on any platform where the C codegen uses the `vec_type` template instead of GCC native vector extensions (e.g., ARM64 macOS/Linux).

## Solution

Add the four arithmetic operators (`+`, `-`, `*`, `/`) to the `vec_type` struct in `src/tl_templates/cpp/common.h`.

## Verification

- A minimal reproducer (`reproducer_vec_type.py`) is included
- Verified on ARM64 macOS (M2) and ARM64 Linux (Neoverse-N1)
- All existing CPU `c` target tests pass after the fix

## Related

- PR #1901 introduced the initial vec_type definitions
- PR #2767 fixes a separate Metal codegen issue

## Summary

Adds element-wise arithmetic support to the `vec_type` C++ template for `+`, `-`, `*`, `/` and their compound-assignment forms. This fixes CPU C codegen failures when AutoVectorize emits expressions using vector aliases such as `float4` and `int32_t4`, including on ARM64 platforms.

Adds a minimal `vec_add` reproducer that compiles and validates a 256-element float32 vector addition through the C backend.

## Validation

- Verified on ARM64 macOS and ARM64 Linux.
- Existing CPU C-target tests pass.
- Reproducer validates output against the Torch reference within `1e-5`.

## C++ style / lint notes

- The PR changes C++ code but does not appear to touch rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) may report advisory findings for the added operators; these are warning-only and should not block merging absent a clear API, FFI, or maintainability risk.

### DIFF
--- a/reproducer_vec_type.py
+++ b/reproducer_vec_type.py
@@ -1,0 +1,37 @@
+"""
+Minimal reproducer for the missing vec_type arithmetic operators in common.h.
+
+Without the fix, compiling a CPU kernel with `c` target fails:
+  error: no match for 'operator-' (operand types are 'float4' and 'float4')
+
+Requires: tilelang, torch
+"""
+
+import tilelang
+import tilelang.language as T
+import torch
+
+
+@T.prim_func
+def vec_add(
+    A: T.Tensor((256,), "float32"),
+    B: T.Tensor((256,), "float32"),
+    C: T.Tensor((256,), "float32"),
+):
+    for i in T.Parallel(256):
+        C[i] = A[i] + B[i]
+
+
+def main():
+    f = tilelang.compile(vec_add, target="c")
+    A = torch.randn(256)
+    B = torch.randn(256)
+    C = torch.zeros(256)
+    f(A, B, C)
+    ref = A + B
+    assert (C - ref).abs().max().item() < 1e-5
+    print("OK: vec_add compiled and ran successfully")
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


3rdparty/cutlass
SiriusNEO, line 1: Is the TVM & CUTLASS change necessary? In my understanding, they are not related to this PR

GY-Bai:
Thanks for the review! The 3rdparty submodule changes were a local git issue (dirty submodules from branch switching). I've force-pushed clean commits — now only common.h and the reproducer are changed. Sorry for the noise!

SiriusNEO (approved):

PR #2777, opened 2026-07-27
[BugFix] Resolve partial scalar reduce barrier participation

KellyFrog:
`T.reduce` will produce wrong barrier counts in `SM >= 90`.

This can be reproduced with

```python
"""Reproduce a conditional 128-thread barrier generated by TileLang."""

from __future__ import annotations

from typing import Any

import tilelang
import tilelang.language as T

BLOCK_THREADS = 128
FRAGMENT_THREADS = 64
ROWS = 4
WIDTH = 512
VECTOR_SIZE = 8


@tilelang.jit(
    out_idx=1,
    target="cuda",
    pass_configs={
        tilelang.PassConfigKey.TL_DISABLE_DATA_RACE_CHECK: True,
        tilelang.PassConfigKey.TL_DISABLE_THREAD_STORAGE_SYNC: True,
    },
)
def make_repro_kernel() -> Any:
    def x_layout(i: int, j: int) -> tuple[int, int]:
        # j // 8 is in [0, 64), so only threads 0..63 own x_frag.
        thread_id = j // VECTOR_SIZE
        local_id = i * VECTOR_SIZE + j % VECTOR_SIZE
        return thread_id, local_id

    @T.prim_func
    def fragment_reduce(
        x: T.Tensor((ROWS, WIDTH), "float32"),
        out: T.Tensor((ROWS, WIDTH), "float32"),
    ) -> None:
        with T.Kernel(1, threads=BLOCK_THREADS):
            x_frag = T.alloc_fragment((ROWS, WIDTH), "float32")
            sum_frag = T.alloc_fragment((ROWS,), "float32")

            T.annotate_layout(
                {
                    x_frag: T.Fragment(x_frag.shape, forward_fn=x_layout),
                    sum_frag: T.Fragment(
                        sum_frag.shape,
                        forward_fn=lambda i, rep: (rep, i),
                        replicate=FRAGMENT_THREADS,
                    ),
                }
            )

            # T.Parallel -> T.reduce_sum -> T.Parallel is the trigger.
            for i, j in T.Parallel(ROWS, WIDTH):
                x_frag[i, j] = x[i, j]

            T.reduce_sum(x_frag, sum_frag, dim=1)

            for i, j in T.Parallel(ROWS, WIDTH):
                out[i, j] = x_frag[i, j] / sum_frag[i]

    return fragment_reduce


def main() -> None:
    kernel = make_repro_kernel()


if __name__ == "__main__":
    main()
```

Now this issue is fixed by introducing intermediate representation in pass `LowerTileOp`.

For convince, reduces that use non-uniform registers layouts in a warp (e.g. using registers only in threads that `tx % 2 == 0`) or use registers in non-consecutive warps (e.g. `tx [0, 32)` and `tx [64, 92)`) will now be rejected by the compiler.

Extra tests over `T.reduce` are added.

2 bugs are observed but not fixed in current PR:

- The same issue also occurs in `SM < 90`, where the default sync method is `__syncthreads` and will require all threads in a block to participate in `T.reduce`.
- Artificial warp specialization (e.g. `if tx < 32: T.reduce(A) elif tx < 64 T.reduce(B)`) calling `T.reduce` at the same time will interfere each other's barriers as the current code hardcodes barrier 1 and 2 in `AllReduce`.

They are expected to be fixed in other 2 PRs.

## Summary

- Fixes incorrect `T.reduce` barrier counts on SM90+ for reductions involving only part of a scalar fragment.
- Resolves participating threads during lowering and supports warp-aligned contiguous partial-CTA reductions.
- Rejects non-uniform warp layouts, non-consecutive warp allocations, invalid thread ranges, and multiple partial scalar reductions that would collide on shared barriers/workspace.
- Updates CUDA barrier generation and keeps ROCm interfaces compatible.
- Adds extensive correctness and rejection tests, including coverage for partial warps, offsets, disjoint groups, and pre-SM90 behavior.

## C++ style / lint notes

- C++ sources and public lowering interfaces are modified; no changes to `docs/developer_guide/cpp_style.md` are indicated.
- The C++ API Style Audit (warning only) remains advisory. Any TLCPP003/TLCPP004 findings should be treated as warning-only unless they reveal a concrete API or maintainability risk.
- Correctness limitations remain for SM<90 block-wide synchronization and artificial warp specialization with concurrent reductions.


testing/python/language/test_tilelang_language_reduce.py
```diff
@@ -58,6 +64,253 @@ def _reduce_op(T, op, src, dst, dim, batch=1):
...
+        env={**os.environ, "TILELANG_PRE_SM90_PARTIAL_REDUCE_CHILD": "1"},
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, (
+        f"partial scalar reduction did not complete on pre-SM90 CUDA (exit code {result.returncode}):\n{result.stderr}"
+    )
+
+
+if __name__ == "__main__" and os.environ.get("TILELANG_PRE_SM90_PARTIAL_REDUCE_CHILD"):
```
SiriusNEO: Why should we introduce this env var in test?

SiriusNEO (changes requested):

LeiWang1999 (dismissed):

LeiWang1999 (commented):
Overall LGTM, but likely we should fix batch reduce as well.

LeiWang1999 (dismissed):

src/backend/common/op/reduce.h
```diff
@@ -75,6 +78,101 @@ inline Fragment ComputeReducerLayout(const Fragment &src_layout, int dim) {
...
+  ICHECK_LT(end, *block_min + *block_extent)
+      << "tl.reduce: scalar AllReduce participating thread range ends after "
+         "the CTA thread bounds";
+
+  int64_t warp_size = 32;
+  if (auto warp_size_attr = target->GetAttr<Integer>("thread_warp_size")) {
+    warp_size = warp_size_attr.value()->value;
+  }
+  ICHECK_EQ((base - *block_min) % warp_size, 0)
+      << "tl.reduce: partial scalar AllReduce requires a warp-aligned "
```
SiriusNEO, line 164: Should this check use base % warp_size instead? If `(tx >= 16 && tx < 80)`, both `base` and `block_min` are 16, so the current check incorrectly treats the range as aligned. But thread 16 is in the middle of a physical warp, while the later full-mask shuffle requires the range to start at `0, 32, 64`, etc. Consider the following partial reduce:

```python
@T.prim_func
  def kernel(
      x: T.Tensor((1, 512), "float32"),
      out: T.Tensor((1,), "float32"),
  ):
          tx = T.get_thread_binding()

          x_frag = T.alloc_fragment((1, 512), "float32")
          sum_frag = T.alloc_fragment((1,), "float32")

          T.annotate_layout({
              x_frag: T.Fragment(
                  x_frag.shape,
                  forward_fn=lambda i, j: (16 + j // 8, j % 8),
              ),
              sum_frag: T.Fragment(
                  sum_frag.shape,
              ),
          })

          for i, j in T.Parallel(1, 512):
              x_frag[i, j] = x[i, j]

          if tx >= 16 and tx < 80:
              T.reduce_sum(x_frag, sum_frag, dim=1)
```

KellyFrog: My bad.

Fixed.

SiriusNEO (approved):
LGTM

PR #2787, opened 2026-07-27
[Metal] Cap threadgroup size at 512 for Apple GPU GEMM

GY-Bai:
Apple M1/M2 GPU registers per core: ~208KB vs NVIDIA: 256KB.
Larger threadgroups cause register spilling to device memory,
significantly degrading performance.

This caps threads at 512 in the Metal GEMM lowering path, following:
- MLX: `tgp_size = WM*WN*32 ≤ 256`
- llama.cpp: `N_SIMDWIDTH = 32`

Only affects GEMM lowering for Metal target (`gemm_metal.py`).

## Summary

- In `tilelang/metal/op/gemm/gemm_metal.py` (`GemmMetal.lower`), the Metal GEMM lowering path now emits a `warnings.warn` when `thread_nums` is greater than `512`, noting that larger Metal GEMM threadgroups may cause register spilling on Apple GPUs.
- The implementation does **not** clamp/truncate `thread_nums`; the warning is emitted after `thread_nums = thread_bounds.extent` and before `self.policy.compute_warp_partition(...)`, so warp partitioning still uses the original `thread_nums`.
- Added debugging and validation tooling for Metal GEMM:
  - `capture_msl.py`: registers a post-processor callback to write generated MSL source to `/tmp/debug_metal.msl`.
  - `metal_gemm_bench.py`: correctness smoke test + performance benchmark comparing TileLang (Metal) vs PyTorch MPS.
  - `metal_test_adaptive.py`: explores candidate tile sizes based on shared-memory fit and runs correctness checks vs `torch.mm`.
  - `metal_test_clean.py`: correctness check for a fixed GEMM size and attempts to compile the generated MSL shader.


tilelang/metal/op/gemm/gemm_metal.py
```diff
@@ -29,6 +29,12 @@ def lower(
         mbar_phase_expr: tir.PrimExpr | None = None,
     ):
         thread_nums = thread_bounds.extent
+        # Apple GPU: cap threadgroup size at 512 to avoid register spilling.
+        # M1/M2 registers per core: ~208KB vs NVIDIA: 256KB.
+        # Larger threadgroups cause register pressure and performance degradation.
+        # Reference: MLX max threads = WM*WN*32 ≤ 256; llama.cpp N_SIMDWIDTH=32.
+        if thread_nums > 512:
+            thread_nums = 512
```
LeiWang1999: Silently truncating `thread_num` during compilation is problematic. Users should explicitly specify this value in their TileLang program, rather than having the compiler adjust it without notification. however if thread_nums > 512 can lead to register spill theoretically, we can report a warning for user.

LeiWang1999 (changes requested):

GY-Bai:
Thanks for the review and guidance! That's a fair point — silently truncating thread_nums is not the right approach. Let me study how similar constraints are handled elsewhere in the repo (e.g., warnings, assertions, or letting the user configure explicitly) and revise accordingly. I'll push an update in a few hours.

GY-Bai:
Replaced by #2799 — clean branch without submodule contamination.

PR #2805, opened 2026-07-29
[BugFix] Fix two-instance modeling in ThreadSync cross-thread race checks

LJC00118:
## Summary

- Fixed ThreadSync cross-thread race modeling by introducing/using `FreshenMutableReads`, improving constraint renaming/substitution/merging semantics, and refining `is_assume`/assumption handling to avoid incorrect “two-instance” cross-thread equivalence checks.
- Strengthened barrier hoisting decisions under divergent/non-uniform control flow by adding a proof-based “block-uniform condition” check for `if` hoist candidates and refining condition/thread-variable property tracking.
- Improved cross-thread conflict detection by adjusting constraint preparation/instance substitution for disjoint-pointer and cross-thread hazard proofs, and refining same-index conflict short-circuiting.
- Reworked parallel-loop race verification to model “other iteration” constraints via renamed constraint-set merging and updated equality proof logic accordingly.
- Expanded ThreadSync regression coverage for cases where thread-private accesses must not sync (including modulo-`if`, non-uniform if/else branches, loop cases), plus new assertions around barrier insertion and unorderable-hazard reporting with premise/assume/bind interactions.
- Minor doc maintenance: added `disjointness` to the spelling wordlist.

## C++ style / lint notes

- This PR changes C++ implementation/declarations (not documented C++ style rules) and does **not** modify `docs/developer_guide/cpp_style.md`.
- The CI step **“C++ API Style Audit (warning only)”** may emit advisory findings (e.g., TLCPP003/TLCPP004) due to header surface changes (adding `tvm::tl::FreshenMutableReads` and `Constr::FreshenReads`, and related constraint visitor/set behavior). These are warning-only and should be treated separately from correctness/build/test outcomes.

### DIFF
--- a/src/transform/common/constr_visitor.h
+++ b/src/transform/common/constr_visitor.h
@@ -27,23 +27,25 @@ namespace tvm::tl {
  * \brief Replace every mutable read in an expression with a fresh, independent
  *        variable.
  *
- * `Analyzer::Bind` installs a rewrite `var -> value`, so a value reading
- * mutable state would equate two reads across a store this does not track, and
- * make two instances agree where two threads read two different registers. A
- * fresh unknown per read only ever weakens equalities.
+ * `Analyzer::Bind` installs a rewrite `var -> value`, so a definition reading
+ * mutable state would outlive a store this does not track, and would make two
+ * instances agree where two threads read two different registers.
  */
 class FreshenMutableReads : public tirx::ExprMutator {
 public:
   using tirx::ExprMutator::operator();
 
 private:
+  /*!
+   * \brief A fresh variable for one occurrence of \p e, never reused for a
+   *        structurally equal one.
+   *
+   * An opaque call need not return the same value twice (`f() - f()` is not
+   * zero), and two reads of one location may be separated by a store. Sharing
+   * would assert an equality that need not hold.
+   */
   PrimExpr Fresh(const PrimExpr &e) {
-    auto it = memo_.find(e);
-    if (it != memo_.end())
-      return it->second;
-    tirx::Var var("free" + std::to_string(memo_.size()), e.dtype());
-    memo_.emplace(e, var);
-    return var;
+    return tirx::Var("free" + std::to_string(count_++), e.dtype());
   }
 
   PrimExpr VisitExpr_(const tirx::BufferLoadNode *op) override {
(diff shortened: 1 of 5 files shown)


src/transform/thread_storage_sync.cc
```diff
@@ -949,12 +993,25 @@ struct TileLangThreadSyncPlanner : public ConstrVisitor {
...
+      // and a plain __syncthreads() is right. Only a proof counts, not the
+      // heuristic.
+      if ((condition_prop.depends_on_runtime && !is_block_uniform) ||
+          (condition_prop.requires_hoist && !proven_uniform)) {
         LOG(WARNING)
             << "[ThreadSync] Hoisting sync from inside if to before if. "
             << "Condition is not safe for in-if sync: " << op->condition;
+        LOG(WARNING) << "[ThreadSync] The hoisted barrier no longer separates "
+                        "the accesses it was inserted for, as both ends of the "
+                        "conflict are inside the branch, so the race remains. "
```
SiriusNEO: For this case, we just throw another warning here. Can we fix it?

src/transform/common/constr_visitor.h
```diff
@@ -15,9 +17,58 @@
...
+  }
+  PrimExpr VisitExpr_(const tirx::ReduceNode *op) override {
+    return Fresh(ffi::GetRef<PrimExpr>(op));
+  }
+  PrimExpr VisitExpr_(const tirx::CallNode *op) override {
+    // Opaque or stateful calls are unknowns; pure calls recurse so that their
+    // pure sub-structure survives.
+    if (tirx::SideEffect(ffi::GetRef<PrimExpr>(op)) >
+        tirx::CallEffectKind::kPure)
+      return Fresh(ffi::GetRef<PrimExpr>(op));
```
SiriusNEO, line 66: This line calls `Fresh` for every Stateful/kOpaque op. Consider a stateful external function:
```bash
v = call_extern("next_value") - call_extern("next_value");
```
`call_extern` is `kOpaque`, so the two calls may return different values (e.g. 7 and 3, making `v == 4`). However, because the two CallNodes are structurally equal, `Fresh()` maps both to the same variable, producing: `v = free0 - free0;  // always 0`.

This can make the analyzer prove false equalities and potentially miss a parallel-write race. Non-pure calls should receive a fresh variable per occurrence, rather than per structurally equal expression.

src/transform/thread_storage_sync.cc
LJC00118: fixed

src/transform/common/constr_visitor.h
LJC00118: I dropped the shared struct memo, now each occurrence gets its own variable.

SiriusNEO (approved):

PR #2339, opened 2026-06-05
[TileOP] Add SM70 GEMM FMA fallback

cklxx:
## Summary

This reopens the SM70 GEMM fallback work from closed PR #2279 on top of current `main`.

The main change is a scalar `cuda.fma` GEMM implementation for Volta cases that SM70 tensor-core MMA cannot legally cover, such as BF16 operands or shapes outside the SM70 MMA tile constraints. Supported FP16 Volta MMA cases still dispatch to the existing SM70 MMA path.

Compared with #2279, this version removes the controversial `copy.cc` Volta fragment staging path entirely. There is no fragment -> shared -> fragment fallback and no `tl_frag_copy` staging buffer. Fragment cast copy remains on the existing normal copy lowering path.

## Changes

- Add `GemmFMA` and register it as `cuda.fma`.
- Route unsupported Volta GEMM combinations to `cuda.fma` at dispatch time.
- Keep aligned FP16 SM70 SS/RS cases on `cuda.mma`.
- Add full-region assertions for fragment A/B/C in `GemmFMA`.
- Use explicit local accumulation in the FMA lowering.
- Use separate A/B staging dtypes (`a_dtype`, `b_dtype`) instead of assuming one input dtype.
- Add dispatch and SM70 kernel coverage tests.

## Validation

On V100 (`Tesla V100-SXM2-32GB`, compute 7.0):

- `cmake --build build -j$(nproc)` in `~/tilelang-sm70-copy`: passed.
- `~/tilelang/.venv/bin/python -m pytest testing/python/cuda/test_cuda_mma_sm75_dispatch.py -q`: 8 passed.
- Explicit `sm_70` source+cubin checks for BF16 SS, BF16 RS, and FP16 fragment-copy RS:
  - BF16 SS/RS source has no `mma_sync_sm70<...>` and no `tl_frag_copy`.
  - FP16 fragment-copy RS source has no `tl_frag_copy` and still uses legal FP16 SM70 MMA.
- BF16 SS SASS check: `FFMA=128`, `HMMA=0`.
- ARLE AOT smoke using `gen_tilelang_aot.py` for `gated_delta_rule_chunk_a_sm70`: cubin generated, source has no BF16 SM70 MMA, SASS `HMMA=0`, `FFMA=256`.

The runtime CUDA pytest file was also invoked on the V100, but the local PyTorch install reports the NVIDIA driver as too old and those three GPU runtime tests were skipped before execution.

## Summary by CodeRabbit

* **New Features**
[...]

PR #2811, opened 2026-07-30
[CUDA] Extend the GEMM FMA fallback to SM75

Chennesxu:
## Summary

On SM75 (Turing), a `T.gemm` whose dtype combination has no native Turing `mma.sync` atom crashes instead of computing. FP32 and FP64 compile but trap when the kernel launches, leaving the CUDA context in an error state:

```text
cute/arch/mma_sm80.hpp:315: Assertion `0 && "Attempting to use SM80_16x8x8_F32TF32TF32F32_TN without CUTE_ARCH_MMA_SM80_ENABLED"` failed.
```

BF16 fails during nvcc compilation:

```text
src/tl_templates/cuda/instruction/mma.h(167): error: static assertion failed with "tl::mma_sync: unsupported configuration"
```

The FP32, FP64, and BF16 cases in `test_tilelang_kernel_gemm.py` all hit these failures on an SM75 GPU. The device-side asserts from FP32 and FP64 also cause later tests in the same process to fail.

## Root Cause

On Turing, `SelectInst` routed GEMMs to the plain MMA path even when their dtype combination had no native SM75 atom. FP32, FP64, and BF16 therefore selected SM80-only atoms.

#2339 added the `cuda.fma` correctness fallback for GEMMs without a usable tensor-core MMA path, but only connected it to Volta.

## Fix

Add `AllowTuringMma` alongside the Volta check and select the existing `cuda.fma` implementation when SM75 has no native atom for the requested dtype combination.

Unlike the Volta check, the Turing check is intentionally dtype-only. The SM75 emitter supports the same operand scopes and transpose combinations as the generic MMA path, so scope-based routing would incorrectly demote valid FP16 `sr`/`rs` cases to the fallback. Besides being slower, the fallback accumulates serially, and FP16 accumulation loses accuracy over long K.

## Tested

* Added `test_tilelang_kernel_sm75_gemm_fma.py` following the SM70 pattern from #2339. Compile-only checks pinned to an explicit `sm_75` target, so they run on any CUDA runner, assert that FP32, FP64, and BF16 lower through the fallback while FP16 stays on MMA. Execution checks gated on compute capability 7.5 compare the results against PyTorch.
* Reverting the `SelectInst` change makes the fallback tests fail with the errors above, while the FP16 test still passes.
* Verified on a TITAN RTX (`sm_75`, Turing) with CUDA 12.4: `test_tilelang_kernel_gemm.py` and `test_tilelang_kernel_gemm_sm75.py` pass, with FP16, INT8, and INT4, including the `sr`/`rs` fragment cases, keeping their native MMA lowering.

## Summary

- Extended SM75 GEMM instruction selection to use the existing `cuda.fma` fallback for unsupported FP32, FP64, and BF16 combinations.
- Preserved native Turing MMA lowering for supported FP16 and integer cases.
- Added SM75 dispatch, compile-only lowering, and hardware correctness tests.

## C++ style / lint notes

- This PR does not modify `docs/developer_guide/cpp_style.md` or public C++ APIs.
- The repository’s “C++ API Style Audit (warning only)” CI step remains applicable; no correctness or build issues are implied by advisory style findings.

### DIFF
--- a/src/cuda/op/gemm.cc
+++ b/src/cuda/op/gemm.cc
@@ -125,6 +125,24 @@ bool Use2CtaRequested(const GemmNode &op) {
   return false;
 }
 
+// Native SM75 mma.sync atoms (see tl_templates/cuda/instruction/mma.h).
+// Dtype-only on purpose: SM75 MMA handles the same scopes and transposes as
+// the generic path, so scope checks here would wrongly demote f16 to FMA.
+bool AllowTuringMma(const GemmNode &op) {
+  DataType a = op.a_->dtype;
+  if (a != op.b_->dtype) {
+    return false;
+  }
+  if (a == DataType::Float(16)) {
+    return op.c_->dtype == DataType::Float(16) ||
+           op.c_->dtype == DataType::Float(32);
+  }
+  if ((a.is_int() || a.is_uint()) && (a.bits() == 8 || a.bits() == 4)) {
+    return op.c_->dtype == DataType::Int(32);
+  }
+  return false;
+}
+
 void FatalWgmmaUnavailable(const GemmNode &op, Target target) {
   LOG(FATAL) << "T.wgmma_gemm() requires Hopper WGMMA lowering, but "
                 "constraints were not satisfied. Got target="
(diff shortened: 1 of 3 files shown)


cklxx (approved):

cklxx:
good job

Chennesxu:
> good job

thanks!

LeiWang1999 (approved):

PR #2825, opened 2026-07-31
[CUDA][Reduce] Restore exact thread image counting

KellyFrog:
First see & merge https://github.com/tile-ai/tvm/pull/61 .

Created safe fallback for counting threads in `src/backend/common/op/reduce.h` when determinating named barrier count.

## Summary

- Restores exact CUDA thread-image counting for reduction named barriers.
- Uses Z3 to count distinct participating thread IDs.
- Falls back to the bounding span when Z3 returns `unknown`.
- Rejects unsatisfiable or non-contiguous thread images.
- Updates the `3rdparty/tvm` submodule. Merge `tile-ai/tvm#61` first.

## C++ style / lint notes

- The PR changes C++ code in `src/backend/common/op/reduce.h`.
- It does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step is not affected.
- The changes introduce no known correctness, build, or test issues. Advisory style warnings remain non-blocking.

### DIFF
--- a/src/backend/common/op/reduce.h
+++ b/src/backend/common/op/reduce.h
@@ -10,6 +10,7 @@
 #include "op/reduce.h"
 #include "support/check.h"
 #include <tvm/ir/cast.h>
+#include <tvm/ir/with_context.h>
 #include <tvm/runtime/logging.h>
 
 #include "layout/layout.h"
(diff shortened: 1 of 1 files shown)


LeiWang1999:
Closing this for now, as the current PR implementation does not further improve the current kernel's performance and would add unnecessary pressure on the solver. We can revisit this in the future if our related build infrastructure becomes more formalized.

PR #2829, opened 2026-07-31
[Fix] Discover flat CUDA includes for NVRTC

### LINKED ISSUE #2828: [BUG] Linux NVRTC misses flat pip CUDA CCCL includes
### Required prerequisites

- [x] I have read the documentation.
- [x] I searched the issue tracker and pull requests for `NVRTC include cccl` and `NVRTC CUDA include layout`; no directly related report was found.

### TileLang version

`0.1.12` at `bdb769ae` on Linux x86_64.

### Problem

The NVRTC backend always derives its Linux CCCL include path from:

```text
$CUDA_HOME/targets/x86_64-linux/include/cccl
```

Some pip-installed CUDA toolkits instead use a flat layout:

```text
$CUDA_HOME
├── include
│   └── cccl
└── targets/<arch>-linux/include
    └── cccl
```

The two branches are alternative (and sometimes overlaid) layouts. With a flat-only pip installation, TileLang passes the non-existent target-specific CCCL directory to NVRTC and does not pass the existing `$CUDA_HOME/include/cccl`. Headers such as `<cuda/std/...>` therefore cannot be resolved.

### Reproduction

This synthetic reproduction is independent of the host CUDA installation and applies the include calculation currently used on Linux:

```python
import platform
import tempfile
from pathlib import Path

with tempfile.TemporaryDirectory() as root:
    cuda_home = Path(root)
    (cuda_home / "include/cccl").mkdir(parents=True)
    target_arch = "sbsa-linux" if platform.machine() in ("aarch64", "arm64") else "x86_64-linux"
    old_cccl_option = cuda_home / "targets" / target_arch / "include" / "cccl"

    print(f"flat_cccl exists={(cuda_home / 'include/cccl').is_dir()}")
    print(f"old_cccl_option={old_cccl_option} exists={old_cccl_option.is_dir()}")
```

Observed on x86_64:

```text
flat_cccl exists=True
old_cccl_option=/var/tmp/.../targets/x86_64-linux/include/cccl exists=False
```

### Expected behavior

NVRTC include discovery should use flat `include/` and `include/cccl` trees when present, retain system `targets/<arch>-linux/include` discovery, support installations containing both layouts, and preserve the existing Windows and missing-layout fallback behavior.

morluto:
Fixes #2828.

### Problem

On Linux, the NVRTC backend always derived CCCL headers from `$CUDA_HOME/targets/<arch>-linux/include/cccl`. Flat pip CUDA installations instead provide `$CUDA_HOME/include` and `$CUDA_HOME/include/cccl`, so the generated NVRTC options omitted the existing CCCL tree and pointed at a non-existent target-specific path.

A synthetic flat tree reproduced the mismatch on the base revision:

```text
flat_include=/var/tmp/.../include exists=True
flat_cccl=/var/tmp/.../include/cccl exists=True
old_cccl_option=/var/tmp/.../targets/x86_64-linux/include/cccl exists=False
```

### Change

Discover existing flat and host-target include roots and their CCCL subdirectories, preserving their deterministic order. Installations that overlay both layouts include both. Windows retains its flat include behavior, and a missing Linux layout retains the previous fallback candidate list so NVRTC still owns the final missing-header diagnostic.

CUDA versions before 13 continue to receive the corresponding legacy `cuda/std` include for each discovered base include root.

### Regression coverage

The new synthetic-layout tests construct temporary directory trees for:

- flat Linux pip installs;
- x86_64 target-specific system installs;
- combined flat and SBSA target-specific installs;
- missing Linux layouts and their legacy fallback;
- the existing flat Windows layout.

The tests never inspect the host CUDA installation and would fail on the base implementation because it has no flat CCCL discovery path.

### Validation

- Focused pytest launcher removed the main checkout's editable `_Finder` and `ScikitBuildRedirectingFinder` entries from `sys.meta_path`, then printed `tilelang.__file__` before calling `pytest.main(["testing/python/jit/test_tilelang_jit_nvrtc_include_paths.py", "-q"])`. It resolved `/home/morluto/dev/tilelang-agent-nvrtc/tilelang/__init__.py` and reported 5 passed.
- `python -m pre_commit run --files tilelang/jit/adapter/nvrtc/include_paths.py tilelang/jit/adapter/nvrtc/libgen.py testing/python/jit/test_tilelang_jit_nvrtc_include_paths.py` — all hooks passed.

## Summary

- Added deterministic NVRTC CUDA include discovery for Linux flat pip layouts and target-specific toolkit layouts.
- Included existing `cccl` subdirectories for each discovered include root.
- Preserved Windows flat-layout handling and legacy fallback paths when Linux directories are missing.
- Preserved legacy `cuda/std` includes for CUDA versions before 13.
- Updated NVRTC compilation to use the discovered include paths.
- Added regression tests for flat, target-specific, combined, missing, and Windows layouts.
- Targeted tests and pre-commit hooks pass.

### DIFF
--- a/testing/python/jit/test_tilelang_jit_nvrtc_include_paths.py
+++ b/testing/python/jit/test_tilelang_jit_nvrtc_include_paths.py
@@ -1,0 +1,56 @@
+from pathlib import Path
+
+from tilelang.jit.adapter.nvrtc.include_paths import discover_cuda_include_paths
+
+
+def _make_include_tree(cuda_home: Path, relative_paths: list[str]) -> list[str]:
+    paths = []
+    for relative_path in relative_paths:
+        path = cuda_home / relative_path
+        path.mkdir(parents=True)
+        paths.append(str(path))
+    return paths
+
+
+def test_discovers_flat_pip_include_layout(tmp_path):
+    expected = _make_include_tree(tmp_path, ["include", "include/cccl"])
+
+    assert discover_cuda_include_paths(str(tmp_path), machine="x86_64") == expected
+
+
+def test_discovers_target_specific_system_include_layout(tmp_path):
+    expected = _make_include_tree(
+        tmp_path,
+        ["targets/x86_64-linux/include", "targets/x86_64-linux/include/cccl"],
+    )
+
+    assert discover_cuda_include_paths(str(tmp_path), machine="x86_64") == expected
+
+
+def test_discovers_flat_and_target_specific_include_layouts(tmp_path):
+    expected = _make_include_tree(
+        tmp_path,
+        [
+            "include",
+            "include/cccl",
+            "targets/sbsa-linux/include",
+            "targets/sbsa-linux/include/cccl",
+        ],
+    )
+
+    assert discover_cuda_include_paths(str(tmp_path), machine="aarch64") == expected
+
+
+def test_preserves_legacy_paths_when_cuda_layout_is_missing(tmp_path):
+    assert discover_cuda_include_paths(str(tmp_path), machine="x86_64") == [
+        str(tmp_path / "include"),
+        str(tmp_path / "targets/x86_64-linux/include"),
+        str(tmp_path / "targets/x86_64-linux/include/cccl"),
+    ]
+
+
+def test_preserves_flat_windows_include_layout(tmp_path):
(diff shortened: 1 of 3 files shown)


lucifer1004 (approved):

lucifer1004:
Thanks for the work!

Issue #2827, opened 2026-07-31
[Feature] Add arena packing for TMEM allocations


Rachmanino:
## Motivation

Each `T.alloc_tmem` currently lowers to an independent hardware allocation. Since allocation sizes are rounded to supported column counts, several logical buffers can exceed the 512-column limit even when their combined footprint fits.

## Proposal

Pack compatible logical TMEM buffers with the same scope, lifetime, and CTA group into shared physical arenas, while preserving the dtype, shape, and layout of each buffer. Emit one allocate/deallocate per arena and lower each logical buffer to `arena_base + column_offset`.

For example, logical buffers requiring 384, 4, and 8 columns should remain separate in TileLang source but may share one 512-column physical allocation. Packing should not increase total TMEM usage or change explicit deallocation semantics.

PR #2831, opened 2026-07-31
[CUDA] Pack logical TMEM buffers into shared `tcgen05.alloc` arenas

Rachmanino:
#2827

## Problem

`LowerSharedTmem` gives every `T.alloc_tmem` buffer its own `tcgen05.alloc`, rounded up to a power of two with a minimum of 32 columns. Narrow buffers therefore each burn a full 32-column allocation even when they need 4.

That is not just waste — it makes some kernels inexpressible. A 384-column accumulator (three 128-column pipeline stages) next to three 4-column block-scale buffers needs **396** columns, but as four separate allocations it asks the hardware for `512 + 32 + 32 + 32 = 608` of the **512** columns a CTA has. Today nothing reports this; the kernel just compiles and misbehaves on device.

The workaround has been to declare one wide TMEM buffer and slice scale factors out of it by hand, passing column offsets manually — which is what a compiler should be doing.

CUTLASS solves the same problem with a bump allocator over one allocation (`cutlass::detail::find_tmem_tensor_col_offset`, used by `sm100_blockscaled_mma_warpspecialized.hpp` to place `accumulator → tCtSFA → tCtSFB`), and DeepGEMM hand-writes the equivalent layout.

## Approach

Plan the allocations before emitting them. `PlanTmemArenas` packs logical TMEM buffers into shared physical allocations, and each buffer's column offset inside the allocation it shares is folded into every address it forms.

Two addressing paths carry TMEM addresses today, and both are handled:

- `BufferLoad` on a TMEM buffer — used by `tcgen05.ld`/`tcgen05.st` — gains the offset in the encoded coordinate.
- The `(base address Var, offset)` argument pairs of the `tl.ptx_tcgen05_*` intrinsics (`_mma_ss`, `_mma_ts`, `_mma_blockscaled_ss`, `cp_warpx4`), which all codegen to `*(uint32_t*)base + offset`, get the offset added to the second half of the pair.

A base address that reaches anything else is rejected with a diagnostic rather than silently addressing the start of the allocation.

**Packing policy.** A buffer joins an existing allocation only when that *strictly* reduces the total number of allocated columns. Each placement therefore costs at most the buffer's own standalone allocation, so:

- packing can never grow a kernel's TMEM footprint, and
- a kernel whose buffers already fit lowers exactly as it did before.

Widest-first placement keeps wide buffers on 32-column boundaries.

**Alignment.** A buffer at least 32 columns wide starts on a 32-column boundary — the same alignment `tcgen05.alloc` would have given it alone — so sharing an allocation cannot leave an accumulator or a TMEM-resident A operand less aligned than it is today. Narrower buffers use the 4-column granularity of `tcgen05.cp.32x128b.warpx4`, the boundary CUTLASS and DeepGEMM place these very operands on.

**Lifetimes.** A buffer with an explicit `T.deallocate_tmem` keeps its own allocation: releasing it early would release whatever shares it. Every other buffer in a block is live for the whole block, so this needs no liveness analysis to be correct. It does leave value on the table — buffers with disjoint live ranges could reuse the same columns outright instead of only filling each other's power-of-two padding — which is marked as a TODO on the planner.

**New diagnostic.** A block that allocates more than 512 columns is now reported at compile time (unless it manages a lifetime by hand, where the peak is not the sum).

## Effect on existing kernels

Measured by capturing `device_kernel.cu` for every TMEM example under baseline and under this change (separate `TILELANG_CACHE_DIR`s, so no cache hit crosses versions) and diffing:

| example | before | after |
| --- | --- | --- |
| `gemm_mxfp8_blockscaled_1d1d` | 3 allocations, 320 columns | 2 allocations, **288 columns** |
| `grouped_gemm_mxfp8_blockscaled_1d1d` | 3 allocations, 320 columns | 2 allocations, **288 columns** |
| `fp8_fp4_gemm_1d1d_sm100` | 3 allocations, 320 columns | 2 allocations, **288 columns** |
| `mha_fwd_bshd`, `gqa_fwd_bshd` | 128 + 128 + 64 / 128 + 128 | same allocations at the same addresses; only the order the declarations and deallocs are emitted in changed |
| `gemm_tcgen5mma{,_ws,_ws_persistent,_ws_clc}` | 1 allocation | byte-identical generated CUDA |

All of them still pass their own correctness checks.

The whole difference in the generated CUDA for the block-scaled GEMM is:

```diff
-  __shared__ __align__(16) uint sfa_data[1];
-    tl::tmem_allocate<true>((&(sfa_data[0])), 32);
-    tl::tcgen05_cp<true>(..., (*reinterpret_cast<uint32_t*>(sfa_data)) + 0);
+    tl::tcgen05_cp<true>(..., (*reinterpret_cast<uint32_t*>(sfb_data)) + 8);
-    tl::tcgen05mma_blockscaled_ss<...>(..., (*reinterpret_cast<uint32_t*>(sfa_data)) + 0, (*reinterpret_cast<uint32_t*>(sfb_data)) + 0);
+    tl::tcgen05mma_blockscaled_ss<...>(..., (*reinterpret_cast<uint32_t*>(sfb_data)) + 8, (*reinterpret_cast<uint32_t*>(sfb_data)) + 0);
-    tl::tmem_deallocate<true>((&(sfa_data[0])), 32);
```

One fewer address word, one fewer alloc/dealloc pair, and the MMA loads one base address instead of two — so this removes a shared-memory load rather than adding one.

Wall-clock is not a usable signal on the machine I measured on (shared with another tenant): three interleaved rounds of the block-scaled GEMM gave baseline 3032 / 2931 / 1303 TFLOP/s against 2824 / 3031 / 1324, i.e. ±7% within a round and 2.3x across rounds. The code-level facts above are the claim; the timings only confirm nothing catastrophic.

## Tests

New `testing/python/transform/test_tilelang_transform_tmem_arena_packing.py`:

- 396 columns across four buffers become one 512-column allocation at offsets 0 / 384 / 388 / 392, matching DeepGEMM's hand-written layout column for column
- block-scale factors share an allocation while the accumulator keeps its own, because packing all three would cost more
- equal-cost cases keep separate allocations
- allocations are issued widest first (PTX forbids a later `tcgen05.alloc` from being larger)
- the never-worse invariant over six shape combinations
- explicit `T.deallocate_tmem` keeps its own allocation
- a dynamic (pipeline-stage) coordinate still carries the constant offset
- an over-budget block is reported

New GPU test in `test_tilelang_language_tmem_copy.py`: a store/load roundtrip through two buffers sharing one allocation. The *wide* buffer's roundtrip is the assertion that matters — the narrow one would still round-trip if its store and load dropped the offset together, but its store would land on the wide buffer's columns.

Existing coverage, on a B200:

- `testing/python/transform/`: 380 passed, 7 skipped
- `tcgen05` GEMM / sliced-operand / TMEM-copy / atom-MMA language tests plus the bf16-TS and int8 TMEM kernel tests: 58 passed, 1 skipped
- examples run end to end: mxfp8 and grouped mxfp8 block-scaled GEMM, `mha_fwd_bshd`, `gqa_fwd_bshd`, `fp8_fp4_gemm_1d1d_sm100`, and the three `gemm_sm100` variants

The new tests were validated by mutation: dropping the column offset in `LowerSharedTmem` fails the GPU roundtrip and three of the planning tests.

## Notes

- The planning step lives inside `LowerSharedTmem` rather than in a separate pass, since buffer collection, column sizing and allocation emission are all already there. It is a self-contained function and can be split out if a second consumer appears.
- Packing relies on the convention that a `tl.ptx_tcgen05_*` intrinsic passes a TMEM operand's base address immediately followed by its offset. That convention is now documented at the point where it is depended on, and a base address reaching any other context is a hard error.

### DIFF
--- a/src/cuda/transform/lower_shared_tmem.cc
+++ b/src/cuda/transform/lower_shared_tmem.cc
@@ -2,11 +2,17 @@
  *  \file lower_shared_tmem.cc
  *  \brief Convert shared.tmem buffers to plain shared + ptx init, and do
  *         coordinate translation (from logical address to physical address)
+ *
+ *  Logical buffers are not one-to-one with tcgen05.alloc calls: PlanTmemArenas
+ *  packs several into one allocation when that saves columns, and shifts every
+ *  address a packed buffer forms by where it starts in the allocation.
  */
 #include "cuda/target_utils.h"
 #include "op/builtin.h"
 #include "support/check.h"
 #include "tvm/ir/type.h"
+#include <algorithm>
+#include <string>
 #include <tvm/arith/analyzer.h>
 #include <tvm/ir/cast.h>
 #include <tvm/tirx/analysis.h>
@@ -17,6 +23,8 @@
 #include <tvm/tirx/stmt_functor.h>
 #include <tvm/tirx/transform.h>
 #include <unordered_set>
+#include <utility>
+#include <vector>
 
 namespace tvm {
 namespace tl {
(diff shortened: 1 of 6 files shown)


Yongqi-Zhuo:
LGTM cc @LeiWang1999

Issue #2844, opened 2026-08-02
[BUG] rocm ci disabled since may 21, 2026 for tileang due to lack of AMD machines contributed


functionstackx:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

amd

### System information

mi355 & mi325

### Problem description

hi @andyluo7 and @chunfangamd, can u provide tilelang team
- at least 1  mi355 (gfx950) machine and
- at least 1 mi300/mi325 (gfx942)  machine 

for CI?

Since May 21, (3 months ago), ROCm CI on TileLang has been disabled https://github.com/tile-ai/tilelang/pull/2241  despite many important kernels on AMD platform being written in TileLang (for example https://github.com/sgl-project/sglang/pull/24033/changes )

### Reproducible example code

The Python snippets:

```python

```

### Traceback

```pytb

```

andyluo7:
@functionstackx , working on re-enabling it on 1 mi300x machine first. PR #2863 and #2864 filed as the first step.

functionstackx:
@andyluo7 Thanks but i am not sure what your PR does considering there is still no AMD machines hooked up to TileLang github repo https://github.com/LeiWang1999/tilelang/blame/74caac19592abe95b6decfb83f58a2f96c5e2fe7/.github/workflows/ci.yml

andyluo7:
AMD General

It will be hooked up soon when I receive the token from Lei Wang.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2863, opened 2026-08-04
[Testing] Gate CUDA-only tests so non-CUDA backends can run the suite

andyluo7:
## Summary

A large number of tests under `testing/python` import CUDA-only internals directly (for example `tilelang.cuda.pipeline.CUDAPassPipelineBodyPrologue`) or exercise NVIDIA-specific features — TMA, LDGSTG, mbarrier, warp specialization, tcgen05, tfloat32.

`src/cuda/CMakeLists.txt` compiles only a small "ALWAYS" subset and then `return()`s when `USE_CUDA=OFF`, so those FFI symbols are never registered and the tests fail with:

```
AttributeError: module 'tilelang.cuda._ffi_api' has no attribute 'AnnotateDeviceBoundTmaCopies'
```

**This is not ROCm-specific.** The same tests fail on any build without the CUDA backend, including Metal and CPU-only builds. It surfaced on ROCm only because that is where we were running.

This PR adds `@tilelang.testing.requires_cuda` to the affected tests, matching the convention already used in ~527 places (decorator placed above `@pytest.mark.parametrize`, consistent with existing usage).

## Portability fixes (keep coverage instead of skipping)

Three cases were genuinely backend-dependent rather than CUDA-only, so they are made target-aware rather than gated:

- **`test_tilelang_transform_layout_inference.py`** — CUDA emits `__launch_bounds__(N, 1)`, HIP emits `__launch_bounds__(N)`. A small helper matches both.
- **`test_tilelang_transform_pipeline_planning.py`** — `PipelinePlanning` deliberately strips pipeline annotations on ROCm targets other than gfx950 (see the `TargetIsRocm`/`TargetIsGfx950` guard in `src/transform/pipeline_planning.cc`), so the test now asserts the documented sequential fallback there.
- **`test_tilelang_kernel_gemm.py`** — the two `f32f32f32` cases use `T.tfloat32`, an NVIDIA-only dtype with no CDNA lowering, so `requires_cuda_or_cdna` is narrowed to `requires_cuda`.

## Stale RDNA tests

Three tests in `test_tilelang_rocm_target.py` still assert that gfx1200 is rejected, but gfx12 is now supported (`get_arch` accepts generations `(11, 12)`). They cannot reach the guard any more, because `target_is_rdna` only ever reports gfx11/gfx12 — gfx1030 routes to CDNA. They now force an unsupported generation via `monkeypatch`, which is the only remaining way to exercise that path.

`test_rdna_device_model_rejects_gfx12_before_device_probe` is renamed to `..._rejects_unsupported_generation_...`, since the old name asserts something that is no longer true.

## Expected impact on the existing CUDA CI: none

Every gate added here is the *weakest sufficient* one. `requires_cuda` is satisfied on any machine with a CUDA device, so nothing that runs on the current NVIDIA runner today stops running.

This was deliberate for the `sm_100a` cast-rounding tests in particular: they are compile-only (`enable_device_compile=False`) and need no Blackwell hardware, so `requires_cuda` keeps them running. Gating them on `requires_cuda_compute_version_ge(10, 0)` would have silently dropped them from CI.

I could only verify directly on ROCm hardware, so this PR's own CI run is the confirmation for the CUDA and Metal legs.

## Test plan

Verified on 8x MI300X (gfx942), ROCm 7.2 container, torch `2.14.0.dev+rocm7.2`, running the exact `ci.yml` invocation:

- [x] Before: 105 failures across `testing/python` (63 in the default set, 42 in `transform`)
- [x] After: **1791 passed, 1222 skipped, 0 failed** in ~4m10s
- [x] No loss of passing tests — the passing count went *up*, because some failures were collateral from crashed xdist workers
- [x] Reproduced identically across repeated runs (no flakiness)
- [x] `pre-commit run --all-files` green
- [x] End-to-end under the real GitHub Actions harness on a self-hosted gfx942 runner: https://github.com/andyluo7/tilelang/actions/runs/30876119186

Part of the work to re-enable ROCm CI (#2844). A follow-up PR re-enables the matrix entry once a runner is registered.

## Summary

- Gated CUDA-specific tests with `tilelang.testing.requires_cuda`.
- Covered TMA, LDGSTG, mbarrier, warp specialization, tcgen05, tfloat32, and NVIDIA-specific tooling.
- Added CUDA-only requirements to tfloat32 GEMM tests.
- Updated launch-bounds checks for CUDA and HIP syntax.
- Added ROCm software-pipelining fallback behavior.
- Updated RDNA tests for supported gfx11 and gfx12 generations.
- ROCm results improved to 1,791 passed, 1,222 skipped, and 0 failed.
- Pre-commit checks passed.

## C++ style / lint notes

- This PR does not change C++ code, C++ style documentation, CI, or lint tooling.
- The C++ API Style Audit is not affected.

### DIFF
--- a/testing/python/components/test_storage_rewrite_detect_inplace.py
+++ b/testing/python/components/test_storage_rewrite_detect_inplace.py
@@ -51,6 +51,7 @@ def _get_device_kernel_script(detect_inplace: bool) -> str:
     return artifact.kernel_source
 
 
+@tilelang.testing.requires_cuda
 def test_storage_rewrite_detect_inplace_toggle():
     script_off = _get_device_kernel_script(detect_inplace=False)
     script_on = _get_device_kernel_script(detect_inplace=True)
--- a/testing/python/issue/test_tilelang_issue_2123.py
+++ b/testing/python/issue/test_tilelang_issue_2123.py
@@ -60,6 +60,7 @@ def test_issue_2123_atomic_load_lower_access_ptr_direct():
     _assert_access_ptr_lowered(lowered)
 
 
+@tilelang.testing.requires_cuda
 def test_issue_2123_atomic_load_lower_access_ptr_pipeline():
     target = tvm.target.Target("cuda", host="llvm")
     func = issue_2123_atomic_load_repro(4).with_attr("global_symbol", "main")
--- a/testing/python/issue/test_tilelang_issue_2529.py
+++ b/testing/python/issue/test_tilelang_issue_2529.py
@@ -33,6 +33,7 @@ def _lower_source(n, block, annotate_linear=False):
     return artifact.kernel_source
 
 
+@tilelang.testing.requires_cuda
 def test_tma_store_selection_uses_copy_bounds():
     full_tile_source = _lower_source(1024, 256)
     # The unannotated rank-1 tail exercises descriptor layout inference.
--- a/testing/python/issue/test_tilelang_issue_2554.py
+++ b/testing/python/issue/test_tilelang_issue_2554.py
@@ -5,6 +5,7 @@ import tilelang.language as T
 import tilelang.testing
 
 
+@tilelang.testing.requires_cuda
 def test_runtime_unknown_sign_vector_negative_index_load():
     @T.prim_func
     def main(A: T.Tensor((1024,), T.float32), B: T.Tensor((4, 4), T.float32)):
@@ -30,6 +31,7 @@ def test_runtime_unknown_sign_vector_negative_index_load():
     torch.testing.assert_close(b, expected)
 
 
+@tilelang.testing.requires_cuda
 def test_runtime_unknown_sign_vector_negative_index_store():
     @T.prim_func
     def main(B: T.Tensor((4, 4), T.float32), A: T.Tensor((1024,), T.float32)):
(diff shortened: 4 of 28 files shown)


Rachmanino:
Really appreciate that. Please let me know whenever you wanna kick off the CI workflow.

andyluo7:
Thanks @Rachmanino — please go ahead and kick off the CI workflow whenever convenient. The CUDA and Metal legs are the parts I couldn't verify locally (I only have gfx942 hardware), so that run is the real check on this PR.

On the two CodeRabbit comments — I'd push back on both:

**`test_tma_store_partial_tile_runtime`** already gets the mark transitively: `requires_cuda_compute_version` calls `*requires_cuda.marks()` internally and its docstring says "This also marks the test as requiring a cuda support." It already skips correctly on ROCm today, so adding the decorator would be a no-op.

**`test_non_cuda_target_skip`** — the name is misleading. It uses a CPU *target*, but it calls `tl.cuda.transform.LowerLDGSTG()`, a CUDA-backend pass that isn't compiled when `USE_CUDA=OFF`. It's one of the tests that fails with `AttributeError: module 'tilelang.cuda._ffi_api' has no attribute 'LowerLDGSTG'`, so removing the gate would reintroduce that failure.

That said, it highlights something real: these tests need the CUDA backend *built*, not a CUDA *device*. `requires_cuda` is a proxy. If you'd prefer a precise `requires_cuda_backend` helper (checking symbol availability rather than device presence), I'm happy to add it — it'd be more accurate for the ~40 compile-only tests here, but it's new public API so I didn't want to introduce it unilaterally.

LeiWang1999 (approved):

PR #2866, opened 2026-08-04
[Debug][TIR] Drive Pass Visualizer with PassInstrument

LeiWang1999:
## Summary

- Replace the Pass Visualizer duplicate CUDA prologue list with a TVM `PassInstrument` attached to the canonical prologue.
- Keep structure-tree reports synchronized with target-dependent and pass-config-dependent compilation behavior.

## Changes

- Capture before/after structure trees for each top-level pass while folding nested implementation passes into their parent stage.
- Run semantic checks and the canonical CUDA prologue with normalized PrimFunc and JIT pass configurations from the analyzed kernel.
- Add stable `PassInfo` names for Python-defined passes visible in the instrumented report.
- Remove the fixed `build_pass_stages` helper and update the public exports, documentation, and regression coverage.
- Cover real pipeline observation, conditional pass configuration, nested callbacks, failure cleanup, and HTML/text emission.

## Validation

- `./format.sh`
- `python -m pytest testing/python/debug/test_pass_visualizer.py testing/python/transform/test_tilelang_transform_decouple_type_cast.py -x -q` (26 passed, 1 skipped)
- `python -m pytest testing/python/debug/test_lower_trace.py testing/python/debug/test_pass_diff.py -x -q` (76 passed)

## Notes

- The viewer remains focused on `CUDAPassPipelineBodyPrologue`; full-pipeline tracing remains available through `TL_LOWER_TRACE`.

## Summary

- Refactored the Pass Visualizer to use `StructureTreePassInstrument` on the canonical `CUDAPassPipelineBodyPrologue`.
- Captured before/after structure trees for top-level passes.
- Grouped nested implementation passes under their parent stages.
- Applied normalized JIT and `PrimFunc` pass configurations from the analyzed kernel.
- Added stable names for Python-defined passes.
- Removed `build_pass_stages`.
- Added `capture_structure`, `PassStructureRecord`, and `StructureTreePassInstrument` exports.
- Added explicit names for several TileLang passes.
- Updated documentation and CLI descriptions to identify CUDA support and canonical pipeline tracing.
- Added regression coverage for conditional ordering, nested callbacks, change detection, failure cleanup, dynamic passes, and pass configurations.

## Validation

- 26 tests passed, 1 skipped.
- 76 tests passed.
- Formatting and targeted test suites completed successfully.

### DIFF
--- a/tilelang/tools/pass_visualizer/core.py
+++ b/tilelang/tools/pass_visualizer/core.py
@@ -355,14 +355,12 @@ class PassStructureRecord:
 
     name: str
     sequence: int
-    before_ir: str
-    after_ir: str
     before_lines: list[str]
     after_lines: list[str]
 
     @property
     def changed(self) -> bool:
-        return self.before_ir != self.after_ir
+        return self.before_lines != self.after_lines
 
 
 @dataclass
@@ -371,7 +369,6 @@ class _ActivePass:
 
     name: str
     sequence: int | None
-    before_ir: str | None
     before_lines: list[str] | None
 
 
@@ -417,12 +414,11 @@ class StructureTreePassInstrument:
             frame = _ActivePass(
                 name=name,
                 sequence=self._next_sequence,
-                before_ir=str(mod),
                 before_lines=before_lines,
             )
             self._next_sequence += 1
         else:
-            frame = _ActivePass(name=name, sequence=None, before_ir=None, before_lines=None)
+            frame = _ActivePass(name=name, sequence=None, before_lines=None)
         self._stack.append(frame)
 
     def run_after_pass(self, mod: tvm.IRModule, info):
@@ -440,14 +436,11 @@ class StructureTreePassInstrument:
         if frame.sequence is None:
             return
 
-        assert frame.before_ir is not None
         assert frame.before_lines is not None
         self.records.append(
             PassStructureRecord(
                 name=name,
                 sequence=frame.sequence,
-                before_ir=frame.before_ir,
-                after_ir=str(mod),
                 before_lines=frame.before_lines,
                 after_lines=capture_structure(mod),
             )
--- a/tilelang/tools/pass_visualizer/core.py
+++ b/tilelang/tools/pass_visualizer/core.py
@@ -1,9 +1,9 @@
-"""Core helpers for the pass visualizer: load a user TileLang kernel, build the
-CUDA lowering pass pipeline, and render a PrimFunc's SBlock structure tree.
+"""Core helpers for the pass visualizer: load a user TileLang kernel, capture
+real compiler-pass executions, and render a PrimFunc's SBlock structure tree.
 
-This mirrors ``CUDAPassPipelineBodyPrologue`` in ``tilelang/cuda/pipeline.py``,
-running each prologue pass (through LayoutInference and the post-LayoutInference
-lowering passes) so the structure tree can be captured before/after every stage.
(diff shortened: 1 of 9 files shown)


tilelang/tools/pass_visualizer/viewer.py
```diff
@@ -178,8 +177,24 @@ def build_pass_data(path: str, factory: str | None, target: str, kwargs: dict[st
...
+    pass_configs = normalize_pass_configs(pass_configs)
+
+    # Semantic checks are part of the real pre-lower path but are not lowering
+    # stages. Run them under the same config without adding them to the browser.
+    with tvm.transform.PassContext(opt_level=3, config=pass_configs), resolved_target:
+        PreLowerSemanticCheck(mod)
+
+    instrument = M.StructureTreePassInstrument()
+    with tvm.transform.PassContext(opt_level=3, config=pass_configs, instruments=[instrument]), resolved_target:
+        mod = CUDAPassPipelineBodyPrologue(mod, resolved_target)
```
SiriusNEO, line 194: So this tool is CUDA-only for now?

tilelang/tools/pass_visualizer/viewer.py
```diff
@@ -115,11 +117,8 @@ def _highlight(line: str) -> str:
 
 
 def _capture_tree(mod) -> list[str]:
-    """Render core.inspect_structure(mod) into a list of text lines."""
-    buf = io.StringIO()
-    with contextlib.redirect_stdout(buf):
-        M.inspect_structure(mod)
-    return buf.getvalue().splitlines()
+    """Compatibility wrapper around the core structure renderer."""
+    return M.capture_structure(mod)
```
SiriusNEO: What's the point of this wrapper

SiriusNEO (commented):
Generally I think it's good to replace the original impl with the PassInstrument approach. Left some comments

tilelang/tools/pass_visualizer/core.py
```diff
@@ -396,6 +341,123 @@ def inspect_structure(mod: tvm.IRModule) -> None:
...
+        if is_top_level:
+            before_lines = capture_structure(mod)
+            if self.input_lines is None:
+                self.input_lines = before_lines
+            frame = _ActivePass(
+                name=name,
+                sequence=self._next_sequence,
+                before_ir=str(mod),
+                before_lines=before_lines,
+            )
```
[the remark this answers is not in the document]
LeiWang1999: Fixed in 59c38ed4. `StructureTreePassInstrument` now stores only the captured structure-tree lines, and `PassStructureRecord.changed` compares those same snapshots. This removes the full-TVMScript copies and keeps the badge aligned with the diff shown by the viewer. I also added a regression test for the structure-based `changed` result.

tilelang/tools/pass_visualizer/viewer.py
LeiWang1999: Yes, the viewer is CUDA-only for now, which preserves the scope of the pre-refactor implementation. I made that boundary explicit in the README/tutorial and CLI help, and added a clear error for non-CUDA targets. Backend-specific pipeline dispatch can be added separately when another backend is supported.

tilelang/tools/pass_visualizer/viewer.py
LeiWang1999: It was only a compatibility forwarding layer left over after moving structure capture into `core.py`. I removed the wrapper, call `M.capture_structure` directly, and updated the test to cover the core API.

SiriusNEO (approved):

Issue #2877, opened 2026-08-05
[Feature Request] tcgen05.ld/st support for M=64 TMEM fragment layouts (half-subpartition tiles)


Rachmanino:
## Summary

`tcgen05.ld`/`tcgen05.st` cannot be lowered for **M=64** TMEM fragments.
`ExpandTcgen05Layout` assumes every TMEM tile covers all 128 datapaths
contiguously, in two places:

```cpp
// src/layout/tcgen05_layout.cc -- the maximal contiguous serialized run
cute::Layout inv_prefix = cute::RightInverse(serialized_tmem_tile);
int64_t chunk = cute::AsConst(cute::Size(inv_prefix));
if (chunk % 128 != 0 || size % chunk != 0)
  return Tcgen05CopyPlan(nullptr);            // (a) Layout F is rejected here

// src/layout/tcgen05_layout.cc:208 -- 16-DP atoms are always widened to 32
int64_t ndup = 32 / Tcgen05AtomDatapaths(meta);   // (b)
```

PTX **Layout F** — the half-subpartition fragment that `make_tmem_frg_a` emits for
a 1SM M=64 operand — occupies only the **low 16 datapaths of each 32-DP
sub-partition**, so its contiguous serialized runs are 16 datapaths long. That
trips `chunk % 128 != 0` at (a), before `ndup` at (b) is even consulted, and the
copy fails with *"Failed to find a suitable instruction for tcgen05.st. Check your
layout."* Supporting it needs both gates relaxed plus a genuinely non-duplicated
16-DP issue.

## Repro

```python
import tilelang, tilelang.language as T
from tilelang.cuda.intrinsics import make_mma_swizzle_layout

def gemm_frag_detour(M, N, K):
    @T.prim_func
    def main(A: T.Tensor((M, K), T.bfloat16), B: T.Tensor((N, K), T.bfloat16),
             C: T.Tensor((M, N), T.float32)):
        with T.Kernel(1, threads=128) as bx:
            a_s = T.alloc_shared((M, K), T.bfloat16)
            b_s = T.alloc_shared((N, K), T.bfloat16)
            T.annotate_layout({a_s: make_mma_swizzle_layout(a_s)})
            T.copy(A, a_s); T.copy(B, b_s)
            with T.sblock():
                T.reads(); T.writes()
                a_t = T.alloc_tmem((M, K), T.bfloat16)
                c_t = T.alloc_tmem((M, N), T.float32)
[...]

PR #2880, opened 2026-08-05
[Layout] Support half-subpartition (M=64) TMEM tiles in tcgen05.ld/st

Rachmanino:
A PTX Layout F fragment -- the 1SM M=64 shape make_tmem_frg_a/c selects -- occupies only the LOW 16 datapaths of each 32-datapath sub-partition, so its four warp groups sit 32 datapaths apart with nothing in between. ExpandTcgen05Layout assumed every tile covers 128 contiguous datapaths, so no instruction could be planned for one: an M=64 TS GEMM could not get its A operand into TMEM at all, dying with "Failed to find a suitable instruction for tcgen05.st". 
Tracks #2877.

No new PTX and no new atoms. The 32dp wrappers already issue the 16x{64,128,256}b shapes twice, once per half, so this adds wrappers that stop after the first issue -- and whether a tile wants one form or the other is a property of the TILE, so the planner derives it rather than the metas carrying a flag: right_inverse's maximal contiguous run is 128 when the fragment fills every datapath and exactly the atom's own extent when it is Layout F. One meta per width still covers both, and the plan reports the datapaths one warp ends up covering -- the number the wrapper is named after, which codegen prints straight into tcgen05_{ld,st}_NNdp*Nx.

### Solution

Replication is one explicit stamp -- what a single wrapper call covers -- rather than a blocked product whose complement happens to enumerate the gaps in the order the wrapper wants them. That coincidence holds only for a dense tile: for Layout F the complement's fastest gap is the unoccupied datapaths above the atom, so it would pack the warps in at stride 16 instead of one sub-partition apart. Spelling the stamp out costs a few lines and serves both.

Mapping it back to logical indices takes two inverses, because two kinds of tile arrive. inv_prefix spans the maximal contiguous chunk -- one whole issue for a full-datapath tile, and valid even where the tile is not injective, as a pack::16b tile is not. A Layout F issue reaches four datapath groups 32 apart so no contiguous prefix covers it, but such a tile IS injective, and left_inverse is then a genuine physical -> logical map with no contiguity requirement. Either way the placement is derived from the tile, not assumed of it.

Which datapaths of a sub-partition a warp covers is read off the tile too -- the run along the datapath axis alone, clamped to the 32 a warp owns -- so an atom too wide for it (32dp32b against Layout F) falls out as a non-division rather than needing a case of its own, and the answer does not depend on whether the tile's columns count values or b32 slots.

### Two things a 16-bit Layout F operand gets wrong if taken from the old code

`pack::16b` was judged by "does the fragment cover its codomain". Layout F leaves gaps on the datapath axis, not inside its columns, and was misread as half-filled -- emitting the modifier for a tile whose columns are full. TmemFragmentNeedsPack16b now asks the intended question directly: is every column of every occupied datapath full?

A partial sub-partition cannot be planned in value columns. Only the width-1 32dp32b atom is insensitive to the unit; the wider atoms pick their b32 column with a LANE bit, so stepping their columns by a value column splits each register's pair across two columns, transposing K with the column's low bit and silently permuting the A operand as [0,2,1,3] within every group of four. The stamp therefore steps the atom by a whole b32 column and carries the packed partner as its own fastest value axis. Only how many tile columns share a b32 column has to be passed in: a value-column layout cannot say that about itself.

Verified on B200: an M=64 TS GEMM with A staged through registers into TMEM now matches torch for K in {64,128,256,512} (and M=128 is unchanged); recovering the operand the MMA actually read is the identity, where it was the permutation above. Layout F register->TMEM->register round trips are bit-exact for fp32 and packed bf16 over one and two warpgroups; the fragments are proven against the CUTLASS Copy_Traits DstLayout formulas symbolically, for all lanes/registers/repetitions/warps at once; full-datapath tiles still lower to the identical 32dp instruction.

## Summary

- Add half-subpartition (`M=64`) TMEM tile support for `tcgen05.ld/st`.
- Derive Layout F expansion, packing detection, datapath coverage, replication, issue planning, and inverse mappings from tile layouts.
- Add 16-datapath load and store wrappers for 64-, 128-, and 256-bit transfers.
- Add synchronization fences and correct chained-segment column advancement.
- Preserve full-datapath lowering and reject unsupported configurations.
- Keep legacy six-argument `tcgen05.ld/st` calls on 32-datapath behavior.

## Validation

- Add M=64 TS GEMM tests for K=64, 128, 256, and 512.
- Validate M=128 lowering, Layout F register-to-TMEM round trips, packed-pair handling, and CUTLASS `Copy_Traits` mappings.
- Test invalid atom, chunk, repetition, issue, and datapath configurations.
- Validate CuTeDSL TMEM copy and GEMM correctness.

## C++ style / lint notes

- The PR changes C++ APIs but does not change `docs/developer_guide/cpp_style.md`.
- Treat TLCPP003/TLCPP004 findings as advisory unless they indicate a clear API, FFI, or maintainability risk.

### DIFF
--- a/src/layout/tcgen05_layout.cc
+++ b/src/layout/tcgen05_layout.cc
@@ -2,21 +2,17 @@
  * \file layout/tcgen05_layout.cc
  * \brief tcgen05.ld/st data-movement shapes as CuTe TV atoms.
  *
- * Each atom below is the CUTLASS ``Copy_Traits<SM100_TMEM_LOAD_*>`` TV
- * layout (cute/atom/copy_traits_sm100.hpp) written over (datapath, b32
- * column) coordinates, cross-checked against the PTX data-movement-shape
- * figures
+ * Each atom is the CUTLASS ``Copy_Traits<SM100_TMEM_LOAD_*>`` TV layout
+ * over (datapath, b32 column), matching the PTX data-movement shapes
  * (https://docs.nvidia.com/cuda/parallel-thread-execution/#tcgen05-memory-layout).
- * Loads and stores share one data-movement shape per width.
+ * Loads and stores share one shape per width.
  */
 
 #include "support/check.h"
 #include <algorithm>
-#include <array>
 #include <tvm/ffi/reflection/registry.h>
 #include <tvm/tirx/op.h>
 #include <tvm/tirx/stmt_functor.h>
-#include <vector>
 
 #include "layout.h"
 #include "tcgen05_layout.h"
@@ -27,6 +23,9 @@ namespace tl {
 using namespace tirx;
 using tvm::ffi::Array;
 
+const cute::Layout kDpOnly = cute::Layout::Parse("(1,1):(1,0)");
+const cute::Layout kColOnly = cute::Layout::Parse("(1,1):(0,1)");
+
 namespace {
 
 IterVar MakeIterVar(std::string name, Range dom) {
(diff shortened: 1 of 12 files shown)


src/layout/tcgen05_layout.cc
```diff
@@ -151,41 +153,62 @@ void Tcgen05CopyPlanNode::RegisterReflection() {
...
-  // right_inverse's stride chain stops at the tile's first serialized gap,
-  // so its size is the maximal contiguous chunk = one tcgen05 issue.
-  // inv_prefix: serialized chunk -> flat logical tile
-  // E.g., inv_prefix = 8192:3, chunk = 8192, num_issues = 3
+  // A full-datapath issue is exactly the tile's maximal contiguous run, and
+  // the divide's rest mode iterates the issues after it.  A Layout F issue is
+  // not contiguous at all -- four datapath groups 32 apart -- so it is
+  // assembled explicitly below and always covers the tile in one go.
+  // inv_prefix: serialized chunk -> flat logical tile.  E.g. 8192:3.
   cute::Layout inv_prefix = cute::RightInverse(serialized_tmem_tile);
```
Yongqi-Zhuo: When Layout F is the case, this will not work. E.g., you have `tmem_tile = ((16,4),col):((1@0,32@0),1@1)`, and here the dp mode is serialized with stride 1, yielding `seralized_tmem_tile = ((16,4),col):((1,32),128)`. And you immediately see the problem: there is a gap `2:16` here. And calling `right_inverse` can only give you `16:1`, which ignores `col` completely, and is not the number of elements per issue that what you want.

So here, I recommend you to perform a larger refactor. You can compute only the projection on the columns, i.e., `composition(kColOnly, tmem_tile) = ((16,4),col):((0,0),1)`, and calling `right_inverse` on it will give you the number of columns per issue.

Yongqi-Zhuo (dismissed):
Good feature addition; some refactors can be done in `tcgen05_layout.cc`.

src/layout/tcgen05_layout.cc
Rachmanino: Thanks — adopted, and it turned out cleaner than a local fix. cols_per_issue now comes from right_inverse(composition(kColOnly, tmem_tile)) exactly as you suggested: Layout F's gaps live only on the datapath axis, so the column projection is gap-free and the derivation needs no case split. The datapath footprint is collected mode-wise, and the two sides join into one chunk that also serves as the rest-domain tiler and the dense physical→logical map, so the partial_subpartition ? size : ... special case and the meaningless right_inverse(serialized_tile) are both gone.

Rachmanino:
@Yongqi-Zhuo Thanks for constructive suggestions. I have refactored the code based on your advice and made some adaptation in CuTeDSL backend since then. Plz take a look again

Yongqi-Zhuo (approved):

Yongqi-Zhuo:
LGTM cc @LeiWang1999

PR #2874, opened 2026-08-04
[CI] Re-enable ROCm CI on a gfx942 runner

andyluo7:
> **Draft — blocked on #2872.** Kept in draft deliberately: the `tests` job is
> skipped for draft PRs, so CI stays quiet until the blocker lands. Opening it
> now so it can be reviewed alongside #2872 rather than a day later.

## Summary

Reverts the temporary disable from #2241. An 8x MI300X (gfx942) runner is registered on this repository and runs the ROCm 7.2 toolchain in a container, so the existing `Nightly-ROCm-7.2` toolkit value applies unchanged.

Three changes, 7 lines total:

**1. Re-enable the matrix entry.**

**2. Drop `--ignore=./python/runtime`** — that directory passes (1 passed, 2 skipped). The ignore was stale.

**3. Drop `--ignore=./python/transform`** — its 42 failures were CUDA-only tests missing a `requires_cuda` gate, fixed in #2863. Ignoring the directory also discarded **215 passing tests**.

## On the runner label

The entry uses `[self-hosted, amd, gfx942]` rather than the previous `[self-hosted, amd, gpu]`.

A single `gpu` pool cannot distinguish CDNA generations, and several behaviours legitimately differ between them:

| Behaviour | gfx942 (CDNA3) | gfx950 (CDNA4) |
|---|---|---|
| Software pipelining | stripped (sequential fallback) | enabled |
| fp8 `e4m3` encoding | FNUZ | OCP |
| `__launch_bounds__` | `(N)` | `(N)` — differs from CUDA's `(N, 1)` |

With one shared label, results would vary depending on which machine picked up the job. An arch-specific label keeps that deterministic and leaves room for a gfx950 entry later.

**This is not load-bearing** — the runner is registered with `self-hosted, amd, gpu, gfx942`, so it matches either naming. Happy to keep `gpu` if you prefer; just say so and I'll change it.

## Blocked on #2872

#2872 fixes a ROCm allreduce correctness bug (`tl.sync_warp` lowers to nothing on HIP, so #2865's fix for #2628 never applied there). Without it the ROCm leg fails on 3 tests and `--maxfail=3` aborts the run.

**Merge order: #2872, then this.** I'll rebase, re-verify on the runner, and mark ready for review once that lands.

## Test plan

Verified on the registered runner (8x MI300X gfx942, ROCm 7.2, torch `2.14.0.dev+rocm7.2`) with the exact post-change invocation — no `--ignore` flags, `--maxfail=3`, `-n8`:

[...]

PR #2843, opened 2026-08-02
[BugFix] Handle vectorized SelectNode in codegen_cuda

jjppp:
Implement cuda codegen for vectorized SelectNode by overriding `visitExpr_(const SelectNode *op, std::ostream &os)` for `CodeGenTileLangCUDA`.
This correctly handles the case when the condition expression in a SelectNode is vectorized.

Fixes #2682 

## Summary

- Added CUDA code generation for vectorized `SelectNode` expressions.
- Generated lane-wise conditional selection for vectorized branches and conditions.
- Preserved base `CodeGenC` generation for scalar and non-vector selections.
- Added a regression test for vectorized `T.Select`, fixing issue `#2682`.

## C++ style / lint notes

- The PR changes C++ code and adds a `SelectNode` override.
- The implementation does not change rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit remains warning-only.
- No correctness or build issue is indicated. Advisory TLCPP003/TLCPP004 findings should not block merge unless they identify a clear API, FFI, or maintainability risk.

PR #2895, opened 2026-08-06
[BugFix] Scalarize vectorized bool with more than 4 lanes

### LINKED ISSUE #2206: CUDA codegen LogFatals on `boolx8` from sparse-MLA bwd kernel on B300/sm_103 (TileLang 0.1.9)
### Summary

A `T.alloc_fragment([N], "bool")` in a non-trivial kernel fails to compile on B300 (sm_103) with TileLang 0.1.9 + CUDA 13.1. The vectorize pass produces a `boolx8` IR type and CUDA codegen has no mapping for it, hitting `LogFatal` in `CodeGenTileLangCUDA::PrintType`.

### Environment

| | |
|---|---|
| GPU | NVIDIA B300 SXM6 (sm_103, compute capability 10.3) |
| Driver | 580.126.09 |
| CUDA | 13.1 (nvcc 13.1.r13.1) |
| TileLang | 0.1.9 (`pip install tilelang`) |
| PyTorch | 2.11.0a0+eb65b36914.nv26.02 |
| Base image | `nvcr.io/nvidia/nemo-automodel:26.04.00` |

### Reproducer

The vendored DeepSeek V4 sparse-MLA backward kernel from NVIDIA-NeMo/Automodel PR [#2076](https://github.com/NVIDIA-NeMo/Automodel/pull/2076) (head `8d82c158`) reliably hits this on every shape we tried.

```bash
git clone https://github.com/NVIDIA-NeMo/Automodel.git
cd Automodel && git checkout 8d82c158
pip install -e . --no-deps
pip install tilelang tile_kernels
```

```python
# repro.py
import sys
sys.path.insert(0, '.')
import torch
import tilelang
print("device:", torch.cuda.get_device_name(0))
print("compute capability:", torch.cuda.get_device_capability(0))
print("tilelang:", tilelang.__version__)

from nemo_automodel.components.models.deepseek_v4.kernels.tilelang_sparse_mla_bwd import bwd

# Shape from a real DSv4-Pro 16-layer training step
kernel = bwd(B=1, S=160, S_kv=160, H=8, D=128, topk=64, sm_scale=0.125)
```

### Output

```
device: NVIDIA B300 SXM6 AC
compute capability: (10, 3)
tilelang: 0.1.9
2026-05-14 10:57:34  [TileLang:tilelang.jit.kernel:INFO] (kernel.py:133): TileLang begins to compile kernel `sparse_mqa_bwd_kernel` with `out_idx=[-3]`
[10:57:36] : Fatal: Cannot convert type boolx8 to CUDA type
FAILED: InternalError
Cannot convert type boolx8 to CUDA type
```

### Stack trace (abridged)

```
tvm.error.InternalError: Cannot convert type boolx8 to CUDA type
  File ".../tilelang/engine/lower.py", line 235, in device_codegen
    device_mod = tvm.ffi.get_global_func(global_func)(device_mod, target)
  File "<unknown>", line 0, in tvm::codegen::BuildTileLangCUDA(...)
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::AddFunction(...)
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::VisitStmt_(AllocateNode*)
  ...
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::VisitStmt_(BufferStoreNode*)
  File "<unknown>", line 0, in tvm::codegen::CodeGenC::PrintExpr(PrimExpr&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::VisitExpr_(CastNode*, ostream&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenC::PrintExpr(PrimExpr&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenC::VisitExpr_(NENode*, ostream&)
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::PrintVecBinaryOp(...)
  File "<unknown>", line 0, in tvm::codegen::CodeGenTileLangCUDA::PrintType(DataType, ostream&)
  File "<unknown>", line 0, in tvm::runtime::detail::LogFatal::~LogFatal()
```

The `PrintVecBinaryOp` → `PrintType("boolx8")` chain confirms vectorize has already merged 8 lanes; the failure is purely in the type printer.

### Root cause analysis (best guess)

In `tilelang_sparse_mla_bwd.py:169`:

```python
mask = T.alloc_fragment([BS], "bool")  # BS = 8
...
for bi_i in T.Parallel(BS):
    mask[bi_i] = ValidMask[by, s_i, i_i * BS + bi_i] != 0
...
acc_p[h_i, bi_i] = T.if_then_else(mask[bi_i], ..., ...)
```

The lower pass vectorizes the `Parallel(BS)` loop into a single 8-lane op with element type `bool`, producing IR-level type `boolx8`. CUDA C++ has no `<8 x i1>` vector type (LLVM IR has it; CUDA does not), so `CodeGenTileLangCUDA::PrintType` LogFatals when the type table lookup misses.

We didn't manage to reduce this to a smaller standalone reproducer (a bare `T.alloc_fragment([8], "bool")` + `T.if_then_else` compiles fine). The trigger appears to require the surrounding `T.Pipelined` + serial-and-parallel loops + `T.if_then_else` over the mask in a buffer store. The full vendored kernel is the smallest reliable trigger we found.

We have not personally reproduced this on sm_90 / sm_100. The fact that NVIDIA-NeMo/Automodel PR #2076 ships passing CI runs on H100 (sm_90) suggests the vectorize cost model and/or codegen path differs between targets, with sm_103 hitting a path the codegen doesn't cover.

### Workaround

Change the storage type to `uint8`:

```diff
-mask = T.alloc_fragment([BS], "bool")
+mask = T.alloc_fragment([BS], "uint8")
-mask[bi_i] = ValidMask[by, s_i, i_i * BS + bi_i] != 0
+mask[bi_i] = T.Cast("uint8", ValidMask[by, s_i, i_i * BS + bi_i] != 0)
-valid_count[0] += T.if_then_else(mask[bi_i], 1, 0)
+valid_count[0] += T.if_then_else(mask[bi_i] != 0, 1, 0)
-acc_p[h_i, bi_i] = T.if_then_else(mask[bi_i], ...)
+acc_p[h_i, bi_i] = T.if_then_else(mask[bi_i] != 0, ...)
```

Semantically equivalent (CUDA `bool` is 1 byte; storing 0/1 in `uint8` and comparing `!= 0` produces the same machine code). Verified end-to-end by training DSv4-Pro 16-layer cut for 20 steps on 2× p6-b300.

### Suggested fixes (in TileLang)

Either of these would fix it without users having to patch downstream kernels:

1. **Codegen fallback**: in `CodeGenTileLangCUDA::PrintType`, map `boolx{N}` → `uchar{N}` (and adjust `PrintVecBinaryOp` to emit byte-vector compares writing back to the byte buffer).
2. **Vectorize-pass guard**: when the element type is `bool`, skip vectorize pass for CUDA targets, or rewrite the fragment to `uint8` early in lowering.

Option 1 is more general (handles any future user-written `T.alloc_fragment(..., "bool")`).

### Cross-reference

Downstream report and workaround: [NVIDIA-NeMo/Automodel#2076 (comment)](https://github.com/NVIDIA-NeMo/Automodel/pull/2076#issuecomment-4449637023)

### LINKED ISSUE #2568: [BUG][Fuzzer][ice-on-valid-code] `T.reduce_max`/`reduce_min`/`reduce_bitand` on a `bool` buffer aborts with an internal assert instead of compiling
### Required prerequisites

0.1.11

### System information

```
TileLang: 0.1.11 (tag v0.1.11, commit cd37ed5fc35ae7a60a1277c8eb49028174ac51e6)
Python:   3.12.3
PyTorch:  2.12.1+cu130 (CUDA 13.0)
GPU:      NVIDIA H100 80GB HBM3 (sm_90)
OS:       Linux
```

The defect is in the shared, target-independent reduce identity selection (a host-side C++ pass), so it is **not** Hopper-specific — it surfaces from the `tilelang.compile` call alone, before any device code runs, as a catchable Python `ValueError` (a TVM `Check failed`, not a segfault). Reproduced on the H100 above; arch-independent by inspection of the shared lowerer (not separately run on other GPUs).

### Problem description

`T.reduce_max`, `T.reduce_min`, or `T.reduce_bitand` over a **`bool`** fragment (`dim=1`, a 32×32 tile, `threads=128`) hard-aborts at compile time:

```
ValueError: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
```

The same reduce compiles fine on `int32` and `float32`, and the intended bool reductions — `T.any_of` (logical OR, the bool analogue of `max`) and `T.all_of` (logical AND, the bool analogue of `min`/`bitand`) — compile and run correctly over the same bool data (see the controls in the repro). The magic number `-9223372036854775808` is `INT64_MIN`, the result of converting the float `±INFINITY` reduction identity to `int64` before a bool `{0,1}` assert. This is present since the shared reduce lowerer was introduced (not a regression — see Additional context).

<details>
<summary>Full dtype × reduce-op matrix captured on the H100 (same 32×32 bool tile)</summary>

```
torch 2.12.1+cu130  tilelang 0.1.11  NVIDIA H100 80GB HBM3 (9, 0)
FAIL bool     reduce_max:    Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
FAIL bool     reduce_min:    Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
FAIL bool     reduce_bitand: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
OK   int32    reduce_max
OK   float32  reduce_max
OK   bool     any_of (logical OR ≡ bool max):  0/32 mismatches vs torch
OK   bool     all_of (logical AND ≡ bool min): 0/32 mismatches vs torch
```

So the crash is isolated to the max/min/bitand *identity selection* for bool; int/fp reductions and the intended `any_of`/`all_of` bool reductions are unaffected. (`T.reduce_bitor`/`reduce_bitxor`/`reduce_sum` on a bool buffer fail with a *different* error — `Cannot convert type boolx8 to CUDA type`, a separate bool-vectorization/codegen issue tracked at #2206 — and are out of scope here so they aren't conflated with this identity crash.)
</details>

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import torch

M, N = 32, 32

def reduce_kernel(dtype, func):
    @T.prim_func
    def main(A: T.Tensor((M, N), dtype), Out: T.Tensor((M,), dtype)):
            A_fr = T.alloc_fragment((M, N), dtype)
            B_fr = T.alloc_fragment((M,), dtype)
            T.copy(A, A_fr)
            func(A_fr, B_fr, dim=1)
            T.copy(B_fr, Out)
    return main

# DEFECT: reduce_max / reduce_min / reduce_bitand on a bool buffer crash at COMPILE time.
for name, func in [("reduce_max", T.reduce_max), ("reduce_min", T.reduce_min),
                   ("reduce_bitand", T.reduce_bitand)]:
    try:
        tilelang.compile(reduce_kernel("bool", func), out_idx=[-1], target="cuda")
        print(f"bool {name}: COMPILED ok (unexpected)")
    except Exception as e:
        print(f"bool {name}: CRASH: {type(e).__name__}: {str(e).strip().splitlines()[-1]}")

# CONTROL A: the same reduce on int32 / float32 compiles fine.
for dt in ["int32", "float32"]:
    tilelang.compile(reduce_kernel(dt, T.reduce_max), out_idx=[-1], target="cuda")
    print(f"{dt} reduce_max: COMPILED ok")

# CONTROL B: the intended bool reductions T.any_of (== max) / T.all_of (== min/bitand)
# compile AND run correctly over the same bool data.
def logical_kernel(op):
    @T.prim_func
    def main(A: T.Tensor((M, N), "bool"), Out: T.Tensor((M,), "bool")):
        with T.Kernel(M, threads=128) as bx:
            A_sh = T.alloc_shared((N,), "bool")
            for j in T.Parallel(N):
                A_sh[j] = A[bx, j]
            Out[bx] = op(A_sh)
    return main

torch.manual_seed(0)
a = (torch.rand(M, N, device="cuda") < 0.5)
for nm, op in [("any_of", T.any_of), ("all_of", T.all_of)]:
    k = tilelang.compile(logical_kernel(op), out_idx=[-1], target="cuda")
    out = k(a).cpu()
    ref = a.any(dim=1).cpu() if nm == "any_of" else a.all(dim=1).cpu()
    print(f"bool {nm}: COMPILED+RAN ok, mismatches={(out != ref).sum().item()}/{M}")
```

Observed output on the H100:

```
bool reduce_max: CRASH: ValueError: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
bool reduce_min: CRASH: ValueError: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
bool reduce_bitand: CRASH: ValueError: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
int32 reduce_max: COMPILED ok
float32 reduce_max: COMPILED ok
bool any_of: COMPILED+RAN ok, mismatches=0/32
bool all_of: COMPILED+RAN ok, mismatches=0/32
```

The working control is in the same script: the identical reduce over `int32`/`float32` compiles, and `T.any_of`/`T.all_of` run with 0/32 mismatches vs `torch`. Only the bool `reduce_max`/`reduce_min`/`reduce_bitand` init-value selection crashes — so this is not a harness or front-end artifact.

### Traceback

```pytb
Captured on the H100 for `T.reduce_max` over a bool fragment:


File ".../tilelang/cuda/pipeline.py", ... in CUDAPassPipelineBodyPrologue
    mod = tilelang.transform.LowerTileOp()(mod)
...
tvm::tl::ReduceOpNode::Lower(const LowerArgs&, arith::Analyzer*)
tvm::tl::backend::ReduceLowerer<tvm::tl::cuda::Reduce>::Lower(...)
tvm::tl::backend::reduce::MakeInitValue(const ReduceOpNode&, int)
tvm::PrimExpr tvm::tirx::make_const<float, void>(DataType, float, Span)
tvm::PrimExpr tvm::tirx::MakeConstScalar<float>(DataType, float, Span)
3rdparty/tvm/src/ir/expr.cc, in tvm::IntImm::IntImm(DataType, int64_t, Span)
ValueError: Check failed: (value == 0 || value == 1) is false: -9223372036854775808 exceeds range of bool
```

### Expected behavior

`T.reduce_max` / `T.reduce_min` / `T.reduce_bitand` over a `bool` buffer should compile and produce the well-defined bool result — `reduce_max` ≡ logical OR (`any`), `reduce_min` ≡ logical AND (`all`), `reduce_bitand` ≡ logical AND for bool — bit-identical to what `T.any_of` / `T.all_of` already compute over the same data. At minimum it should be rejected with an actionable diagnostic rather than an internal `IntImm` range assertion leaking `-9223372036854775808`.

Grounded in TileLang's own behavior: the same reduce families compile fine for `int32`/`float32` (the int/uint branches in `MakeInitValue` supply correct identities); the combine ops (`Max`/`Min`/`&`) are well-defined on bool; and the project ships `T.any_of`/`T.all_of` ([`tilelang/language/logical.py#L12-L87`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/logical.py#L12-L87)), which are exactly these bool results and run with 0 mismatches in the control — so only the init value is missing a bool case.

### Additional context

**Root cause.** The shared GPU reduce lowering computes the reduction identity in `reduce::MakeInitValue` ([`src/backend/common/op/reduce.h#L103-L148`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/backend/common/op/reduce.h#L103-L148)). It branches only on `dst_dtype.is_int()` / `dst_dtype.is_uint()` and otherwise assumes the float path. In this TVM fork `bool` is its own type code (`is_bool()` is `code() == kBool`, [`data_type.h#L150`](https://github.com/tile-ai/tvm/blob/ec7f7bd92c185d8c0b84a5dc17709a0860aa5714/include/tvm/runtime/data_type.h#L150)), distinct from `is_int()`/`is_uint()` ([`data_type.h#L210-L212`](https://github.com/tile-ai/tvm/blob/ec7f7bd92c185d8c0b84a5dc17709a0860aa5714/include/tvm/runtime/data_type.h#L210-L212)), so for a bool destination both are false and each branch falls into the float `else`: `isMax` → `make_const(bool, -INFINITY)` ([`reduce.h#L112-L118`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/backend/common/op/reduce.h#L112-L119)), `isMin` → `make_const(bool, INFINITY)` ([`#L120-L126`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/backend/common/op/reduce.h#L120-L127)), `isBitAnd` → `make_const(bool, -INFINITY)` ([`#L130-L136`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/backend/common/op/reduce.h#L130-L137)). Each is `make_const<float>(bool, ±INFINITY)`; the infinity converts to `int64` (`INT64_MIN`) and fails the bool `{0,1}` assert in `IntImm` ([`expr.cc#L67-L69`](https://github.com/tile-ai/tvm/blob/ec7f7bd92c185d8c0b84a5dc17709a0860aa5714/src/ir/expr.cc#L67-L69)). There is no bool guard on the path — not at the surface entry ([`reduce_op.py#L253-L266`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/reduce_op.py#L253-L266)), not in `ReduceOpNode::Lower` ([`src/op/reduce.cc#L189-L191`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/op/reduce.cc#L189-L191)), nor in `MakeInitValue`. Separately, the `isBitAnd` bool identity is also semantically wrong even aside from the crash: bitwise-AND's identity is all-ones (`true = 1`), not `-INFINITY`.

**Suggested fix.** Add an explicit `bool is_bool = dst_dtype.is_bool();` gate in `MakeInitValue` with bool-domain identities mirroring `any_of`/`all_of`: `isMax` → `0` (any_of identity), `isMin` and `isBitAnd` → `1` (all_of identity). The combine ops in `MakeReduce`/`MakeUpdate` already operate correctly on bool (`Max`/`Min`/`&`), so fixing only the init value makes `reduce_max` ≡ `any_of` and `reduce_min`/`reduce_bitand` ≡ `all_of`, without any NaN/INF const reaching `IntImm`; it also fixes the wrong bitand identity. A good regression test asserts each bool reduce equals the corresponding `any_of`/`all_of` over random bool tiles.

<details>
<summary>Fix status — the value-fix removes this crash but exposes the separate #2206 codegen gap; a clean-reject alternative flips cleanly</summary>

Verified: the `is_bool` value-fix above **does eliminate the documented `-9223372036854775808` `IntImm` range assert** — but a clean compile is then blocked by the *separate* `Cannot convert type boolx8 to CUDA type` codegen defect (#2206), which the baseline crash was masking. So the value-fix turns this crash into that different, already-tracked codegen crash rather than a clean compile.

If a full bool-reduce compile path (past #2206) isn't in scope yet, the minimal-risk alternative is a clean diagnostic at the reduce surface or in `MakeInitValue`, e.g.

```cpp
ICHECK(!op.dst->dtype.is_bool())
    << "tl.reduce " << op.type->type << " is not supported for bool buffers; "
    << "use T.any_of (max) / T.all_of (min/bitand) instead.";
```

which I verified replaces the internal `IntImm` assert with that actionable message; `int32`/`float32` controls still compile. The value-fix is the correct end state once #2206's bool codegen is closed; until then the clean diagnostic avoids leaking `INT64_MIN`.
</details>

**Provenance.** Present since the shared GPU tile-op lowerers were consolidated into `src/backend/common/op/reduce.h` in [#2163](https://github.com/tile-ai/tilelang/pull/2163) (merged 2026-05-07); `MakeInitValue` has never had a bool branch (`is_bool` appears 0 times in the file at the pinned commit). **Not a regression** — the bool `reduce_max`/`min`/`bitand` path has crashed since it could first be expressed; the missing identity predates the refactor in the pre-share location. Reproduced fresh on an H100 (sm_90), with int/fp and `any_of`/`all_of` controls.

**Distinct from existing reports.** I searched the open and closed tracker and found no report of the bool reduce identity-selection crash in `MakeInitValue`. Nearest neighbors are different mechanisms: the same op family (`reduce_bitand`) on a **float** buffer fails later, in the integer `&`-fold in `MakeReduce` (an unsupported-LHS `InternalError`), not here in the init value; the `boolx8` bool codegen gap (#2206) is the separate out-of-scope failure noted above; and the reduce layout/thread-count crashes (keep-dim rank clash, non-power-of-two width) are shape/thread-driven, not dtype-driven.

jjppp:
* Scalarize vectorized bool with more than 4 lanes to avoid ICE during downstream codegen
* Use dedicated init values for boolean reductions

Fixes #2568, fixes #2206 

## Summary

- Scalarize vectorized boolean values with more than four lanes to prevent CUDA codegen failures.
- Use boolean identity values for reductions:
  - `0` for `max` and logical OR.
  - `1` for `min`, bitwise AND, and logical AND.
- Preserve existing initialization behavior for integer and floating-point reductions.
- Add CUDA regression tests for boolean reductions and coverage for `int32` and `float32` reductions.

## C++ style / lint notes

- The PR changes C++ code in `reduce.h` and `vectorize_loop.cc`.
- No changes to the rules documented in `docs/developer_guide/cpp_style.md` are indicated.
- The “C++ API Style Audit (warning only)” CI step is relevant if configured for these files.
- No correctness or build issue is reported from style warnings. Advisory TLCPP003/TLCPP004 findings should not block merge unless they indicate a new API, FFI, or maintainability risk.

### DIFF
--- a/src/backend/common/op/reduce.h
+++ b/src/backend/common/op/reduce.h
@@ -206,8 +206,9 @@ inline void CheckAllReduceWidth(int reducing_threads, int scale,
 
 inline PrimExpr MakeInitValue(const ReduceOpNode &op, int vsize = 1) {
   auto dst_dtype = op.dst->dtype;
-  auto is_int = dst_dtype.is_int();
+  bool is_int = dst_dtype.is_int();
   bool is_uint = dst_dtype.is_uint();
+  bool is_bool = dst_dtype.is_bool();
   auto bits = dst_dtype.bits();
 
   PrimExpr scalar;
@@ -218,6 +219,8 @@ inline PrimExpr MakeInitValue(const ReduceOpNode &op, int vsize = 1) {
       scalar = make_const(op.dst->dtype, SignedMin(bits));
     } else if (is_uint) {
       scalar = make_const(op.dst->dtype, 0);
+    } else if (is_bool) {
+      scalar = make_const(op.dst->dtype, 0);
     } else {
       scalar = make_const(op.dst->dtype, -INFINITY);
     }
@@ -226,6 +229,8 @@ inline PrimExpr MakeInitValue(const ReduceOpNode &op, int vsize = 1) {
       scalar = make_const(op.dst->dtype, SignedMax(bits));
     } else if (is_uint) {
       scalar = make_const(op.dst->dtype, UnsignedMax(bits));
+    } else if (is_bool) {
+      scalar = make_const(op.dst->dtype, 1);
     } else {
       scalar = make_const(op.dst->dtype, INFINITY);
     }
@@ -236,6 +241,8 @@ inline PrimExpr MakeInitValue(const ReduceOpNode &op, int vsize = 1) {
       scalar = make_const(op.dst->dtype, -1);
     } else if (is_uint) {
       scalar = make_const(op.dst->dtype, UnsignedMax(bits));
+    } else if (is_bool) {
+      scalar = make_const(op.dst->dtype, 1);
     } else {
       scalar = make_const(op.dst->dtype, -INFINITY);
     }
--- a/src/transform/vectorize_loop.cc
+++ b/src/transform/vectorize_loop.cc
@@ -301,7 +301,17 @@ public:
   }
 
   PrimExpr VisitExpr(const PrimExpr &e) final {
-    return ExprFunctor::VisitExpr(e);
+    if (need_scalarize_) {
+      return e;
+    }
+    PrimExpr ret = ExprFunctor::VisitExpr(e);
+    // Vectorized boolean types with more than 4 lanes (e.g., boolx8)
+    // are not representable, scalarize for backend compatibility
+    if (ret.dtype().is_vector_bool() && ret.dtype().lanes() > 4) {
+      need_scalarize_ = true;
+      return e;
+    }
+    return ret;
   }
 
   PrimExpr VisitExpr_(const AddNode *op) final {
--- a/testing/python/issue/test_tilelang_issue_2568.py
+++ b/testing/python/issue/test_tilelang_issue_2568.py
@@ -1,0 +1,87 @@
+"""Regression tests for GitHub issue #2568.
+
+This test asserts that the bool reductions compile, run, and produce results
+identical to the equivalent logical reductions over the same data. The compile
+path also exercises the wide-bool-vector handling that used to abort codegen
+with ``Cannot convert type boolx8 to CUDA type``.
+"""
+
+import pytest
+import torch
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+
+
+M, N = 32, 32
+THREADS = 128
+
+
+def _make_reduce_kernel(dtype, reduce_fn):
+    @T.prim_func
+    def main(A: T.Tensor((M, N), dtype), Out: T.Tensor((M,), dtype)):
+        with T.Kernel(1, 1, threads=THREADS) as (bx, by):
+            A_fr = T.alloc_fragment((M, N), dtype)
+            B_fr = T.alloc_fragment((M,), dtype)
+            T.copy(A, A_fr)
+            reduce_fn(A_fr, B_fr, dim=1)
+            T.copy(B_fr, Out)
+
+    return main
+
+
+def _make_logical_kernel(logical_op):
+    @T.prim_func
+    def main(A: T.Tensor((M, N), "bool"), Out: T.Tensor((M,), "bool")):
+        with T.Kernel(M, threads=THREADS) as bx:
+            A_sh = T.alloc_shared((N,), "bool")
+            for j in T.Parallel(N):
+                A_sh[j] = A[bx, j]
+            Out[bx] = logical_op(A_sh)
+
+    return main
+
+
+def _make_bool_tile():
+    torch.manual_seed(0)
+    a = torch.rand(M, N, device="cuda") < 0.5
+    # Deterministic rows so both outcomes of the logical reductions occur:
+    # row 0 is all-true (any_of/all_of -> True), row 1 is all-false (-> False).
+    a[0] = True
+    a[1] = False
+    return a
+
+
+@pytest.mark.parametrize(
+    ("reduce_fn", "logical_op", "torch_reduce"),
+    [
+        (T.reduce_max, T.any_of, lambda a: a.any(dim=1)),
+        (T.reduce_min, T.all_of, lambda a: a.all(dim=1)),
+        (T.reduce_bitand, T.all_of, lambda a: a.all(dim=1)),
+    ],
+    ids=["reduce_max==any_of", "reduce_min==all_of", "reduce_bitand==all_of"],
+)
+@tilelang.testing.requires_cuda
+def test_bool_reduce_matches_logical_reduction(reduce_fn, logical_op, torch_reduce):
+    a = _make_bool_tile()
+
+    out = tilelang.compile(_make_reduce_kernel("bool", reduce_fn), out_idx=[-1], target="cuda")(a)
+    logical_out = tilelang.compile(_make_logical_kernel(logical_op), out_idx=[-1], target="cuda")(a)
+    torch_ref = torch_reduce(a)
+
+    assert out.dtype == torch.bool
+    assert torch.equal(out, logical_out)
+    assert torch.equal(out, torch_ref)
+
+
+@pytest.mark.parametrize("dtype", ["int32", "float32"])
+@tilelang.testing.requires_cuda
+def test_non_bool_reduce_max_still_compiles(dtype):
+    # Compile-time control from the issue: the int/float reduce path must be
+    # unaffected by the bool identity fix.
+    tilelang.compile(_make_reduce_kernel(dtype, T.reduce_max), out_idx=[-1], target="cuda")
+
+
+if __name__ == "__main__":
+    tilelang.testing.main()


jjppp:
I think there might be multiple (at least two different) ways to support vectorized bools (e.g., `boolx8`).
1. Scalarize for specific lanes
2. Extend codegen to delegate to existing types (e.g., `int8x8`) via bit-packing

For me, the first approach seems most straightforward and simplest to implement.
I'm not a kernel expert, so I'm not certain about its performance impact (though it is not originally supported, ;-P).
Would be glad to hear some thoughts on this.

---

Some notes on the CI failure: since #2874 we have ROCm CI re-enabled, and #2889 fixes an ROCm bug that is missed by #2843

andyluo7:
@jjppp , the failing ROCm check here isn't caused by this PR. test_tilelang_issue_2682.py::test_vectorized_select is currently failing on main for everyone on gfx942 (1 failed, 1830 passed in your run) — it's a HIP codegen gap from 
#2843. #2889 fixes it and is green on the gfx942 runner. Once that merges, please rebase and the AMD leg should go clean.

PR #2898, opened 2026-08-06
[BugFix][CUDA] Lower tl.__tan to tanf for float32

### LINKED ISSUE #2879: [BUG] `T.__tan` on `float32` silently returns inaccurate values
### Required prerequisites

0.1.13

### System information

Python 3.12.3, torch 2.12.1+cu130, CUDA 13.2 (nvcc 13.2.78); On NVIDIA A100 (sm_80).

### Problem description

`T.__tan(x)` on `float32` silently returns values off by orders of magnitude, instead of a bounded fast-math approximation like the rest of the `__` family (`__sinf`/`__cosf`/`__expf` all match torch to ~1e-6). `__tanf`'s error is unbounded near the poles. 200k `randn` points vs `torch.tan`:

| call | max abs err | mean | err > 1e-3 |
|---|---|---|---|
| `__tanf` (current) | **1791.3** | 1.2e-2 | 861 / 200000 |
| `tanf` (intended) | 0 | 0 | 0 |

### Reproducible example code

The Python snippets:

```python
import torch, tilelang
import tilelang.language as T

@tilelang.jit(out_idx=[1])
def tan_kernel(N: int):
    @T.prim_func
    def main(A: T.Tensor((N,), "float32"), B: T.Tensor((N,), "float32")):
            for i in T.Parallel(N):
                B[i] = T.__tan(A[i])   # emits __tanf(...)
    return main

a = torch.randn(1024, device="cuda", dtype=torch.float32)
print((tan_kernel(1024)(a) - torch.tan(a)).abs().max().item())
```

### Traceback

```pytb
None — compiles and runs, results silently wrong and deterministic.
```

### Expected behavior

`T.__tan` on fp32 should satisfy the same accuracy bar the repo's fastmath tests apply to `__exp`/`__log` (`torch.testing.assert_close(..., rtol=1e-3, atol=1e-3)`), or fall back to accurate `tanf` per the project's own documented intent (see below) — not silently return values orders of magnitude off near the poles.

### Additional context

Root cause: `tl.__tan` uses `CUDAFastMath` → `__tanf` (`src/cuda/codegen/codegen_cuda.cc:4618-4622`). The intended fix `CUDAFastMathTan` (fp32 → `tanf`, with the comment "`__tanf` ... too deviant from numpy tan ... use just `tanf`") is **dead code** at `codegen_cuda.cc:175-193` and `intrin_rule_cuda.cc:72-90` — ROCm wires its equivalent (`HIPFastMathTan`, `intrin_rule_hip.cc:223`), CUDA forgot. Same mapping also makes fp16/bf16 `tan` emit non-existent `htan` (only `htanh` exists) → compile failure.

Tests stay green because `run_fastmath_mathop_test` compares `__tan` against `expected = b_fastmath` (self-comparison); only `__exp`/`__log` are validated vs torch.

Dino1844:
## Summary

`T.__tan` on `float32` lowers to the CUDA fast-approximate intrinsic `__tanf`, which is wildly inaccurate near the tan poles (x ≈ ±π/2). This PR dispatches `tl.__tan` through the already-written-but-never-wired `CUDAFastMathTan` so it emits the accurate `tanf` instead.

Closes #2879.

## Problem

`src/cuda/codegen/codegen_cuda.cc` dispatched `tl.__tan` via `CUDAFastMath` → `__tanf`. Measured on A100 (200k points vs `torch.tan`):

| call | max abs err | mean | err > 1e-3 |
|---|---|---|---|
| `__tanf` (current) | **1791.3** | 1.2e-2 | 861 / 200000 |
| `tanf` (accurate) | 0 | 0 | 0 |

The intended fix already existed: `CUDAFastMathTan` (`codegen_cuda.cc:175`) returns `tanf` for fp32 — its comment even says "`__tanf` seems to produce some values too deviant from numpy tan ... use just `tanf` instead" — but it was never referenced. The ROCm backend wires its equivalent (`HIPFastMathTan`); the CUDA path did not.

## Fix

```cpp
} else if (op->op.same_as(tl::__tan())) {
    CUDAFastMath math_func;      // before: fp32 -> "__tanf"
    CUDAFastMathTan math_func;   // after:  fp32 -> "tanf"
```

## Tests

- `testing/python/math/test_math_fast_math.py` / `.../fastmath/test_mathops_fastmath.py`: the fastmath tests now assert `__tan` emits `tanf` (not `__tanf`).
- New `test_tan_lowers_to_accurate_tanf_without_fastmath`: compiles `T.__tan` without `--use_fast_math`, asserts the generated source uses `tanf`, and checks numerical correctness vs `torch.tan` over `linspace(-1.5707, 1.5707)` (rtol/atol 1e-6) — i.e. all the way to the poles, where `tanf` is bit-exact vs torch.

## Scope note

This PR implements the fp32 fix tracked in #2879: `T.__tan` on `float32` now
lowers to the accurate `tanf` instead of the fast approximate `__tanf`. The
fp16/bf16 compile side of the same issue is handled separately in #2894 (the
`htan` wrappers in `common.h`), which I noticed is already in flight, so I
have deliberately scoped this PR to the fp32 half rather than overlapping
with it.

## Ecosystem context

This is the same bug family reported upstream:
- **apache/tvm#19546** (filed by the TileLang maintainer): "Standard TIR math intrinsics should not lower to fast-math `__*f` by default" — its suggested fix table lists `tirx.tan → tanf`, exactly this change.
- **getkeops/keops#333**: `--use_fast_math` silently degrades transcendental accuracy and should not be a default.
- PyTorch has no per-element fastmath; Triton has none at all (triton-lang/triton#1535). The ecosystem norm is "standard math = accurate by default; fast approximation only via explicit opt-in".

## Verification

- Built from source (CUDA 13.2, A100 sm_80); `pytest testing/python/math/test_math_fast_math.py testing/python/fastmath/test_mathops_fastmath.py` → 67 passed.
- Regression guard verified: reverting the codegen change makes the new assertions fail (`__tanf` appears in the generated source).
- ruff check / ruff format / clang-format clean on all touched files.

## Summary

- Lower CUDA `tl.__tan` for `float32` through `CUDAFastMathTan`, which emits accurate `tanf` instead of approximate `__tanf`.
- Update fast-math tests to expect `tanf` for `tan`.
- Add regression coverage for generated CUDA source and numerical accuracy against `torch.tan` near tangent poles.
- Limit the fix to `float32`; `float16` and `bfloat16` handling remains tracked separately.

## C++ style / lint notes

- The C++ change does not alter public APIs or documented rules in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) is relevant when CI runs it. This change introduces no correctness or build issues.

### DIFF
--- a/testing/python/fastmath/test_mathops_fastmath.py
+++ b/testing/python/fastmath/test_mathops_fastmath.py
@@ -228,7 +228,8 @@ def run_fastmath_mathop_test(mathop_name, mathop_func, M=128, N=128, block_M=32,
     print("FAST_MATH=True:")
     # Strip the __ prefix for checking in the CUDA source
     cuda_mathop_name = mathop_name.lstrip("_")
-    check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=True)
+    expect_fastmath = cuda_mathop_name != "tan"
+    check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=expect_fastmath)
 
     # Test numerical correctness
     torch_dtype = dtype.as_torch()
--- a/testing/python/fastmath/test_mathops_fastmath.py
+++ b/testing/python/fastmath/test_mathops_fastmath.py
@@ -228,7 +228,7 @@ def run_fastmath_mathop_test(mathop_name, mathop_func, M=128, N=128, block_M=32,
     print("FAST_MATH=True:")
     # Strip the __ prefix for checking in the CUDA source
     cuda_mathop_name = mathop_name.lstrip("_")
-    expect_fastmath = cuda_mathop_name != "tan"
+    expect_fastmath = cuda_mathop_name != "tan"  # __tan uses accurate tanf, not __tanf
     check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=expect_fastmath)
 
     # Test numerical correctness
--- a/testing/python/math/test_math_fast_math.py
+++ b/testing/python/math/test_math_fast_math.py
@@ -228,7 +228,7 @@ def run_fastmath_mathop_test(mathop_name, mathop_func, M=32, N=32, block_M=32, b
     print("FAST_MATH=True:")
     # Strip the __ prefix for checking in the CUDA source
     cuda_mathop_name = mathop_name.lstrip("_")
-    expect_fastmath = cuda_mathop_name != "tan"
+    expect_fastmath = cuda_mathop_name != "tan"  # __tan uses accurate tanf, not __tanf
     check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=expect_fastmath)
 
     # Test numerical correctness
@@ -361,30 +361,6 @@ def test_fastmath_versions(name, func):
     """Test that __exp, __exp10, __log, __log2, __log10, __tan, __cos, __sin generate fastmath CUDA code"""
     run_fastmath_mathop_test(name, func, dtype=T.float32)
     print(f"✓ {name} test passed")
-
-
-@tilelang.testing.requires_cuda
-def test_tan_lowers_to_accurate_tanf_without_fastmath():
-    n = 4096
-
-    @T.prim_func
-    def main(A: T.Tensor((n,), "float32"), B: T.Tensor((n,), "float32")):
-        with T.Kernel(1, threads=128):
-            for i in T.Parallel(n):
-                B[i] = T.__tan(A[i])
-
-    kernel = tilelang.compile(
-        main,
-        out_idx=[1],
-        target="cuda",
-        pass_configs={tilelang.PassConfigKey.TL_ENABLE_FAST_MATH: False},
-    )
-    source = kernel.get_kernel_source()
-    assert "tanf(" in source, f"expected accurate tanf in source, got:\n{source}"
-    assert "__tanf" not in source, f"__tan must not emit __tanf, got:\n{source}"
-
-    a = torch.linspace(-1.5707, 1.5707, n, device="cuda", dtype=torch.float32)
-    torch.testing.assert_close(kernel(a), torch.tan(a), rtol=1e-6, atol=1e-6)
 
 
 if __name__ == "__main__":
--- a/testing/python/math/test_math_fast_math.py
+++ b/testing/python/math/test_math_fast_math.py
@@ -228,7 +228,8 @@ def run_fastmath_mathop_test(mathop_name, mathop_func, M=32, N=32, block_M=32, b
     print("FAST_MATH=True:")
     # Strip the __ prefix for checking in the CUDA source
     cuda_mathop_name = mathop_name.lstrip("_")
-    check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=True)
+    expect_fastmath = cuda_mathop_name != "tan"
+    check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=expect_fastmath)
 
     # Test numerical correctness
     torch_dtype = dtype.as_torch()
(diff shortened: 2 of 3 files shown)


testing/python/fastmath/test_mathops_fastmath.py
```diff
@@ -228,7 +228,8 @@ def main(
     print("FAST_MATH=True:")
     # Strip the __ prefix for checking in the CUDA source
     cuda_mathop_name = mathop_name.lstrip("_")
-    check_fastmath_usage(source_fastmath, cuda_mathop_name, expect_fastmath=True)
+    expect_fastmath = cuda_mathop_name != "tan"
```
LeiWang1999: an unexpected change, is the right fix to improve check_fastmath_usage?

testing/python/math/test_math_fast_math.py
```diff
@@ -362,5 +363,29 @@ def test_fastmath_versions(name, func):
     print(f"✓ {name} test passed")
 
 
+@tilelang.testing.requires_cuda
+def test_tan_lowers_to_accurate_tanf_without_fastmath():
```
LeiWang1999: would be good for us to improve check_fastmath_usage instead introduce another test.

testing/python/fastmath/test_mathops_fastmath.py
Dino1844: Thanks for the review. I benchmarked all fastmath intrinsics — every one matches the accurate version except `__tanf`, which deviates significantly while accurate `tanf` matches torch exactly. This also aligns with your comment in `CUDAFastMathTan` about `__tanf` deviating from numpy.

So the other ops can keep fastmath, and only tan needs a special case. Is special-casing tan here acceptable, or would you prefer other cleaner way? Happy to adjust the style if you'd prefer.

testing/python/math/test_math_fast_math.py
Dino1844: I will integrate this precision test into the `check_fastmath_usage`. Thanks for your advice.

testing/python/fastmath/test_mathops_fastmath.py
LeiWang1999: special-casing tan in check_fastmath_usage sounds good to me, other cleaner way is also welcome :)

LeiWang1999:
Thanks for the detailed investigation and the well-written PR — the error measurements and the ecosystem survey are genuinely appreciated. After discussion, however, we've decided not to take this change, and I'd like to explain why.

## `T.__tan` is the explicit fast-math opt-in, and should stay aligned with CUDA C

The `__`-prefixed family in TileLang maps 1:1 to CUDA's `__` intrinsics by design: `T.__exp` → `__expf`, `T.__sin` → `__sinf`, and `T.__tan` → `__tanf`. A user who writes `T.__tan` has explicitly asked for the hardware fast path, with exactly the semantics (and documented accuracy caveats) of CUDA C. Making it silently emit the accurate `tanf` would:

- break that contract,
- make `T.__tan` identical to `T.tan` at codegen level — i.e. a dead intrinsic, and
- remove the only way to get per-op fast tan, which is precisely the fine-grained control the `__` family exists to provide.

Users who need accuracy near the poles should use `T.tan`, which already lowers to the accurate `tanf`.

## The cited precedents address the opposite problem

`CUDAFastMathTan` (and ROCm's `HIPFastMathTan`, and apache/tvm#19546) all deal with **standard** `tir.tan`: in old TVM, standard math ops were lowered to fast-math `__*f` **by default**, and `CUDAFastMathTan` was the tan-specific exemption from that policy, because `__tanf`'s pole error broke numpy-reference tests. That's "the user didn't ask for fast math but got it" — the opposite of `T.__tan`, where the user explicitly did. That policy has since been fixed (standard ops now dispatch through `CUDAMath`), which is also why `CUDAFastMathTan` has been dead code here: its reason to exist is gone. Wiring it to the explicit intrinsic inverts the original intent.

## A side note from review

`CUDAFastMathTan` has no `bfloat16` branch (unlike `CUDAMath`), so this change would make `T.__tan` on bf16 emit an empty function name — the generated code degenerates to `(x)`, a silent identity. That's another sign the class wasn't designed for this dispatch site.

## What we'll do instead

The footgun raised in #2879 is legitimate — `__tanf` is the only member of the `__` family with unbounded error, and that deserves to be visible. So:

1. Keep `T.__tan` → `__tanf`, but add a prominent docstring warning: `__tanf` is implemented in hardware as `__sinf(x) * (1 / __cosf(x))` and its error is **unbounded** near the poles (x ≈ ±π/2) — use `T.tan` if you need accuracy there.
2. Add an `ICHECK(!func_name.empty())` guard at the `tl.__tan` dispatch site so unsupported dtypes fail loudly at compile time instead of generating wrong code.

You're very welcome to re-scope this PR to that documentation + guard change if you'd like — otherwise we'll push it as a follow-up. Closing this one, and re-scoping #2879 as a documentation issue. Thanks again for the thorough report — the measurements in it are exactly what made the right call clear.

PR #2902, opened 2026-08-06
[Example] Generalize TCGEN05 GEMMs for Thor (sm110a) and add Stream-K scheduling

xinhao-luo:
## Summary

This PR generalizes the TCGEN05 GEMM examples to support NVIDIA Thor/SM110a while preserving SM100 compatibility, and adds a 2-CTA Stream-K scheduler for workloads with under-filled persistent waves.

## Changes

- Rename `examples/gemm_sm100` to `examples/gemm_tcgen05`, as these kernels rely on TCGEN05 capabilities rather than being exclusive to SM100.
- Generalize the architecture guard used by `warpgroup_reg_alloc` and `warpgroup_reg_dealloc` to support Thor/SM110a targets.
- Add a 2-CTA TCGEN05 Stream-K scheduler for GEMMs with under-filled persistent waves and sufficiently long K dimensions.

The Stream-K implementation splits the under-filled tail wave along the K dimension and distributes its work across the resident clusters. Partial accumulators are stored in a workspace and combined before the epilogue, while the remaining full waves continue to use ordinary data-parallel persistent scheduling.

## Performance

The kernels were tuned independently for each shape on an NVIDIA Thor T5000. The following table reports BF16 GEMM latency in milliseconds; the best result in each row is shown in **bold**.

| M | N | K | cuBLAS | Static Persistent | CLC | Pipelined CLC | Stream-K |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 256 | 2048 | 4096 | **0.2331** | 0.3428 | 0.3312 | 0.3337 | 0.3082 |
| 256 | 2048 | 8192 | **0.2277** | 0.2406 | 0.2372 | 0.2450 | 0.2392 |
| 256 | 2048 | 16384 | 0.5166 | 0.5510 | 0.5131 | 0.5055 | **0.4692** |
| 1024 | 1024 | 1024 | 0.0860 | 0.0889 | **0.0854** | **0.0854** | 0.0908 |
| 2048 | 2048 | 2048 | **0.2451** | 0.2513 | 0.2553 | 0.2610 | 0.2721 |
| 4096 | 4096 | 4096 | 1.4441 | 1.4719 | **1.4054** | 1.4057 | 1.7721 |
| 2560 | 512 | 8192 | 0.7082 | 0.5243 | 0.5349 | 0.5335 | **0.4050** |
| 2560 | 512 | 16384 | 1.2433 | 1.2011 | 1.4068 | 1.3981 | **0.8351** |
| 2560 | 512 | 32768 | 2.2219 | 2.4162 | 2.6746 | 2.6666 | **1.5638** |

The tuning space covered:

- `block_M × block_N`: `128 × 128`, `128 × 256`, and `256 × 128` when valid
- `block_K`: `64` and `128`
- Pipeline stages: `3`, `4`, and `6`
- TMA store enabled and disabled
- Valid swizzle-group sizes
- Pipelined CLC stages: `2`, `3`, and `4`

Stream-K is not universally faster, but it consistently performs best for the tested `2560 × 512 × long-K` workloads.

These shapes contain 20 cluster tiles. With six resident 2-CTA clusters on Thor, ordinary persistent scheduling produces waves of `6 + 6 + 6 + 2`, leaving only two clusters active during the final long-K wave. Stream-K splits the remaining two tiles along K and distributes their work across all six clusters, reducing tail underutilization.

For these long-K cases, Stream-K achieves up to **1.75×** speedup over cuBLAS and **1.44×** over the best non-Stream-K implementation.

## Summary

- Generalizes GEMM examples from SM100 to TCGEN05 while preserving SM100 compatibility.
- Renames the example directory and updates architecture documentation.
- Adds persistent and 2-CTA Stream-K GEMM examples for long-K workloads.
- Broadens `warpgroup_reg_alloc` and `warpgroup_reg_dealloc` architecture guards.
- Adds CLI configuration, validation, benchmarking, and generated-source output across the examples.
- Stream-K stores partial accumulators in workspace buffers and combines them before the epilogue.

## Performance

On NVIDIA Thor T5000, Stream-K achieved up to 1.75× the cuBLAS performance and 1.44× the best non-Stream-K implementation for tested `2560 × 512 ×` long-K workloads. Stream-K does not improve all workloads.

## C++ style / lint notes

- The PR changes `src/tl_templates/cuda/intrin.h`.
- It does not change `docs/developer_guide/cpp_style.md`.
- No C++ API Style Audit (warning only) CI change is reported.
- No correctness, build, or test issue is reported. Advisory style warnings, if any, are non-blocking unless they indicate an API, FFI, or maintainability risk.


src/tl_templates/cuda/intrin.h
```diff
@@ -4,6 +4,13 @@
 #include "cutlass/arch/reg_reconfig.h"
 #include "cutlass/cutlass.h"
 
+#if defined(CUDA_CTA_RECONFIG_ACTIVATED) ||                                    \
+    (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900) &&                       \
+     (defined(__CUDA_ARCH_SPECIFIC__) ||                                       \
```
Rachmanino, line 9: I think the original CUDA_CTA_RECONFIG_ACTIVATED macro is defined internally in CUTLASS. Why we need to introduce such change?

Rachmanino (commented):
Thanks for helping us validate these kernels on sm110!
1. In my view, simply renaming the `gemm_sm100` folder to `gemm_tcgen05`, along with a dedicated gemm_tcgen5mma_ws_persistent_streamk example would be better, or there's too much duplication in `gemm_tcgen05` and `gemm_sm100`
2. plz double check the comment I left in `intrin.h`

src/tl_templates/cuda/intrin.h
xinhao-luo: Thanks for the comment! CUTLASS's `CUDA_CTA_RECONFIG_ACTIVATED` in `reg_reconfig.h` still doesn't cover `sm110a` (I just checked CUTLASS `main`), so `warpgroup_reg_alloc/dealloc` would hit the `static_assert` on `sm110a`. The new guard uses the generic `__CUDA_ARCH_SPECIFIC__` / `__CUDA_ARCH_FAMILY_SPECIFIC__` macros instead, so no CUTLASS change is needed. Alternatively, We can add `sm110a` support to CUTLASS upstream and bump the dependency — happy to do that if you prefer.

xinhao-luo:
@Rachmanino  

Thanks for the review!
1. I've split the Stream-K kernel into a dedicated example: `gemm_tcgen5mma_ws_persistent_streamk.py`.
2. Regarding `intrin.h`, I've replied in the inline thread.

examples/gemm_tcgen05/gemm_tcgen5mma_ws_persistent.py
```diff
@@ -145,8 +146,15 @@ def gemm_persistent_2cta(
...
     accum_dtype,
     num_stages,
+    group_size,
     use_tma_store,
 ):
+    """Persistent 2-CTA TCGEN05 GEMM with warp specialization.
+
+    A 2-CTA cluster splits each output tile in N: each CTA loads half of B and
+    the TCGEN05 MMA accumulates both halves into tensor memory, so the
+    epilogue only runs in one of the two CTAs.
```
Rachmanino: "the epilogue only runs in one of the two CTAs" is not correct. Actually both CTAs in a cluster get half of the epilogue results in TMEM. I personally think that there's no need to modify any docstring in this file.

Rachmanino (changes requested):

src/tl_templates/cuda/intrin.h
Rachmanino: Got it, thanks

examples/gemm_tcgen05/gemm_tcgen5mma_ws_persistent.py
xinhao-luo: Thanks for the correction. You are right. I removed the added docstrings from the existing persistent example and restored its original comment style. I also aligned the new Stream-K example with the same style.

Rachmanino (approved):

Rachmanino:
cc @LeiWang1999

PR #2904, opened 2026-08-06
[Test][CUDA] Gate issue 2883 regression on CUDA

### LINKED ISSUE #2903: [BUG] Issue 2883 CUDA regression fails on ROCm-only builds
### Problem

`testing/python/issue/test_tilelang_issue_2883.py::test_large_parallel_layout_is_injective` explicitly lowers with a CUDA target, but the test is collected on ROCm-only builds. The current main ROCm job fails before completing the suite:

```text
AttributeError: module 'tilelang.cuda._ffi_api' has no attribute 'AnnotateDeviceBoundTmaCopies'
```

The failure is visible in the gfx942 job: https://github.com/tile-ai/tilelang/actions/runs/31115687512/job/92665328476

Result: `1 failed, 1902 passed, 1236 skipped`.

### Root cause

The regression added in #2899 always constructs `tvm.target.Target("cuda")` and enters the CUDA lowering pipeline. CUDA-specific C++ registrations are absent from a ROCm-only build, so `AnnotateDeviceBoundTmaCopies` is not registered.

This is the same test-classification problem addressed in #2863, which gates CUDA-only regressions with `tilelang.testing.requires_cuda`.

### Expected behavior

The regression should continue to run on CUDA CI and skip when TileLang is built without CUDA support.

### Proposed fix

Add `@tilelang.testing.requires_cuda` to `test_large_parallel_layout_is_injective`.

### Environment

- Current main: `af475304`
- Runner: gfx942
- Workflow: Nightly ROCm 7.2, Python 3.12

morluto:
Fixes #2903.

## Problem

`test_large_parallel_layout_is_injective` explicitly lowers with a CUDA target but is collected on ROCm-only builds. Since those builds do not register CUDA-specific C++ passes, the current main ROCm job fails with:

```text
AttributeError: module 'tilelang.cuda._ffi_api' has no attribute 'AnnotateDeviceBoundTmaCopies'
```

The failing gfx942 job completed with `1 failed, 1902 passed, 1236 skipped`: https://github.com/tile-ai/tilelang/actions/runs/31115687512/job/92665328476

## Change

Add `@tilelang.testing.requires_cuda` to the regression added in #2899. This keeps the test active on CUDA CI and skips it when TileLang is built without CUDA, following the convention established in #2863.

## Validation

- `pytest -q testing/python/issue/test_tilelang_issue_2883.py` on the existing ROCm build: `1 skipped`
- `pre-commit run --files testing/python/issue/test_tilelang_issue_2883.py`: passed
- `git diff --check`: passed

## Summary

- Added `@tilelang.testing.requires_cuda` to the Issue 2883 regression test.
- The test runs on CUDA-enabled builds and skips on ROCm-only builds.
- ROCm validation, pre-commit checks, and `git diff --check` passed.

### DIFF
--- a/testing/python/issue/test_tilelang_issue_2883.py
+++ b/testing/python/issue/test_tilelang_issue_2883.py
@@ -10,6 +10,7 @@ import tilelang.testing
 from tilelang import tvm
 
 
+@tilelang.testing.requires_cuda
 def test_large_parallel_layout_is_injective():
     stride = T.dynamic("stride")
 


morluto:
GitHub Actions appears to be having transient service issues at the moment. The previous Reminder Bot and Metal checks both failed during `Set up job` with `Service Unavailable` while resolving action download information, before any repository code ran. The CUDA and ROCm jobs were then cancelled by fail-fast. I have retriggered the checks, and the new run is currently pending.

SiriusNEO:
Thanks for your contribution! But I think we should enable the multi-backend compilation in this test case instead of requiring CUDA. Fixed in https://github.com/tile-ai/tilelang/pull/2914 so close this for now

PR #2909, opened 2026-08-06
[BugFix] Reject layouts on shared buffers with non-constant extents

### LINKED ISSUE #2906: [BUG] `makeBufferWithLayout` segfaults on a symbolic shared-buffer extent instead of hitting its own `ICHECK`
**Version:** v0.1.11 (`VERSION` = 0.1.11), commit `607a9144cf375e1176ad487dd5b0e3d87c2cb79a`
**Platform:** Linux, CUDA 13.0, NVIDIA RTX 4000 Ada Generation (sm_89), Python 3.10, torch 2.12.1+cu130
**Still present on `main`:** the two loops below are unchanged as of `af4753045895`
(source inspected; the reproduction itself was run on `607a9144`).

## Summary

`makeBufferWithLayout` in `src/transform/lower_tile_op.cc` dereferences the result of
`.as<IntImmNode>()` without a null check when computing a shared buffer's replication
factor. A symbolic (non-constant) extent therefore crashes the compiler with SIGSEGV.

The loop immediately after it guards the identical condition with an `ICHECK`, so the
crash appears to be an oversight rather than intended behaviour — the diagnostic that was
written for this case can never fire, because the unguarded loop runs first.

## Location

`src/transform/lower_tile_op.cc`, inside `if (IsSharedBuffer(buffer))`:

```cpp
Array<PrimExpr> buffer_shape = buffer->shape;
int buffer_extent = 1;
int layout_extent = 1;
for (size_t i = 0; i < buffer_shape.size(); i++) {
  auto shape = buffer_shape[i].as<IntImmNode>();
  buffer_extent *= shape->value;                     // <-- no null check
}
for (size_t i = 0; i < layout_shape.size(); i++) {
  auto shape = layout_shape[i].as<IntImmNode>();
  ICHECK(shape) << "Layout output shape must be constant integer, but got: "
                << layout_shape[i];                  // <-- guarded
  layout_extent *= shape->value;
}
```

If `buffer_shape[i]` is not an `IntImmNode`, `.as<IntImmNode>()` returns `nullptr` and
`shape->value` dereferences it.

## Reproduction

Run this **from a file**, not `python -c` — TVMScript's `@T.prim_func` parser calls
`inspect.getsourcelines`, which has no source under `-c`, so the parser fails first and the
crash never surfaces.

```python
import tilelang
import tilelang.language as T

m = T.dynamic("m", "int32")
N = 256

@tilelang.jit(out_idx=[1])
def build():
    @T.prim_func
    def k(A: T.Tensor((m, N), "float16"), B: T.Tensor((m, N), "float16")):
            s = T.alloc_shared([m, N], "float16")
            # element counts match, so the layout.cc:678 product check passes and
            # execution reaches makeBufferWithLayout with a symbolic shape
            T.annotate_layout({s: Layout([m, N], lambda i, j: [i, j])})
            T.copy(A[0:m, :], s)
            T.copy(s, B[0:m, :])
    return k

build()
```

**Actual:** process dies with SIGSEGV (exit `-11`), no diagnostic.

**Expected:** the existing message, i.e.
`Layout output shape must be constant integer, but got: m`

## Notes

A dynamic shared extent *without* an explicit layout annotation compiles fine — that path
never reaches `makeBufferWithLayout`, so the bug only surfaces when a layout annotation
forces the buffer through the remap path.

`T.dynamic("m", "int32")` and `T.symbolic("m")` behave identically here — both segfault with
the same stack. That is expected from `tilelang/language/symbolics.py`, where `symbolic` is a
`@deprecated(..., "v0.1.9")` alias that calls `dynamic(name, dtype)` and returns the same
`tirx.Var`. The reproduction uses `T.dynamic` as the non-deprecated spelling.

## Suggested fix

Guard the first loop the same way as the second, and report which buffer failed:

```cpp
for (size_t i = 0; i < buffer_shape.size(); i++) {
  auto shape = buffer_shape[i].as<IntImmNode>();
  ICHECK(shape) << "Shared buffer '" << buffer->name
                << "' extent must be a constant integer, but got: " << buffer_shape[i];
  buffer_extent *= shape->value;
}
```

Happy to send a PR if that shape looks right.

sohampanda000:
## Summary

Fixes #2906.

`T.annotate_layout` on a shared buffer whose extent is not a compile-time constant crashed the compiler with SIGSEGV. `makeBufferWithLayout` derives the replication factor as `buffer_extent / layout_extent`, reading each dimension with `buffer_shape[i].as<IntImmNode>()->value`. A symbolic extent makes `.as<IntImmNode>()` return null, and the unguarded dereference kills the process. The `ICHECK` in the loop immediately below guards the identical condition for the layout output shape, but can never fire, because the unguarded loop runs first.

## Fix

Add the check in `layout_inference.cc` beside the injectivity check introduced by #2719, rather than in `makeBufferWithLayout`, so the failure surfaces during layout inference with a message naming the buffer and the offending dimension. The predicate is the same one the crashing code relies on (`.as<IntImmNode>()`), so passing the guard guarantees the downstream dereference is safe rather than approximating it.

Only the combination is rejected. The trigger is one cell of a 2×2, and the other three are left working:

| shared extent | `annotate_layout` | before | after |
|---|---|---|---|
| static `(BM, N)` | yes | builds + runs | unchanged |
| static `(BM, N)` | no | builds + runs | unchanged |
| symbolic `(m, N)` | no | builds + runs | unchanged |
| symbolic `(m, N)` | **yes** | **SIGSEGV** | `ValueError` |

A symbolic shared extent on its own is a supported construct — it lowers to CUDA dynamic shared memory (`extern __shared__`, size passed at launch, loop bounds computed from the symbol) and executes correctly. What cannot be expressed over a non-constant extent is a *layout*, since both the replication factor and the index map require a known extent. The fix is scoped to that.

## Tests

Added `testing/python/issue/test_tilelang_issue_2906.py`:

* the crash is rejected with a `ValueError` naming the buffer and dimension;
* a symbolic shared extent **without** a layout annotation still compiles, executes, and returns the expected result — this guards against over-rejecting a case that works today;
* a static tile under a dynamic grid is unaffected.

Verified locally against `main` (`af47530`, which contains #2719), built with CUDA on sm_89: the tests fail with SIGSEGV (`rc=139`) without the change and pass with it. `bash format.sh --files ...` is clean on both changed files.

## Summary

- Prevents a compiler SIGSEGV when `T.annotate_layout` targets a shared buffer with a non-constant extent.
- Raises `ValueError` with the buffer and offending dimension before unsafe dereferencing.
- Preserves symbolic shared buffers without layout annotations.
- Preserves static tiles under dynamic grids.

## Tests

- Added CUDA regression tests for invalid annotated symbolic shared buffers.
- Added coverage for symbolic shared buffers without layouts.
- Parameterized static shared-tile coverage for annotated and unannotated layouts.

## C++ style / lint notes

- The C++ change does not modify public APIs or documented rules in `docs/developer_guide/cpp_style.md`.
- No CI or lint tooling changes are included.
- The “C++ API Style Audit (warning only)” remains advisory. Warning-only findings do not affect correctness, build, or test validation.

### DIFF
--- a/src/transform/layout_inference.cc
+++ b/src/transform/layout_inference.cc
@@ -749,6 +749,26 @@ private:
                         buffer->shape, &analyzer_, Integer(anchor_bits),
                         Integer(GetElementStorageBits(buffer->dtype)));
           if (IsSharedBuffer(buffer)) {
+            // A layout is an index map over a known extent, and buffer
+            // remapping later derives the replication factor as buffer_extent /
+            // layout_extent from IntImm values (makeBufferWithLayout in
+            // transform/lower_tile_op.cc). A symbolic extent supplies neither,
+            // so reject it here instead of dereferencing a null IntImmNode
+            // downstream. A shared buffer with a symbolic extent is otherwise
+            // supported; only carrying a layout on one is not.
+            for (size_t dimension_index = 0;
+                 dimension_index < buffer->shape.size(); ++dimension_index) {
+              if (!buffer->shape[dimension_index].as<IntImmNode>()) {
+                TVM_FFI_THROW(ValueError)
+                    << "Invalid layout for shared buffer `" << buffer->name
+                    << "`: its shape must be a compile-time constant to carry "
+                       "a "
+                    << "layout, but dimension " << dimension_index << " is `"
+                    << buffer->shape[dimension_index]
+                    << "`. Allocate the buffer with a static "
+                    << "tile extent, or drop the T.annotate_layout on it.";
+              }
+            }
             arith::IterMapResult injectivity =
                 resolved_layout->DetectInjective();
             if (!injectivity->errors.empty()) {
--- a/testing/python/issue/test_tilelang_issue_2906.py
+++ b/testing/python/issue/test_tilelang_issue_2906.py
@@ -1,0 +1,89 @@
+"""Regression test for GitHub issue #2906.
+
+A shared buffer whose extent is not a compile-time constant cannot carry a
+layout: the layout is an index map over a known extent, and buffer remapping
+derives a replication factor from ``IntImm`` values. Passing such a buffer to
+``T.annotate_layout`` previously dereferenced a null ``IntImmNode`` in
+``makeBufferWithLayout`` and killed the process with SIGSEGV, so the existing
+``ICHECK`` on the layout's own output shape could never fire.
+
+It must instead be rejected with a ``ValueError`` naming the buffer and the
+non-constant dimension. A symbolic shared extent *without* a layout annotation
+is supported and must keep working.
+"""
+
+import pytest
+import torch
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+from tilelang.layout import Layout
+
+N = 256
+BM = 16
+
+REJECT_MATCH = r"shared buffer `[^`]*`.*compile-time constant.*dimension 0"
+
+
+def _kernel(shared_shape, layout_shape, annotate):
+    m = T.dynamic("m")
+
+    @T.prim_func
+    def main(A: T.Tensor((m, N), "float16"), B: T.Tensor((m, N), "float16")):
+        with T.Kernel(T.ceildiv(m, BM), threads=128) as bx:
+            s = T.alloc_shared(shared_shape, "float16")
+            if annotate:
+                T.annotate_layout({s: Layout(layout_shape, lambda i, j: [i, j])})
+            start = bx * BM
+            T.copy(A[start : start + BM, :], s)
+            T.copy(s, B[start : start + BM, :])
+
+    return main
+
+
+@tilelang.testing.requires_cuda
+def test_symbolic_shared_extent_with_layout_is_rejected():
+    """The regression: symbolic extent + annotate_layout used to SIGSEGV."""
+    m = T.dynamic("m")
+
+    @T.prim_func
+    def main(A: T.Tensor((m, N), "float16"), B: T.Tensor((m, N), "float16")):
+        with T.Kernel(1, threads=128):
+            s = T.alloc_shared((m, N), "float16")
+            T.annotate_layout({s: Layout([m, N], lambda i, j: [i, j])})
+            T.copy(A[0:m, :], s)
+            T.copy(s, B[0:m, :])
+
+    with pytest.raises(ValueError, match=REJECT_MATCH):
+        tilelang.compile(main, out_idx=[1])
+
+
+@tilelang.testing.requires_cuda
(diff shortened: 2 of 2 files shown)


sohampanda000:
Thanks for the review — both points applied:

- Renamed the loop variable to `dimension_index` per the C++ style guide.
- Tightened `REJECT_MATCH` in the test to also require `dimension 0`, so the assertion verifies the diagnostic actually identifies the offending dimension rather than just matching the general message.

Re-verified after the change against `main` (`af47530`) on sm_89: the tests still fail with SIGSEGV (`rc=139`) without the fix and pass with it, and `bash format.sh --files ...` is clean on both files.

sohampanda000:
Applied — `test_static_shared_tile_still_works` is now parameterized over `annotate=True/False`, so both cases actually execute rather than the docstring overstating coverage. Assertions unchanged. Renamed to drop `_with_layout_` from the name, since it now covers both.

Re-verified on sm_89 against `main` (`af47530`): 4 passed (was 3), and still SIGSEGV (`rc=139`) without the fix. `format.sh` clean.

Re the earlier `REJECT_MATCH` point — that one was already applied in the previous push; it now reads `r"shared buffer \`[^\`]*\`.*compile-time constant.*dimension 0"`, so the assertion requires the diagnostic to name the offending dimension.

sohampanda000:
@LeiWang1999 — quick reminder on this one when you have a moment, since you reviewed and merged #2719, which added the injectivity check this sits directly beside in `layout_inference.cc`.

`T.annotate_layout` on a shared buffer with a non-constant extent segfaults in `makeBufferWithLayout`. The guard goes next to your injectivity check, so the failure surfaces during layout inference with a message naming the buffer and dimension, rather than reaching the code that assumes `IntImm` extents.

Scope is deliberately narrow — only the layout + symbolic-extent combination is rejected. A symbolic shared extent on its own still lowers to CUDA dynamic shared memory and runs correctly, and there is a test asserting that so the fix cannot over-reject.

Verified against `main` on sm_89: SIGSEGV without the change, 4 tests passing with it.

One correction to an earlier version of this comment: I said CI was green, which was inaccurate. Only CodeRabbit and pre-commit.ci have run. The `CI` workflow has never executed on this branch — as a first-time contributor the run appears to need maintainer approval. If someone could approve the workflow, that would be the most useful next step, and I will fix whatever it surfaces.

sepcnt:
We also hit the same path with runtime-length shared scans and found that some symbolic layouts can be supported rather than rejected. https://github.com/tile-ai/tilelang/pull/3233

sohampanda000:
Superseded by #3233.

PR #2917, opened 2026-08-07
[BugFix][CPU] Compute scalar GEMM products in accum dtype instead of input dtype

Dino1844:
## Summary

`GemmScalar` — the CPU/LLVM fallback for `T.gemm` — evaluates each scalar multiply in the **input** dtype and only then widens the already-rounded product to the accumulator dtype:

```python
C_buf[c0 + i, c1 + j] += T.cast(
    A_buf[...] * B_buf[...],   # fp16 x fp16 -> fp16 (rounded)
    accum_dtype,
)
```

For fp16 inputs this diverges from CUDA/ROCm mma, which computes fp16 products exactly (the 22-bit result fits in fp32's 24-bit mantissa) and accumulates in fp32. It also diverges from TileLang's own CUDA FMA fallback (`tilelang/cuda/op/gemm/gemm_fma.py`), whose docstring documents the correct convention: *"Casts widen operands to `accum_dtype` before multiplication so narrow input dtypes stay numerically sound (e.g. BF16 to FP32 accumulation on Volta). This path favors correctness and coverage over peak throughput."*

## Problem

Rounding each product to fp16 before widening has two consequences:

1. **Accuracy.** Every product carries an extra ~2^-11 relative rounding (fp32 accumulation is ~2^-24), which dominates the error for realistic K. Measured on normal-range inputs: the fp16 GEMM deviates from exact-product accumulation by up to ~1e-2 absolute (K=256), whereas exact products are accurate to fp32-accumulation level (~1e-5).

2. **Overflow -> NaN/Inf.** Products with |a·b| > 65504 (fp16 max) overflow to ±inf; mixed-sign infs then produce NaN inside the fp32 accumulator. Measured with |a|,|b| ~ 600: 100% of outputs were NaN/Inf before the fix, while exact products stay finite (they fit in fp32).

## Fix

Cast both operands to `accum_dtype` before multiplying (exact for fp16/bf16/fp8 products, all of which fit exactly in fp32):

```python
C_buf[c0 + i, c1 + j] += T.cast(A_buf[...], accum_dtype) * T.cast(B_buf[...], accum_dtype)
```

This matches mma semantics and the convention already established in `gemm_fma.py`.

## Verification

fp16 `T.gemm` on `target="c"` (compiled and executed): normal-range inputs match an exact-product fp64 reference to fp32-accumulation level (max abs err 2e-5 vs ~1e-2 before); large-magnitude inputs (products > 65504) no longer produce NaN/Inf. Existing CPU GEMM tests (`test_tilelang_cpu_tgemm.py`, `test_tilelang_cpu_gemm.py`) pass.

## Summary

- Fixed CPU/LLVM scalar `T.gemm` accumulation.
- Cast both operands to `accum_dtype` before multiplication.
- Prevented fp16 product rounding and fp16 overflow to `Inf` or `NaN`.
- Aligned CPU behavior with CUDA/ROCm MMA and existing CUDA FMA fallback semantics.
- Improved fp16 GEMM accuracy and preserved existing CPU GEMM test coverage.

### DIFF
--- a/tilelang/cpu/op/gemm/gemm_scalar.py
+++ b/tilelang/cpu/op/gemm/gemm_scalar.py
@@ -47,8 +47,10 @@ class GemmScalar(GemmBase):
                 T.clear(C_buf)
             for i, j, k in T.grid(M, N, K):
                 C_buf[c0 + i, c1 + j] += T.cast(
-                    A_buf[a0 + (k if trans_A else i), a1 + (i if trans_A else k)]
-                    * B_buf[b0 + (j if trans_B else k), b1 + (k if trans_B else j)],
+                    A_buf[a0 + (k if trans_A else i), a1 + (i if trans_A else k)],
+                    accum_dtype,
+                ) * T.cast(
+                    B_buf[b0 + (j if trans_B else k), b1 + (k if trans_B else j)],
                     accum_dtype,
                 )
 


Dino1844:
Hi @LeiWang1999 and maintainers,

While auditing TileLang I focused on the CPU backend, and wanted to confirm the current support status and direction before investing effort, to avoid overlapping with community work and wasting your review time.

Observation: The CPU backend (target="c" / target="llvm") is marked Experimental in the README. From my testing:
1. Some core ops are unimplemented: T.cumsum, T.atomic_add, finalize_reducer, gemm_sp fail with Unresolved call (CUDA has ~14 ops, CPU 5);
2. Vectorization is incomplete: T.exp/min/max/cast etc. fail to compile in T.Parallel, vec_type only implements +-*/;
3. Some semantics diverge from CUDA/torch, e.g. NaN / -0.0 in min/max.

Questions:
1. Is CPU an actively developed direction? Any roadmap or in-progress work?
2. If contributions are welcome, what to prioritize — unimplemented ops, or vectorization/math support?

Thanks for your time! If CPU isn't a near-term priority, please let me know and I'll adjust accordingly.

LeiWang1999 (approved):

PR #2922, opened 2026-08-08
[BugFix] Keep loop-invariant stores during vectorize planning

penguin-wwy:
VectorizePlanner::ComputeBufferVectorSize returns vector_size=1 for stores whose address is invariant within the vector boundary, because vectorizing them produces a broadcast store: every lane reads the same old value and
stores to the same address, so the last lane silently wins (accumulation pattern).
However, Plan() then bucketed such entries by address invariance alone and moved them to the local/fragment bucket, which the simple-case strategy (GCD(memory_min, non_cast_call_node_min)) ignores — the constraint
was dropped and the loop was vectorized anyway.

## Summary

- Fixed vectorization planning to preserve loop-invariant global and shared-buffer stores as memory constraints.
- Added a CUDA regression test for invariant-store accumulation.
- Verified results against `Init + a.sum(dim=1)` using float16 tensors and toleranced comparison.

## C++ style / lint notes

- The PR changes C++ code but does not modify the rules in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit is not relevant because the PR adds no public C++ API.
- No correctness or build issues are identified. Any audit findings would be warning-only.

## Tests

- Added `vectorize_invariant_store_accumulate(M, K)`.
- Added `test_vectorize_invariant_store_accumulate()`.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -284,7 +284,7 @@ public:
           depends_on_loop_var = !IsExprInvariantInVectorBoundary(
               elem_offset, inner_for_->loop_var, vector_size_, analyzer_);
         }
-        if (depends_on_loop_var) {
+        if (depends_on_loop_var || info.is_store) {
           memory_min = arith::ZeroAwareGCD(memory_min, info.vector_size);
           has_global_or_shared_buffer = true;
         } else {
--- a/testing/python/language/test_tilelang_language_vectorize.py
+++ b/testing/python/language/test_tilelang_language_vectorize.py
@@ -114,6 +114,52 @@ def test_vectorize_invariant_index():
 
 
 @tilelang.jit
+def vectorize_invariant_store_accumulate(M, K):
+    @T.prim_func
+    def main(
+        A: T.Tensor[(M, K), T.float16],  # noqa: F821
+        Init: T.Tensor[(M,), T.float16],  # noqa: F821
+        B: T.Tensor[(M,), T.float16],  # noqa: F821
+    ):
+        with T.Kernel(M // 128, threads=128) as bx:
+            tx = T.get_thread_binding(0)
+            row = bx * 128 + tx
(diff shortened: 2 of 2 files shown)


src/transform/loop_vectorize.cc
```diff
@@ -284,7 +284,7 @@ class VectorizePlanner : public arith::IRMutatorWithAnalyzer {
           depends_on_loop_var = !IsExprInvariantInVectorBoundary(
               elem_offset, inner_for_->loop_var, vector_size_, analyzer_);
         }
-        if (depends_on_loop_var) {
+        if (depends_on_loop_var || info.is_store) {
```
SiriusNEO, line 287: This check is only in the global/shared scope. Consider the following case:
```python
acc = T.alloc_local((1,), T.float32)
  for i in T.vectorized(4):
      acc[0] += A[row, i]
```
This loop-invarient store to a local/fragment seems to have the same problem. Maybe we should fix them too.

SiriusNEO (changes requested):

testing/python/language/test_tilelang_language_vectorize.py
```diff
@@ -113,6 +113,52 @@ def test_vectorize_invariant_index():
...
+        with T.Kernel(M // 128, threads=128) as bx:
+            tx = T.get_thread_binding(0)
+            row = bx * 128 + tx
+            B[row] = Init[row]
+            # B[row] does not depend on the vectorized loop variable k, so
+            # vectorizing this loop turns the accumulate into a broadcast
+            # store: every lane reads the same old B[row], and the last
+            # lane's store wins, dropping all addends but A[row, K-1].
+            for k in T.vectorized(K):
+                B[row] = B[row] + A[row, k]
```
SiriusNEO, line 133: In my opinion, this should either report an error, or throw a warning and don't vectorize it. Vectorization should not be together with an accumulation statement.

PR #2924, opened 2026-08-09
[Bugfix] Carry memory_order through vectorized atomic_add

### LINKED ISSUE #2688: [BUG][Fuzzer][wrong-code] Vectorized `T.atomic_add` silently downgrades the requested `memory_order` to relaxed
### Required prerequisites

0.1.12 (reproduced this session on the released `0.1.12` wheel). Source on `main` re-checked at `d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6` (and the offending line is unchanged at current `main` HEAD `c5e53076`).

### System information

NVIDIA L40S (`sm_89`), CUDA 12.8, Python 3.13, tilelang 0.1.12. The defect is in target-independent transform/codegen (the atomic-add vectorizer and its CUDA codegen), so it is not specific to one GPU.

### Problem description

A tile-region `T.atomic_add(Out, Src, memory_order="acq_rel")` on `float16`/`bfloat16` (or `float32`) silently produces a **relaxed** atomic whenever the add auto-vectorizes to `AtomicAddx2`/`AtomicAddx4`, even though the *identical* call lowers scalar-wise — and correctly emits the requested ordering — when it does not vectorize. Whether it vectorizes is decided by whether the per-thread inner-loop extent shares a factor ≥2 with the atomic vector width (`GCD(vector_width, extent) ≥ 2`, i.e. the extent is divisible by 2 for the fp16/bf16 `x2` path or by 2/4 for the fp32 `x4` path) — a tiling detail unrelated to memory ordering. No error or warning is raised at any stage; the kernel compiles and runs, and the requested memory ordering is just gone.

Concretely, the same `T.atomic_add(Out[i], X[i], memory_order="acq_rel")` in a `T.Parallel` loop emits (both verified this session, fp16, `threads=32`):

| per-thread extent | lowering | emitted CUDA | ordering |
|---|---|---|---|
| 3 (`N=96`) — `GCD(2,3)=1`, no vectorize | scalar | `AtomicAdd(&Out[...], Src[...], 4)` | acq_rel (id 4) — **honored** |
| 2 (`N=64`) — `GCD(2,2)=2`, vectorizes | `AtomicAddx2` | `AtomicAddx2(&Out[tid*2], *(uint1*)(Src+tid*2))` | (no arg) → **relaxed** |

So identical public-API input yields different memory ordering depending purely on an invisible vectorization choice driven by the tile shape. This is not a regression — the vectorized atomic path has reconstructed the wide op from two operands since it first shipped.

<details><summary>Emitted atomic lines, both paths (this session, 0.1.12)</summary>

```
extent 3 (no vectorize): AtomicAdd((&(Out[((int)threadIdx.x)])), Src[((int)threadIdx.x)], 4);
   -> honors acq_rel (trailing ", 4)"): True
extent 2 (vectorizes)  : AtomicAddx2((&(Out[(((int)threadIdx.x) * 2)])), *(uint1*)(Src + (((int)threadIdx.x) * 2)));
   -> honors acq_rel (trailing ", 4)"): False   (silently relaxed)
```

</details>

### Reproducible example code

```python

def mk(N, threads):
    @T.prim_func
    def k(X: T.Tensor((N,), "float16"), Out: T.Tensor((N,), "float16")):
        with T.Kernel(1, threads=threads):
            # identical tile-region atomic_add on both paths; only N/threads differ
            T.atomic_add(Out[0:N], X[0:N], memory_order="acq_rel")   # acq_rel == id 4
    return k

def emitted_atomic(N, threads):
    src = tilelang.compile(mk(N, threads), out_idx=-1, target="cuda").get_kernel_source()
    return [l.strip() for l in src.splitlines() if "AtomicAdd" in l][0]

# SCALAR lowering (1 elem/thread): honors acq_rel -> "AtomicAdd(..., 4)"
print(emitted_atomic(130, 130))   # -> AtomicAdd(&Out[...], X[...], 4);   ("...4)" present)

# VECTORIZED lowering (4 elem/thread, SAME call): drops it -> "AtomicAddx2(...)" no order arg
print(emitted_atomic(256, 64))    # -> AtomicAddx2(&Out[...], *(uint1*)(X+...));   (order gone -> relaxed)
```

### Traceback

No traceback — the kernel compiles and runs to completion on both paths; the vectorized path silently emits a weaker (relaxed) atomic than requested, with no error or warning.

### Expected behavior

A vectorized `T.atomic_add` should carry the same requested `memory_order` its scalar sibling does — the wide device helpers already accept and honor a `memory_order` argument (they branch to explicit ordered PTX for a non-relaxed order), so the requested `acq_rel` seems computable on the vectorized path too, not just the scalar one. Grounding: the compiler already emits `AtomicAdd(..., 4)` for the identical tile-region call when it lowers scalar-wise, so honoring the order is the behavior the code already exhibits on one of its two lowerings of the same input.

(One doc note for the maintainers: the `atomic_add` docstring currently says "the tile-region path ignores `memory_order`", but the scalar tile-region lowering demonstrably *does* honor it, so that line reads as stale — the intent appears to be to honor the order, which makes the vectorized drop the defect. Reconciling the docstring either way would remove the ambiguity.)

### Additional context

**Root cause.** The atomic-add vectorizer does not carry the operation's memory-order operand when it rewrites a tile-region atomic add into its wide (x2/x4) form, so the wide atomic is generated without any ordering and silently falls back to relaxed. In detail: the tile-region lowering attaches the requested order as the third operand of the scalar `atomic_add_elem_op` ([`atomic_add.cc#L185-L192`](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/cuda/op/atomic_add.cc#L185-L192), pushing `op.GetMemoryOrder()` — [`atomic_reduce.h#L47-L54`](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/op/atomic_reduce.h#L47-L54)); the vectorizer then reconstructs the wide op from **only** the destination and source and discards that operand ([`vectorize_loop.cc#L629-L664`](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/transform/vectorize_loop.cc#L629-L664) — the return is `Call(op->dtype, GetVectorizedAtomicOp(vector_size), {dst, src})`, two operands only).

<details><summary>Why the dropped operand changes the emitted ordering</summary>

The CUDA codegen for both the scalar and the wide atomics appends the order only when it is present as a third operand:

- scalar [`AtomicAdd` — codegen_cuda.cc#L4597-L4608](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/cuda/codegen/codegen_cuda.cc#L4597-L4608): `AtomicAdd(dst, src)` then `if (op->args.size() > 2) stream << ", " << args[2];`
- wide [`AtomicAddx2`/`AtomicAddx4` — codegen_cuda.cc#L4620-L4643](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/cuda/codegen/codegen_cuda.cc#L4620-L4643): identical `if (op->args.size() > 2)` guard.

After the vectorizer drops the operand, `op->args.size()` is 2, so the guard is false and no order is emitted. The wide device helper then takes its default: [`AtomicAddx2(half_t*, ValType, int memory_order = relaxed)` — atomic.h#L557-L577](https://github.com/tile-ai/tilelang/blob/d36ec37c596ff8e587fc67fcf1e88bcb90cc60d6/src/tl_templates/cuda/atomic.h#L557-L577) branches on the order — relaxed → a plain `atomicAdd`; non-relaxed → explicit ordered PTX (`tl_atomic_add_v2_f16` with fences). So the requested `acq_rel` (id 4) becomes relaxed (id 0), silently. `AtomicAddx4` and the `float32` wide wrappers share the same default-relaxed / branch shape, so `fp32 → AtomicAddx4` is the same mechanism (I exhibited the `fp16 → AtomicAddx2` face; the x4 face is inferred from the identical guard and helper, not separately run).

</details>

**Suggested fix (proposed, not built).** One direction that fits the existing wiring is to carry the memory-order operand through the wide reconstruction — append the original op's third operand (when present) to the `{dst, src}` argument list in the vectorizer — so the wide `AtomicAddx2`/`AtomicAddx4` codegen and device helpers receive the order exactly as the scalar path does. The codegen and device helpers already accept it (the `if (op->args.size() > 2)` guard and the `memory_order` parameter are in place); only the vectorizer's two-operand reconstruction appears to drop it.

**Distinct from #2382/#2574/#2573.** #2382 (closed, fixed by #2414) is the *same* vectorizer function but a different defect — the wide helper bit-reinterprets instead of converting when the source dtype differs from the destination; its fix touched only `atomic.h` and never the operand reconstruction, so the memory-order drop is untouched (`{dst, src}` still stands at `main`). #2574 is an *illegal* memory order on `atomic_load`/`atomic_store` reaching a device assert (a missing frontend legality check; the order is carried, not dropped). #2573 is an `int64` `atomic_max`/`atomic_min` compile failure with no memory-order involvement. I searched the open and closed tracker and found no existing report of the memory-order drop on the vectorized `atomic_add` path.

**Provenance.** Not a regression — the vectorized atomic path has reconstructed the wide op from two operands (`{addr_dst, addr_src}` / `{dst, src}`) since it first shipped: it existed in the pre-refactor `atomicadd_vectorize.cc`, was relocated into `vectorize_loop.cc` (with `MutateAtomicAddExpr_`) by [#1677](https://github.com/tile-ai/tilelang/pull/1677) (`5feb2253`, merged 2026-01-16), and the line was last moved by the #2166 refactor (`431c85f1`, 2026-05-12). The `memory_order` argument for atomics arrived alongside it in [#1676](https://github.com/tile-ai/tilelang/pull/1676) (`b27fb928`, merged 2026-01-16, an ancestor of #1677), so from the moment a non-relaxed order could be requested the vectorized path has never carried it. (Origins verified via `git log`/`git blame` and `gh pr view`; the exact PR that first added the two-operand form pre-refactor is not pinned beyond "no later than #1677".)

**Reach.** Triggering requires a tile-region `T.atomic_add` (a region argument such as `Out[0:N]`, which is what routes to the wide helpers) with a non-relaxed `memory_order` *and* a per-thread inner-loop extent divisible by the dtype's atomic vector width (2 for fp16/bf16, 2-or-4 for fp32), so the add auto-vectorizes. A non-relaxed order is the rare ingredient — `relaxed` is the default, so a program only hits this if it explicitly asks for ordering; the divisible extent is common (the 128-multiple tiles typical of reduction/scatter kernels split to an even per-thread extent), so once a kernel does request ordering on a region it will usually take the dropping path. `grep -rn` over `examples/` and `testing/` this session: the shipped examples that pass `memory_order=` all pass `"relaxed"` (`examples/norm/layernorm.py:127-128`, `examples/flash_attention/example_gqa_bwd_tma_reduce_varlen.py:476`), so the downgrade is a no-op for them. The atomic tests that use a non-relaxed order (`testing/python/language/test_tilelang_language_atomic.py:104`, `memory_order="release"`; and `:46`) index a single element (`B[idx_i, idx_j]`), which routes to the scalar/extern `AtomicAdd` path — the one that honors the order — never the region/vectorized path; and `test_atomic_addx2` calls `T.atomic_addx2` directly, which has no `memory_order` parameter. So no example or test exercises the region + non-relaxed-order + even-count intersection, and the value-comparison tests would not catch an ordering downgrade regardless (they check a sum, which has no cross-thread ordering dependency).

arcusbuilds:
Fixes #2688.

## Cause

`vectorize_loop.cc` rebuilt the wide atomic from two operands:

```cpp
return Call(op->dtype, GetVectorizedAtomicOp(vector_size), {dst, src});
```

The region lowering puts `memory_order` in `args[2]`. Codegen appends it only under
`if (op->args.size() > 2)`, so the emitted `AtomicAddx2(dst, src)` picks up the
`= memory_order_relaxed` default on the device function. Legal C++, no diagnostic anywhere.
Everything downstream already accepted the operand, including `num_inputs = 3`; the vectorizer
was under-filling it.

The fix carries the trailing operands through, starting at index 2 because the two-argument form
is legal, and copying rather than visiting because the order is lane-invariant.

## Why atomic.h changes too

Carrying the order makes inline PTX reachable that had never been assembled: the non-relaxed
branches were constant-folded away while the order always defaulted to relaxed. Three of them do
not build.

1. **fp16 used `.v2.f16`**, whose vector operands require sm_90, and fp16 vectorizes to x2 on
   every target. Now uses packed `f16x2`, available from sm_60 and emitting identical SASS on
   sm_90, so fp16 needs no arch guard at all and the per-half pack/unpack goes away.
2. **bf16 was missing `.noftz`**, which ptxas requires for `.bf16`. Syntax error on every target,
   sm_90 included.
3. **bf16 has no ordered atomic below sm_90 at any width**, so unlike fp16 there is nothing to
   fall back to. The sm_90 path is guarded, with `__threadfence()` around a relaxed bf16x2
   atomicAdd below it: two-way where release and acquire need one-way, so conservative rather
   than lossy.

New guards use `__CUDA_ARCH__` rather than the `__CUDA_ARCH_LIST__` spelling elsewhere in this
file, which fails a `-gencode` multi-arch build with `C1012`. The remaining uses guard
declarations, where `__CUDA_ARCH__` is undefined in the host pass, so those stay as they are.

Docstrings for `atomic_add`, `atomic_max` and `atomic_min` claimed the region path ignores
`memory_order`, which is no longer true.

## Testing

Six cases covering acquire, release and acq_rel on fp16 and bf16, since each ordering is a
separately spelled asm string. Each asserts on the generated CUDA, then runs the kernel.

RTX 2000 Ada (sm_89), CUDA 13.3: **56 passed, 3 skipped, 1 failed**. The failure is
`test_tma_atomic_add`, which fails identically on unmodified `main` (TMA needs sm_90 and that
test has no arch guard) and uses no `memory_order`.

Negative control: reverting only the two source files, keeping the new tests, and rebuilding
makes them fail with `AtomicAddx2((&(Out[...])), *(uint1*)(X + ...));`. Restoring and rebuilding
makes them pass.

Every PTX spelling was checked against sm_75 through sm_90 with ptxas. There is no sm_90
hardware here, so those branches are verified to assemble rather than to run.

## One untested path

`AtomicAddx2Ret(bfloat16_t*)` takes the same guard through the shared helper, and
`AtomicAddx4Ret` routes through it. This is the only part with no test, and it cannot have one:
nothing reachable instantiates the bf16 Ret overloads with a non-relaxed order, since
`GetVectorizedAtomicOp` returns only the non-ret ops, `T.atomic_addx2` / `T.atomic_addx4` take no
`memory_order`, and region-form `return_prev=True` raises `NotImplementedError`. Guarded anyway
so the file stays consistent and the path is already correct when `return_prev` is implemented.
Verified by instantiating the template directly: without the guard it fails on sm_80/86/89 with
`Instruction 'atom.add.bf16.global' requires .target sm_90 or higher`, and passes on all four
targets with it.

## Summary

- Preserve trailing operands, including `memory_order`, when vectorizing `T.atomic_add` to `AtomicAddx2` or `AtomicAddx4`.
- Use packed `f16x2` PTX for vector fp16 atomics.
- Add `.noftz` to vector bf16 PTX.
- Add fenced bf16 atomics for architectures below sm_90.
- Update `atomic_add`, `atomic_max`, and `atomic_min` documentation.
- Add six CUDA tests for acquire, release, and acq_rel ordering on fp16 and bf16.

## C++ style / lint notes

- The PR changes C++ code in `src/tl_templates/cuda/atomic.h` and `src/transform/vectorize_loop.cc`.
- The available summary does not identify changes to rules in `docs/developer_guide/cpp_style.md`.
- Review the “C++ API Style Audit (warning only)” CI step for advisory findings.
- Treat TLCPP003/TLCPP004 findings as warning-only unless they indicate a new API, FFI, or maintainability risk.

### DIFF
--- a/src/tl_templates/cuda/atomic.h
+++ b/src/tl_templates/cuda/atomic.h
@@ -143,29 +143,27 @@ TL_DEVICE void tl_atomic_add_bf16(unsigned short &ret, unsigned long long addr,
   }
 }
 
-TL_DEVICE void tl_atomic_add_v2_f16(unsigned short &ret_x,
-                                    unsigned short &ret_y,
-                                    unsigned long long addr,
-                                    unsigned short val_x, unsigned short val_y,
-                                    int memory_order) {
+// Packed f16x2 rather than `.v2.f16`: both encode the same pairwise add, but
+// `atom` with vector operands requires sm_90 while the packed 32-bit form is
+// available from sm_60, so this keeps ordered fp16 wide atomics working on
+// every supported target.
+TL_DEVICE void tl_atomic_add_f16x2(unsigned int &ret, unsigned long long addr,
+                                   unsigned int val, int memory_order) {
   if (IsReleaseMemoryOrder(memory_order)) {
-    asm volatile(
-        "atom.release.gpu.global.add.noftz.v2.f16 {%0,%1}, [%2], {%3,%4};"
-        : "=h"(ret_x), "=h"(ret_y)
-        : "l"(addr), "h"(val_x), "h"(val_y)
-        : "memory");
+    asm volatile("atom.release.gpu.global.add.noftz.f16x2 %0, [%1], %2;"
+                 : "=r"(ret)
+                 : "l"(addr), "r"(val)
+                 : "memory");
   } else if (IsAcquireLikeMemoryOrder(memory_order)) {
-    asm volatile(
-        "atom.acquire.gpu.global.add.noftz.v2.f16 {%0,%1}, [%2], {%3,%4};"
-        : "=h"(ret_x), "=h"(ret_y)
-        : "l"(addr), "h"(val_x), "h"(val_y)
-        : "memory");
+    asm volatile("atom.acquire.gpu.global.add.noftz.f16x2 %0, [%1], %2;"
+                 : "=r"(ret)
+                 : "l"(addr), "r"(val)
+                 : "memory");
   } else if (IsAcqRelLikeMemoryOrder(memory_order)) {
-    asm volatile(
-        "atom.acq_rel.gpu.global.add.noftz.v2.f16 {%0,%1}, [%2], {%3,%4};"
-        : "=h"(ret_x), "=h"(ret_y)
-        : "l"(addr), "h"(val_x), "h"(val_y)
-        : "memory");
+    asm volatile("atom.acq_rel.gpu.global.add.noftz.f16x2 %0, [%1], %2;"
+                 : "=r"(ret)
+                 : "l"(addr), "r"(val)
+                 : "memory");
   }
 }
 
@@ -175,20 +173,23 @@ TL_DEVICE void tl_atomic_add_v2_bf16(unsigned short &ret_x,
                                      unsigned short val_x, unsigned short val_y,
                                      int memory_order) {
   if (IsReleaseMemoryOrder(memory_order)) {
-    asm volatile("atom.release.gpu.global.add.v2.bf16 {%0,%1}, [%2], {%3,%4};"
-                 : "=h"(ret_x), "=h"(ret_y)
-                 : "l"(addr), "h"(val_x), "h"(val_y)
-                 : "memory");
+    asm volatile(
+        "atom.release.gpu.global.add.noftz.v2.bf16 {%0,%1}, [%2], {%3,%4};"
+        : "=h"(ret_x), "=h"(ret_y)
+        : "l"(addr), "h"(val_x), "h"(val_y)
+        : "memory");
   } else if (IsAcquireLikeMemoryOrder(memory_order)) {
-    asm volatile("atom.acquire.gpu.global.add.v2.bf16 {%0,%1}, [%2], {%3,%4};"
-                 : "=h"(ret_x), "=h"(ret_y)
-                 : "l"(addr), "h"(val_x), "h"(val_y)
-                 : "memory");
+    asm volatile(
+        "atom.acquire.gpu.global.add.noftz.v2.bf16 {%0,%1}, [%2], {%3,%4};"
+        : "=h"(ret_x), "=h"(ret_y)
+        : "l"(addr), "h"(val_x), "h"(val_y)
+        : "memory");
   } else if (IsAcqRelLikeMemoryOrder(memory_order)) {
-    asm volatile("atom.acq_rel.gpu.global.add.v2.bf16 {%0,%1}, [%2], {%3,%4};"
-                 : "=h"(ret_x), "=h"(ret_y)
-                 : "l"(addr), "h"(val_x), "h"(val_y)
-                 : "memory");
+    asm volatile(
+        "atom.acq_rel.gpu.global.add.noftz.v2.bf16 {%0,%1}, [%2], {%3,%4};"
+        : "=h"(ret_x), "=h"(ret_y)
+        : "l"(addr), "h"(val_x), "h"(val_y)
+        : "memory");
   }
 }
 
@@ -597,16 +598,10 @@ TL_DEVICE void AtomicAddx2(half_t *ref, ValType val,
     // Since atomicAdd does not support memory order, atomic_ref does not
     // support vectorized atomic operation we can only inline ptx code here
     // Note: Vectorized atomic operations only support global space
-    // Note: for 16-bit value, we need to reinterpret_cast the value to unsigned
-    // short and use "h" register in assembly
-    unsigned short add_val_x_cast = tl_atomic_detail::PackBits16(add_val.x);
-    unsigned short add_val_y_cast = tl_atomic_detail::PackBits16(add_val.y);
-    unsigned long long ref_addr = reinterpret_cast<unsigned long long>(ref);
-    unsigned short ret_val_x_cast;
-    unsigned short ret_val_y_cast;
-    tl_atomic_detail::tl_atomic_add_v2_f16(ret_val_x_cast, ret_val_y_cast,
-                                           ref_addr, add_val_x_cast,
-                                           add_val_y_cast, memory_order);
+    unsigned int ret_val;
+    tl_atomic_detail::tl_atomic_add_f16x2(
+        ret_val, reinterpret_cast<unsigned long long>(ref),
+        *reinterpret_cast<const unsigned int *>(&add_val), memory_order);
   }
 }
 
@@ -618,16 +613,11 @@ AtomicAddx2Ret(half_t *ref, ValType val,
   if (tl_atomic_detail::IsRelaxedMemoryOrder(memory_order)) {
     return atomicAdd(reinterpret_cast<half2 *>(ref), add_val);
   } else {
-    unsigned short add_val_x_cast = tl_atomic_detail::PackBits16(add_val.x);
-    unsigned short add_val_y_cast = tl_atomic_detail::PackBits16(add_val.y);
-    unsigned long long ref_addr = reinterpret_cast<unsigned long long>(ref);
-    unsigned short ret_val_x_cast;
-    unsigned short ret_val_y_cast;
-    tl_atomic_detail::tl_atomic_add_v2_f16(ret_val_x_cast, ret_val_y_cast,
-                                           ref_addr, add_val_x_cast,
-                                           add_val_y_cast, memory_order);
-    return half2(tl_atomic_detail::UnpackBits16<__half>(ret_val_x_cast),
-                 tl_atomic_detail::UnpackBits16<__half>(ret_val_y_cast));
+    unsigned int ret_val;
+    tl_atomic_detail::tl_atomic_add_f16x2(
+        ret_val, reinterpret_cast<unsigned long long>(ref),
+        *reinterpret_cast<const unsigned int *>(&add_val), memory_order);
+    return *reinterpret_cast<const half2 *>(&ret_val);
   }
 }
 
@@ -670,6 +660,35 @@ TL_DEVICE __nv_bfloat162 ToBfloat162(float *val) {
 }
 
 #if (defined(__CUDA_ARCH_LIST__) && (__CUDA_ARCH_LIST__ > 750))
+namespace tl_atomic_detail {
+// Below sm_90 there is no ordered bf16 atomic add in any width: `atom` with
+// vector types requires sm_90, and so does `atom.add.bf16` itself, so the
+// pair cannot be decomposed into ordered scalar atomics the way the fp16 path
+// does. Emulate the ordering with fences around a relaxed bf16x2 atomicAdd
+// instead. `__threadfence()` lowers to `fence.sc.gpu` (MEMBAR.SC.GPU), i.e.
+// device scope -- the same scope as the `.gpu` qualifier on the ordered PTX
+// above -- and is a two-way barrier, strictly stronger than the one-way
+// barrier release/acquire require. So this is conservative rather than lossy:
+// it just costs more than the sm_90+ instruction would. One consequence worth
+// knowing: for seq_cst this fallback is *stronger* than the sm_90 path, which
+// maps seq_cst onto `atom.acq_rel` like the rest of this file does.
+TL_DEVICE __nv_bfloat162 tl_atomic_add_v2_bf16_fenced(__nv_bfloat162 *addr,
+                                                      __nv_bfloat162 val,
+                                                      int memory_order) {
+  const bool acq_rel = IsAcqRelLikeMemoryOrder(memory_order);
+  if (acq_rel || IsReleaseMemoryOrder(memory_order)) {
+    // Prior writes become visible before the add does.
+    __threadfence();
+  }
+  __nv_bfloat162 prev = atomicAdd(addr, val);
+  if (acq_rel || IsAcquireLikeMemoryOrder(memory_order)) {
+    // Later reads cannot be hoisted above the add.
+    __threadfence();
+  }
+  return prev;
+}
+} // namespace tl_atomic_detail
+
 template <typename ValType>
 TL_DEVICE void AtomicAddx2(bfloat16_t *ref, ValType val,
                            int memory_order = int(cuda::memory_order_relaxed)) {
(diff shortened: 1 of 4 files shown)


src/tl_templates/cuda/atomic.h
```diff
@@ -654,6 +674,10 @@ TL_DEVICE void AtomicAddx2(bfloat16_t *ref, ValType val,
     tl_atomic_detail::tl_atomic_add_v2_bf16(ret_val_x_cast, ret_val_y_cast,
                                             ref_addr, add_val_x_cast,
                                             add_val_y_cast, memory_order);
+#else
+    tl_atomic_detail::tl_atomic_add_v2_bf16_fenced(
+        reinterpret_cast<__nv_bfloat162 *>(ref), add_val, memory_order);
+#endif
```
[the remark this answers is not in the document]
arcusbuilds: Done in the latest push. `tl_atomic_add_v2_bf16_fenced` now returns the previous value (`atomicAdd` already produced it, it was just discarded), and `AtomicAddx2Ret(bfloat16_t*)` is guarded the same way as `AtomicAddx2`. `AtomicAddx4Ret` routes through it, so it is covered too.

One correction to the research above: `atom.add.noftz.v2.bf16` and `atom.add.noftz.bf16x2` both do exist and both assemble on sm_90. I checked each spelling against ptxas 13.3 on sm_75/80/86/89/90. What is true is that neither works below sm_90, which is why the fenced fallback is needed at all.

Verified by instantiating the template directly, since no Python API can reach this path: without the guard it fails on sm_80/86/89 with `Instruction 'atom.add.bf16.global' requires .target sm_90 or higher`; with it, all four targets pass. Noted in the PR description as the one change without test coverage.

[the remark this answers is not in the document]
arcusbuilds: Also done, with one scoping note.

Both new guards now use `__CUDA_ARCH__`. I reproduced the failure you are pointing at rather than taking it on faith: with `-gencode arch=compute_80,code=sm_80 -gencode arch=compute_90,code=sm_90`, the `__CUDA_ARCH_LIST__` spelling gives `fatal error C1012: unmatched parenthesis`, and the same build passes with `__CUDA_ARCH__`. Single-arch sm_80/86/89/90 all still select the right branch.

I did not convert the other `__CUDA_ARCH_LIST__` uses in this file, because they are not equivalent. Lines 28, 46, 446, 609, 727, 756 and 799 guard *declarations* (`struct` specializations, function templates), and `__CUDA_ARCH__` is undefined during the host compilation pass, so those declarations would disappear from it. `__CUDA_ARCH__` is correct specifically for the two sites here, which guard statements inside a `__device__` function body. That is also the dominant convention in this repo: 154 uses across 15 files in `src/tl_templates/cuda/`, against 23 uses of `__CUDA_ARCH_LIST__` in 3.

Converting the declaration-level guards would need a different fix and is outside this bugfix.

src/tl_templates/cuda/atomic.h
```diff
@@ -119,55 +119,56 @@ TL_DEVICE void tl_atomic_add_bf16(unsigned short &ret, unsigned long long addr,
...
+
+TL_DEVICE void tl_atomic_add_v2_bf16(unsigned short &ret_x,
+                                     unsigned short &ret_y,
+                                     unsigned long long addr,
+                                     unsigned short val_x, unsigned short val_y,
+                                     int memory_order) {
   if (IsReleaseMemoryOrder(memory_order)) {
     asm volatile(
-        "atom.release.gpu.global.add.noftz.v2.f16 {%0,%1}, [%2], {%3,%4};"
+        "atom.release.gpu.global.add.noftz.v2.bf16 {%0,%1}, [%2], {%3,%4};"
```
SiriusNEO, line 177: Why use bf16 inside a f16 helper here

SiriusNEO (changes requested):

src/tl_templates/cuda/atomic.h
arcusbuilds: That asm is inside `tl_atomic_add_v2_bf16`, not the f16 helper. The diff is misleading here: two separate changes landed in one hunk. `tl_atomic_add_v2_f16` was deleted and replaced by the packed `tl_atomic_add_f16x2` above it (vector-operand `atom` needs sm_90, while the packed 32-bit `f16x2` form works from sm_60), and the pre-existing `tl_atomic_add_v2_bf16` gained `.noftz`, which ptxas requires for `.bf16`. Git happened to align the removed f16 signature against the bf16 signature, so the hunk reads as if an f16 helper switched to bf16 asm. In the file itself they're separate functions: [`tl_atomic_add_f16x2` (L126)](https://github.com/tile-ai/tilelang/blob/62b7c5eb13ad6e501bd744fdf2adfdefd9918fe5/src/tl_templates/cuda/atomic.h#L126) is all f16, and [`tl_atomic_add_v2_bf16` (L146)](https://github.com/tile-ai/tilelang/blob/62b7c5eb13ad6e501bd744fdf2adfdefd9918fe5/src/tl_templates/cuda/atomic.h#L146) is all bf16.

SiriusNEO: I see, my bad

SiriusNEO (commented):
It seems that only vectorized AtomicAdd is correctly fixed with`memory_order` passed, like atomic_addx2 or x4. Should we fix for all atomic add cases?

arcusbuilds:
@SiriusNEO The other paths already carry the order. Scalar `T.atomic_add`, `atomic_max`, and `atomic_min` pass `memory_order` through on both the extern and tile-region paths, and max/min never reach the vectorizer, since `GetVectorizedAtomicOp` maps only the add ops. #2688 is specifically about a supplied order being silently dropped when the region form auto-vectorizes, and that is the bug this PR fixes.

The explicit `T.atomic_addx2` / `T.atomic_addx4` intrinsics are a different case: they take no `memory_order` parameter at the language level, so there is no supplied order to drop; they have always been relaxed. Wiring the parameter through is an API addition (the device helpers already accept it), and it would make the currently unreachable ordered `Ret` paths reachable, which needs its own tests. I'll file a separate issue to track that, so this PR stays scoped to the wrong-code bug.

arcusbuilds:
@SiriusNEO  gentle ping, I think this is ready, Let me know your thoughts, and if I've missed anything I'll work on it.

SiriusNEO (approved):

Issue #2936, opened 2026-08-10
[BUG] `T.gemm` outputs NaN when kernel has multiple thread blocks


One-sixth:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### What version of TileLang are you using?

0.1.13 (also reproduced on 0.1.12)

### System information

- TileLang version: 0.1.13 (also reproduced on 0.1.12)
- GPU: NVIDIA RTX 4060 Laptop (Ada Lovelace, Driver 610.88)
- CUDA: 13.0
- OS: Windows 11
- Python: 3.13

## Description

`T.gemm` produces incorrect results (all NaN / garbage values) when the `T.Kernel` grid has more than 1 block in any dimension. The same `T.gemm` call works correctly with a single-block grid `(1, 1)`.

## Expected Behavior

`T.gemm` should produce the same result regardless of the number of thread blocks. Multi-block execution should give correct results (within float32 precision, max_diff ~0.016).

## Actual Behavior

With `T.Kernel(1, 16, threads=64)`, all `T.gemm` outputs are incorrect — either NaN, extremely large garbage values (~1e38), or wrong values. With `T.Kernel(1, 1, threads=64)`, the same computation is correct.

The kernel does execute and write to the output tensor (output values differ from initialization), but the computed values are garbage.

## Additional Observations

- The bug occurs with `threads=64` (2 warps). Also tested with `threads=128` (4 warps) — same result.
- Matrix size: `[16, 64] @ [64, 16] = [16, 16]`
- `T.Parallel` and `T.copy` work correctly in multi-block mode; only `T.gemm` is affected.
- The kernel **does execute** (output is not the initial value), but `T.gemm` produces wrong results.
- **First call returns NaN, subsequent calls return stable-but-wrong values** (e.g. ~2.3e33). This suggests uninitialized shared memory or register state within `T.gemm`'s internal implementation — the first execution reads garbage (NaN), subsequent executions read the same garbage from the previous run (stable but wrong).
- The issue is likely within `T.gemm`'s internal workspace initialization, not the user-facing shared memory buffers.

## Workaround

Replacing `T.gemm` with manual matrix multiplication using `T.Parallel` + sequential loops works correctly in multi-block mode:

```python
for i, j in T.Parallel(BC, BC):
[...]

PR #2939, opened 2026-08-11
Warn when a T.gemm accumulator is not provably initialized

RyanL2:
## What

Adds `VerifyGemmAccumInit`, a compile-time diagnostic that warns when a `T.gemm` accumulator cannot be shown to be initialized before use.

```
Warning: Uninitialized T.gemm accumulator: 1 accumulator(s) read before being written
  [1] `c_f` is passed to T.gemm without being initialized
  --> kernel.py:40:1
T.gemm accumulates into C, so an uninitialized accumulator adds into stale register
contents. Add `T.clear(C)` before the gemm, or pass `clear_accum=True`. ...
```

## Why

`T.gemm` accumulates into its C operand and `clear_accum` defaults to `False`, so a fragment that is never written is read while uninitialized. That is undefined behaviour: it silently produces correct results on some architectures and NaN on others.

This came out of #2936, where a kernel with no accumulator initialization returned NaN on sm_89. On sm_86 the same kernel returns correct results in every configuration I tested, which is what makes this failure mode expensive to diagnose — nothing is wrong with the compiler, and the symptom does not reproduce on the machine you are debugging on. The generated CUDA tells the whole story:

```c
float c_f[4];                                    // declared, no initializer
...
tl::mma_sync<...>(c_f, A_local, B_local);        // reads c_f at ki == 0
```

## How

The pass walks the body in execution order before `LowerTileOp` — after that point `tl.gemm` is expanded into MMA intrinsics and the accumulator's identity is no longer recoverable. It tracks which buffers have been definitely written, and reports any `tl.gemm` whose C operand is unwritten *and* whose `clear_accum` argument is the literal `false`. Findings are aggregated into a single warning, following `VerifyParallelLoop`.

A write is recognised from the access mask on a `tl.region` argument (so `T.clear`/`T.fill`, a `T.copy` destination, and a prior gemm on the same buffer all count), or from a direct `BufferStore` (a `T.Parallel` loop body).

The rule is deliberately permissive: a write anywhere earlier in execution order counts, including inside a loop or conditional the gemm is not part of. A false positive on correct code is worse than missing an exotic case.

The pipelined `clear_accum=(k == 0)` idiom is quiet automatically — `k == 0` is not a literal `false`, so it is never reported. No pattern-matching on that idiom was needed.

## Testing

- Five unit tests: warns when uninitialized; silent with `T.clear`; silent with `clear_accum=True`; enabled by default; disabled via `tl.disable_gemm_accum_init_check`.
- **False-positive sweep over real in-tree GEMM kernels** — `testing/python/kernel/test_tilelang_kernel_gemm{,_batched,_simt,_with_stride}.py` and `examples/gemm/test_example_gemm.py`: **22 passed, 0 warnings**.
- End-to-end through the full JIT pipeline: the #2936 reproducer warns; the same kernel with `T.clear` or `clear_accum=True` does not; `@tilelang.jit(pass_configs={PassConfigKey.TL_DISABLE_GEMM_ACCUM_INIT_CHECK: True})` suppresses it.
- Regression check: `testing/python/transform/` — 316 passed, 87 skipped. `testing/python/language/` — 989 passed, 170 skipped. **Zero accumulator warnings across either suite.** Five pre-existing failures in the language suite (`get_warp_info`, `let`, `ptr`, `tma_1d`) reproduce identically on `origin/main` without this change; they are unrelated and hardware-specific to my sm_86 box.

## Notes

The check is **on by default**, unlike the data-race check. It only fires when `clear_accum` is literally false and no write to the accumulator was seen anywhere earlier, so correct kernels do not trip it — and the sweep above found no in-tree kernel that does. A diagnostic that was off by default would not have helped the reporter of #2936, which is the point of the pass. Happy to flip it to opt-in if you would rather it match the race check.

`T.gemm_sp` has the same `clear_accum` semantics (`args[11]`) and is not covered here, to keep this reviewable. Glad to follow up.

## Summary

Adds `VerifyGemmAccumInit`, a compile-time diagnostic for `T.gemm` accumulators that lack a provable prior initialization.

- Tracks writes from `T.clear`, `T.fill`, `T.copy` destinations, prior GEMMs, and direct `BufferStore` operations.
- Warns when the C operand is unwritten and `clear_accum` is literally `false`.
- Aggregates findings into one warning.
- Runs before `LowerTileOp` in CPU, CUDA, Metal, ROCm, and WebGPU pipelines.
- Enables the check by default.
- Supports disabling through `tl.disable_gemm_accum_init_check` and `PassConfigKey.TL_DISABLE_GEMM_ACCUM_INIT_CHECK`.
- Leaves `clear_accum=True`, prior initialization, and non-literal expressions such as `clear_accum=(k == 0)` unreported.
- Does not cover `T.gemm_sp`.
- Adds regression tests for supported initialization paths, warning aggregation, and default and disabled configurations.

## C++ style / lint notes

The PR changes C++ files but does not change rules documented in `docs/developer_guide/cpp_style.md`. The C++ API Style Audit (warning only) may report advisory TLCPP003/TLCPP004 findings. These findings are separate from correctness, build, and test results and should not block the PR unless they indicate a clear API, FFI, or maintainability risk.

## Testing

GEMM kernel tests and transform/language suites pass with no new accumulator warnings. Five unrelated pre-existing language-suite failures reproduce on the target branch.

### DIFF
--- a/src/op/builtin.cc
+++ b/src/op/builtin.cc
@@ -31,6 +31,7 @@ TVM_REGISTER_PASS_CONFIG_OPTION(kLayoutVisualizationEnable, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kLayoutVisualizationFormats, ffi::String);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDeviceCompileFlags, ffi::Array<ffi::String>);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableDataRaceCheck, Bool);
+TVM_REGISTER_PASS_CONFIG_OPTION(kDisableGemmAccumInitCheck, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kDisableLoopUnswitching, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kLoopUnswitchingAllowNonTrivialElse, Bool);
 TVM_REGISTER_PASS_CONFIG_OPTION(kIfStmtBindingInlineReplayableBinds, Bool);
--- a/src/op/builtin.h
+++ b/src/op/builtin.h
@@ -80,6 +80,15 @@ static constexpr const char *kLayoutVisualizationFormats =
 static constexpr const char *kDeviceCompileFlags = "tl.device_compile_flags";
 static constexpr const char *kDisableDataRaceCheck =
     "tl.disable_data_race_check";
+
+/*!
+ * \brief Disable the T.gemm accumulator-initialization check.
+ *
+ * The check is enabled by default; it only fires when clear_accum is
+ * literally false and no write to the accumulator precedes the gemm.
+ */
+static constexpr const char *kDisableGemmAccumInitCheck =
+    "tl.disable_gemm_accum_init_check";
 static constexpr const char *kDisableThreadStorageSync =
     "tl.disable_thread_storage_sync";
 static constexpr const char *kForceLetInline = "tl.force_let_inline";
--- a/src/transform/verify_gemm_accum_init.cc
+++ b/src/transform/verify_gemm_accum_init.cc
@@ -1,0 +1,193 @@
+/*!
+ * \file verify_gemm_accum_init.cc
+ *
+ * Warn when a T.gemm accumulator is not provably initialized before use.
+ *
+ * T.gemm accumulates into its C operand, so a fragment that is never written
+ * before the first T.gemm is read while uninitialized. That is undefined
+ * behaviour: it silently yields correct results on some architectures and NaN
+ * on others (see issue #2936). This pass reports it at compile time.
+ */
+
+#include "../op/gemm.h"
+#include "../op/region.h"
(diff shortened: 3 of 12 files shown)


LeiWang1999:
Thanks for the contribution! However, the approach feels a bit too gemm-specific to me. The root cause is more general — a non-global-scope buffer being read before any write to it — and I'd prefer to see a more general pass that captures that. Maybe we can open another PR to address this properly, closed for now.

RyanL2:
Makes sense, thanks. I checked the existing tile-op access-region infrastructure and the general form looks feasible before LowerTileOp and tile ops already expose their read/write regions.

So I’ll rework this into a general non-global-buffer read-before-write verifier. I’m planning to keep it conservative/index-insensitive so it can remain default-on: any prior write to the buffer counts as initialization, and opaque pointer/address escapes will be treated as writes rather than risking false positives. I’ll reuse the config/test plumbing from #2939 and open a separate PR.

Issue #2897, opened 2026-08-06
[Feature Request][RFC] Redesign reducers as ownership-safe deferred reductions


LeiWang1999:
## Required prerequisites

- [x] I searched the issue tracker and did not find an existing design proposal covering first-class reducer update semantics. This RFC is related to, but broader than, the concrete wrong-code reports linked below.

## Motivation

TileLang needs a way to express **deferred reductions**: contributions may be produced across multiple `T.Parallel` loops and multiple `T.Pipelined` iterations, each thread should accumulate local partials first, and cross-thread communication should happen only once at the end.

This is useful for kernels such as the in-tree GEMV reducer example:

```python
acc = T.alloc_reducer((block_M,), accum_dtype, op="sum", replication="all")
T.clear(acc)

for k_tile in T.Pipelined(...):
    # load A_frag and x_frag
    for i, j in T.Parallel(block_M, block_N):
        acc[i] += A_frag[i, j] * x_frag[j]

T.finalize_reducer(acc)
```

Using `T.reduce_sum` once per tile would introduce a collective per pipeline iteration. Materializing the entire reduction domain and reducing once would be too expensive. A reducer is therefore useful, but its semantic purpose should be **local monoid accumulation followed by one deferred collective**, not a user-selected fully replicated fragment layout.

The current implementation conflates these two concepts. That has produced silent wrong-code, unclear ownership rules, and special cases throughout layout inference and verification.

This RFC proposes making reducer updates first-class operations, defining a compositional correctness lowering, and treating subgroup ownership analysis as an optional optimization rather than a requirement for correctness.

Inspected against `main` at `fef9f760920fa5e81596e08696ee3c6e8e1c18ce`.

## Current model

The public API currently returns an ordinary fragment buffer and attaches reducer metadata to the surrounding block:

```python
acc = T.alloc_reducer(shape, dtype, op="sum", replication="all")
T.fill(acc, 0)
acc[i] += value
T.finalize_reducer(acc)
```

Relevant implementation choices:

1. [`alloc_reducer`](https://github.com/tile-ai/tilelang/blob/fef9f760920fa5e81596e08696ee3c6e8e1c18ce/tilelang/language/allocate.py#L283-L319) allocates a `local.fragment` buffer and records string metadata for `op` and `replication`.
[...]

PR #2881, opened 2026-08-05
[BugFix] Deduplicate replicated reducer updates

KellyFrog:
An issue with replicated reducer fragment buffers will occur when loaded with `T.Parallel`. This can be reproduced with:

```python
@tilelang.jit(
    out_idx=None,
    target="cuda",
    pass_configs=PASS_CONFIGS,
)
def reduce_with_finalize_reducer_tl():
    @T.prim_func
    def _reduce_with_finalize_reducer(
        x: T.Tensor((BD,), T.float32),
        out: T.Tensor((1,), T.float32),
    ) -> None:
        with T.Kernel(1, threads=THREADS):
            x_frag = T.alloc_fragment((BD,), T.float32)
            T.copy(x, x_frag)

            total = T.alloc_reducer((1,), T.float32, replication="all")
            T.fill(total, 0.0)
            for j in T.Parallel(BD):
                total[0] += x_frag[j]
            T.finalize_reducer(total)

            if T.get_thread_binding() == 0:
                out[0] = total[0]

    return _reduce_with_finalize_reducer
```

This produces cuda codes like:

```cuda
  float x_frag[1];
  float total[1];

  x_frag[0] = x[threadIdx.x & 7]; // !?
  total[0] = 0;
  total[0] += x_frag[0]; // !?
  AllReduce<SumOp, 128>(total[0]);
```

Reducer fragment buffers have long lifespan and can interact with multiple `T.Parallel`s and buffers, making its layout hard to analyze.

For conveneince and correctness, this issue is fixed by only enabling the first replicate (e.g. `rep == 0`) to write to reducer buffers. Now the following code will be generated, leaving sementics with `T.Parallel` unchanged:

```cuda
  float x_frag[1];
  float total[1];

  x_frag[0] = x[threadIdx.x & 7];
  total[0] = 0;
  if ((threadIdx.x >> 3) == 0) {
    total[0] += x_frag[0];
  }
  AllReduce<SumOp, 128>(total[0]);
```

## Summary

- Guard stores to fully replicated reducer buffers so only the canonical replica (`REP=0`) updates each buffer.
- Pass replicated reducer buffers through `PartitionLoop` and `LowerParallelLoop`.
- Preserve reducer buffer lookup order and verify buffer identity with `Buffer::same_as`.
- Add CUDA regressions for short and mixed parallel extents.
- Mark the reducer partition hook as temporary in the documentation.

## C++ style / lint notes

[...]

PR #2940, opened 2026-08-11
[Lang] Reducer v2: first-class deferred reduction epochs with planned physical lowering

### LINKED ISSUE #2408: [BUG][Fuzzer][wrong-code]  `T.alloc_reducer(replication="all")` + `T.finalize_reducer` silently returns `floor(threads / block_N)` × the correct sum instead of `A @ x` when `block_N` doesn't divide the thread count
### Required prerequisites

`0.1.11+cuda.gita4399e4c` (commit `a4399e4c`).

### System information

- GPU: NVIDIA A10G (`sm_86`)
- CUDA 12.8, PyTorch 2.8.0

### Problem description

A reducer allocated with `T.alloc_reducer(..., replication="all")` and finalized with `T.finalize_reducer` silently returns a result that is `floor(threads / block_N)` times too large when the reduce-loop tile width `block_N` does not divide the thread count — no error, every element uniformly scaled.

When `block_N` does not divide `threads`, the work in the `T.Parallel(block_M, block_N)` loop ends up replicated across `floor(threads / block_N)` thread groups that each compute the full sum, but `finalize_reducer` reduces over *all* threads with no check that the replica count is 1 — so the duplicate groups are summed in, multiplying the result. (Source: [`layout_reducer.cc`](https://github.com/tile-ai/tilelang/blob/a4399e4c/src/transform/layout_reducer.cc#L224) lays the reducer out fully-replicated; [`finalize_reducer.h`](https://github.com/tile-ai/tilelang/blob/a4399e4c/src/backend/common/op/finalize_reducer.h#L42-L53) reduces over the full thread extent; [`reduce.h`](https://github.com/tile-ai/tilelang/blob/a4399e4c/src/tl_templates/cuda/reduce.h#L223) sums all lanes.)

Trigger boundary (`M=N=512`, `block_M=64`, `threads=128`, fp16; median kernel/reference ratio on `|ref|>1` elements):

| `block_N` | `floor(threads/block_N)` | ratio, `replication="all"` | ratio, `replication="none"` (control) |
|-----------|--------------------------|----------------------------|----------------------------------------|
| 56        | 2                        | **2.00** (wrong)           | 1.00 (correct)                         |
| 48        | 2                        | **2.00** (wrong)           | 1.00 (correct)                         |
| 40        | 3                        | **3.00** (wrong)           | 1.00 (correct)                         |
| 24        | 5                        | **5.00** (wrong)           | 1.00 (correct)                         |
| 64        | 2                        | 1.00 (correct)             | —                                      |
| 120       | 1                        | 1.00 (correct)             | —                                      |
| 128       | 1                        | 1.00 (correct)             | —                                      |

The over-count is uniform across every output element (min/max element-wise ratio both within rounding of the factor). It only appears when `threads % block_N != 0` and `floor(threads/block_N) > 1`: `block_N=64` divides 128 cleanly (the layout then 2D-tiles rows × columns instead of pure column-replication, so there is no duplicate full-sum group) and is correct. `replication="none"` on the identical tile is correct for all widths, isolating the defect to the `replication="all"` / `finalize_reducer` path.

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import torch

M, N = 512, 512
block_M, block_N, threads = 64, 56, 128   # block_N=56 does NOT divide threads=128
dtype, accum_dtype = "float16", "float32"

@tilelang.jit
def gemv(A, x):
    A: T.Tensor((M, N), dtype)
    x: T.Tensor((N,), dtype)
    o = T.empty((M,), dtype)
    with T.Kernel(T.ceildiv(M, block_M), threads=threads) as bm:
        acc = T.alloc_reducer(block_M, accum_dtype, replication="all")
        T.clear(acc)
        for bn in T.Pipelined(T.ceildiv(N, block_N), num_stages=2):
            a_s = T.alloc_shared((block_M, block_N), dtype)
            T.copy(A[bm * block_M, bn * block_N], a_s)
            a_f = T.alloc_fragment((block_M, block_N), dtype)
            T.copy(a_s, a_f)
            x_f = T.alloc_fragment(block_N, dtype)
            T.copy(x[bn * block_N], x_f)
                acc[i] += a_f[i, j] * x_f[j]
        T.finalize_reducer(acc)
        T.copy(acc, o[bm * block_M])
    return o

torch.manual_seed(0)
A = torch.randn(M, N, device="cuda", dtype=torch.float16)
x = torch.randn(N, device="cuda", dtype=torch.float16)
out = gemv(A, x).float()
ref = A.float() @ x.float()
print("ratio (median):", float((out / ref)[ref.abs() > 1].median()))
# prints ~2.0 == floor(threads/block_N) == floor(128/56); should be 1.0
```

The kernel is the upstream [`examples/gemv/example_gemv.py` `gemv_alloc_reducer`](https://github.com/tile-ai/tilelang/blob/a4399e4c/examples/gemv/example_gemv.py#L230-L253) (its own example) with `block_M=64`, `threads=128`, and `block_N=56`.

The generated finalize reduces all 128 threads regardless of `block_N`:

```cpp
for (int __finred_0 = 0; __finred_0 < 64; ++__finred_0) {
  o_reducer[__finred_0] = tl::AllReduce<tl::SumOp, 128, 1, 0>::run(
      o_reducer[__finred_0], (&(((float*)workspace)[0])));
}
```

with the reducer accumulated under a `threadIdx.x % 56` lane mapping (active threads `< 112 = 2*56`), so two full-sum groups are summed → 2×.

### Traceback

```pytb
Silent
```

### Expected behavior

`out == A @ x`. The reduction is well-defined for any `block_N`, and the API places no documented constraint on `block_N` relative to `threads`, so the expected result is the correct sum — `finalize_reducer` accounting for the replica count (dividing the over-counted sum down, or summing only one replica group) would make every `block_N` correct.

If instead `block_N` dividing `threads` is meant to be a requirement of this lowering path, the path should reject a violating config at compile time rather than miscompile it — which is how the loud sibling already behaves: the fragment loop partitioner computes the same replicate count `num_thread / thread_extent` at [`loop_partition.cc`](https://github.com/tile-ai/tilelang/blob/a4399e4c/src/transform/loop_partition.cc#L217-L218), and a plain `T.reduce_sum` whose reduction width is smaller than the thread count raises an ICHECK there. The `replication="all"` reducer path reaches the analogous "replica groups > 1" situation through `FullyReplicated` + `finalize_reducer` but has no equivalent guard, so it fails silently instead.

Either way (correct result, or a compile-time error) is fine; the silent `k × (A@x)` is the part that isn't. Which of the two to adopt is a design call for the maintainers.

### Additional context

- The `replication="all"` reducer path is opt-in: `T.alloc_reducer(..., replication=None)` defaults to `"none"` ([`allocate.py` L289-L291](https://github.com/tile-ai/tilelang/blob/a4399e4c/tilelang/language/allocate.py#L289-L291)), so this only affects kernels that explicitly request `replication="all"`. The in-tree users of that path are the `gemv_alloc_reducer` example above and `testing/python/language/test_tilelang_language_reduce.py`, all of which use a `block_N` that divides the thread count (so the example's own test, at `block_N=128`/`threads=256`, does not exercise the non-dividing case).
- Introduced by [#757](https://github.com/tile-ai/tilelang/pull/757) ("Introduce `alloc_reducer` to separate inter and intra warp reduction"). The review there did not raise the replicate-count / `block_N`-divisibility case.
- Version note: on `0.1.11` (`a4399e4c`) this is the silent over-count above; earlier releases (≈`0.1.9`) may instead raise an error on some of these `block_N` widths via a different (layout div-by-zero) code path, so the observable symptom can differ by version even though the reducer lowering is the same.

LeiWang1999:
Fixes #2408. Design discussion: #2897. Supersedes the stopgap in #2881 (and the prototype direction of #2901).

## Problem

v1 implemented reducers as `local.fragment` buffers with a special layout, so no compiler pass owned the contribution-multiplicity semantics:

- loop-layout replication silently multiplied contributions — 128 threads summing 8 elements returned **576 instead of 36** (#2408);
- `FinalizeReducer` had to guess the reduction width from the buffer's `ReplicateExtent`, but the information needed (the update sites' loop layouts) was already lost by then;
- every stage carried buffer-identity special cases to compensate: `LayoutReducer` forced layouts before inference, `ParallelOp` skipped rep=ALL buffers, `LowerTileOp` reverse-looked-up `var_remap_` to collect `fully_replicated_reducer_buffers` for `PartitionLoop`, the vectorizer special-cased reducer stores. Missing any one of them silently changed contribution counts.

## Design

The reducer becomes a **first-class deferred reduction epoch**. Core invariant: `T.reducer_update` is the single source of contribution multiplicity; **physical layout can never change how many times a logical contribution is combined**.

```python
acc = T.alloc_reducer(shape, dtype, op="sum", seed=None)
T.reducer_init(acc)
for i, k in T.Parallel(M, K):
    T.reducer_update(acc[i], contribution)   # once per dynamic logical iteration
result = T.alloc_fragment(shape, dtype)
T.finalize_reducer(acc, result)              # out-of-place
```

- **`local.reducer` virtual scope + three first-class ops.** The reducer is opaque to all fragment machinery (default-deny by scope dispatch, no exemption lists). Ordinary reads/writes, `T.clear`/`T.fill`, aliasing, and in-place finalize are compile errors.
- **`VerifyReducerEpoch`** (early): lifecycle state machine + access rules, diagnostics point at user-written code.
- **`ReducerPlanAndMaterialize`** (after `LayoutInference`, before `LowerTileOp`): loop layouts are **read-only inputs** — the planner decides storage, communication, and execution multiplicity only, and materializes everything into ordinary IR. First version implements the canonical **FullParticipant wide plan**: identity-initialized per-participant partials, canonical-replica-guarded combine stores, participant-wide `AllReduce`, seed combined exactly once post-collective, copy into the destination fragment. Narrow plans (compact storage, sub-block collectives) are follow-ups that must fall back to exactly this lowering when their proofs fail.
- **Generic multiplicity marker** `tl.parallel_multiplicity`: `PartitionLoop` lowers it to a `REP == 0` guard knowing nothing about reducers (retires the `fully_replicated_reducer_buffers` compatibility hook and its TODO).
- **`VerifyReducerConsumed`** (after `LowerTileOp`): any surviving v2 construct is a compile error — a missed pass can never mean silently-wrong code.

## Legacy compatibility & v1 removal

**`CanonicalizeLegacyReducer`** rewrites v1 syntax into v2 ops on a strict whitelist (anything unrecognized is a compile error):

- `T.clear`/`T.fill` → `reducer_init`; non-identity fills on idempotent combines (max/min — documented in v1 as "clampers") are forwarded as one-time seeds; non-zero sum fills are rejected (their v1 behavior was participant-count-dependent);
- `acc[i] += v` / `acc[i] = T.max(acc[i], v)` → `reducer_update` (combine op and indices must match);
- in-place `T.finalize_reducer(acc)` → out-of-place finalize with a synthesized destination; post-finalize reads are redirected; `batch` annotations are forwarded end-to-end.

With the shim in place, the entire v1 lowering path is **deleted**: `LayoutReducer`, the `PartitionLoop`/`LowerParallelLoop` reducer-buffer plumbing, the `ParallelOp` rep=ALL special cases, and the vectorizer's reducer parameter chain. There is now exactly one reducer lowering path; the legacy syntax is a detachable shim on top of it (emits `DeprecationWarning`).

## Narrow physical plans

`ReducerPlanAndMaterialize` analyzes each epoch and, when a set of structural proofs succeeds, replaces the wide baseline with a **narrow plan**: compact per-thread partial storage (the update-site loop layout with reduction axes projected out via the same `ComputeReducerLayout` machinery `T.reduce` uses), unguarded updates (every replica accumulates exactly its assigned iterations), and a collective that combines only the thread-expression splits sourced from reduction axes:

- `Parallel(M) -> acc[M]` (and `Parallel(M){serial(K)}`): zero splits → **zero collectives/barriers/workspace** (the LocalComplete case);
- `Parallel(8) -> acc[0]` over 128 threads: `AllReduce<8>` in 16 independent replication groups instead of `AllReduce<128>`;
- `Parallel(M, K) -> acc[M]`: `AllReduce<K-group width>` instead of the full block.

Proof obligations (any failure deterministically falls back to the wide plan, never to an error): direct var-to-dim update indices, replica-safe contribution values, full-block thread coverage, a single power-of-two reduce step, structurally equal plans across update sites, and destination-layout containment (`ProveFragmentContains`). When the destination's free-level inferred layout mismatches but the destination is provably unconstrained (finalize-written, read only by copies to global) and the induced storage is one slot per thread, the planner replaces the destination layout instead of falling back.

The pass config **`tl.reducer_force_baseline`** disables narrow plans; differential tests assert forced-baseline and auto plans agree numerically. Plan decisions and fallback reasons are DLOG'd.

## Testing (H100, kernel cache disabled)

- New `test_tilelang_language_reducer_v2.py` (22 tests): the 36-vs-576 regression, deferred accumulation across serial tiles, multiple update sites, max/min, int32, seed-applied-once, narrow-plan codegen assertions (zero-collective LocalComplete, projected `AllReduce<8>`/`AllReduce<16>`), forced-baseline differentials, and 7 lifecycle/access diagnostics.
- Existing `test_tilelang_language_reduce.py` (93 tests) passes **unchanged** through the legacy shim — numerics, batched-AllReduce codegen assertions, sm_80 named barriers, and error-message tests.
- Pipelined GEMV example (legacy syntax, `T.Pipelined` + multi-tile accumulation) numerically verified through the shim.
- Copy suite passes (exercises the code paths the deleted plumbing lived in).

## Follow-ups (independently switchable, always falling back to the wide plan)

- Destination-layout steering during inference (multi-slot induced layouts, e.g. GEMV with 128 rows over 256 threads, currently fall back to wide).
- Multi-step collectives, batched narrow finalize, seed support in narrow plans.
- Per-thread footprint policy tuning for the wide plan (currently a hard 256-slot limit with a diagnostic).
- Reducer state transforms (e.g. FlashAttention online-softmax rescale) as a verified combine-endomorphism extension.

## Summary

- Introduces Reducer v2 with deferred reducer epochs and first-class initialization, update, and finalize operations.
- Adds lifecycle and access verification passes.
- Adds `ReducerPlanAndMaterialize` with narrow and wide physical plans.
- Uses deterministic wide-plan fallback and supports `tl.reducer_force_baseline`.
- Replaces reducer-specific partitioning hooks with `tl.parallel_multiplicity`.
- Converts legacy reducer syntax through `CanonicalizeLegacyReducer`.
- Removes the Reducer v1 lowering path and updates all backend pipelines.
- Fixes contribution over-counting for replicated reducers, including issue `#2408`.
- Adds coverage for numerical correctness, seeds, combine operations, code generation, baseline differentials, and diagnostics.

## C++ style / lint notes

- The PR changes C++ implementation and public C++ APIs.
- The “C++ API Style Audit (warning only)” CI step may report advisory findings for the new APIs.
- Warning-only findings are separate from correctness, build, and test failures.
- Do not block merge for TLCPP003 or TLCPP004 unless a finding identifies a clear API, FFI, or maintainability risk.


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2956, opened 2026-08-13
[Transform] Add VerifyBufferInit, a general buffer-initialization check

RyanL2:
# [Transform] Warn when a buffer is read before anything writes it

Follow-up to #2939, closed as too gemm-specific. This is the general check
suggested there: a non-global-scope buffer read at a point where nothing has
written it. Motivating case is #2936 — a `T.gemm` accumulator nothing
initializes, correct on one machine and NaN on another.

`VerifyBufferInit` walks each `PrimFunc` in execution order, warns, and returns
the function unchanged. It never fails a build.

It runs after `CanonicalizeLegacyReducer` / `VerifyReducerEpoch` and before
`PipelinePlanning` and `LowerTileOp`. That slot matters: tile operators still
carry their access regions, so writes from `T.copy`, `T.clear` and `T.fill` are
visible rather than opaque, and loop bodies are still in source order rather
than rotated across pipeline stages. Rebased onto #2940 — same point in the
sequence that `LayoutReducer` used to occupy.

## Why `GetReadBeforeWriteRegions()`

The one new piece of op-level API is a virtual on `TileOperatorNode` defaulting
to `GetAccessRegions().reads`, overridden only by `GemmNode` / `GemmSPNode`.

`GetAccessRegions()` answers **may-read** and is correctly conservative for the
dependency analysis it was built for. This check needs **must-read**: does the
op consume the region's prior contents before establishing a value of its own?

Gemm is where the two diverge. `GemmNode::GetAccessRegions()` keeps C in the
read set unless the clear is *proven* (`if (!is_one(clearAccum_))`), which is
right for pipelining but means the common `clear_accum=(k == 0)` idiom — neither
literal true nor false — reports C as read with nothing writing it beforehand.
That idiom is all over `examples/`, so a check consuming
`GetAccessRegions().reads` would false-positive on much of the tree.
`test_silent_for_the_pipelined_clear_accum_idiom` is the regression guard:
swapping the verifier back fails that test and only that test.

The override lives in the gemm ops, where the knowledge already lives, so the
verifier stays op-agnostic. Existing consumers of `GetAccessRegions()` are
untouched. The reducer v2 ops from #2940 are a useful check on that claim since
they landed after this was written: they inherit the default, need no override,
and are silent.

(Not to be confused with the `kAccessReadWrite` mask in `GemmNode::Gemm` —
`GemmNode` overrides `GetAccessRegions()` and never consults it.)

## Deliberate imprecision

The pass detects "nothing writes this buffer at all" and nothing finer — no
index-level or path-sensitive reasoning, so it will not catch partial
initialization. Four rules follow:

- **Scope-dependent ordering.** For per-thread storage a write counts if it
  appears earlier in the body — source order is that thread's execution order.
  Shared memory ignores order and asks only whether some *other* operation
  writes the buffer, because a warp-specialized kernel routinely places the
  producer branch after the consumer branch and coordinates through barriers.
  Excluding the reading operator's own write is what keeps #2936 reported.
- **Calls that may write are assumed to**, via `TCallEffectKind`; an op with no
  registered effect kind is assumed to write.
- **Exempt scopes.** `global` is the caller's responsibility; `shared.barrier`
  and `shared.cluster_barrier` hold mbarrier state the pass does not model.
  Parameters are exempt at any scope for the same reason as `global`.
- **A store establishes nothing until its value is evaluated**, so
  `x[i] = x[i] + 1` and the carry-forward `idx = T.if_then_else(cond, i, idx)`
  are findings. The store is also discounted as its own writer, so shared and
  per-thread scopes reach that conclusion alike.

**Known limitation**: a read the loop *guards* out of its own first iteration
(`if k > 0: C[k] = f[0]`) is reported though it never reads uninitialized
storage. Clearing it needs path sensitivity the pass does not attempt. An
*unguarded* loop-carried read is reported and correctly so. No guarded case
appears anywhere in `examples/`.

## Default-on, and the evidence for it

Enabled by default, with `tl.disable_buffer_init_check` to opt out.
`VerifyParallelLoop` is the precedent that had to default off, so the default
here is backed by measurement:

- **`testing/python/language`** (997 passing): zero warnings, pass/fail counts
  identical to the same tree without this change.
- **All 285 `.py` under `examples/`** (104 are `example_*.py`): **three warnings
  in two places**, both real read-before-writes, no known false positive. The
  last commit fixes all three; afterwards, zero warnings and zero import
  failures, re-measured on top of #2940.

Neither finding is new — `lse_max_local` dates to #46 (Jan 2025),
`cur_batch_idx` to #515 (May 2025) — and **no runtime failure is claimed for
either.**

**`example_mha_inference.py`** never initializes `lse_max_local` and passes it
to `T.reduce_max(..., clear=False)`, which marks the destination as a read. It
is visible in the generated CUDA. On an RTX 3060 Ti the current form is
bit-identical to both fixes across five launches; seeding `+inf` instead turns
all 524288 outputs into NaN. Benign here is a property of the machine, not the
code. Note the pass found this through the *default* interface — reduce needs no
override, and the check is not gemm-shaped.

**`example_grouped_gemm_fwd.py`** and its `_bwd` counterpart carry
`cur_batch_idx` forward through itself with no initializer. The third file in
that directory, `example_grouped_gemm_fwd_ptr.py`, writes `cur_batch_idx = 0`
before the same loop and is silent — a second author, writing the same loop ten
months later (#1923), initialized it unprompted, which is the best evidence
this is an oversight rather than an idiom. One red herring worth naming: both
files fail their own `torch.allclose`, and that is **not** caused by this.
Adding the initializer leaves an unchanged 0.25 max abs error over ~3.4k of
1572864 elements, so the mismatch is pre-existing and untouched here.

The fix commit is three lines and is deliberately last — drop it without
touching anything else if you would rather those files go through their own
review.

## Tests

`testing/python/transform/test_tilelang_transform_verify_buffer_init.py`, 28
tests: the #2936 gemm kernel and its three silent variants; shared, `local` and
extern-filled buffers; global-scope and scoped-parameter reads; legacy and v2
reducer idioms; stores that read their own destination in both scope classes.
Two tests pin the scope-ordering rule from opposite sides — a shared buffer
written only by a later branch is silent, a fragment written only after its read
still warns — and the first fails under uniform order-sensitivity while the
second fails under uniform whole-function suppression. The rest cover
aggregated reporting and the config wiring end-to-end through `tilelang.compile`.

## Open questions

1. **Default-on** — one-line change to `should_enable_buffer_init_check` if you
   would rather see the check land quietly first.
2. **The gemm hint.** The analysis is op-agnostic, but the *wording* of a report
   checks for `GemmNode` to suggest `T.clear(C)` / `clear_accum=True`. If that
   reads as too gemm-aware, dropping to the generic message costs nothing
   structurally.

Closes the thread from #2939. Related: #2936.

## Summary

- Added `VerifyBufferInit`, a warning-only transform that detects reads from non-global buffers before writes reach them.
- Added scope-aware analysis for must-read and may-read regions, calls, opaque pointer escapes, shared memory, loops, and source spans.
- Added `GetReadBeforeWriteRegions()` with GEMM and GEMM-SP overrides.
- Enabled the pass across CPU, CUDA, Metal, ROCm, and WebGPU pipelines.
- Added `tl.disable_buffer_init_check` for opt-out control.
- Added 28 tests covering analysis, reporting, configuration, and pipeline behavior.
- Initialized buffers in three examples. All 285 example scripts now produce zero warnings.

## C++ style / lint notes

- The PR changes C++ files and adds a public C++ API.
- Review `docs/developer_guide/cpp_style.md` for applicable API and maintainability guidance.
- No correctness, build, or test issue is identified in the supplied summary.

### DIFF
--- a/tilelang/cpu/pipeline.py
+++ b/tilelang/cpu/pipeline.py
@@ -7,6 +7,7 @@ import tilelang
 from tilelang.backend.pass_pipeline.pipeline_utils import (
     LayoutVisual,
     allow_vectorize,
+    should_enable_buffer_init_check,
     should_enable_race_check,
     should_force_let_inline,
 )
@@ -27,6 +28,13 @@ def CPUPassPipelineBody(mod: IRModule, target: Target) -> IRModule:
     mod = tilelang.transform.Simplify()(mod)
     mod = tilelang.transform.CanonicalizeLegacyReducer()(mod)
     mod = tilelang.transform.VerifyReducerEpoch()(mod)
+    # Warn on buffers that are read before anything writes them.
+    # Runs after the reducer passes above, so legacy reducers have been
+    # canonicalized, and before PipelinePlanning and LowerTileOp, while
+    # loop bodies are still in source order and tile ops still declare
+    # their access regions.
+    if should_enable_buffer_init_check():
+        mod = tilelang.transform.VerifyBufferInit()(mod)
 
     mod = tilelang.transform.IfStmtBinding()(mod)
     mod = tilelang.transform.Simplify()(mod)
--- a/tilelang/cpu/pipeline.py
+++ b/tilelang/cpu/pipeline.py
@@ -7,7 +7,6 @@ import tilelang
 from tilelang.backend.pass_pipeline.pipeline_utils import (
     LayoutVisual,
     allow_vectorize,
-    should_enable_buffer_init_check,
     should_enable_race_check,
     should_force_let_inline,
 )
@@ -33,8 +32,7 @@ def CPUPassPipelineBody(mod: IRModule, target: Target) -> IRModule:
     # canonicalized, and before PipelinePlanning and LowerTileOp, while
     # loop bodies are still in source order and tile ops still declare
     # their access regions.
-    if should_enable_buffer_init_check():
-        mod = tilelang.transform.VerifyBufferInit()(mod)
+    mod = tilelang.transform.VerifyBufferInit()(mod)
 
     mod = tilelang.transform.IfStmtBinding()(mod)
     mod = tilelang.transform.Simplify()(mod)
--- a/examples/flash_decoding/example_mha_inference.py
+++ b/examples/flash_decoding/example_mha_inference.py
@@ -134,7 +134,7 @@ def flashattn(batch, heads, seqlen_q, seqlen_kv, dim, is_causal, block_M, block_
                 ],
                 lse_local,
             )
-            T.reduce_max(lse_local, lse_max_local, dim=0, clear=False)
+            T.reduce_max(lse_local, lse_max_local, dim=0, clear=True)
             for k in T.Pipelined(num_split):
                 T.copy(lse_local[k, :], lse_local_split)
                 for i in T.Parallel(block_M):
--- a/examples/grouped_gemm/example_grouped_gemm_bwd.py
+++ b/examples/grouped_gemm/example_grouped_gemm_bwd.py
@@ -33,6 +33,7 @@ def grouped_gemm_fwd(
 
         m_start_padded = bx * block_M
 
+        cur_batch_idx = 0
         for i in range(batch_count):
             in_cur_batch_idx = m_start_padded >= batch_padded_offsets[i]
             cur_batch_idx = T.if_then_else(in_cur_batch_idx, i, cur_batch_idx)
--- a/examples/grouped_gemm/example_grouped_gemm_fwd.py
+++ b/examples/grouped_gemm/example_grouped_gemm_fwd.py
@@ -82,6 +82,7 @@ def grouped_gemm(
 
         m_start_padded = bx * block_M
 
+        cur_batch_idx = 0
         for i in range(batch_count):
             in_cur_batch_idx = m_start_padded >= batch_padded_offsets[i]
             cur_batch_idx = T.if_then_else(in_cur_batch_idx, i, cur_batch_idx)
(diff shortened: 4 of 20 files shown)


LeiWang1999:
Thanks, would be better to remove claude co-author in commit message :)

RyanL2:
@coderabbitai resume

tilelang/cpu/pipeline.py
```diff
@@ -27,6 +28,13 @@ def CPUPassPipelineBody(mod: IRModule, target: Target) -> IRModule:
     mod = tilelang.transform.Simplify()(mod)
     mod = tilelang.transform.CanonicalizeLegacyReducer()(mod)
     mod = tilelang.transform.VerifyReducerEpoch()(mod)
+    # Warn on buffers that are read before anything writes them.
+    # Runs after the reducer passes above, so legacy reducers have been
+    # canonicalized, and before PipelinePlanning and LowerTileOp, while
+    # loop bodies are still in source order and tile ops still declare
+    # their access regions.
+    if should_enable_buffer_init_check():
```
LeiWang1999: Suggestion: read the config inside the C++ pass instead of gating in Python

  Could we drop should_enable_buffer_init_check() and have the pass check its own config? The pass function already receives the PassContext:

  auto pass_func = [=](PrimFunc f, IRModule m, PassContext ctx) {
    if (ctx->GetConfig<Bool>(kDisableBufferInitCheck, Bool(false)).value()) {
      return f;
    }
    ...
  };

LeiWang1999 (commented):
LGTM, some comment left.

tilelang/cpu/pipeline.py
RyanL2: Sounds good. Will do.

LeiWang1999 (approved):

Issue #1729, opened 2026-01-26
[Feature Request] Improve layout when small fragments cause excessive thread replication


LJC00118:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

### Motivation

TileLang code:
```python
import tilelang
import tilelang.language as T


@tilelang.jit
def get_qwq():
    @T.prim_func
    def main(
        A: T.Tensor[(2, 2560), T.float32],
        B: T.Tensor[(2, 2560), T.float32],
        C: T.Tensor[(2, ), T.float32]
    ):
        with T.Kernel(1, threads=256):
            tx = T.get_thread_binding(0)

            C_local = T.alloc_fragment((2, ), T.float32)
            T.copy(C, C_local)

            for i, j in T.Parallel(2, 2560):
                if C_local[i] >= 0:
                    B[i, j] = A[i, j]

    return main


kernel = get_qwq()
print(kernel.get_kernel_source())
```

generated code:
```cuda
#include <tl_templates/cuda/gemm.h>
#include <tl_templates/cuda/copy.h>
#include <tl_templates/cuda/reduce.h>
#include <tl_templates/cuda/ldsm.h>
#include <tl_templates/cuda/threadblock_swizzle.h>
#include <tl_templates/cuda/debug.h>
#ifdef ENABLE_BF16
#include <tl_templates/cuda/cuda_bf16_fallbacks.cuh>
#endif

extern "C" __global__ void main_kernel(const float* __restrict__ A, float* __restrict__ B, const float* __restrict__ C);
extern "C" __global__ void __launch_bounds__(256, 1) main_kernel(const float* __restrict__ A, float* __restrict__ B, const float* __restrict__ C) {
  float C_local[1];
  C_local[0] = C[(((int)threadIdx.x) & 1)];
  if ((((int)threadIdx.x) >> 1) == 0) {
    #pragma unroll
    for (int i = 0; i < 640; ++i) {
      float4 v_ = *(float4*)(A + (((((int)threadIdx.x) & 1) * 2560) + (i * 4)));
      tl::store_global_128_conditional((&(B[(((((int)threadIdx.x) & 1) * 2560) + (i * 4))])), (*(uint4 *)(&(v_))), (0x0p+0f/*0.000000e+00*/ <= C_local[0]));
    }
  }
}
```

[...]

PR #2960, opened 2026-08-13
[Layout][Inference] Add IO-aware cost model for free-mode selection

LeiWang1999:
## Summary

- Add an IO-aware policy for choosing among complete free-mode layout-inference attempts, addressing the thread-collapse pathology in #1729 and removing attempt-order tie-breaking for equal-register layouts.
- Score global-memory behavior from the in-tree CuTe layout algebra without materializing a production `(thread, slot) -> address` table or evaluating TVM expressions in the hot loop.
- Enable IO-aware selection by default while retaining `tl.layout_cost_model="register-count"` as the legacy register-count escape hatch.

## Design

For each connected free-mode component, layout inference still tries every member as the inference root. The new `LayoutCostModel` ranks each complete attempt lexicographically by estimated memory cost and then fragment register slots.

```text
issue = steps * repeat * threads * lane_bytes
bw    = repeat * coalescing_segments * segment_bytes
statement_cost = max(issue, bw)
attempt_cost   = sum(statement_cost)
```

- Pack `(logical coordinates, replica)` into `[thread, slot]` using the fragment's canonical mapping.
- Recover a plain CuTe layout with `LayoutFromTileLang`, check the partial `RightInverse` for bijectivity, and compose it with each global address layout.
- Flatten the resulting layouts into constant `(extent, stride)` modes once, then evaluate the warp/step scoring loop with plain `int64_t` arithmetic.
- Derive the widest sustainable vector width from coalesced slot/thread modes, sharing `MaxVectorLoadBits` with the vectorizer so the model and generated code use the same width policy.
- Count distinct memory segments per vector step and warp; replicated stores only charge the replica-zero lanes.
- Parameterize warp size from the target and keep segment geometry behind one binding point.
- Cache statement scores by operator and structurally equal layout so equivalent attempts reuse their result.

Statements outside the supported affine/constant geometry are charged a conservative worst case. An opaque or non-bijective candidate therefore cannot win by being cheaper to analyze.

The change also canonicalizes non-injective read-only broadcast fragments to full replication. Written fragments keep the stricter behavior because replication changes store semantics.

## Verification and tooling

- Add eight constructed case groups with structured golden snapshots for both IO-aware and legacy policies.
- Add lowered device-TIR vector anchors to check that the winning layout's predicted width is actually emitted.
- Add a Python CuTe mirror and an independent NumPy enumeration oracle for differential testing.
- Audit the C++ mode arithmetic against the temporary full-enumeration oracle before removing that oracle and its pass config; keep `run.py --cute` as the standing differential guard.
- Add debug-only inference/cost diagnostics and a repository skill documenting the Layout/Fragment, inference, and CuTe conventions.

## Validation

- `cmake --build build -j32`
- `python maint/layout_inference/run.py` — 32/32 model and variant checks passed
- `python maint/layout_inference/run.py --cute` — 88/88 CuTe-vs-NumPy statement checks passed
- `python maint/layout_inference/run.py --anchor` — 16/16 lowered vector-width anchors passed
- Layout-related Python test selection — 69 passed, 2 skipped
- C++ mode-arithmetic audit across the layout-relevant test corpus and harness — zero disagreements before oracle removal
- Repository pre-commit hooks passed

## Risk and follow-up

- `tl.layout_cost_model="register-count"` restores the historical register-count ordering.
- Unsupported expressions, symbolic geometry, sub-byte accesses, and failed CuTe recovery receive a conservative worst-case score.
- The model intentionally covers global-memory issue depth and coalescing, not shared-memory bank conflicts, cache reuse, latency, or occupancy.
- Segment granularity is currently a 128-byte approximation; hardware calibration of model-disagreement cases remains follow-up work.


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #2965, opened 2026-08-14
[BugFix][Hopper][Blackwell] Read the WGMMA/UMMA K-panel stride from the layout instead of the operand extent

### LINKED ISSUE #2964: [BUG][wrong-code] A shared `T.gemm` operand sliced along MN silently reads the wrong K panel when `block_K` spans more than one swizzle atom
## Summary

Passing a shared buffer **sliced along MN** as a `T.gemm` operand silently computes the wrong result whenever `block_K` spans more than one swizzle atom. No warning, no assertion, no illegal access — the kernel reads real data from the wrong place, so it just returns wrong numbers.

The same GEMM written with one shared buffer per slice is correct, which makes the slice/no-slice pair a clean A/B.

## Reproduction

`main@6af825e0`, H200 (sm_90a), CUDA 13.2, torch 2.13.

```python
import torch

import tilelang
import tilelang.language as T

M, N, K = 64, 128, 128
BM, BN, BK, W0 = 64, 128, 128, 64  # split BN into W0 + (BN - W0) = 64 + 64


def build(sliced: bool):
    @T.prim_func
    def main(
        a: T.Tensor((M, K), "bfloat16"),
        b: T.Tensor((N, K), "bfloat16"),
        c: T.Tensor((M, N), "bfloat16"),
    ):
            a_s = T.alloc_shared((BM, BK), "bfloat16")
            c0 = T.alloc_fragment((BM, W0), "float32")
            c1 = T.alloc_fragment((BM, BN - W0), "float32")
            T.clear(c0)
            T.clear(c1)
            if sliced:
                # One shared buffer, consumed as two row slices.
                b_s = T.alloc_shared((BN, BK), "bfloat16")
            else:
                # Same math, one buffer per slice.
                b_s0 = T.alloc_shared((W0, BK), "bfloat16")
                b_s1 = T.alloc_shared((BN - W0, BK), "bfloat16")
            for ko in T.Pipelined(T.ceildiv(K, BK), num_stages=0):
                T.copy(a[by * BM, ko * BK], a_s)
                if sliced:
                    T.copy(b[bx * BN, ko * BK], b_s)
                    T.gemm(a_s, b_s[0:W0, :], c0, transpose_B=True)
                    T.gemm(a_s, b_s[W0:BN, :], c1, transpose_B=True)
                else:
                    T.copy(b[bx * BN, ko * BK], b_s0)
                    T.copy(b[bx * BN + W0, ko * BK], b_s1)
                    T.gemm(a_s, b_s0, c0, transpose_B=True)
                    T.gemm(a_s, b_s1, c1, transpose_B=True)
            T.copy(c0, c[by * BM, bx * BN])
            T.copy(c1, c[by * BM, bx * BN + W0])

    return tilelang.compile(main, out_idx=[2], target="cuda")


torch.manual_seed(0)
a = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
b = torch.randn(N, K, device="cuda", dtype=torch.bfloat16)
ref = a.float() @ b.float().T

for sliced in (False, True):
    err = (build(sliced)(a, b).float() - ref).abs().max().item()
    print(f"{'one buffer, two slices' if sliced else 'two buffers      ':22s}  max_err = {err:.4f}")
```

```
two buffers             max_err = 0.1248     <- bf16 accumulation noise
one buffer, two slices  max_err = 43.3982    <- wrong
```

## Scope

Both operands and both slice axes, `block_M = block_N = 128`, slicing 64 + 64:

| operand | `block_K=64` | `block_K=128` | `block_K=256` |
|---|---|---|---|
| K-major A (`A_s[0:64, :]`) | ok | **err ≈ 113** | **err ≈ 121** |
| K-major B (`B_s[0:64, :]`, `transpose_B=True`) | ok | **err ≈ 101** | **err ≈ 107** |
| MN-major A (`A_s[:, 0:64]`, `transpose_A=True`) | ok | ok | ok |
| MN-major B (`B_s[:, 0:64]`) | ok | ok | ok |

Independent of `threads`, `num_stages`, warp specialization, and dtype (fp16 and bf16 both). MN-major operands are unaffected — that branch of the offset formula does not use the term below.

`block_K = 64` is correct only by accident: at bf16 that is exactly one 128B swizzle atom, so the buggy term is multiplied by zero (see below).

## Root cause

`tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py`, `wgmma_ss_atom` / `wgmma_rs_atom`:

```python
B_offset = (ki // bk_atom_size) * n_dim * b_swizzle_atom_elems + ...
```

When `block_K` exceeds one swizzle atom, the shared layout splits the K axis into panels and stores them **panel-major**. A `(128, 128)` bf16 tile decodes as

```
Sw<3,3,3> o 0 o (128,(64,2)):(64,(1,8192))
                              ^^^^ panel stride = 128 rows x 64 elems
```

so advancing to the next K panel steps by **the buffer's** MN extent. The formula instead reconstructs that step from `n_dim`, the extent of *this operand*. The two coincide only when the operand covers the whole buffer; for a slice `n_dim` shrinks while the panel spacing does not.

Concretely, for the repro above (bf16, 64-element atom, `bk_atom_size = 4`, so `ki // 4` selects the panel):

| operand | slice origin (correct) | `ki >= 4` should read | actually reads |
|---|---|---|---|
| `b_s[0:64, :]` | elem 0 | 8192 = row 0, panel 1 | 4096 = **row 64, panel 0** |
| `b_s[64:128, :]` | 4096 ✓ | 12288 = row 64, panel 1 | 8192 = **row 0, panel 1** |

Emitted CUDA — the panel step is 8192 bytes where it should be 16384:

```cpp
tl::initialize_wgmma_descriptor<1, 1, 64>(desc_b_1, (char*)&b_s[0] + 8192);  // origin: correct
tl::wgmma_ss<...>(..., desc_b_1 + (((ki >> 2) * 8192 + (ki & 3) * 32) >> 4), ...);
//                                            ^^^^ should be 16384
```

Two consequences worth calling out:

- The **slice origin is correct** (`slice_byte_offset` is derived from the layout). Only the panel step is wrong, so the first K atom of every operand is right and everything past it is wrong — the error is a partial-garbage mix rather than an obvious blow-up.
- The two slices read **each other's rows**, all within the buffer, so there is no out-of-bounds signal.

The same expression is duplicated in five more places, all with the same defect:

- `wgmma_sp_macro_generator.py` — sparse Hopper, `wgmma_ss` and `wgmma_rs`
- `tcgen05_macro_generator.py` — Blackwell, `tcgen05_ss_atom`, `tcgen05_ts_atom`, `tcgen05_blockscaled_atom`

## Why nothing caught it

`tilelang/language/gemm_op.py` used to assert `A_offset[-2] == 0`, i.e. reject an operand with a non-zero offset on the second-to-last axis. That assertion was dropped when the region-based lowering landed, on the premise that "every lowering consumes the complete operand BufferRegions". That is half true: the *origin* is consumed from the region correctly, but the *panel stride* is still reconstructed from emitter-level tile dims. The guard was removed and no equivalent check replaced it.

Existing coverage misses it on both backends:

- `test_tilelang_kernel_gemm_with_stride.py` does slice a shared operand, but at `block_K = 32` (single atom) and on the pre-Hopper `mma` path.
- `test_tilelang_language_tcgen05_sliced_operands.py` slices only along **K**, which is exactly the axis where the panel term is unused. MN slicing is validated for atom alignment but never numerically.

## Environment

- tilelang `main@6af825e0`
- H200 (sm_90a), nvcc 13.2, driver 595.71.05, torch 2.13.0+cu130
- Also reproduces with `TL_DISABLE_WARP_SPECIALIZED` both on and off

I have a fix and will open a PR shortly.

bigSheep123:
Closes #2964.

A shared operand sliced along MN silently computed wrong results whenever `block_K` spanned more than one swizzle atom:

```python
T.gemm(a_s, b_s[0:64, :], c0, transpose_B=True)   # wrong past the first K atom
```

## The bug

When `block_K` exceeds one swizzle atom, the shared layout splits the K axis into panels and stores them **panel-major**. A `(128, 128)` bf16 tile decodes as

```
Sw<3,3,3> o 0 o (128,(64,2)):(64,(1,8192))
                              ^^^^ panel stride = 128 rows x 64 elems
```

so stepping to the next K panel moves by **the buffer's** MN extent. The atom offset formulas instead reconstructed that step from the operand's own extent:

```python
B_offset = (ki // bk_atom_size) * n_dim * b_swizzle_atom_elems + ...
```

Those coincide only when the operand covers its whole buffer. For a slice, `n_dim` shrinks while the panel spacing does not, so the panel step came out half (or a quarter, …) of the true value.

The slice **origin** was already layout-derived and correct, so only K atoms past the first were affected — the two halves of a split buffer read each other's rows, in bounds, with no diagnostic. Hence a silent wrong result rather than a crash.

`block_K == 64` was correct only by accident: at 16-bit dtypes that is exactly one 128B atom, so `ki // k_atom_size` is always zero and the bad multiplier was multiplied away. MN-major operands never reach the term at all.

## The fix

Take the panel stride from the decoded layout instead of reconstructing it. `cute.restrict` already returns the sublayout carrying it, and slicing does not change a stride — the number was sitting right there:

```python
k_panel_elems = None
if k_major:
    if isinstance(tile[k_idx].shape, tuple) and len(tile[k_idx].shape) == 2:
        k_panel_elems = tile[k_idx].stride[-1]
```

Exposed as `k_panel_elems` on `WGMMADescriptorParams` / `TCGEN05DescriptorParams`. It is `None` when the layout has a single K panel or the operand is MN-major — exactly the cases where the term is multiplied by zero — and those keep the previous expression, so their behaviour is untouched.

The same defect was duplicated across all seven atom emitters, all fixed here:

- `wgmma_macro_generator.py` — dense Hopper `wgmma_ss_atom` / `wgmma_rs_atom`
- `wgmma_sp_macro_generator.py` — sparse Hopper `wgmma_ss` / `wgmma_rs`
- `tcgen05_macro_generator.py` — Blackwell `tcgen05_ss_atom` / `tcgen05_ts_atom` / `tcgen05_blockscaled_atom`

Emitted CUDA for the repro, before → after:

```cpp
desc_b_1 + (((ki >> 2) *  8192 + (ki & 3) * 32) >> 4)   // before: operand extent
desc_b_1 + (((ki >> 2) * 16384 + (ki & 3) * 32) >> 4)   // after:  buffer extent
```

## Validation

H200 (sm_90a), nvcc 13.2, torch 2.13.0+cu130.

**The whole-buffer path is provably unchanged.** The risk in reading a stride off the layout is breaking the path that already worked, so I checked the invariant directly: across 120 K-major layouts (fp16 / bf16 / fp8_e4m3 / fp32 × 5 MN extents × 6 K extents), for **both** descriptor builders, `k_panel_elems` is either `None` or *exactly equal* to the old reconstruction — zero disagreements. The change is a strict generalization, not a replacement. Multi-panel layouts only arise under 128B swizzle; 32B/64B are always single-panel and take the fallback.

I also confirmed the `None` contract has no hole: every K-axis slice that spans more than one panel keeps the `(atom, panels)` structure with the correct buffer-derived stride, including panel-unaligned origins such as `K[64:192]`.

**New numeric test** — `test_tilelang_kernel_gemm_sliced_operand.py` cross-checks a sliced operand against one-buffer-per-slice over operand × majorness × `block_K`, asserting the two agree bit-for-bit. Before this change it fails on exactly the four K-major `block_K > 64` combinations and passes the other eight; after, all 12 pass.

**Regression suites**, all green (167 passed, 21 skipped):
`kernel/test_tilelang_kernel_gemm{,_with_stride,_batched}.py`, `tilelibrary/test_tilelang_tilelibrary_gemm_sp.py`, `issue/test_tilelang_issue_2928.py`, `language/test_tilelang_language_wgmma_gemm.py`, `language/test_tilelang_language_tcgen05_sliced_operands.py`, `cuda/test_tilelang_cuda_tcgen05_operand_layout.py`.

### One caveat on the Blackwell changes

I have no SM100 hardware, so the three `tcgen05` sites carry **no numeric verification** — every `tcgen05` numeric test skips here. What I could establish for them:

- the descriptor-parameter invariant above holds for `compute_umma_descriptor` too;
- a layout-level assertion added to `test_tilelang_cuda_tcgen05_operand_layout.py` pins the buffer-derived panel stride for an atom-aligned MN slice (hardware-independent, and it fails without the fix);
- the fallback expressions are statically in scope in all three functions, `ts_atom` and `blockscaled_atom` included.

I included them because it is one formula copied seven times and leaving four of the seven fixed seemed worse than fixing all of them. Happy to split the `tcgen05` half into its own PR if you would rather land it behind real SM100 validation.

Existing coverage missed this on both backends, for the record: `test_tilelang_kernel_gemm_with_stride.py` does slice a shared operand but at `block_K = 32` on the pre-Hopper `mma` path, and `test_tilelang_language_tcgen05_sliced_operands.py` slices only along K — the one axis where this term is unused.

## Summary

Fixes incorrect WGMMA and UMMA results when shared GEMM operands are sliced along MN and `block_K` spans multiple swizzle atoms.

- Derives K-panel spacing from decoded shared-memory layouts.
- Centralizes decoding in `decode_k_panel_elems`.
- Uses `k_panel_stride(mn_extent)` in dense and sparse Hopper WGMMA and Blackwell TCGEN05 emitters.
- Rejects unsupported multi-panel layouts instead of reconstructing unsafe strides.
- Preserves existing behavior for single-panel and MN-major layouts.
- Adds regression tests for sliced operands, operand types, majorness, multiple `block_K` values, and TCGEN05 descriptor spacing.
- Adds hardware-independent tests for valid and rejected layout forms.

Hopper paths passed numerical validation on H200. Blackwell changes have layout-level validation but no numeric SM100 validation.

### DIFF
--- a/testing/python/kernel/test_tilelang_kernel_gemm_sliced_operand.py
+++ b/testing/python/kernel/test_tilelang_kernel_gemm_sliced_operand.py
@@ -1,0 +1,178 @@
+"""A shared operand that is a slice of a wider buffer must match the same GEMM
+written with one buffer per slice.
+
+Regression coverage for the WGMMA K-panel stride: a K-major operand whose
+``block_K`` spans more than one swizzle atom is stored as (MN, (atom, panels)),
+and stepping from one K panel to the next moves by the *buffer's* MN extent. The
+offset formulas used to reconstruct that step from the operand's own extent,
+which is smaller for a slice, so ``T.gemm(A_s, B_s[0:64, :], ...)`` silently read
+the wrong panel for every ``ki`` past the first atom. ``block_K == 64`` (bf16 =
+one 128B atom, single panel) never exercised the term and stayed correct.
+"""
+
+import pytest
+import torch
+
+import tilelang
+import tilelang.language as T
+import tilelang.testing
+
+ACCUM = T.float32
+
+
+def _matmul_sliced(M, N, K, block_M, block_N, block_K, dtype, split, operand, transposed, sliced):
+    """Split one GEMM into two half-width ones over a shared operand.
+
+    ``operand`` selects which operand is split ("A" splits M, "B" splits N),
+    ``transposed`` gives that operand's ``transpose_`` flag (which decides whether
+    the split axis is the buffer's leading or trailing one), and ``sliced`` picks
+    the shape under test (one buffer consumed as two slices) versus the reference
+    (one buffer per half).
+    """
+    trans_a = operand == "A" and transposed
+    trans_b = operand == "B" and transposed
+    a_shape = (K, M) if trans_a else (M, K)
+    b_shape = (N, K) if trans_b else (K, N)
+    split_full = block_N if operand == "B" else block_M
+    rest = split_full - split
+
+    def a_tile(extent):
+        return (block_K, extent) if trans_a else (extent, block_K)
+
+    def b_tile(extent):
+        return (extent, block_K) if trans_b else (block_K, extent)
+
+    @T.prim_func
+    def main(
+        A: T.Tensor(a_shape, dtype),
+        B: T.Tensor(b_shape, dtype),
+        C: T.Tensor((M, N), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), T.ceildiv(M, block_M), threads=128) as (bx, by):
+            C0 = T.alloc_fragment((split, block_N) if operand == "A" else (block_M, split), ACCUM)
+            C1 = T.alloc_fragment((rest, block_N) if operand == "A" else (block_M, rest), ACCUM)
+            T.clear(C0)
+            T.clear(C1)
+
+            if operand == "B":
+                A_s = T.alloc_shared(a_tile(block_M), dtype)
+                if sliced:
+                    B_s = T.alloc_shared(b_tile(block_N), dtype)
+                else:
+                    B_s0 = T.alloc_shared(b_tile(split), dtype)
+                    B_s1 = T.alloc_shared(b_tile(rest), dtype)
+            else:
+                B_s = T.alloc_shared(b_tile(block_N), dtype)
+                if sliced:
+                    A_s = T.alloc_shared(a_tile(block_M), dtype)
+                else:
+                    A_s0 = T.alloc_shared(a_tile(split), dtype)
+                    A_s1 = T.alloc_shared(a_tile(rest), dtype)
+
+            for ko in T.Pipelined(T.ceildiv(K, block_K), num_stages=0):
+                if operand == "B":
+                    if trans_a:
+                        T.copy(A[ko * block_K, by * block_M], A_s)
+                    else:
+                        T.copy(A[by * block_M, ko * block_K], A_s)
+                    if sliced:
+                        if trans_b:
+                            T.copy(B[bx * block_N, ko * block_K], B_s)
+                            T.gemm(A_s, B_s[0:split, :], C0, trans_a, trans_b)
+                            T.gemm(A_s, B_s[split:block_N, :], C1, trans_a, trans_b)
+                        else:
+                            T.copy(B[ko * block_K, bx * block_N], B_s)
+                            T.gemm(A_s, B_s[:, 0:split], C0, trans_a, trans_b)
+                            T.gemm(A_s, B_s[:, split:block_N], C1, trans_a, trans_b)
+                    else:
+                        if trans_b:
+                            T.copy(B[bx * block_N, ko * block_K], B_s0)
+                            T.copy(B[bx * block_N + split, ko * block_K], B_s1)
+                        else:
+                            T.copy(B[ko * block_K, bx * block_N], B_s0)
+                            T.copy(B[ko * block_K, bx * block_N + split], B_s1)
+                        T.gemm(A_s, B_s0, C0, trans_a, trans_b)
+                        T.gemm(A_s, B_s1, C1, trans_a, trans_b)
+                else:
+                    if trans_b:
+                        T.copy(B[bx * block_N, ko * block_K], B_s)
+                    else:
+                        T.copy(B[ko * block_K, bx * block_N], B_s)
+                    if sliced:
+                        if trans_a:
+                            T.copy(A[ko * block_K, by * block_M], A_s)
+                            T.gemm(A_s[:, 0:split], B_s, C0, trans_a, trans_b)
+                            T.gemm(A_s[:, split:block_M], B_s, C1, trans_a, trans_b)
+                        else:
+                            T.copy(A[by * block_M, ko * block_K], A_s)
+                            T.gemm(A_s[0:split, :], B_s, C0, trans_a, trans_b)
+                            T.gemm(A_s[split:block_M, :], B_s, C1, trans_a, trans_b)
+                    else:
+                        if trans_a:
+                            T.copy(A[ko * block_K, by * block_M], A_s0)
+                            T.copy(A[ko * block_K, by * block_M + split], A_s1)
+                        else:
+                            T.copy(A[by * block_M, ko * block_K], A_s0)
+                            T.copy(A[by * block_M + split, ko * block_K], A_s1)
+                        T.gemm(A_s0, B_s, C0, trans_a, trans_b)
+                        T.gemm(A_s1, B_s, C1, trans_a, trans_b)
+
+            if operand == "B":
+                T.copy(C0, C[by * block_M, bx * block_N])
+                T.copy(C1, C[by * block_M, bx * block_N + split])
+            else:
+                T.copy(C0, C[by * block_M, bx * block_N])
+                T.copy(C1, C[by * block_M + split, bx * block_N])
+
+    return main
+
+
+def run_sliced_operand(M, N, K, block_M, block_N, block_K, dtype, split, operand, transposed):
+    torch_dtype = getattr(torch, dtype)
+    trans_a = operand == "A" and transposed
+    trans_b = operand == "B" and transposed
+
+    torch.manual_seed(0)
+    a = torch.randn((K, M) if trans_a else (M, K), device="cuda", dtype=torch_dtype)
+    b = torch.randn((N, K) if trans_b else (K, N), device="cuda", dtype=torch_dtype)
+    ref = (a.T if trans_a else a).float() @ (b.T if trans_b else b).float()
+
+    outputs = {}
+    for sliced in (False, True):
+        kernel = tilelang.compile(
+            _matmul_sliced(M, N, K, block_M, block_N, block_K, dtype, split, operand, transposed, sliced),
+            out_idx=[2],
+            target="cuda",
+            pass_configs={tilelang.PassConfigKey.TL_DISABLE_WARP_SPECIALIZED: True},
+        )
+        outputs[sliced] = kernel(a, b).float()
+
+    # The two shapes issue the same instructions over the same values, so they
+    # agree bit-for-bit; the fp32 reference only bounds the accumulation error.
+    torch.testing.assert_close(outputs[True], outputs[False], rtol=0, atol=0)
+    torch.testing.assert_close(outputs[True], ref, rtol=1e-2, atol=1)
+
+
+@tilelang.testing.requires_cuda
+@tilelang.testing.requires_cuda_compute_version_ge(9, 0)
+@pytest.mark.parametrize("operand", ["A", "B"])
+@pytest.mark.parametrize("transposed", [False, True])
+@pytest.mark.parametrize("block_K", [64, 128, 256])
+def test_gemm_sliced_shared_operand(operand, transposed, block_K):
+    """block_K 128/256 put more than one swizzle atom on the K axis."""
+    run_sliced_operand(
+        M=256,
+        N=256,
+        K=512,
+        block_M=128,
+        block_N=128,
+        block_K=block_K,
+        dtype="bfloat16",
+        split=64,
+        operand=operand,
+        transposed=transposed,
+    )
+
+
+if __name__ == "__main__":
+    tilelang.testing.main()
--- a/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py
+++ b/tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py
@@ -80,6 +80,14 @@ class WGMMADescriptorParams:
     buffer; passed to ``T.increase_descriptor_offset`` after building the
     descriptor from the buffer base. ``0`` for a whole-buffer / base-origin
     operand. Computed by :func:`compute_gmma_descriptor` from the CuTe layout."""
+    k_panel_elems: int | None = None
+    """Element stride between consecutive K swizzle-atom panels (K-major only),
+    read off the decoded layout. ``None`` when the layout has a single K panel,
+    or for an MN-major operand, in which case the ``ki // k_atom_size`` term the
+    offset formulas multiply it into is always zero / unused. Callers must not
+    reconstruct this as ``mn_extent * swizzle_atom_elems``: for an operand that
+    is a slice of a wider buffer (``B[64:128, :]``) the panels stay spaced by the
+    *buffer's* MN extent, not the operand's."""
 
 
 def compute_gmma_descriptor(tl_layout, buffer, transposed: bool, micro_size_k: int = 16, region=None) -> WGMMADescriptorParams:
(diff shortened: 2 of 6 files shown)


bigSheep123:
Thanks — both review points taken in 03c8eea6.

**The silent `None` fallback (valid, and worse than it looked).** You are right that the `None` → `mn_extent * swizzle_atom_elems` fallback is only sound for a single-panel layout and that nothing checked it. I had only verified the precondition *empirically* (`make_swizzled_layout` never produced a `None` for a multi-panel layout across a 120-layout grid), which says nothing about a custom layout supplied via `T.annotate_layout`. That is the same shape as the bug this PR fixes — a formula precondition assumed but not enforced — so it deserved a real guard, not a docstring.

The decode now lives in one shared `decode_k_panel_elems` and **asserts** when a swizzled operand spans more than one panel without exposing a constant panel stride. Panel count is rounded up so a K extent that is not a whole number of atoms cannot pass as single-panel.

While adding the guard I found a second hole in my own accept condition, which I think is the more serious of the two: reading `stride[-1]` as the panel step is only valid once the leading sub-mode is confirmed to be the contiguous atom. On a panels-first `(2, atom):(panel_stride, 1)` ordering it would have returned `1` — a *silently wrong* step rather than a refusal. Now checked, and covered by a test that asserts the refusal.

**The accessor (taken).** `k_panel_stride(mn_extent)` on both params dataclasses; the eleven restatements of the ternary are gone.

New `testing/python/cuda/test_tilelang_cuda_wgmma_operand_layout.py` exercises the decode directly and hardware-independently: canonical multi-panel strides, the single-panel / MN-major `None` cases, and each refusal path (panels-first, flat multi-atom, three-sub-mode K, partial-atom extent).

Re-verified that the tightening rejects nothing it previously accepted: the whole-buffer invariant is unchanged (120 K-major layouts × 2 builders, 60 single-panel, 0 disagreements), every multi-panel K slice still decodes to the buffer-derived stride, and the suite set is green (174 passed, 40 skipped).

One placement question for the maintainers: `decode_k_panel_elems` is layout-generic but currently lives in `wgmma_macro_generator.py` and is imported by `tcgen05_macro_generator.py`, which adds a Blackwell → Hopper module edge. Happy to move it under `tilelang/layout/` if you would rather not have that dependency.

bigSheep123:
Rebased onto `main` to refresh CI. No conflicts — none of the 11 upstream commits touched the files this PR changes. Re-verified on the new base (H200, sm_90a): the sliced-operand tests, both descriptor unit-test files, the gemm/wgmma suites and the sparse suite all pass, the whole-buffer invariant is still 0 disagreements across 120 K-major layouts × 2 builders, and the issue repro stays at accumulation noise.

Two things worth flagging for whoever picks this up.

### The red `CuTeDSL Examples` check is a hang in two examples this PR does not touch

It timed out at the 2h job limit. What the log shows:

- the examples suite ran at normal speed and reached **96% in 3m22s** (the job's recent successful runs take 2–4 min end to end), then produced no further output for 1h55m before the runner killed it;
- 86 collected, 83 completed. Exactly two tests started and never reported:
  - `examples/deepseek_v32/test_tilelang_example_deepseek_v32.py::test_example_sparse_mla_fwd_pipelined`
  - `examples/gemm_fp8/test_example_gemm_fp8.py::test_example_tilelang_gemm_fp8`

Both pass locally on this branch on Hopper, in 25s and 6s respectively. This PR adds no files under `examples/` and changes only descriptor-offset arithmetic, which is O(1) at compile time and cannot hang; and 83 of the 86 examples — including `test_example_gemm`, flash-attention, grouped-gemm and warp-specialize, all of which lean on the WGMMA descriptor path — passed with the change in place. A wrong descriptor step fails, it does not stall.

My guess is contention: the job runs 8 xdist workers against one GPU, and those two are among the heaviest. But that is a guess — I could not reproduce the stall, and there is no OOM or CUDA error in the log. For context the job's recent history is 5 cancelled / 5 skipped / 2 success, so it does not look reliable independently of this PR. If it trips again after the rebase, I would appreciate a maintainer's read on whether it is a known flake.

### CI does not cover the Blackwell half

The CUDA job runs on an H100, so the four `tcgen05` numeric tests skip there, exactly as they do on my hardware. The three `tcgen05` sites in this PR therefore have **no numeric verification anywhere** — only the layout-level assertion added to `test_tilelang_cuda_tcgen05_operand_layout.py`, the shared-decode unit tests, and the descriptor-parameter invariant.

I grouped them with the Hopper fix because it is one formula copied across seven emitters and leaving four of seven fixed seemed worse than fixing all of them. If you would rather land the Blackwell part behind real SM100 validation, say so and I will split it into its own PR — the Hopper half stands alone.

tilelang/cuda/intrinsics/macro/wgmma_macro_generator.py
```diff
@@ -30,6 +30,51 @@ def _min_leaf_stride(stride) -> int:
...
+        panel_stride = stride[-1]
+        if stride[0] == 1 and shape[0] == swizzle_atom_elems and isinstance(panel_stride, int):
+            return panel_stride
+    if swizzle_mode.is_none():
+        # Panels are a swizzle concept; an unswizzled operand has none.
+        return None
+    # Round up: a K extent that is not a whole number of atoms still spans two
+    # panels, and must not slip through as "single panel".
+    n_panels = -(-int(cute.size(k_mode)) // swizzle_atom_elems)
+    assert n_panels <= 1, (
```
LeiWang1999, line 70: The guard design here is exactly right, and I verified each arm behaves as documented:

- the panels-first misread case (`(panels, atom):(panel_stride, 1)`) is rejected by the `stride[0] == 1 and shape[0] == swizzle_atom_elems` check rather than silently returning `1`;
- a K extent that is not a whole number of atoms ceils to 2 panels and trips the assert instead of slipping through as single-panel;
- and most importantly, a swizzled multi-panel K mode that isn't canonical **asserts loudly instead of falling back to the extent-based reconstruction** — falling back would quietly reintroduce the exact wrong-code path this PR removes. Refusing is the right call.

One probe beyond the PR's matrix: `block_K = 192` (bf16 → **three** K panels, a non-power-of-two count, `ki // atom` multiplier reaching 2) — both K-major combos pass the sliced-vs-separate-buffers bit-for-bit check on an H100. The decode generalizes past the 2/4-panel cases the tests pin.

testing/python/kernel/test_tilelang_kernel_gemm_sliced_operand.py
```diff
@@ -0,0 +1,178 @@
...
+    torch.testing.assert_close(outputs[True], outputs[False], rtol=0, atol=0)
+    torch.testing.assert_close(outputs[True], ref, rtol=1e-2, atol=1)
+
+
+@tilelang.testing.requires_cuda
+@tilelang.testing.requires_cuda_compute_version_ge(9, 0)
+@pytest.mark.parametrize("operand", ["A", "B"])
+@pytest.mark.parametrize("transposed", [False, True])
+@pytest.mark.parametrize("block_K", [64, 128, 256])
+def test_gemm_sliced_shared_operand(operand, transposed, block_K):
```
LeiWang1999, line 161: Suggest adding `192` here: bf16 gives three K panels — a non-power-of-two panel count with the `ki // k_atom_size` multiplier reaching 2, which {64, 128, 256} never produces. I ran both K-major combos at `block_K=192` (M=256, N=256, K=384) on an H100 through this file's own `run_sliced_operand` and they pass bit-for-bit — cheap to pin.

LeiWang1999 (commented):
Reviewed with independent verification on an H100 (sm_90a). **Excellent PR — the analysis in the description is accurate down to the detail, and everything I could re-derive or re-run checked out. LGTM.**

What I reproduced and verified locally (`TILELANG_DISABLE_CACHE=1` throughout):

- **Negative control**: with only the test files applied (fix reverted), `test_tilelang_kernel_gemm_sliced_operand.py` fails on **exactly** the four claimed combos — K-major (`A`/untransposed, `B`/transposed) × `block_K` ∈ {128, 256} — and passes the other eight.
- **With the fix**: all 62 tests across the three touched/added test files pass (12 sliced-operand numeric cases, the `decode_k_panel_elems` unit tests, the tcgen05 layout pins).
- **Beyond the PR's matrix**: a `block_K = 192` probe (three K panels — non-power-of-two, panel multiplier reaching 2) passes both K-major combos bit-for-bit; details in the inline comment.
- **Regressions**: `kernel/test_tilelang_kernel_gemm.py`, `language/test_tilelang_language_wgmma_gemm.py`, `issue/test_tilelang_issue_2928.py` — 57 passed, 2 skipped.
- **Code-level**: all seven emitter sites are consistent (sparse A reads its panel stride off the *compressed* layout, matching the old expression's frame of reference), and the `None` contract of `k_panel_stride()` only engages where the `ki // k_atom_size` factor is provably zero.

On the tcgen05 caveat: I also have no SM100 hardware, so those three sites stay numerically unverified here too — but the shared decode, the descriptor-invariant sweep, and the hardware-independent layout pin make the risk about as contained as it can be without silicon. Fixing all seven copies of the formula was the right call.

One reviewer's note for anyone re-running the validation: a warm kernel cache **silently masks this fix** in a dev tree (the cache key doesn't cover the Python macro-generator sources, so kernels cached from the buggy build keep being served). `TILELANG_DISABLE_CACHE=1` before A/B-ing — it cost me a confused half hour.

PR #3036, opened 2026-08-17
[BugFix][CUDA] Support int4x2 and uint4x2 codegen

### LINKED ISSUE #2994: [BUG][Fuzzer][ice-on-valid-code] Width-2 auto-vectorized `int4`/`uint4` store crashes with `Cannot convert type int4x2 to CUDA type!` instead of compiling
### Required prerequisites

0.1.13 (latest release)

### System information

- TileLang 0.1.13, Python 3.13, Linux.
- GPU: NVIDIA L40S (compute capability 8.9, sm_89). The defect is in the CUDA type printer (`src/cuda/codegen/codegen_cuda.cc`), which is architecture-independent — the crash is at code-generation time, before any arch-specific lowering.

### Problem description

A plain `T.copy` between two `int4` (4-bit) buffers aborts at compile time with

```
InternalError: Cannot convert type int4x2 to CUDA type!
```

The kernel is legal: `int4` is a supported element dtype, and `T.copy` auto-vectorizes the transfer (here to a width-2 vector, `int4x2`). The CUDA type printer's `int4` branch has cases for vector widths 4, 8, 16, 32, and 64, but no case for width 2, so a width-2 vectorized `int4` transfer falls through to a fatal `LOG(FATAL)` instead of printing a type.

The width chosen is a function of the loop shape, so the crash is data-shape-dependent: 256 elements over 128 threads vectorizes to width 2 and crashes; the same copy over 64 or 32 threads vectorizes to width 4 (`int16_t`) or 8 (`int`) and compiles. A scalar (non-vectorized) `int4` copy also compiles and round-trips correctly, and the identical vectorized `T.copy` on the sibling `int8` dtype compiles — so this is specific to the missing width-2 `int4` type mapping, not to `T.copy` or to sub-word dtypes in general. The crash is **not `T.copy`-specific**: an element-wise `T.Parallel` assignment over the same `int4` buffer at the same thread count also aborts with the identical `int4x2` error, and the unsigned sibling `uint4` crashes the same way — see §17. It is the auto-vectorizer folding an `int4` store loop to width 2, times the missing `int4x2` case in the CUDA type printer.

### Reproducible example code

```python

N = 256

# int4 (4-bit) buffers. T.copy auto-vectorizes; at 128 threads over 256
# elements the transfer becomes width-2 (int4x2), which the CUDA type
# printer has no case for -> compile-time crash.
@T.prim_func
def crash(A: T.Tensor((N,), "int4"), C: T.Tensor((N,), "int4")):
        T.copy(A, C)

# Control: identical copy, scalar (threads=1, serial loop) -> not vectorized,
# uses the scalar int4 load/store helpers, compiles and round-trips.
@T.prim_func
def ok(A: T.Tensor((N,), "int4"), C: T.Tensor((N,), "int4")):
    with T.Kernel(1, threads=1) as bx:
        for i in T.serial(N):
            C[i] = A[i]

packed = torch.randint(-128, 128, (N // 2,), dtype=torch.int32).to(torch.int8).cuda()

k = tilelang.compile(ok, out_idx=[1])
out = k(packed)
print("control (scalar) round-trips:", torch.equal(out.view(torch.int8), packed))  # -> True

k = tilelang.compile(crash, out_idx=[1])   # -> InternalError: Cannot convert type int4x2 to CUDA type!
```

<details><summary>Boundary: which vector widths compile vs crash (same kernel, thread count varied)</summary>

| threads | int4 elems/thread → vector | CUDA type printed | result |
|---|---|---|---|
| 32  | `int4x8`  | `int`      | compiles |
| 64  | `int4x4`  | `int16_t`  | compiles |
| 128 | `int4x2`  | (no case)  | **crash** |

The width-2 case is the only one in the 2–64 range with no mapping. The `int8` sibling at the same shape compiles and round-trips.

</details>

### Traceback

```
[..] : Fatal: Cannot convert type int4x2 to CUDA type!
InternalError: Cannot convert type int4x2 to CUDA type!
```

### Expected behavior

The width-2 `int4` copy should compile, the same as widths 4/8/16/32/64 already do. `int4x2` is 8 bits and can be printed the same way width-4 (`int4x4` → `int16_t`) and width-8 (`int4x8` → `int`) are printed as a wider integer holding the packed nibbles; the width-4 path already round-trips correctly on the same kernel. Alternatively, if width-2 vectorization of `int4` is not intended to be supported, the copy vectorizer could avoid choosing width 2 for `int4`, so the type printer is never asked to emit `int4x2`.

### Additional context

**Root cause.** `PrintType` in the CUDA code generator does not handle a width-2 vector of the 4-bit `int4` dtype, so it emits a fatal error instead of a type. In [`CodeGenTileLangCUDA::PrintType`, `case 4:`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L952-L978) the `int4` branch enumerates `lanes()` 4, 8, 16, 32, and 64 and otherwise hits [`LOG(FATAL) << "Cannot convert type ... to CUDA type!"`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L978); there is no `lanes() == 2` case. The other sub-word integer branch (`case 8:` for `int8`) instead `break`s and falls through to the generic width-2..4 tail (`os << lanes`), which is why `int8` at the same shape compiles while `int4` aborts.

**Suggested fix.** Add a `lanes() == 2` case to the `int4` branch of `PrintType` (a width-2 `int4` is a single byte; print it alongside the [existing width-4 case](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L961)). If width-2 `int4` vectorization is not intended to be supported, the alternative is to gate the copy vectorizer away from width 2 for `int4`.

**Provenance.** The `int4` branch's missing width-2 case predates the [#2297](https://github.com/tile-ai/tilelang/pull/2297) refactor (merged 2026-06-02) that relocated this printer into TileLang's own `codegen_cuda.cc`; #2297 only moved the code, it did not write the width enumeration. The gap has been present since the `int4` type mapping was added; reproduced on 0.1.13, and the branch is unchanged on `main`.

**Dedup.** I searched the open and closed tracker and found no existing report of this defect. Distinct from [#2526](https://github.com/tile-ai/tilelang/issues/2526): that is a runtime `misaligned address` crash from an unconditional 16-byte load emitted for the CUDA `int4` **vector type** (four `int32`s) over a `make_tensor` view, rooted in `loop_vectorize.cc`. This is a compile-time type-printer crash for the 4-bit `int4` **dtype** vectorized to width 2.

**Reach.** `int4` is an accepted element dtype (`tilelang/language/dtypes.py`: `"int4": "Int4"`, `int4 = dtype("int4")`) and `T.copy` is a core op documented to move whole buffers; the crash is also reachable via a plain `T.Parallel` element-wise store (§17), so it is not gated on `T.copy`. In this repo's shipped code the actual `int4` *dtype* is otherwise unused as a buffer element type: `grep -rn '"int4"' examples testing` at v0.1.13 returns a single hit — a docstring mention in `examples/dequantize_gemm/quantize/lop3.py` (`"int4"` as a *quantization format string*, decoded via lop3 into fp16, not a `T.Tensor(..., "int4")` buffer). The other `int4` sites (`examples/bitnet-1.58b/...`, `examples/dequantize_gemm/example_dequant_gemm_fine_grained.py`) store int4 *values packed inside int8 buffers* and never instantiate an `int4`-dtype buffer or copy one. So **no shipped example or test exercises a vectorized `int4`-dtype copy** — verified by running (PART-1, §17): there is no example to run that hits this path, which is why CI is green. Given an actual `int4` buffer, the crashing width-2 vectorization is selected by the common default of one element more than a byte per thread (256 elements / 128 threads); other thread counts dodge it by landing on width 4 or 8.

**Impact.** The trigger is narrow: an `int4` buffer whose copy loop shape happens to vectorize to exactly width 2 (e.g. 256 elements over 128 threads), while nearby thread counts land on the mapped widths 4/8. When it fires it is a compile-time `LOG(FATAL)` — the kernel fails to build loudly and immediately, corrupting no data and never reaching a running workload silently; a data-shape change (thread count) makes it appear or disappear but the failure is always a hard abort, never a wrong value. Fixing it closes the width-2 hole in the `int4` type printer so that the full 2–64 vectorization range compiles as widths 4/8/16/32/64 already do.

<details><summary>§17 Generalization record</summary>

**Two-level root.**
- **SOURCE-level root:** `CodeGenTileLangCUDA::PrintType`, `case 4:` (the 4-bit `int` dtype-code branch) in `src/cuda/codegen/codegen_cuda.cc`. It enumerates `lanes()` ∈ {4, 8, 16, 32, 64} and any other lane count falls through to `LOG(FATAL) << "Cannot convert type ... to CUDA type!"`. There is no `lanes()==2` case and (unlike `case 8:`/`case 16:`) no `break` into the generic width-2..4 tail (`os << lanes`). The sign bit is not consulted, so signed and unsigned share the hole.
- **OPERATOR-level root:** the auto-**vectorizer** (`loop_vectorize.cc`) folding an `int4` store loop to width 2 whenever the loop shape (elems/thread) lands on 2. Any op whose store loop the vectorizer folds to `int4x2` reaches the printer hole — this is not tied to `T.copy`.

**PART 1 — example/doc-claim run result.** The draft's only Reach claim is a *negative* one (no shipped example exercises a vectorized `int4`-dtype copy → CI green). Verified by running: `grep -rn '"int4"' examples testing` at v0.1.13 returns exactly one hit, a docstring in `examples/dequantize_gemm/quantize/lop3.py` where `"int4"` is a **quantization format string** (decoded via lop3 into fp16), not a `T.Tensor(..., "int4")` buffer. The other `int4` mentions (`examples/bitnet-1.58b/...`, `examples/dequantize_gemm/example_dequant_gemm_fine_grained.py`) pack int4 values inside `int8` buffers and never instantiate an `int4`-dtype buffer. There is therefore **no shipped example that runs this path** — nothing to execute that would hit or dodge the bug; the negative Reach claim is confirmed by the source search, not by a passing example. (The draft's own repro and its scalar control were re-run at 0.1.13, below.)

**Found-first cell (re-run at v0.1.13).** `int4` buffer, `T.copy(A, C)`, `threads=128`, `N=256` → `int4x2` → `InternalError: Cannot convert type int4x2 to CUDA type!`. Still reproduces.

**4-axis sweep (each a fresh process, fresh cache, at v0.1.13; observed, not asserted):**

|---|---|---|---|
| baseline | `int4` `T.copy`, threads=128, N=256 | `raised: Cannot convert type int4x2 to CUDA type!` | — (found-first) |
| related-type (sign) | `uint4` `T.copy`, threads=128, N=256 | `raised: Cannot convert type uint4x2 to CUDA type!` | **same** — `case 4:` ignores sign at width 2 |
| related-operator | `int4` **`T.Parallel`** element-wise store, threads=128, N=256 | `raised: Cannot convert type int4x2 to CUDA type!` | **same** — same vectorizer→printer path, not `T.copy`-specific |
| related-type (scope) | `int4` `T.copy` **via `alloc_shared`**, threads=128, N=256 | `raised: Cannot convert type int4x2 to CUDA type!` | **same** — scope is not a mitigating knob |
| related-source (sibling case) | `int8` `T.copy`, threads=128, N=256 | `compiled`; round-trips `True`; generated source emits `char2` | **distinct** — `case 8:` `break`s into the `os << lanes` tail (→ `char2`); the hole is `case 4:`-only |
| related-type (width up) | `int4` `T.copy`, threads=64, N=256 (→ `int4x4`) | `compiled` | N/A — mapped branch (`lanes()==4`→`int16_t`); width-2-specific hole |
| related-type (width up) | `int4` `T.copy`, threads=32, N=256 (→ `int4x8`) | `compiled` | N/A — mapped branch (`lanes()==8`→`int`) |
| vectorization off | `int4`, threads=1, serial loop | `compiled` + round-trips `True` | N/A — scalar path never asks `PrintType` for a vector `int4` |
| similar-logic (latent) | `bool`/`int1` `T.copy`, threads=64 N=256 and threads=128 N=512 | `compiled` (both) — source shows scalar `bool`/`char`, no wide-lane vector | **latent sibling, unreached** — see below |

**Reframe.** Broadened from "`T.copy` of `int4`" toward the class **"any auto-vectorized store of an `int4`/`uint4` buffer that folds to width 2"**: signed and unsigned both crash, and `T.Parallel` (not just `T.copy`) reaches it, so the trigger is the *vectorizer picking width 2 for a 4-bit dtype*, not the op. Kept the width-2 boundary specific (widths 4/8/16/32/64 all compile). One fix (add `lanes()==2` to `case 4:`) closes it for all these faces.

**Similar-logic sibling `case 1:` (int1/bool) — same-shaped source hole, could NOT reach via this surface.** [`case 1:`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L935-L951) has the identical construct: it enumerates `lanes()` ∈ {8, 16, 32} and every other lane count (incl. 2 and 4) falls through to the same `LOG(FATAL)` with no `break`. It is a real source-level twin of the `int4` hole. But I could not reach it through the copy/parallel vectorizer at v0.1.13: bool copies at shapes that would demand width 2/4 compiled, and the generated source used scalar `bool`/`char` (the vectorizer does not select an unmapped bool lane count here). Recorded as a **latent same-shaped hole, not independently reproduced** — honest: it shares the source pattern but I have no live trigger for it via `T.copy`/`T.Parallel`.

**Same root as sibling draft B045↔B113.** B113 is a near-duplicate FACE of this same `int4x2` `PrintType` gap (same `case 4:` missing `lanes()==2`, same `LOG(FATAL)`), reached via a different surface. Same root cause — NOT merged here; recorded per project policy so the two faces are traceable to one fix.

**Distinct from #2526** (per Dedup above): #2526 is a runtime misaligned-address crash from a 16-byte load over the CUDA `int4` vector type (four `int32`s), rooted in `loop_vectorize.cc` — a different mechanism (runtime, wrong address) from this compile-time type-printer fall-through. Distinct, not filed here.

</details>

### LINKED ISSUE #2995: [BUG][Fuzzer][ice-on-valid-code] A 4-bit (`int4`/`uint4`) access resolving to `lanes == 2` crashes codegen with "Cannot convert type int4x2 to CUDA type"
### Required prerequisites

0.1.13

### System information

TileLang 0.1.13; CUDA 12.8; NVIDIA L40S (sm_89).

### Problem description

A plain element-wise `T.copy` of a 4-bit buffer (`int4` **or** `uint4`) aborts compilation at the CUDA codegen step with `Fatal: Cannot convert type int4x2 to CUDA type!`, at the ordinary default `threads=128`. `int4` is a first-class dtype with a dedicated scalar packed load/store path, and the same copy compiles and round-trips correctly at other thread counts — so this is a codegen gap, not an unsupported operation.

The vectorizer distributes the 256 elements over 128 threads as 2 `int4` per thread, producing an `int4x2` (8-bit) vector access. `CodeGenTileLangCUDA::PrintType`'s 4-bit `case 4:` has branches for `lanes == 4, 8, 16, 32, 64` but none for `lanes == 2`, so the 2-lane access falls through to the `LOG(FATAL)` else-arm — for **both** the signed (`int4`) and unsigned (`uint4`) arms of that case. Only `lanes == 2` is affected: `threads=64` (4 int4/thread → `int4x4`) and `threads=32` (8 int4/thread → `int4x8`) both compile; the trigger is independent of `N` (any `N / threads == 2` fires it). Not a regression — see Provenance.

### Reproducible example code

```python

N = 256

# Trigger: element-wise copy of an int4 buffer at the common default threads=128.
# The vectorizer packs 2 int4 per thread -> an int4x2 access, which
# CodeGenTileLangCUDA::PrintType has no branch for -> LOG(FATAL).
@T.prim_func
def trigger(A: T.Tensor((N,), "int4"), C: T.Tensor((N,), "int4")):
        T.copy(A, C)

# Control: same copy, threads=32 -> 8 int4/thread -> int4x8, which PrintType
# DOES handle. Compiles and round-trips (any lane count except 2 works).
@T.prim_func
def control(A: T.Tensor((N,), "int4"), C: T.Tensor((N,), "int4")):
    with T.Kernel(1, threads=32) as bx:
        T.copy(A, C)

try:
    tilelang.compile(trigger, out_idx=[1])
    print("trigger: COMPILED")
except Exception as e:
    print("trigger CRASH:", str(e).splitlines()[-1])   # -> Cannot convert type int4x2 to CUDA type!

k = tilelang.compile(control, out_idx=[1])
packed = torch.randint(-128, 128, (N // 2,), dtype=torch.int32).to(torch.int8).cuda()
out = k(packed)
ob = out.view(torch.int8) if out.dtype != torch.int8 else out
print("control:", "COMPILED + round-trips" if torch.equal(ob, packed) else "WRONG VALUE")
```

Output:

```
trigger CRASH: Cannot convert type int4x2 to CUDA type!
control: COMPILED + round-trips
```

### Traceback

```
[..] : Fatal: Cannot convert type int4x2 to CUDA type!
tilelang error: InternalError: Cannot convert type int4x2 to CUDA type!
```

### Expected behavior

The copy should compile, like it already does at every other lane count. An `int4x2` is exactly 8 bits, i.e. one byte — the same width `PrintType` already emits `signed char`/`char` for scalar `int4`. Adding a `lanes == 2` branch to the 4-bit case (emitting the 1-byte type the scalar path uses) would let the 2-lane access codegen the same way the 4/8/16/32/64-lane accesses already do.

### Additional context

**Root cause.** `CodeGenTileLangCUDA::PrintType` does not handle a 2-lane 4-bit vector: its 4-bit [`case 4:`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L952) enumerates `lanes == 4, 8, 16, 32, 64` (starting at [L961](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L961)) but has no `lanes == 2`, so an `int4x2` produced by the vectorizer falls through to the [`LOG(FATAL)` else-arm](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L978).

<details><summary>Boundary: which thread count → which lane count, and which compile</summary>

Over 256 `int4` elements, the vectorizer packs `256 / threads` elements per thread:

| threads | int4 / thread | vector type | result |
|---|---|---|---|
| 128 | 2 | `int4x2` | **crash** (`Cannot convert type int4x2`) |
| 64 | 4 | `int4x4` | compiles |
| 32 | 8 | `int4x8` | compiles |
| 1 (serial) | 1 | scalar | compiles, round-trips |

Only `lanes == 2` is missing from the enumeration; every neighbor is handled.

</details>

**Suggested fix.** Add a `lanes == 2` arm to the 4-bit `case 4:` at `codegen_cuda.cc:961` alongside the existing `lanes == 4`..`64` arms, emitting the 1-byte type the scalar branch uses (`signed char`/`char`) (not verified end-to-end).

**Provenance.** The 4-bit vector `PrintType` branches (and the missing `lanes == 2`) predate 0.1.9 — present since the initial codebase migration (`#10`); origin before that is unverified.

**Dedup.** I searched the open and closed tracker and found no existing report of this `int4x2` codegen gap.

**Reach.** `int4` is a documented, first-class dtype and int4 quantization is a common idiom — the `dequantize_gemm` example family targets 4-bit weights (`examples/dequantize_gemm/`: `example_dequant_gemm_w4a8.py`, `example_dequant_gemv_fp16xint4.py`, `example_dequant_gemm_fine_grained.py`; the `int4` value appears only as an `out_dtype` *literal* in `examples/dequantize_gemm/quantize/lop3.py:1112`, not as a buffer dtype). **Ran** `example_dequant_gemm_fine_grained.py` verbatim on 0.1.13 (default `test_run_dequantize_gemm()` config) — it **compiles and passes** (`profiler.assert_allclose` OK). It dodges this path because it declares `storage_dtype = T.int8` (verified: line 34) and unpacks 4-bit values arithmetically; `example_dequant_gemv_fp16xint4.py` likewise uses `storage_dtype = T.int8` (line 20). Neither declares an `int4` buffer, so none emits a raw `int4` vector access and CI stays green. The trigger is a raw `int4` buffer copied at `threads=128`; the crash fires whenever the per-thread lane count resolves to exactly 2 (here, `N / threads = 2`). No shipped example exercises the raw-`int4` path.

**Impact.** The trigger is narrow: only a raw `int4` vector access that resolves to exactly `lanes == 2` (every other lane count, and the scalar path, compile), which the packed-int8/lop3 examples never emit. When it fires it is a compile-time `LOG(FATAL)` at codegen — loud and caught immediately, blocking that one kernel from building while corrupting no output and never reaching a running workload silently. Adding the `lanes == 2` arm closes this boundary in the 4-bit `PrintType` enumeration so a legal `int4` copy is no longer refused.

<details><summary>§17 Generalization record</summary>

**Two-level root.**
- **SOURCE-level.** `CodeGenTileLangCUDA::PrintType`, integer `case 4:` at [codegen_cuda.cc:952](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L952). The narrow-integer cases (`case 1`, `case 4`) enumerate lane counts with an explicit `if/else if` chain and a `LOG(FATAL)` else-arm ([L978](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L978)), rather than a range predicate (`lanes <= N`) the way `case 16`/`case 32` do. Any legal lane count the vectorizer emits that is not in the enumeration falls through to FATAL. `case 4` enumerates `{4,8,16,32,64}` — the hole is `lanes == 2`.
- **OPERATOR-level.** Op-independent: the `int4x2` vector type is produced by the *vectorizer* from `N / threads`, not by `T.copy`. Any data-movement / elementwise op over a raw `int4` buffer that resolves to 2 elems/thread routes through the same `PrintType` FATAL (confirmed with `T.Parallel` below).

**4-axis sweep (each a real cell, fresh python process, fresh cache, run on 0.1.13 / L40S):**

|---|---|---|---|
| found-first | `int4` copy `threads=128`, `N=256` → `int4x2` | **crash** `Cannot convert type int4x2` | — |
| related-type | `uint4` copy `threads=128` → `uint4x2` | **crash** `Cannot convert type uint4x2` | **SAME** — `case 4` uint arm, identical missing `lanes==2` hole |
| related-type | `int8` copy `threads=128` → `int8x2` | compiles + round-trips OK | **distinct** — `case 8` else-arm emits `char` (a fallback, not `LOG(FATAL)`) |
| related-type | `bool` copy `threads=128` → 2-lane bool | compiles | **distinct** — routes via `is_vector_bool()` → `ushort2` (L919), never reaches integer `case 1`; the `case 1` `{8,16,32}` hole is unreachable this way |
| related-source | `int4` copy `N=64 threads=32` → `int4x2` | **crash** `Cannot convert type int4x2` | **SAME** — proves root is `lanes==2`, not the specific `N` |
| related-operator | `int4` `T.Parallel` elementwise write `threads=128` | **crash** `Cannot convert type int4x2` | **SAME** — op-independent; root is `PrintType` codegen, not `T.copy` |

Original adjacent-lane cells (all **distinct**, the enumeration handles them): `threads=64`→`int4x4` compiles (`lanes==4` present); `threads=32`→`int4x8` compiles+round-trips (`lanes==8` present); `threads=1`→scalar compiles+round-trips.

**Class.** The class is "narrow-integer `PrintType` lane-enumeration with a FATAL fall-through": it fires for **both `int4` and `uint4`** at `lanes==2` (same case, both arms), so the title/problem is reframed from the single `int4` face to the 4-bit face. It does **not** generalize to `int8` (fallback else-arm, no FATAL) or to `bool` (different code path). The 1-bit `case 1` shares the same fragile idiom (enumerates `{8,16,32}`, FATAL else, missing `lanes ∈ {2,4}`) but is **unreachable via a `bool` buffer** (bool vectors are diverted to `is_vector_bool`), so it is recorded as a latent same-shape sibling, **not** a confirmed reproducer — **distinct, not filed**.

**Same-root sibling draft.** B045 is a near-duplicate face of the **same** 4-bit `int4x2`/`case 4` gap. Recorded as same-root; **not merged** (kept as distinct drafts).

</details>

SamJSui:
## Summary

Width-2 vectorized `int4` and `uint4` accesses currently fail during CUDA code generation with:

`Cannot convert type int4x2 to CUDA type!`

This change adds the missing one-byte CUDA representation and regression coverage for both signed and unsigned four-bit types.

## Root cause

TileLang can distribute 256 four-bit elements over 128 threads as two elements per thread, producing an `int4x2` or `uint4x2` access.

`CodeGenTileLangCUDA::PrintType` handles four-bit vectors with 4, 8, 16, 32, and 64 lanes, but not 2 lanes. The width-2 access therefore reaches the fatal unsupported-type path.

## Fix

- Lower two-lane four-bit vectors to one physical byte:
  - `int4x2` → `int8_t`
  - `uint4x2` → `uint8_t`
- Map `uint4` to `torch.uint8` for compiled-kernel storage
- Add signed and unsigned codegen and round-trip regression tests
- Add storage-dtype coverage for `uint4`

## Tests

Tested on an NVIDIA RTX 4070 Ti SUPER (`sm_89`).

- Exact base `ddc2c54d2260`: reproduces the `int4x2`/`uint4x2` codegen failure
- Patched commit `4f91c73f75ff`: signed and unsigned copies compile and round-trip
- Generated CUDA uses `int8_t` and `uint8_t`
- Branch-specific tests: 4 passed
- Incremental C++ build: passed
- `bash format.sh --files ...`: passed
- `pre-commit run --all-files`: passed
- `python3 -m pytest testing`

Fixes #2994
Fixes #2995

## Summary

- Added CUDA code generation for packed `int4x2` and `uint4x2` vectors.
- Lowered two-lane vectors to one physical byte using `int8_t` or `uint8_t`.
- Added packed constructors, loads, stores, gathers, broadcasts, lane extraction, and vector assembly.
- Added alignment and divisibility checks for packed vector accesses.
- Scalarized wider unaligned vector loads when required.
- Mapped `uint4` storage to `torch.uint8`.
- Added signed nibble extraction and sign extension.
- Added CUDA-gated regression tests for signed and unsigned vectors, round trips, gathers, broadcasts, lane selection, storage dtypes, and wider vector lanes.
- Preserved scalar behavior and existing support for lane counts 4, 8, 16, 32, and 64.

## Validation

- Tested on an NVIDIA RTX 4070 Ti SUPER (`sm_89`).
- Ran branch-specific tests, an incremental C++ build, formatting, pre-commit checks, and the test suite.

## C++ style / lint notes

- The PR changes CUDA C++ codegen but does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) CI step remains relevant.
- No correctness or build issue is reported from this audit.
- Any TLCPP003/TLCPP004 findings are advisory unless they identify a clear API, FFI, or maintainability risk.


SamJSui:
per-commit.ci has been hanging since the GH outage. Going to close/reopen the PR

SiriusNEO (changes requested):
I think currently this PR contains some stride problems. Considering the following case:
```python
@tilelang.testing.requires_cuda
def test_uint4x2_stride2_gather_runtime():
      @T.prim_func
      def kernel(
            A: T.Tensor((8,), "uint4"),
            B: T.Tensor((2,), "uint4"),
      ):
          with T.Kernel(1, threads=1):
                for i in T.vectorized(2):
                    B[i] = A[i * 2]

compiled = tilelang.compile(kernel)

# Logical values: [1, 2, 3, 4, 5, 6, 7, 0]
source = torch.tensor(
          [0x21, 0x43, 0x65, 0x07],
          dtype=torch.uint8,
          device="cuda",
)
output = torch.zeros((1,), dtype=torch.uint8, device="cuda")

compiled(source, output)

# The kernel should gather A[0] == 1 and A[2] == 3,
# producing the packed byte 0x31.
#
# The current implementation may instead access byte[0] and byte[2],
# producing 0x51.
assert output.item() == 0x31
```

Expected failure with the current implementation: `AssertionError: assert 81 == 49`. Here, 81 == 0x51, while the correct result is 49 == 0x31.

SamJSui:
@SiriusNEO Thanks for the call out! I pushed a fix for strided and unaligned pack loads, by loading each logical nibble.

Added some regression tests and address CodeRabbit's call outs as well.

SiriusNEO (approved):

PR #3045, opened 2026-08-17
[Tool] Add official compile-only CLI

### LINKED ISSUE #2913: [Feature Request] Add TileLang support to Compiler Explorer (Godbolt)
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and did not find an existing issue requesting TileLang support in Compiler Explorer.

### Motivation

[Compiler Explorer](https://godbolt.org/) provides an interactive way to experiment with code and inspect compiler output in the browser. It already supports GPU programming systems such as Triton and CuTe DSL.

Adding TileLang support would make it easier to:

- experiment with small TileLang kernels without setting up a local environment;
- inspect useful compiler outputs and transformations;
- compare TileLang versions or supported targets;
- share reproducible examples through Compiler Explorer links;
- help new users learn TileLang and help contributors investigate compiler behavior.

Issue #1895 discusses pass/IR dumping that could also be useful in a Compiler Explorer context, but this request is broader: making TileLang available as a supported language/compiler on the platform.

### Solution

Add TileLang support to Compiler Explorer so that users can enter a self-contained TileLang example and inspect useful compilation results in the browser.

The implementation approach is intentionally left open. Contributors and maintainers familiar with TileLang, TVM, and Compiler Explorer may be able to identify a better design than any solution proposed upfront.

### Alternatives

Users can currently run TileLang locally and inspect generated output using its existing debugging and inspection facilities. However, that does not provide the same discoverability, convenience, version comparison, or shareable browser-based experience.

### Additional context

- Compiler Explorer: https://godbolt.org/
- Compiler Explorer repository: https://github.com/compiler-explorer/compiler-explorer
- Related TileLang IR-dump request: #1895

Contributions and design suggestions from the community would be very welcome.

LibertychaserUS:
Fixes #2913.

`python -m tilelang.tools.compile_only` lowers a kernel to source without running it. Default `--target c`. `--target cuda` is optional and soft-fails on Metal-style wheels.

`auto` is rejected as a string or as JSON. Bare `{"kind":"cuda"}` is pinned to `sm_80`. `--output_file` is removed before compile so a failed CE rerun can't keep the previous dump.

```
python -I -m pytest testing/python/tools/test_tilelang_tools_compile_only.py -q
```

CE: https://github.com/compiler-explorer/compiler-explorer/pull/9029
infra: https://github.com/compiler-explorer/infra/pull/2307

## Summary

- Adds the official `python -m tilelang.tools.compile_only` CLI.
- Lowers TileLang kernels to inspectable source without running them.
- Defaults to the CPU `c` target.
- Supports optional CUDA targets with validation and soft failure when CUDA FFI is unavailable.
- Rejects `auto` targets and pins bare CUDA targets to `sm_80`.
- Removes stale output before compilation while protecting the input file and symlinks.
- Enables `#line` directives when supported by the installed pass configuration.
- Adds comprehensive CLI, target-resolution, error-handling, line-directive, and optional CUDA tests.
- Supports TileLang integration with Compiler Explorer and related infrastructure.

### DIFF
--- a/tilelang/tools/compile_only.py
+++ b/tilelang/tools/compile_only.py
@@ -198,12 +198,16 @@ def _build_parser() -> argparse.ArgumentParser:
 def cli_main(argv: list[str] | None = None) -> int:
     """CE-shaped CLI (D2): ``--output_file`` + example.py. Default target ``c`` (D3)."""
     args = _build_parser().parse_args(argv)
+    output = Path(args.output_file)
+    # Match CuTe CE wrapper: drop a previous artifact before compile so a
+    # failed run cannot leave yesterday's assembly for CE to read.
+    output.unlink(missing_ok=True)
 
     try:
         module = load_example(args.input_file)
         func = discover_prim_func(module)
         source = compile_kernel_source(func, resolve_target(args.target))
-        Path(args.output_file).write_text(source)
+        output.write_text(source)
     except Exception as exc:
         print(f"{_ERROR_PREFIX}: {exc}", file=sys.stderr)
         return 1
--- a/tilelang/tools/compile_only.py
+++ b/tilelang/tools/compile_only.py
@@ -1,0 +1,151 @@
+"""Compile-only tool: emit inspectable ``lower()`` kernel source without a GPU.
+
+Usage::
+
+    python3 -I -m tilelang.tools.compile_only --output_file out.s example.py
+
+The default target is ``c`` (CPU C). CUDA is optional::
+
+    python3 -I -m tilelang.tools.compile_only --target cuda --output_file out.s example.py
+
+Programmatic API::
+
+    from tilelang.tools.compile_only import compile_kernel_source
+
+    source = compile_kernel_source(func)  # default target c
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib.util
+import sys
+from pathlib import Path
+from types import ModuleType
+
+__all__ = [
+    "DEFAULT_TARGET",
+    "cli_main",
+    "compile_kernel_source",
+    "discover_prim_func",
+    "load_example",
+    "resolve_target",
+]
+
+# D3: explicit CPU C. Never "auto". Bare "cuda" still runs the CUDA normalizer
+# (torch.cuda probe); pin an arch so optional --target cuda does not device-detect.
+DEFAULT_TARGET = "c"
+_PINNED_CUDA_TARGET: dict[str, str] = {"kind": "cuda", "arch": "sm_80"}
+_ERROR_PREFIX = "tilelang compile-only error"
+
+
+def resolve_target(target: str) -> str | dict[str, str]:
+    """Map a CLI/library target string to an explicit lower() target.
+
+    Parameters
+    ----------
+    target : str
+        User-facing target. ``auto`` is rejected. ``cuda`` is pinned to sm_80.
+
+    Returns
+    -------
+    str or dict
+        A TVM target string, or a pinned CUDA target dict.
+    """
+    normalized = target.strip()
+    key = normalized.lower()
+    if not normalized or key == "auto":
+        raise ValueError("target must be explicit; do not use auto")
+    if key == "cuda":
+        return dict(_PINNED_CUDA_TARGET)
+    return normalized
+
+
+def load_example(path: str) -> ModuleType:
+    """Load ``example.py`` as a module (D2). Do not treat it as a launch script."""
+    spec = importlib.util.spec_from_file_location("tilelang_compile_only_input", path)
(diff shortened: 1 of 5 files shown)


LibertychaserUS:
@coderabbitai review

tilelang/tools/compile_only.py
```diff
@@ -0,0 +1,224 @@
...
+    return parser
+
+
+def cli_main(argv: list[str] | None = None) -> int:
+    """CE-shaped CLI (D2): ``--output_file`` + example.py. Default target ``c`` (D3)."""
+    args = _build_parser().parse_args(argv)
+    output = Path(args.output_file)
+    # Match CuTe CE wrapper: drop a previous artifact before compile so a
+    # failed run cannot leave yesterday's assembly for CE to read.
+    output.unlink(missing_ok=True)
```
SiriusNEO, line 242: `output.unlink()` runs before `load_example(args.input_file)`. Therefore, a valid but mistaken invocation such as:

```
python -m tilelang.tools.compile_only --output_file example.py example.py
```

  deletes the input source and then fails to load it, causing irreversible data loss. Please detect identical/aliased paths before unlinking (for example with `Path.resolve()` and
  `os.path.samefile()` when applicable), return a clear error, and add a regression test confirming the input remains intact.

LibertychaserUS: Good catch. `--output_file` now errors if it names the same file as the input (including via symlink), and the source is left alone. Added a regression test.

tilelang/tools/compile_only.py
```diff
@@ -0,0 +1,241 @@
...
+
+    if isinstance(target, str):
+        target = resolve_target(target)
+    # Optional CUDA: Metal wheels lack the transform FFI. Fail clearly instead
+    # of leaking AttributeError from AnnotateDeviceBoundTmaCopies.
+    if _is_cuda_target(target) and not cuda_codegen_available():
+        raise RuntimeError(_CUDA_FFI_MISSING)
+    # Match JITKernel._compile_artifact: LayoutInference reads Target.current().
+    resolved = target if isinstance(target, Target) else Target(target)
+    with tilelang.transform.PassContext(opt_level=3), resolved:
```
penguin-wwy: can pass `TL_EMIT_LINE_DIRECTIVES=True` to emit the line number so that the CE tilelang wrapper can consume it.

LibertychaserUS: compile-only now sets `TL_EMIT_LINE_DIRECTIVES` when this build has the pass, so the C dump includes `#line` for the CE wrapper. Older wheels reject the unknown key, so we skip it there.

LibertychaserUS:
The red CI check doesn’t look like a problem in this diff.

The only failing job is `Test for Python 3.12 with CUDA-auto (on self-hosted-nvidia)` (run 32352550474, attempt 2). It never reaches build or pytest — it dies in `Checkout repository`:

```
git submodule update --init --force --recursive
##[error]fatal: Unable to find current revision in submodule path '3rdparty/composable_kernel'
##[error]The process '/usr/bin/git' failed with exit code 128
```

The main repo fetches and checks out fine; only the submodule update fails. This PR does not touch that pin (`b38bb492`, same as main). The same checkout error hit an unrelated PR on the same runner a few hours earlier (run 32362839081), so this looks like runner-side state. CuTeDSL Examples is skipped only because it `needs` the CUDA job.

Where tests did run, this PR is green: ROCm’s full `testing/python` suite — 2129 passed, 0 failed, including all 23 compile-only tests; Metal, Quick Lint, and pre-commit.ci pass as well. (An earlier revision of this PR did fail one new `#line` test on ROCm; that was fixed in 25f889db and now passes.)

SiriusNEO (dismissed):

LeiWang1999:
Thanks @LibertychaserUS would love to see some doc update about this tool :)

LibertychaserUS:
Thanks for the review @LeiWang1999! Docs added in 4a40cf7a:

- New page `docs/tools/compile_only.md` — quick start, CLI reference, target rules (`c` default, pinned `sm_80` for bare `cuda`, `auto` rejected), error reporting, `#line` directive behavior, and the programmatic API.
- Wired into the Tools table (`docs/tools/index.md`) and the docs toctree (`docs/index.md`).

Every command, output, and error message in the page was run for real on a CPU-only machine against the pip wheel — the same shape as the Compiler Explorer deployment.

One question: would you also like a runnable entry under `examples/` (e.g. `examples/compile_only/`), or an "Examples" section linking to the 25 test cases in `testing/python/tools/test_tilelang_tools_compile_only.py`? Happy to add either.

SiriusNEO (approved):

PR #3047, opened 2026-08-18
[CUDA] Fix FP8 min/max codegen

### LINKED ISSUE #2985: [BUG][Fuzzer][ice-on-valid-code] `T.max`/`T.min`/`T.clamp` on `float8_e4m3`/`float8_e5m2` fail to compile instead of computing the value
### Required prerequisites

0.1.13 (latest release)

### System information

- Installed from the `0.1.13` wheel.
- Python 3.13, linux; PyTorch with CUDA 12.8.
- GPU: NVIDIA L40S (`sm_89`) — where this session's run was done. The defect is in the target-independent binary-op codegen (`MaxNode`/`MinNode` → bare `max(...)`/`min(...)`) plus the `float8` template type, so it is architecture-independent (not arch-gated); only `sm_89` was run this session.

### Problem description

`T.max`, `T.min`, and `T.clamp` applied to `float8`-typed operands fail to compile — nvcc rejects the store with `no operator "=" matches these operands / operand types are: fp8_e4_t = float`. The same kernel with `float16` or `bfloat16` operands compiles and returns the correct value, so this is a `float8` coverage gap, not a "this dtype can't be compared" limitation.

The codegen emits a bare `max(A[i], B[i])` for an fp8-typed `MaxNode`. There is no `max(fp8, fp8) -> fp8` overload: the arguments promote to `float`, so `max(...)` returns `float`, and `tl::float_e4m3_t` / `tl::float_e5m2_t` have no `operator=(float)` (their only converting constructor is from `__nv_bfloat16`, and it is `explicit`), so the assignment `C[i] = max(...)` does not compile.

Both `float8` variants fail identically (`fp8_e4_t = float` for `e4m3`, `fp8_e5_t = float` for `e5m2`), and all three ops share the mechanism: `T.clamp` lowers to nested `T.max`/`T.min`, so it hits the same missing overload. All six (op × dtype) pairs were run this session and all six fail with the identical `no operator "=" matches` store error. Not a regression — see Provenance.

<details>
<summary>Emitted CUDA and the per-op / per-dtype result</summary>

The generated kernel for the `float8_e4m3` `T.max` case:

```cuda
extern "C" __global__ void main_kernel(const fp8_e4_t* A, const fp8_e4_t* B, fp8_e4_t* C) {
  C[((int)threadIdx.x)] = max(A[((int)threadIdx.x)], B[((int)threadIdx.x)]);
}
```

Compile result of `C[i] = T.<op>(A[i], B[i])` (and `T.clamp(A[i], lo, hi)`) per dtype, this session:

| op | float16 | bfloat16 | float8_e4m3 | float8_e5m2 |
|---|---|---|---|---|
| `T.max`   | ok | ok | fail (`fp8_e4_t = float`) | fail (`fp8_e5_t = float`) |
| `T.min`   | ok | ok | fail | fail |
| `T.clamp` | ok | ok | fail | fail |

</details>

### Reproducible example code

```python
N = 128

@tilelang.jit(out_idx=[2])
def maxk(dtype):
    @T.prim_func
    def main(A: T.Tensor((N,), dtype), B: T.Tensor((N,), dtype), C: T.Tensor((N,), dtype)):
        with T.Kernel(1, threads=N):
            i = T.get_thread_binding()
            C[i] = T.max(A[i], B[i])
    return main

# control: float16 compiles and matches torch.maximum
a16 = torch.randn(N, device="cuda").to(torch.float16)
b16 = torch.randn(N, device="cuda").to(torch.float16)
c16 = maxk("float16")(a16, b16)
print((c16.float() - torch.maximum(a16, b16).float()).abs().max().item())   # -> 0.0

# bug: float8_e4m3 fails to compile
a8 = torch.randn(N, device="cuda").to(torch.float8_e4m3fn)
b8 = torch.randn(N, device="cuda").to(torch.float8_e4m3fn)
maxk("float8_e4m3")(a8, b8)   # -> nvcc error: no operator "=" matches these operands (fp8_e4_t = float)
```

`T.min` and `T.clamp` fail the same way; `float8_e5m2` behaves identically to `float8_e4m3`.

### Traceback

```pytb
tvm_kernels.cu(14): error: no operator "=" matches these operands
            operand types are: fp8_e4_t = float
    C[((int)threadIdx.x)] = max(A[((int)threadIdx.x)], B[((int)threadIdx.x)]);
```

### Expected behavior

`T.max`/`T.min`/`T.clamp` on `float8` should compile and return the elementwise result, as they already do for `float16`/`bfloat16` (verified: `float16` returns `max_err = 0.0` on the same kernel). The operation is computable in `float8`: writing the max in `float` and converting back — `C[i] = T.Cast(dtype, T.max(T.Cast("float32", A[i]), T.Cast("float32", B[i])))` — compiles and returns the exact result (`max_err = 0.0`, verified this session), which is the same route TileLang already uses to close comparable `float8` coverage gaps.

### Additional context

**Root cause.** An fp8 `max`/`min` codegens to a bare `max(...)`/`min(...)` call whose `float` result has no way back into the fp8 wrapper type — the wrapper has no `operator=(float)` — so the store fails to compile.

The CUDA codegen prints an fp8-typed `MaxNode`/`MinNode` as a bare `max(...)`/`min(...)` call ([`codegen_c.cc` `PrintBinaryExpr`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L514-L534), reached from the [Max/Min visitors](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L587-L591)), but no `max`/`min` overload returns `float8`: the fp8 operands promote to `float`, so `max(...)` is `float`, and the [`tl::float_e4m3_t` / `float_e5m2_t` wrapper types](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/tl_templates/cuda/common.h#L808) have no `operator=(float)` — their [only converting constructor from a scalar float type is `explicit float_e4m3_t(__nv_bfloat16)`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/tl_templates/cuda/common.h#L814-L817) — so the store `C[i] = max(...)` fails to compile. `T.clamp` lowers to nested `T.max`/`T.min`, so it inherits the same failure.

</details>

**Generalization (tested this session, 0.1.13 / sm_89 / L40S).** This is a *missing-overload* class, not a per-op quirk. All six (`max`/`min`/`clamp`) × (`float8_e4m3`/`float8_e5m2`) pairs fail with the identical store error (`fp8_e4_t = float` / `fp8_e5_t = float`); the `float16`/`bfloat16` controls all compile and match torch; and the `float`-cast route (`T.Cast(dtype, T.max(T.Cast("float32", …), …))`) compiles and returns `max_err = 0.0`. The neighbors share one root — the codegen emits a bare `max`/`min` whose `float` result has no `operator=` on the fp8 wrapper — so the title and description are framed to the whole (op × fp8) class with `max` as the worked example.

**Suggested fix.** One option that matches how the `float16`/`bfloat16` cases already work is a `float8` overload of the emitted symbols that computes in `float` and converts back — `TL_DEVICE tl::float_e4m3_t max(tl::float_e4m3_t a, tl::float_e4m3_t b) { return tl::float_e4m3_t(max(float(a), float(b))); }` (and `min`, for both `e4m3` and `e5m2`) — so the result type matches the store target instead of resolving to `float`. This is the same shape as the `float` cast the manual sibling path above already uses successfully.

**Provenance.** Present since the `float8` wrapper types were wired: the `tl::float_e4m3_t`/`float_e5m2_t` structs (common.h) carry no `operator=(float)` and no `max`/`min` overload, and the codegen has always emitted a bare `max(...)`/`min(...)` for `MaxNode`/`MinNode`. Not a regression — the (`max`/`min`/`clamp`, `float8`) pairs have never compiled on this path.

**Dedup.** I searched the open and closed tracker (and PRs) and found no existing report of the `max`/`min`/`clamp` on `float8` compile failure. Distinct from [#2565](https://github.com/tile-ai/tilelang/issues/2565), which covers the same *class* of gap for the unary math intrinsics (`exp`/`log`/`sqrt`/`erf`/…) — but its `float8` face is a *frontend* failure (`Unresolved call ir.Op ... name="tirx.exp"`; the op never reaches codegen), and its op set is disjoint from the binary `max`/`min`/`clamp` here, which *do* reach codegen and fail at nvcc.

**Reach.** The trigger is `T.max`/`T.min`/`T.clamp` with `float8`-typed operands. The op docstrings type the operands as generic `PrimExpr` with no dtype disclaimer, and `float8_e4m3`/`float8_e5m2` are first-class TileLang dtypes; clamping a value into the fp8 representable range is a standard fp8-quantization idiom. `grep`: the shipped fp8-cast examples that clamp (`examples/cast/example_per_token_cast_to_fp8.py:35`, `example_group_per_split_token_cast_to_fp8.py`, `example_triton_cast_to_fp8.py`) all `T.clamp` a value that is still in the *input* dtype (`float16`/`bfloat16`) and only then `T.copy` into a separate `float8` buffer — so they clamp in the wide type and never apply `max`/`min`/`clamp` to an fp8-typed value, which is why CI is green. The bug fires the moment the clamped/compared value is itself `float8` (e.g. clamping an already-`float8` accumulator, or writing `T.max` directly on `float8` operands).

**Impact.** The trigger is narrow — `T.max`/`T.min`/`T.clamp` applied to an operand whose dtype is itself `float8` (the shipped fp8-cast idiom clamps in the wide type first, so this misses CI). When it fires the failure is a compile-time nvcc error, loud and caught immediately: it blocks that one kernel from building but corrupts no data and cannot reach a running workload silently. Fixing it closes an fp8 coverage gap so a legal `float8` kernel compiles, matching the `float16`/`bfloat16` behavior; it removes a build-time refusal, not a silent wrong result.

Chennesxu:
## Summary

`T.max`, `T.min`, and `T.clamp` failed during nvcc compilation when applied to `float8_e4m3` or `float8_e5m2` operands:

```text
error: no operator "=" matches these operands
    operand types are: fp8_e4_t = float
```

## Root cause

CUDA codegen emits unqualified min/max calls for TIR `MinNode` and `MaxNode` expressions:

```cuda
C[i] = max(A[i], B[i]);
```

The FP8 wrappers promote to float, so the call returns float, but the result cannot be implicitly assigned back to the FP8 wrapper. `T.clamp` lowers to nested max and min operations and fails through the same path.

The per-lane vector fallback emits the same unqualified calls and has the same overload-resolution gap.

## Fix

Add `tl::min` and `tl::max` overloads alongside the FP8 wrapper types:

```cpp
TL_DEVICE float_e4m3_t max(float_e4m3_t lhs, float_e4m3_t rhs) {
  return float_e4m3_t(
      ::fmaxf(static_cast<float>(lhs), static_cast<float>(rhs)));
}
```

Equivalent overloads are provided for min and for `float_e5m2_t`.

The existing unqualified calls resolve these overloads through argument-dependent lookup, so scalar expressions, vector lanes, and nested clamp operations reuse the same implementation without adding FP8-specific branches to CUDA codegen.

`float8_e4m3fn` maps to the same emitted CUDA wrapper as `float8_e4m3` and therefore uses the same overloads.

## Tests

Added coverage for:

* exact runtime results for `T.max`, `T.min`, and `T.clamp` with `float8_e4m3` and `float8_e5m2`;
* vectorized max/min/clamp compilation through the per-lane fallback;
* `float8_e4m3fn` overload resolution.

Verified locally on a TITAN RTX (`sm_75`) with CUDA 12.4. The complete clamp test file passes 11 tests.

Fixes #2985

## Summary

- Fix CUDA code generation for `T.max`, `T.min`, and `T.clamp` with FP8 operands.
- Compute results with `fminf` and `fmaxf`, then convert them back to the FP8 type.
- Add overloads for `float_e4m3_t` and `float_e5m2_t`.
- Add scalar and vectorized tests for supported FP8 formats and operation combinations.

## C++ style / lint notes

- The PR does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit remains warning-only.
- No correctness or build issue is indicated by the reported style changes.

### DIFF
--- a/src/tl_templates/cuda/common.h
+++ b/src/tl_templates/cuda/common.h
@@ -1136,6 +1136,26 @@ template <typename T> TL_DEVICE T fast_min(T a, T b) { return b < a ? b : a; }
 
 template <> TL_DEVICE float fast_min(float a, float b) { return fminf(a, b); }
 
+TL_DEVICE float_e4m3_t max(float_e4m3_t lhs, float_e4m3_t rhs) {
+  return float_e4m3_t(
+      ::fmaxf(static_cast<float>(lhs), static_cast<float>(rhs)));
+}
+
+TL_DEVICE float_e4m3_t min(float_e4m3_t lhs, float_e4m3_t rhs) {
+  return float_e4m3_t(
+      ::fminf(static_cast<float>(lhs), static_cast<float>(rhs)));
+}
+
+TL_DEVICE float_e5m2_t max(float_e5m2_t lhs, float_e5m2_t rhs) {
+  return float_e5m2_t(
+      ::fmaxf(static_cast<float>(lhs), static_cast<float>(rhs)));
+}
+
+TL_DEVICE float_e5m2_t min(float_e5m2_t lhs, float_e5m2_t rhs) {
+  return float_e5m2_t(
+      ::fminf(static_cast<float>(lhs), static_cast<float>(rhs)));
+}
+
 // --- max2 ----------------------------------------------------------------
 
 TL_DEVICE float2 max2(float2 a, float2 b) {
--- a/testing/python/language/test_tilelang_language_clamp.py
+++ b/testing/python/language/test_tilelang_language_clamp.py
@@ -1,3 +1,5 @@
+import pytest
+
 import tilelang.testing
 from tilelang import language as T
 
@@ -111,6 +113,89 @@ def test_clamp():
     run_clamp(1024, 128, T.float32, -0.06, 0.05)
     run_clamp_value_range(1024, 128, T.float16)
     run_clamp_value_range(1024, 128, T.float32)
+
+
+FP8_OPS = {
+    "max": lambda a, b, dtype: T.max(a, b),
+    "min": lambda a, b, dtype: T.min(a, b),
+    "clamp": lambda a, b, dtype: T.clamp(a, T.cast(-1.0, dtype), T.cast(1.0, dtype)),
+}
+
+
+def fp8_operand_kernel(N, block_N, dtype, op):
+    @T.prim_func
+    def main(
+        A: T.Tensor((N,), dtype),
+        B: T.Tensor((N,), dtype),
+        C: T.Tensor((N,), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), threads=block_N) as bx:
+            for i in T.Parallel(block_N):
+                C[bx * block_N + i] = FP8_OPS[op](A[bx * block_N + i], B[bx * block_N + i], dtype)
+
+    return main
+
+
+def run_fp8_min_max_clamp(N, block_N, dtype, op):
+    import torch
+
+    kernel = tilelang.compile(fp8_operand_kernel(N, block_N, dtype, op))
+    torch_dtype = dtype.as_torch()
+
+    # Integers are exactly representable in both float8 formats, so the
+    # reference comparison can be exact.
+    a = torch.randint(-4, 5, (N,), device="cuda").to(torch.float32)
+    b = torch.randint(-4, 5, (N,), device="cuda").to(torch.float32)
+    if op == "max":
+        ref = torch.maximum(a, b)
+    elif op == "min":
+        ref = torch.minimum(a, b)
+    else:
+        ref = a.clamp(-1.0, 1.0)
+
+    # Allocate the output here instead of using out_idx: rebuilding a torch
+    # tensor from an fp8 DLPack tensor is not supported on every torch version.
+    C = torch.empty(N, device="cuda", dtype=torch_dtype)
+    kernel(a.to(torch_dtype), b.to(torch_dtype), C)
+    torch.testing.assert_close(C.float(), ref.to(torch_dtype).float(), atol=0, rtol=0)
+
+
+def fp8_operand_vectorized_kernel(N, block_N, dtype, op, vec=4):
+    @T.prim_func
+    def main(
+        A: T.Tensor((N,), dtype),
+        B: T.Tensor((N,), dtype),
+        C: T.Tensor((N,), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), threads=block_N // vec) as bx:
+            for i in T.Parallel(block_N // vec):
+                for v in T.vectorized(vec):
+                    idx = bx * block_N + i * vec + v
+                    C[idx] = FP8_OPS[op](A[idx], B[idx], dtype)
+
+    return main
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("dtype", [T.float8_e4m3, T.float8_e5m2])
+@pytest.mark.parametrize("op", list(FP8_OPS))
+def test_fp8_min_max_clamp(dtype, op):
+    run_fp8_min_max_clamp(1024, 128, dtype, op)
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("op", list(FP8_OPS))
+def test_fp8_vectorized_min_max(op):
+    # Vectorizing reaches the per-lane fallback, a separate overload-resolution
+    # context from the scalar expression.
+    tilelang.compile(fp8_operand_vectorized_kernel(1024, 128, T.float8_e4m3, op), out_idx=[2], target="cuda")
+
+
+@tilelang.testing.requires_cuda
+def test_fp8_e4m3fn_min_max():
+    # float8_e4m3fn shares the emitted CUDA type with float8_e4m3; clamp covers
+    # both min and max.
+    tilelang.compile(fp8_operand_kernel(1024, 128, T.float8_e4m3fn, "clamp"), out_idx=[2], target="cuda")
 
 
 if __name__ == "__main__":


LeiWang1999:
SGTM overall. One concern is that the changes in `codegen_cuda` introduce FP8-specific branching into otherwise generic scalar and vector lowering. Would it be cleaner to define `min`/`max` overloads alongside the FP8 types instead?

```cuda
namespace tl {

TL_DEVICE float_e4m3_t max(float_e4m3_t lhs, float_e4m3_t rhs) {
  return float_e4m3_t(
      ::fmaxf(static_cast<float>(lhs), static_cast<float>(rhs)));
}

TL_DEVICE float_e4m3_t min(float_e4m3_t lhs, float_e4m3_t rhs) {
  return float_e4m3_t(
      ::fminf(static_cast<float>(lhs), static_cast<float>(rhs)));
}

}  // namespace tl
```

With equivalent overloads for `float_e5m2_t`, the existing generated `max(a, b)` and `min(a, b)` calls should resolve through ADL, including the per-lane vector fallback. This would keep the generic codegen unchanged and place the FP8-specific behavior closer to the type definitions.

Chennesxu:
> SGTM overall. One concern is that the changes in `codegen_cuda` introduce FP8-specific branching into otherwise generic scalar and vector lowering. Would it be cleaner to define `min`/`max` overloads alongside the FP8 types instead?
> 
> ```cuda-c++
> namespace tl {
> 
> TL_DEVICE float_e4m3_t max(float_e4m3_t lhs, float_e4m3_t rhs) {
>   return float_e4m3_t(
>       ::fmaxf(static_cast<float>(lhs), static_cast<float>(rhs)));
> }
> 
> TL_DEVICE float_e4m3_t min(float_e4m3_t lhs, float_e4m3_t rhs) {
>   return float_e4m3_t(
>       ::fminf(static_cast<float>(lhs), static_cast<float>(rhs)));
> }
> 
> }  // namespace tl
> ```
> 
> With equivalent overloads for `float_e5m2_t`, the existing generated `max(a, b)` and `min(a, b)` calls should resolve through ADL, including the per-lane vector fallback. This would keep the generic codegen unchanged and place the FP8-specific behavior closer to the type definitions.

Got it, switched to the FP8 min/max overloads and removed the codegen-specific branches. Thanks for the suggestion!

PR #3060, opened 2026-08-21
[BugFix] Scalarize Select in automatic vectorization

KellyFrog:
Now `T.select` vectorization is disabled to align with `T.if_then_else`.

## Summary

- Disabled automatic vectorization of `T.select`.
- Kept explicit `T.vectorized` loops unchanged.
- Added coverage for data-dependent parallel selects.
- Verified scalar code generation and correctness against `torch.where`.

## C++ style / lint notes

- The change touches C++ code but does not alter rules in `docs/developer_guide/cpp_style.md`.
- No correctness or build issues were identified.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -530,6 +530,13 @@ private:
     return arith::IRMutatorWithAnalyzer::VisitStmt_(node);
   }
 
+  PrimExpr VisitExpr_(const SelectNode *node) final {
+    // Select stays an expression-level ternary. Constrain its vector width
+    // using the same condition-uniformity rule as IfThenElse.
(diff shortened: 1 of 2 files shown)


src/transform/loop_vectorize.cc
```diff
@@ -530,6 +530,19 @@ class VectorizePlanner : public arith::IRMutatorWithAnalyzer {
     return arith::IRMutatorWithAnalyzer::VisitStmt_(node);
   }
 
+  PrimExpr VisitExpr_(const SelectNode *node) final {
```
LeiWang1999, line 533: Can we just leverage `IfThenElseNode` 's implementation?

PR #3065, opened 2026-08-22
[Bugfix] Fix non-divisible nested modulo simplification

### LINKED ISSUE #2955: [BUG][Fuzzer][wrong-code] `B[j % K]` in a 2-D `T.Parallel(M, N)` loop silently reads the wrong element when `K ∤ N` instead of computing `j % K`
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and did not find a report of this defect.

0.1.13

### System information

TileLang 0.1.13 (release commit `8001cc4`, bundled TVM `0e15b27`); Linux, NVIDIA L40S (sm_89), CUDA 12.8. Reproduced on sm_89; the defect is in an architecture-independent lowering pass (`FlattenBuffer` → TVM iter-map simplification), so it is not GPU-specific.

### Problem description

A plain per-column table lookup `B[j % K]` written inside a 2-D `T.Parallel(M, N)` loop returns the wrong `B` element for every row `i > 0` when `K` does not divide `N`. The kernel compiles and runs with no error; the result is silently wrong and deterministic. When `K` divides `N`, or when the inner loop is serial (not part of the 2-D parallel fusion), the same code is correct.

For `for i, j in T.Parallel(16, 64): C[i,j] = A[i,j] + B[j % 7]`, 512 of 1024 outputs are wrong (all rows `i >= 1`). The emitted CUDA indexes `B` as `B[threadIdx.x % 7]` — the flattening maps the 1024 iterations onto `8 x 128` (`i * 128 + threadIdx.x`), so the column is `j = (i*128 + threadIdx.x) % 64 = threadIdx.x % 64`, and the intended index is `(threadIdx.x % 64) % 7`. The inner `% 64` is dropped, so for `threadIdx.x >= 64` the wrong `B` slot is read.

<details><summary>Emitted CUDA (the dropped inner modulo)</summary>

```cuda
extern "C" __global__ void __launch_bounds__(128, 1)
main_kernel(const float* A, const float* B, float* C) {
  #pragma unroll
  for (int i = 0; i < 8; ++i) {
    C[((i * 128) + ((int)threadIdx.x))] =
        (A[((i * 128) + ((int)threadIdx.x))] + B[(((int)threadIdx.x) % 7)]);
    //                                          ^^^^^^^^^^^^^^^^^^^^^^^^^ should be ((threadIdx.x % 64) % 7)
  }
}
```
</details>

### Reproducible example code

```python
import numpy as np, torch, tilelang, tilelang.language as T

M, N = 16, 64

def build(K):
    @T.prim_func
    def main(A: T.Tensor((M, N), "float32"),
             B: T.Tensor((K,), "float32"),
            for i, j in T.Parallel(M, N):
                C[i, j] = A[i, j] + B[j % K]
    return main

A = torch.arange(M * N, dtype=torch.float32, device="cuda").reshape(M, N)

def run(K):
    B = (torch.arange(K, dtype=torch.float32, device="cuda") + 1) * 100
    out = tilelang.compile(build(K), out_idx=[2])(A, B).cpu().numpy()
    ref = A.cpu().numpy() + B.cpu().numpy()[np.arange(N) % K][None, :]   # B[j % K] per column
    return int((out != ref).sum())

print("K=7 (7 does NOT divide N=64):", run(7), "/ 1024 wrong")  # -> 512 / 1024 wrong (all rows i>=1)
print("K=8 (8 divides N=64):       ", run(8), "/ 1024 wrong")  # -> 0            (control)
```

### Traceback

### Expected behavior

`B[j % K]` should read `B[j % K]` for every `(i, j)`, independent of whether `K` divides `N` — as it already does when `K = 8` (divides `N = 64`) and when the inner `j` loop is serial rather than part of the 2-D parallel fusion (both verified correct over the same inputs). Computing `B[j % 7]` correctly seems the natural expectation, since the identical kernel computes the right answer on those sibling configurations.

### Additional context

**Root cause.** TVM's iterator-affine-map simplifier rewrites a *nested* modulo `floormod(floormod(x, c_inner), c_outer)` to `floormod(x, c_outer)`. That identity is sound only when `c_outer | c_inner` — but the rewrite does not check divisibility; instead it gates the fold on whether `PadDividendToDivisor` can *pad* the inner extent up to a multiple of `c_outer`. Padding always succeeds (any extent can be rounded up to a multiple), so it is not a real precondition: it rounds `c_inner=64` up to `70` (a multiple of `7`) and treats that as license to fold. Padding is the wrong test here — it is a mechanism for *splitting* an iterator whose range isn't divisible by a tile (extend the range, mask the extra iterations with a predicate), and it is sound for a monotone quantity (a plain iterator, a `floordiv`) whose range is merely truncated. But `x % 64` is not truncated — it *wraps*: at `x = 64` it returns to `0`, it never takes the padded values `64..69`. Folding on a padded-but-wrapping quantity rewrites the index for the *real* iterations too, so the collapsed `x % 7` is wrong for every `x ≥ 64`. The correct precondition is simply `c_outer | c_inner` (`CanProveDivisible(c_inner, c_outer)`), with no padding — and the same function's `scale != 1` branch already does exactly that (it uses `CanProveDivisible` and bails when it can't prove divisibility); the `scale == 1` branch drops that check and routes through padding instead. `FlattenBuffer` passes every access index through this simplifier at the `Surjective` level (which enables padding), so the legitimate column index `(threadIdx.x % 64) % 7` collapses to `threadIdx.x % 7`.

<details><summary>Where the inner modulo is dropped (pass-by-pass + the folding rule)</summary>

Running the lowering pipeline stage by stage, the `B` subscript is correct as `B[tx % 64 % 7]` up to and including `NarrowDataType`, and becomes `B[tx % 7]` immediately after [`FlattenBuffer`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/transform/flatten_buffer.cc#L246-L258):

```
[NarrowDataType]  C[i*2 + tx//64, tx%64] = A[...] + B[tx % 64 % 7]
[FlattenBuffer]   C_1[i*128 + tx]        = A_1[...] + B_1[tx % 7]
```

`FlattenBuffer` routes each access index through `IterMapSimplify` at the `Surjective` level. The offending rewrite is in `SplitFloorModConst`, whose own comment states the identity it applies — [iter_affine_map.cc#L2018-L2023](https://github.com/tile-ai/tvm/blob/0e15b274bce8b46f971abf5ac390e844aa6acee5/src/arith/iter_affine_map.cc#L2018-L2023):

```cpp
// floormod(floormod(floordiv(iter, lower_factor), c1c2), c1)
// = floormod(floordiv(iter, lower_factor), c1), where c1=rhs
return IterSplitExpr(padded->source,
                     /* lower_factor = */ padded->lower_factor,
                     /* extent = */ rhs,
                     /* scale = */ padded->scale);
```

The dividend is *padded* to a multiple of the divisor (`PadDividendToDivisor`) rather than gated on `c1 | c1c2`. Isolating the simplifier confirms the level dependence (with `tx` an iterator of extent 128):

```
iter_map_simplify([tx % 64 % 7], {tx: [0,128)], check_level=Bijective)   -> [tx % 64 % 7]   # correct
iter_map_simplify([tx % 64 % 7], {tx: [0,128)], check_level=Surjective)  -> [tx % 7]        # drops % 64
```

`Bijective` (level 0) keeps the inner modulo; `Surjective` (level 1, used by `FlattenBuffer`) drops it. The value divergence is 512/1024 (`(tx % 64) % 7 != tx % 7` for `tx >= 64`), matching the runtime wrong-count exactly.
</details>

<details><summary>Scope of the unsound fold: which nested div/mod shapes it affects (numerically ground-truthed)</summary>

I swept the neighboring nested div/mod fold shapes through `iter_map_simplify`, comparing each simplified form against a re-evaluation of the original over the whole domain (a nonzero mismatch count = an unsound fold). One root cause — the divisibility-blind `floormod(floormod(·), ·)` fold — shows three faces; the sibling shapes with correct handling are unaffected:

| nested shape | result | why |
|---|---|---|
| `(x % a) % b` | **unsound** when `b ∤ a` and the domain lets the inner mod wrap (`ext > a`) | the fold under test; `a`, `b` need not be powers of two — `40%7`, `12%5`, `100%9`, `28%6` all misfold |
| `(i·BN + tx) % N % K` | **unsound** when `K ∤ N` | same fold, the real post-fusion flatten shape (this bug) |
| `((x % a) % b) % c` | **unsound**, and it *compounds* | same fold applied recursively drops *both* inner bounds, e.g. `x%64%7%2 → x%2` |
| `(x / a) / b` | sound | `(x/a)/b = x/(ab)` is unconditional for `x ≥ 0` — no divisibility needed |
| `(x % a) / b` | sound | folds only when the domain proves the mod inactive (`x < a`) |
| `(x / a) % b` | sound | folds to `x % (ab) // a`, a correct identity that keeps the structure |
| `(x·s) % m` (scale ≠ 1) | sound | takes the `CanProveDivisible`-guarded branch |

So the defect is precisely the outer-`floormod`-collapsed-to-single-mod fold on the `scale == 1` path; the divisor (`divdiv`), mixed (`moddiv`/`divmod`), and scaled shapes each either need no divisibility precondition or already enforce one, and stay correct — bounding the fix to that one branch.
</details>

**Suggested fix.** The `scale == 1` branch of `SplitFloorModConst` should not decide this fold via `PadDividendToDivisor`. The legality of `floormod(floormod(x, c_inner), c_outer) → floormod(x, c_outer)` depends only on `c_outer | c_inner`, so the branch should gate on `CanProveDivisible(c_inner, c_outer)` and keep the nested modulo when it can't prove divisibility — mirroring what the same function's `scale != 1` branch already does. This is a change to the vendored TVM iter-map simplifier (not verified end-to-end); it is confined to that one branch, and the swept sibling shapes above (div/mixed/scaled) show that branch is the only one that folds without a sound divisibility check, so the fix does not disturb the other folds.

**Provenance.** The rewrite lives in the iter-affine-map padding support inherited from upstream TVM (the non-divisible floordiv/floormod iter-map work, 2022-2023 era); the current line in the bundled fork was last touched by an NFC FFI refactor (#18312, 2025-09-13), not introduced by it. Present since TileLang first vendored this TVM; origin PR in TVM not pinned here.

**Dedup.** I searched the open and closed tracker (`floormod`, `IterMap`, `modulo`, `Parallel index`, `FlattenBuffer`) and found no existing report of this nested-modulo miscompile.

**Reach.** The trigger is `B[j % K]` on the inner index of a 2-D `T.Parallel(M, N)` with `K ∤ N` — a common-idiom lookup: per-column bias/table/rotation indexed by a modulus is a standard kernel pattern and the natural way to express it. The 2-D `T.Parallel` fusion is the default for elementwise/epilogue loops, so the fusion+flatten path is ubiquitous; the extra ingredient is `K ∤ N` (whenever `K` divides `N` the fold is legitimate and the result is correct — e.g. `K=8, N=64` above). The dividing case is common because power-of-two tile widths and power-of-two moduli divide each other, so the bug hides behind the frequent all-powers-of-two configuration — but `K`/`N` need not be powers of two for the bug (any non-dividing pair with the inner modulus actually wrapping misfolds, per the sweep above). `grep -rn "% .*]" examples/ tests/` finds modulo-indexed reads, but the shipped examples use moduli/extents that divide (so they dodge the boundary); no test exercises a non-dividing modulus inside a 2-D `T.Parallel`, which is why CI is green.

**Impact.** The narrow ingredient is a nested modulo whose divisors are non-nesting — `K ∤ N` inside the fused 2-D parallel loop; power-of-two tile widths and moduli divide and stay correct, so the boundary is only crossed by odd/non-dividing table sizes. When it fires it is a silent wrong-code: the kernel compiles and runs with no error, corrupts a whole output region (512/1024 elements here, every row `i >= 1`), and the values are data-dependent on the `B` table contents rather than one noticeable constant — so it can pass a smoke run and slip into a workload undetected. Gating the fold on `c_outer | c_inner` closes this nested-modulo miscompile class in the iter-map simplifier; it does not otherwise change any correctly-folding case.

haoyang9804:
## Summary

- update the bundled TVM iter-map simplifier to preserve a nested modulo when the inner split can wrap and its extent is not divisible by the outer modulus
- retain the valid simplification for non-wrapping and divisible cases
- add mirrored TileLang regression coverage

This prevents `(x % 64) % 7` from being incorrectly rewritten as `x % 7` during `FlattenBuffer`.

Depends on tile-ai/tvm#64.

Fixes #2955.

## Testing

- native build passed
- TVM arithmetic suite: 39 passed
- TileLang arithmetic suite: 39 passed
- Metal codegen retains `(threadIdx.x & 63) % 7`

## Summary

- Updated the bundled TVM revision to preserve nested modulo expressions when divisibility is not proven.
- Prevented `(x % 64) % 7` from simplifying incorrectly to `x % 7`.
- Added regression coverage for divisible and non-divisible extents.
- Native build, TVM arithmetic, TileLang arithmetic, and Metal codegen tests pass.

## C++ style / lint notes

- The PR updates bundled TVM C++ code through a submodule revision.
- It does not modify `docs/developer_guide/cpp_style.md`.
- No new correctness or build issues are reported.
- Any C++ API Style Audit findings are warning-only and do not block the change.

### DIFF
--- a/testing/python/arith/test_arith_iter_affine_map.py
+++ b/testing/python/arith/test_arith_iter_affine_map.py
@@ -218,6 +218,18 @@ def test_compound_floormod_two_regression():
     )
 
 
+def test_nested_floormod_requires_divisible_extents():
+    x = tvm.tirx.Var("x", "int32")
+    flm = tvm.tirx.floormod
+    dom_map = var_dom([(x, 128)])
+
+    non_divisible = flm(flm(x, 64), 7)
+    assert_iter_map_simplify({non_divisible: non_divisible}, dom_map)
+
+    divisible = flm(flm(x, 64), 8)
+    assert_iter_map_simplify({divisible: flm(x, 8)}, dom_map)
+
+
 def test_predicate():
     x = tvm.tirx.Var("x", "int32")
     y = tvm.tirx.Var("y", "int32")


LeiWang1999:
Thanks! The TVM-side fix has been merged via [tile-ai/tvm#65](https://github.com/tile-ai/tvm/pull/65).

This PR now updates `3rdparty/tvm` from the previous PR head (`1b372fa482cf`) to the merged `tilelang_main` commit (`121df796b1b3`). The update is included in commit `f8fd99ef`.

PR #3066, opened 2026-08-22
[BugFix][Transform] Avoid int32 overflow in vector analysis

### LINKED ISSUE #3027: [BUG][Fuzzer][ice-on-valid-code] LoopVectorize's internal index-scale product overflows int32 and aborts compile for a large-stride affine access whose every real index fits int32
### Required prerequisites

0.1.13

### System information

- TileLang `0.1.13` (source tag `v0.1.13`, commit `8001cc4ccf6149382d2019654a19f59c1d4d0482`).
- Python 3.12, Linux.
- GPU: NVIDIA L40S (sm_89). The crash is in a host-side transform pass (`LoopVectorize`), before any GPU codegen, so it is architecture-independent — no GPU is needed to hit it.

### Problem description

A parallel loop containing a large-stride affine global access — a strided load `O[i] = A[i * stride]`, the mirror store `O[i * stride] = A[i]`, or an implicit large buffer row-stride `A[i, 0]` (all confirmed, see §17) — aborts compilation with

```
InternalError: Check failed: value < 1LL << (dtype.bits() - 1) (2400000000 vs. 2147483648)
  : Literal value 2400000000 exceeds maximum of int32
```

even when the kernel is entirely representable in int32 — in the repro below the largest logical index is `3 * 600_000_000 = 1.8e9 < 2^31` and the buffer has `2e9 < 2^31` elements, so no index or extent is out of int32 range. There is no way to compile this kernel as written; a slightly smaller `stride` compiles and gathers the correct values.

The crashing literal (`2400000000`) is **not** any index the kernel computes: it is an intermediate the vectorization pass forms internally while checking whether the index is invariant across a vector boundary. The trigger tracks that intermediate, not the real index range — e.g. `stride=500_000_000, extent=8` (max index `3.5e9`, genuinely above `2^31`) compiles fine, while `stride=600_000_000, extent=4` (max index `1.8e9`, well below `2^31`) crashes. So the crash is decoupled from actual int32-index overflow.

<details>
<summary>trigger boundary — the crash follows the internal scale, not the real max index</summary>

`float32` gather `O[i] = A[i*stride]`, buffer `5e9` elems (same across cells):

| extent | stride | max logical index | result |
|--------|--------|--------------------|--------|
| 4 | 500M | 1.5e9 | COMPILED-OK |
| 4 | 600M | 1.8e9 (< 2^31) | **CRASH** (`2400000000`) |
| 8 | 300M | 2.1e9 | **CRASH** (`2400000000`) |
| 8 | 500M | 3.5e9 (> 2^31) | COMPILED-OK |
| 8 | 600M | 4.2e9 | **CRASH** (`2400000000`) |

The crash literal is constant at `2400000000` regardless of the true max index, and cases with a genuinely out-of-int32 index (3.5e9) compile.
</details>

### Reproducible example code

```python
import tilelang
import tilelang.language as T
import torch

def gather(stride, extent, nel):
    @tilelang.jit(out_idx=[-1])
    def build():
        @T.prim_func
        def main(A: T.Tensor((nel,), "float32"), O: T.Tensor((extent,), "float32")):
            with T.Kernel(1, threads=extent):
                for i in T.Parallel(extent):
                    O[i] = A[i * stride]          # strided gather
        return main
    return build()

# FAIL: every index is a legal int32 (max = 3*600_000_000 = 1.8e9 < 2^31),
# buffer 2e9 elems (< 2^31) -> all values fit int32, yet compilation crashes.
try:
    gather(600_000_000, 4, 2_000_000_000)
    print("FAIL-CASE: COMPILED-OK (unexpected)")
except Exception as e:
    print("FAIL-CASE crash:", str(e).splitlines()[0])
# -> FAIL-CASE crash: ... Literal value 2400000000 exceeds maximum of int32

# CONTROL: identical kernel, stride 500_000_000 -> compiles and gathers correctly.
k = gather(500_000_000, 4, 2_000_000_000)
A = torch.arange(2_000_000_000, dtype=torch.float32, device="cuda")   # A[j] = j
O = k(A)
ref = [float(i * 500_000_000) for i in range(4)]
print("CONTROL:", O.tolist(), "==", ref, "->", O.tolist() == ref)
# -> CONTROL: [0.0, 500000000.0, 1000000000.0, 1500000000.0] == [...] -> True
```

### Traceback

```
InternalError: Check failed: value < 1LL << (dtype.bits() - 1) (2400000000 vs. 2147483648)
  : Literal value 2400000000 exceeds maximum of int32
```

<details>
<summary>abridged C++ frames</summary>

```
tvm::tl::GetVectorizeSize(...)
tvm::tl::VectorizePlanner::ComputeBufferVectorSize(...)
tvm::tl::IsExprInvariantInVectorBoundary(...)
tvm::arith::Analyzer::CanProveEqual(...)
tvm::arith::Analyzer::Simplify(...)
tvm::arith::SumExprNode::Normalize_(...)
tvm::arith::SplitExprNode::NormalizeWithScale(long)
tvm::tir::MakeConstScalar<long>(int32, 2400000000)
tvm::IntImm::IntImm(...)   <- Check failed
```
</details>

### Expected behavior

A strided gather `A[i * stride]` whose every index and buffer extent is within int32 should compile (and, as the control shows, it computes the correct values on a nearby stride). One direction is to perform the vector-boundary invariance arithmetic in a width that cannot overflow (int64) so the internal `stride × scale` intermediate is representable; another is to have the invariance check bail out of vectorization (return "not invariant" / fall back to the minimum vector size) when the simplifier would overflow, rather than aborting the whole compilation. Which is appropriate depends on how the pass is meant to treat wide-index kernels — a call for the maintainers.

### Additional context

**Root cause.** `LoopVectorize`'s vector-boundary invariance check does its index arithmetic in int32 and overflows on a large stride, so it aborts compilation instead of returning a vectorization decision. In [`ComputeBufferVectorSize`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/transform/loop_vectorize.cc#L801) the element offset is built as `elem_offset += transformed_indices[i] * strides[i]` (int32, [L814-L816](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/transform/loop_vectorize.cc#L814-L816)), then passed to [`IsExprInvariantInVectorBoundary`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/transform/loop_vectorize.cc#L830) at each trial vector size. That function substitutes `floordiv(var, vecsize)*vecsize` for the loop var and calls [`analyzer->CanProveEqual`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/transform/loop_vectorize.cc#L1010); canonical simplification then normalizes the split expression via `SumExprNode::NormalizeWithScale`, which multiplies the `stride` coefficient by the boundary scale and constructs an `int32` `IntImm` of the product (`2400000000` here) — tripping the `value < 2^31` guard in the `IntImm` constructor. Because the overflow is in the *internal* scaled coefficient, it fires for indices that are themselves valid int32 (repro: max index `1.8e9`) and does not fire for some genuinely out-of-range indices (`3.5e9`).

**Suggested fix.** In `IsExprInvariantInVectorBoundary`/`ComputeBufferVectorSize` (`loop_vectorize.cc`), either build the `elem_offset` invariance expression in int64 so the `stride × scale` intermediate cannot overflow, or have the invariance check fall back to the minimum vector size when the simplifier would overflow, instead of aborting compilation.

**Provenance.** Present on 0.1.13 as shipped (and on 0.1.9). The invariance-check-based boundary analysis that constructs and simplifies `elem_offset` here was introduced in [#2050](https://github.com/tile-ai/tilelang/pull/2050) ("Replace syntactic loop-var checks with invariance checks", merged 2026-04-17); whether the same input crashed via the prior syntactic check is unverified. On `main` after the #2935 refactor the same `elem_offset += indices[i]*strides[i]` construction feeds `IsExprInvariantInVectorBoundary` → `CanProveEqual`, unchanged, so the path is still present (inspected, not run — only 0.1.13 was executed this session).

**Dedup.** I searched the open and closed tracker and found no existing report of this compile crash. #2560 (int32 wrap of a compile-time-constant index) is a different defect: it is a *silent wrong address* in `FlattenBuffer`, triggered by an index genuinely above `2^31`, whereas this is a *compile-time crash* in `LoopVectorize` that fires for indices within int32.

**Reach.** The trigger is an affine global index `A[i * stride]` (or an implicit large buffer row-stride) — an ordinary kernel idiom (a competent user writes exactly this for a strided read; it is not a fuzzer-only shape). The specific ingredient needed to trip it is a coefficient large enough that `coeff × (vector-boundary scale)` crosses `2^31`; with the default `float32` vectorization the crossing point in the repro is around `stride ≈ 6e8`, so the affected region is large strides into large buffers rather than typical tile-local access. In the v0.1.13 tree `119` example files use `T.Parallel` loops with strided/affine global indices, so the idiom is pervasive, but the shipped examples all use tile-sized strides far below this boundary; none trip it and no test exercises a stride near `2^31` — which is why CI is green. Verified by running the shipped `examples/elementwise/example_elementwise_add.py` verbatim on 0.1.13: it uses this exact `T.Parallel` + affine-global-index idiom (`A[by*block_M, bx*block_N]`, `block_M=block_N=32` into a 1024×1024 buffer) and compiled + passed its own `assert_close` (exit 0) — it dodges because tile-sized strides keep the internal scale far below `2^31`.

**Impact.** The trigger is narrow: it needs a stride large enough that `stride × vector-boundary-scale` crosses `2^31` (≈`6e8` for `float32`), so it is confined to large-stride gathers into large buffers, not tile-local access. When it fires it is a loud compile-time `InternalError` that aborts the build of that one kernel — caught immediately, never a silent wrong result, and it cannot reach a running workload or corrupt any output; the deterministic literal in the message also does not correspond to a real index. Fixing it closes a valid-input class (a strided gather whose every index and extent fits int32) that currently cannot be built.

<details>
<summary>§17 Generalization record</summary>

**Two-level root.**
- *SOURCE-level (which implementation is fragile):* bundled TVM's canonical simplifier, [`SplitExprNode::NormalizeWithScale`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/arith/canonical_simplify.cc#L129) — it builds the accumulated coefficient as `res = res * make_const(dtype, sscale)` where `dtype` is the index's `int32` and `sscale = stride × boundary-scale`; the `IntImm` constructor's `value < 2^31` guard aborts on the product. The TileLang sink that feeds it is `LoopVectorize`'s `elem_offset += transformed_indices[i]*strides[i]` (int32, `loop_vectorize.cc` L814-816) → `IsExprInvariantInVectorBoundary` (L830) → `CanProveEqual` (L1010) → `Analyzer::Simplify`.
- *OPERATOR-level (what correlates):* any op that hands `LoopVectorize` a large-coefficient affine int32 index (explicit `i*stride` OR an implicit large buffer row-stride) and routes it through the vector-boundary invariance check — load, store, and multi-dim indexing all share the same `elem_offset` accumulation and sink.

**Found first.** `float32` gather `O[i] = A[i*stride]`, `extent=4`, `stride=600_000_000`: largest logical index `3 * 600_000_000 = 1.8e9 < 2^31`, buffer `< 2^31` elements, yet compilation aborts with `Literal value 2400000000 exceeds maximum of int32`. The crashing literal `2400000000 = 4 * 600_000_000` is the internal `stride × vector-boundary-scale` product, not any index the kernel computes.

**Class hypothesis.** *Un-narrowed construct*: an int32 `IntImm` is built from a scaled coefficient that is never reduced to a width that can hold it, so the `IntImm` guard aborts on an intermediate whose result would be fine at 64-bit. Same-family (do NOT merge) as filed **#2982** (`T.fill(frag_f32, -3.4028235e38)`), where the same range-check aborts on an un-narrowed literal — different sink (frontend fill constant vs. an internal scaled coefficient inside the vectorizer), different pass, same underlying guard; recorded here only as shared root.

**4-axis sweep (each cell RUN on L40S, 0.1.13, one kernel per fresh process):**

| axis | cell tested | observed result | same-root? |
|------|-------------|-----------------|------------|
| related-type | `int32` buffer, `stride=600M` | **CRASH** — `... 2400000000 exceeds maximum of int32` | same-root (32-bit element → same scale → same product; numeric semantics irrelevant) |
| related-type | `float16` buffer, `stride=600M` | **CRASH** — same literal `2400000000` | same-root (element width changes the scale but a stride still exists that crosses `2^31`) |
| related-type | `float16` buffer, `stride=300M` | **CRASH** — `Can't fetch the lanes of a scalable vector at a compile time` | **distinct-adjacent** — different sink/symptom (scalable-vector lane query), not the int32 `IntImm` guard; not filed |
| related-type | `float32`, `extent=8`, `stride=500M` (real index genuinely `> 2^31`, internal scale stays under) | **COMPILED-OK** (max index 3.5e9) | same-root — proves the crash tracks the *internal scale*, not the real index |
| related-operator | strided **store** `O[i*stride] = A[i]`, `stride=600M` | **CRASH** — same literal `2400000000` | same-root (store path L580/593 shares the `elem_offset` accumulation + sink) |
| related-operator | 2D **implicit row-stride** `A[i, 0]` with row width `600M` | **CRASH** — same literal `2400000000` | same-root — the coefficient can come from a buffer stride, not just an explicit `*`; broadens class beyond the `A[i*stride]` idiom |
| similar-logic | condition-path `if i*600M < N` (sibling `IsExprInvariantInVectorBoundary` site, `CheckConditionVectorized` L628-635) | **COMPILED-OK** | same source function reached, but this shape did not force the split-normalize through the overflowing scale — sibling site, not triggered by this input |

**Reframe.** Class confirmed for load, store, and implicit-row-stride affine indexing through `LoopVectorize` (all same root, same literal); title/Problem generalized from "strided global load" to "large-stride affine index". The `float16`+`stride=300M` "scalable vector" error is a **distinct-adjacent** bug (different sink, same pass) — recorded, not filed. No NEW independently-fileable bug found while sweeping. The condition-path sibling site shares the fragile source but did not reproduce on the shapes tried.

**Example-run (PART 1).** `examples/elementwise/example_elementwise_add.py` (v0.1.13) uses this idiom (`T.Parallel` + affine global index `A[by*block_M, bx*block_N]`) and was run verbatim: compiled and passed `assert_close`, exit 0 — dodges because `block_M=block_N=32` into a 1024×1024 buffer keeps the internal scale far below `2^31`.
</details>

kobecai:
## Motivation

  Fixes #3027.

  TileLang's vectorization analysis may abort while simplifying a valid int32
  index expression. The failure occurs when the canonical simplifier scales an
  internal coefficient and attempts to materialize the result as an int32
  `IntImm`.

  For example:

  ```python
  A[(i - (i // 4) * 4) * 600_000_000]
  ```

  For `i` in `[0, 8)`, this expression only evaluates to:

  ```text
  0, 600M, 1.2B, 1.8B
  ```

  All of these indices fit within int32. During vector-boundary analysis,
  however, the canonical simplifier may internally construct:

  ```text
  4 * 600_000_000 = 2_400_000_000
  ```

  This exceeds the int32 `IntImm` range and aborts compilation inside
  `SplitExprNode::NormalizeWithScale`:

  ```text
  Literal value 2400000000 exceeds maximum of int32
  ```

  The overflowing value is an analyzer intermediate, not an index evaluated by
  the generated kernel.

  ## Root cause

  `IsExprInvariantInVectorBoundary` substitutes an aligned loop variable and
  passes the original int32 expression to `Analyzer::CanProveEqual`.

  Canonical simplification represents affine coefficients internally as int64,
  but eventually materializes a scaled coefficient using the expression's
  original dtype. A large int32 coefficient multiplied by the candidate vector
  width can therefore exceed the int32 `IntImm` range.

  The same problem can also occur in `CanProveIndependent`. Although that check
  does not apply an additional vector-boundary substitution, the input expression
  may already contain a large internal coefficient such as
  `4 * 600_000_000`.

  ## Changes

  This PR reuses the existing `Int64Promoter` to widen risky expressions for
  vectorization analysis only.

  Before invoking the analyzer, it:

  1. visits signed integer scalar subexpressions;
  2. obtains their constant integer bounds;
  3. checks whether scaling those bounds by the analysis scale may exceed the
     source dtype;
  4. promotes a temporary analysis copy to int64 only when needed.

  The prepared analysis expression is used by:

  - `CanProveIndependent`;
  - `IsExprInvariantInVectorBoundary`;
  - `IndicesCanVectorize`.

  The check visits integer subexpressions rather than only examining the root
  expression. This also covers:

  - integer arithmetic inside boolean conditions;
  - cancelling or bounded affine expressions whose intermediate coefficients are
    larger than their final value range;
  - expressions that already contain an overflowing internal coefficient before
    vector-boundary scaling.

  ## Semantics and scope

  The promoted expression is temporary and is not written back to the
  `PrimFunc`.

  Therefore, this change does not alter:

  - BufferLoad or BufferStore index dtypes;
  - the generated kernel's index dtype;
  - `NarrowDataType(32)` behavior;
  - runtime address arithmetic for existing valid int32 kernels.

  The int64 expression is only used by the analyzer and vectorization planner to
  produce an invariance or vector-width decision.

  Expressions whose scaled bounds fit their existing dtype continue through the
  original analysis path without promotion.

  This PR does not change the vectorization policy for strided gathers. A
  large-stride gather that is not vectorizable still falls back to scalar access;
  it simply no longer aborts during analysis.

  ## Tests

  Added host-side transform tests covering:

  - the original large-stride load;
  - the corresponding large-stride store;
  - an implicit large 2-D row stride;
  - a bounded residual expression that reproduces the failure on current `main`;
  - preservation of unit-stride `Ramp` vectorization after analysis-only
    promotion;
  - preservation of a successful large-coefficient boundary-invariance proof.

  The bounded-residual regression fails before this change with:

  ```text
  Literal value 2400000000 exceeds maximum of int32
  ```

  and passes after the change.

  ### Host-side validation

  ```bash
  python -m pytest \
    testing/python/transform/test_tilelang_transform_loop_vectorize.py \
    testing/python/transform/test_tilelang_transform_legalize_vectorized_loop.py \
    -q
  ```

  Result:

  ```text
  7 passed
  ```

  Formatting and static checks:

  ```bash
  python -m pre_commit run --files \
    src/transform/loop_vectorize.cc \
    testing/python/transform/test_tilelang_transform_loop_vectorize.py
  ```

  Result:

  ```text
  All checks passed
  ```

  ### CUDA validation

  CUDA compilation and runtime validation passed.

  The following cases compile and produce the expected results:

  - the original issue #3027 large-stride gather;
  - the large-stride store variant;
  - the bounded-residual regression;
  - existing vectorized index and boundary-invariant index cases.

  The bounded-residual kernel accesses the expected index sequence:

  ```text
  0, 600M, 1.2B, 1.8B, 0, 600M, 1.2B, 1.8B
  ```

  Checklist:

  - [x] Host-side regression tests
  - [x] Existing vectorized-loop legalization test
  - [x] Formatting and static checks
  - [x] CUDA compilation validation
  - [x] CUDA runtime correctness validation

## Summary

- Prevents int32 overflow during TileLang vectorization analysis.
- Uses `Int64Promoter` for temporary int64 analysis in independence, boundary-invariance, and index-vectorization checks.
- Preserves original `PrimFunc` expressions, index dtypes, runtime address arithmetic, and `NarrowDataType(32)` behavior.
- Adds CUDA tests for large-stride loads, stores, implicit row strides, bounded residuals, large offsets, and boundary-invariant indices.

## Validation

- Host-side tests passed.
- Formatting checks passed.
- CUDA compilation passed.
- CUDA runtime validation passed.

## C++ style / lint notes

- The PR changes C++ implementation code but does not touch rules documented in `docs/developer_guide/cpp_style.md`.
- No correctness or build issues are reported. TLCPP003/TLCPP004 findings should not block merge unless they identify a clear API, FFI, or maintainability risk.

### DIFF
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -49,6 +49,13 @@ namespace tl {
 using namespace tirx;
 using namespace ffi;
 
+namespace {
+
+PrimExpr SimplifyExprForAnalyzer(const PrimExpr &expr, int scale,
+                                 arith::Analyzer *analyzer);
+
+} // namespace
+
 /*!
  * \brief Check if buffer strides represent a contiguous (row-major) layout.
  * \param buffer The buffer to check.
@@ -733,8 +740,12 @@ private:
     if (!inner_for_) {
       return;
     }
-    PrimExpr condition = analyzer_->Simplify(cond);
     int condition_vector_size = loop_extent_vector_size_;
+    PrimExpr condition = cond;
+    if (condition_vector_size > 1) {
+      condition =
+          SimplifyExprForAnalyzer(cond, condition_vector_size, analyzer_);
+    }
     while (condition_vector_size > 1 &&
            !IsExprInvariantInVectorBoundary(condition, inner_for_->loop_var,
                                             condition_vector_size, analyzer_)) {
@@ -808,7 +819,10 @@ private:
     // of the whole loop body. To avoid planning a vector size that will be
     // immediately scalarized (and to keep semantics sane for side-effectful
     // calls), require the offset to be invariant within the vector boundary.
-    PrimExpr offset_s = analyzer_->Simplify(offset);
+    PrimExpr offset_s = offset;
+    if (access_vec_size > 1) {
+      offset_s = SimplifyExprForAnalyzer(offset, access_vec_size, analyzer_);
+    }
     while (access_vec_size > 1 &&
            !IndicesCanVectorize(offset_s, inner_for_->loop_var,
                                 inner_for_->extent, access_vec_size,
@@ -1109,6 +1123,11 @@ PrimExpr PrepareExprForAnalyzer(const PrimExpr &expr, int scale,
   return promoter(expr);
 }
 
+PrimExpr SimplifyExprForAnalyzer(const PrimExpr &expr, int scale,
+                                 arith::Analyzer *analyzer) {
+  return analyzer->Simplify(PrepareExprForAnalyzer(expr, scale, analyzer));
+}
+
 } // namespace
 
 bool CanProveIndependent(const PrimExpr &expr, Var var,
@@ -1120,10 +1139,12 @@ bool CanProveIndependent(const PrimExpr &expr, Var var,
     return true;
   }
   // 2. if \forall v_1, v_2, f(v_1) == f(v_2), f is independent with v
-  PrimExpr analysis_expr = PrepareExprForAnalyzer(expr, 1, analyzer);
   Var var_1("_t", var.dtype());
-  PrimExpr expr_1 = Substitute(analysis_expr, {{var, var_1}});
-  if (analyzer->CanProveEqual(analysis_expr, expr_1)) {
+  PrimExpr expr_1 = Substitute(expr, {{var, var_1}});
+  PrimExpr equality = PrepareExprForAnalyzer(expr == expr_1, 1, analyzer);
+  const auto *equality_node = equality.as<EQNode>();
+  ICHECK(equality_node);
+  if (analyzer->CanProveEqual(equality_node->a, equality_node->b)) {
     return true;
   }
   return false;
--- a/src/transform/loop_vectorize.cc
+++ b/src/transform/loop_vectorize.cc
@@ -29,6 +29,7 @@
 #include "arith/int_operator.h"
 #include "arith/ir_visitor_with_analyzer.h"
 #include "backend/common/target_utils.h"
+#include "common/int64_promoter.h"
 #include "common/loop_vectorization_utils.h"
 #include "support/check.h"
 #include <iostream>
@@ -1065,6 +1066,51 @@ int GetVectorizeSize(const For &loop, arith::Analyzer *analyzer,
   return VectorizePlanner(analyzer, layout_map).Plan(loop);
 }
 
+namespace {
+
+// Canonical simplification may multiply an integer subexpression's coefficient
+// by the candidate vector width before it constructs an IntImm. Widen the
+// analysis copy if any such scaled bound can exceed the source dtype. Inspect
+// subexpressions as well as the root so boolean conditions and cancelling
+// affine terms cannot hide a risky integer coefficient.
+bool ScaledIntSubexprMayOverflow(const PrimExpr &expr, int scale,
+                                 arith::Analyzer *analyzer) {
+  ICHECK_GE(scale, 1);
+  bool may_overflow = false;
+  PostOrderVisit(expr, [&](const ObjectRef &obj) {
+    if (may_overflow) {
+      return;
+    }
+    Optional<PrimExpr> opt_subexpr = obj.as<PrimExpr>();
+    if (!opt_subexpr.defined()) {
+      return;
+    }
+    PrimExpr subexpr = opt_subexpr.value();
+    DataType dtype = subexpr.dtype();
+    if (!dtype.is_int() || dtype.bits() >= 64 || dtype.lanes() != 1) {
+      return;
+    }
+
+    arith::ConstIntBound bound = analyzer->const_int_bound(subexpr);
+    const int64_t type_max = (1LL << (dtype.bits() - 1)) - 1;
+    const int64_t type_min = -(1LL << (dtype.bits() - 1));
+    may_overflow = bound->max_value > type_max / scale ||
+                   bound->min_value < type_min / scale;
+  });
+  return may_overflow;
+}
+
+PrimExpr PrepareExprForAnalyzer(const PrimExpr &expr, int scale,
+                                arith::Analyzer *analyzer) {
+  if (!ScaledIntSubexprMayOverflow(expr, scale, analyzer)) {
+    return expr;
+  }
+  Int64Promoter promoter;
+  return promoter(expr);
+}
+
+} // namespace
+
 bool CanProveIndependent(const PrimExpr &expr, Var var,
                          arith::Analyzer *analyzer) {
   // 1. if var doesn't exist, it is independent
@@ -1074,9 +1120,10 @@ bool CanProveIndependent(const PrimExpr &expr, Var var,
     return true;
   }
   // 2. if \forall v_1, v_2, f(v_1) == f(v_2), f is independent with v
+  PrimExpr analysis_expr = PrepareExprForAnalyzer(expr, 1, analyzer);
   Var var_1("_t", var.dtype());
-  auto expr_1 = Substitute(expr, {{var, var_1}});
-  if (analyzer->CanProveEqual(expr, expr_1)) {
+  PrimExpr expr_1 = Substitute(analysis_expr, {{var, var_1}});
+  if (analyzer->CanProveEqual(analysis_expr, expr_1)) {
     return true;
   }
   return false;
@@ -1095,10 +1142,12 @@ bool IsExprInvariantInVectorBoundary(const PrimExpr &expr, Var var,
   //     A[i] = B[i] * C[i//4]
   // if vecsize=4, f(i)=i//4 depends only on i//4
   // Therefore A[i] = B[i] * C[i//4] can be vectorized with vecsize=4
+  PrimExpr analysis_expr =
+      PrepareExprForAnalyzer(expr, target_vectorized_size, analyzer);
   PrimExpr var_aligned =
       floordiv(var, target_vectorized_size) * target_vectorized_size;
-  PrimExpr expr_aligned = Substitute(expr, {{var, var_aligned}});
-  if (analyzer->CanProveEqual(expr, expr_aligned)) {
+  PrimExpr expr_aligned = Substitute(analysis_expr, {{var, var_aligned}});
+  if (analyzer->CanProveEqual(analysis_expr, expr_aligned)) {
     return true;
   }
   return false;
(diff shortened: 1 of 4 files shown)


src/transform/loop_vectorize.cc
```diff
@@ -1065,6 +1066,51 @@ int GetVectorizeSize(const For &loop, arith::Analyzer *analyzer,
...
+  });
+  return may_overflow;
+}
+
+PrimExpr PrepareExprForAnalyzer(const PrimExpr &expr, int scale,
+                                arith::Analyzer *analyzer) {
+  if (!ScaledIntSubexprMayOverflow(expr, scale, analyzer)) {
+    return expr;
+  }
+  Int64Promoter promoter;
```
SiriusNEO, line 1122: Currently I'm not sure the `T.Cast` case is correctly handled. Consider the following case:
```python
idx = T.cast(
    (i - (i // 4) * 4) * 600_000_000,
    "int64",
)
O[i] = A[idx]
```
Here the intermediate index expr will overflow, but since it uses int64 promoter here, it will ignore this overflow, which may case the problem?

kobecai: Thanks for pointing this out. You're right that `Int64Promoter` itself does not recurse through an already-`int64` cast.

  I added this exact case as a regression test. In eager TIRX, it is represented as:

  ```python
  idx: T.int64 = T.Cast("int64", int32_expr)
  output[i] = A[idx]
  ```

  The cast is stored in a Bind, while the buffer index is the bound Var. VectorizePlanner visits the Bind value and binds it to the Analyzer. During that binding, the Analyzer canonicalizes the integer cast and pushes the int64 type into the affine arithmetic before analyzing the later buffer index.

  Therefore, although the original bounded residual expression remains int32, overflow-prone affine coefficients created during analysis are evaluated in int64. This only widens the analysis copy; the original IR remains Cast(int64, int32_expr), so runtime arithmetic semantics are unchanged.

  The new regression test verifies that the original narrow expression is preserved under the cast and that LayoutInference completes without aborting.

  Validation: all 7 dedicated regression tests passed. The full transform suite completed with 465 passed, 37 skipped, and 1 unrelated failure in `test_hoist_broadcast[dtype3]`.

kobecai:
Pushed a follow-up fix in `676b50cf` for the CUDA CI failure.

  The vectorizer was changing a `BufferLoad` from scalar to vector indices without updating the load dtype. This left the load typed as scalar and eventually produced an invalid `Cast(int64, int32x4)` during analyzer simplification.

  The fix reconstructs the `BufferLoad` after vectorizing its indices, so the load dtype is updated to `int32x4` and the widening cast becomes the lane-consistent `Cast(int64x4, int32x4)`.

  I also added a regression test covering the dynamic int32 index + `atomic_addx4` case. The affected sparse MLA backward test passes on GPU, and the local loop-vectorization and related transform tests pass.

PR #3085, opened 2026-08-25
[BugFix] Don't drop syncs from the other if branch

### LINKED ISSUE #3042: [BUG] 屏障误删导致的共享内存的读写冲突
### Required prerequisites

0.1.12

### System information

root@s-113-2-32:/mnt/nas/home/jianpeng.zhu/project/vllm_torch_op/150s/git/tilelang# python3 -m torch.utils.collect_env
<frozen runpy>:128: RuntimeWarning: 'torch.utils.collect_env' found in sys.modules after import of package 'torch.utils', but prior to execution of 'torch.utils.collect_env'; this may result in unpredictable behaviour
Collecting environment information...
PyTorch version: 2.10.0
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.2 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version: 22.1.0git (4.5.0.20260721 7b070f4f757e4199bcefb8b8e743b4000b81e366)
CMake version: version 4.4.0
Libc version: glibc-2.39

Python version: 3.12.11 (main, Jul 31 2025, 06:45:11) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-5.4.0-42-generic-x86_64-with-glibc2.39
Is CUDA available: True
CUDA runtime version: 10.2.89
CUDA_MODULE_LOADING set to: 
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Could not collect
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
Caching allocator config: N/A

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   46 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          160
On-line CPU(s) list:             0-159
Vendor ID:                       GenuineIntel
BIOS Vendor ID:                  Intel(R) Corporation
Model name:                      Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz
BIOS Model name:                 Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz  CPU @ 2.3GHz
BIOS CPU family:                 179
CPU family:                      6
Model:                           106
Thread(s) per core:              2
Core(s) per socket:              40
Socket(s):                       2
Stepping:                        6
CPU(s) scaling MHz:              24%
CPU max MHz:                     3400.0000
CPU min MHz:                     800.0000
BogoMIPS:                        4600.00
Flags:                           fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm3dnowprefetch cpuid_fault epb cat_l3 invpcid_single ssbd mba ibrs ibpb stibp ibrs_enhanced tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb intel_pt avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local wbnoinvd dtherm ida arat pln pts hwp hwp_act_window hwp_epp hwp_pkg_req avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg tme avx512_vpopcntdq rdpid md_clear pconfig flush_l1d arch_capabilities
Virtualization:                  VT-x
L1d cache:                       3.8 MiB (80 instances)
L1i cache:                       2.5 MiB (80 instances)
L2 cache:                        100 MiB (80 instances)
L3 cache:                        120 MiB (2 instances)
NUMA node(s):                    2
NUMA node0 CPU(s):               0-39,80-119
NUMA node1 CPU(s):               40-79,120-159
Vulnerability Itlb multihit:     Not affected
Vulnerability L1tf:              Not affected
Vulnerability Mds:               Not affected
Vulnerability Meltdown:          Not affected
Vulnerability Spec store bypass: Mitigation; Speculative Store Bypass disabled via prctl and seccomp
Vulnerability Spectre v1:        Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:        Mitigation; Enhanced IBRS, IBPB conditional, RSB filling
Vulnerability Srbds:             Not affected
Vulnerability Tsx async abort:   Not affected

Versions of relevant libraries:
[pip3] numpy==1.26.4
[pip3] onnx==1.22.0
[pip3] onnxruntime-gpu==1.17.1+corex.4.5.0.20260721
[pip3] torch==2.10.0+corex.4.5.0.20260721
[pip3] torch_c_dlpack_ext==0.1.5
[pip3] torch_cluster==1.6.0+corex.4.5.0.20260721
[pip3] torch_quiver==0.1.0+corex.4.5.0.20260721
[pip3] torch_scatter==2.1.0+corex.4.5.0.20260721
[pip3] torch_sparse==0.6.16+corex.4.5.0.20260721
[pip3] torchaudio==2.10.0+corex.4.5.0.20260721
[pip3] torchdebug==4.4.0+corex.4.5.0.20260721
[pip3] torchvision==0.25.0+corex.4.5.0.20260721
[pip3] triton==3.2.0+corex.4.5.0.20260721

### Problem description

因为线程发散走到这个分支，这边的代码仅对then分支进行了是否删除屏障的判断，就将then和else的屏障都删除了，但else中如果存在共享内存的读写操作，会因此导致读写冲突

### Reproducible example code

The Python snippets:

```python

```

### Traceback

```pytb

```

### Expected behavior

_No response_

### Additional context

_No response_

haoyang9804:
Fixes #3042.

## What went wrong

`ThreadSync` was checking only the `then` condition when it decided if a barrier had to be moved out of an `if`. If that check said yes, it removed barriers from **both** the `then` and `else` branches.

That is not always right becuase the two branches can have different sets of participating threads. A barrier may be unsafe in `then`, but still be a perfectly valid partial barrier in `else`.

## Small repro

The test uses 128 threads and this branch:

```python
if tx == 0 or tx >= 33:
    l[0] = 0.0
else:
    s[tx] = T.cast(tx, "float32")
    l[0] = s[33 - tx]
```

Only threads `1..32` enter the `else`. They all write shared memory first, then read the value written by another thread (`1 <-> 32`, `2 <-> 31`, etc), so we need a 32-thread partial barrier between those two ops.

Before this fix the pass looked at the `then` condition, decided to hoist, and generated something like:

```python
T.tvm_storage_sync("shared")
if tx == 0 or tx >= 33:
    ...
else:
    s[tx] = ...
    l[0] = s[33 - tx]
```

The barrier is now before both accesses, so it doesnt order anything inside `else`. One warp can read before the other warp writes and get an old/uninitialized value.

## Fix

Check the branches seperately:

- use `op->condition` for `then`
- use `Not(op->condition)` for `else`
- only remove syncs from the branch that actually needs hoisting

After the fix the reprodcuer keeps this in the `else` branch:

```python
s[tx] = ...
T.tvm_storage_sync("shared", barrier_id, 32)
l[0] = s[33 - tx]
```

## When it triggers

It needs this combo:

1. `ThreadSync` finds a cross-thread shared-memory dependency inside one arm of an `IfThenElse`.
2. The two arms have different partial-barrier participation patterns.
3. The old decision based on the `then` condition gets applied to syncs from both arms.

## Tested

- rebuilt with cmake
- full `test_tilelang_transform_thread_sync.py`: `21 passed, 21 skipped` on my Mac
- pre-commit passed

## Summary

Fixes `ThreadSync` barrier handling for `IfThenElse` branches.

- Analyzes each branch with its own condition.
- Removes synchronization only from the branch that requires hoisting.
- Preserves required partial barriers in the `else` branch.
- Prevents shared-memory read/write conflicts.
- Adds masked warp-sync handling for ballot-derived masks.
- Rejects non-warp-uniform synchronization conditions.
- Adds regression coverage for partial barriers, explicit synchronization, and masked warp synchronization.

## Validation

- 21 tests passed.
- 21 tests skipped.
- CMake rebuild passed.
- Pre-commit checks passed.

## C++ style / lint notes

The change touches C++ implementation code but does not modify rules in `docs/developer_guide/cpp_style.md`.

The C++ API Style Audit runs in warning-only mode. Any TLCPP003/TLCPP004 findings are advisory and separate from correctness, build, and test results.

### DIFF
--- a/src/transform/thread_storage_sync.cc
+++ b/src/transform/thread_storage_sync.cc
@@ -1064,8 +1064,7 @@ struct TileLangThreadSyncPlanner : public ConstrVisitor {
       ConditionThreadPropertyChecker checker(&analyzer, env_threads_,
                                              let_var_properties_);
       auto validate_sync_condition = [&](const PrimExpr &branch_condition,
-                                         const char *branch_name,
-                                         const Stmt &branch_body) {
+                                         const char *branch_name) {
         auto condition_prop = checker.AnalyzeExpr(branch_condition);
         bool proven_block_uniform = !condition_prop.is_block_uniform &&
                                     IsBlockUniformCondition(branch_condition);
@@ -1090,16 +1089,14 @@ struct TileLangThreadSyncPlanner : public ConstrVisitor {
                "barrier lowering. Hoisting the barrier before the if would "
                "not order the conflicting accesses inside the branch. "
                "Condition: "
-            << branch_condition << "\nBranch IR:\n"
-            << branch_body;
+            << branch_condition;
       };
 
       if (has_sync_in_then) {
-        validate_sync_condition(op->condition, "then", op->then_case);
+        validate_sync_condition(op->condition, "then");
       }
       if (has_sync_in_else) {
-        validate_sync_condition(tirx::Not(op->condition), "else",
-                                op->else_case.value());
+        validate_sync_condition(tirx::Not(op->condition), "else");
       }
     }
 
@@ -2187,6 +2184,35 @@ private:
         ICHECK(prev_indice_bytes.dtype() == curr_indice_bytes.dtype());
         provably_disjoint =
             analyzer.CanProve(tirx::NE(prev_indice_bytes, curr_indice_bytes));
+        // 对短小展开循环逐次代入迭代值，避免 div/mod 组合让符号证明返回
+        // unknown。
+        if (!provably_disjoint && loop != nullptr &&
+            loop->kind == ForKind::kUnrolled) {
+          const auto *loop_min = loop->min.as<IntImmNode>();
+          const auto *loop_extent = loop->extent.as<IntImmNode>();
+          constexpr int64_t kMaxEnumeratedUnrollExtent = 64;
+          if (loop_min != nullptr && loop_extent != nullptr &&
+              loop_extent->value > 1 &&
+              loop_extent->value <= kMaxEnumeratedUnrollExtent) {
+            provably_disjoint = true;
+            for (int64_t offset = 0; offset < loop_extent->value - 1;
+                 ++offset) {
+              Map<Var, PrimExpr> iteration_sub;
+              iteration_sub.Set(
+                  loop->loop_var,
+                  make_const(loop->loop_var.dtype(), loop_min->value + offset));
+              PrimExpr prev_at_iteration =
+                  Substitute(prev_indice_bytes, iteration_sub);
+              PrimExpr curr_at_iteration =
+                  Substitute(curr_indice_bytes, iteration_sub);
+              if (!analyzer.CanProve(
+                      tirx::NE(prev_at_iteration, curr_at_iteration))) {
+                provably_disjoint = false;
+                break;
+              }
+            }
+          }
+        }
       } else {
         try {
           auto prev_min = analyzer.Simplify(
--- a/src/transform/thread_storage_sync.cc
+++ b/src/transform/thread_storage_sync.cc
@@ -437,10 +437,12 @@ private:
 struct ConditionThreadProperty {
   bool depends_on_runtime{false};
   bool is_block_uniform{true};
+  bool requires_hoist{false};
 
   void Merge(const ConditionThreadProperty &other) {
     depends_on_runtime = depends_on_runtime || other.depends_on_runtime;
     is_block_uniform = is_block_uniform && other.is_block_uniform;
+    requires_hoist = requires_hoist || other.requires_hoist;
   }
 };
 
@@ -448,7 +450,7 @@ struct ConditionThreadProperty {
  * \brief Analyze whether an if-condition is runtime-dependent and/or uniform
  * across all threads in a block.
  *
- * For branch-local sync validation we care about two independent properties:
+ * For sync hoisting decisions we care about two independent properties:
  *
  * 1. Does the condition depend on runtime values such as memory loads?
  * 2. Even if it does, is it still block-uniform, i.e. identical for every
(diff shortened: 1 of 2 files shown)


src/transform/thread_storage_sync.cc
```diff
@@ -991,22 +991,28 @@ struct TileLangThreadSyncPlanner : public ConstrVisitor {
...
+        // A partial barrier is unsafe when the participating thread count is
+        // unknown. Keep a plain __syncthreads() only for proven block-uniform
+        // conditions.
+        return (condition_prop.depends_on_runtime && !is_block_uniform) ||
+               (condition_prop.requires_hoist && !proven_uniform);
+      };
+
+      bool hoist_then = !syncs_in_then.empty() && must_hoist(op->condition);
+      bool hoist_else =
+          !syncs_in_else.empty() && must_hoist(tirx::Not(op->condition));
```
SiriusNEO: In my understanding, here it only checks whether the else branch is a consecutive group of 32. If we have `tx=1...32` (which is not warp-aligned), it will be hoisted also.

Consider the following example. Do you think we should fix this case?

```python
  @T.prim_func
  def main():
      s = T.alloc_buffer((64,), "float32", scope="shared")
      out = T.alloc_buffer((128,), "float32", scope="global")
      tx = T.launch_thread("threadIdx.x", 128)
      if tx == 0 or tx >= 33:
          out[tx] = 0.0
      else:
          s[tx] = T.cast(tx, "float32")
          out[tx] = s[33 - tx]
```

haoyang9804: Yes, we should fix it. Good catch.
Checking for 32 consecutive threads is not enough; the participant set also needs to be warp-aligned. In the current patch, `tx=1..32` is actually kept rather than hoisted and lowered to a partial `bar.sync`, which is unsafe. I’ll require warp-uniform participation for partial barriers, report unsupported cases instead of hoisting them, and update the tests with aligned and misaligned cases.

Does this approach make sense to you?

SiriusNEO:
@regression-perf

PR #3087, opened 2026-08-25
[Bugfix] Fix wait parity for explicit mbarriers in pipelined loops

Yongqi-Zhuo:
A synchronous T.gemm(mbar=...) inside T.Pipelined waited on the wrong mbarrier parity. InjectSoftwarePipeline annotated every copy/im2col/gemm call with the ring parity (k / num_stages) % 2, but that law belongs to the compiler-generated pipeline_mbar ring: an explicit mbarrier whose arrive happens inside a tile op is deliberately excluded from ExpandPipelineBarriers, so it remains a single slot that flips once per iteration. The auto-emitted wait asked for a parity that had already completed, returned immediately, and the next iteration's copies overwrote a shared-memory stage the still-running MMA was reading (wrong results observed at num_stages >= 3).

Annotate only the ops that consume the phase -- those that auto-emit a paired arrive+wait on an explicit mbarrier: sync gemm with an mbar, and copies carrying an explicit "barrier" that are not is_tma_copy -- and annotate them with the logical iteration parity (k - min) % 2, written against the original loop var. EmitImpl's existing per-block Substitute(loop_var -> normalized_access_index) specializes the expression for every prologue/body/epilogue clone, since the tirx mutators already rewrite PrimExprs inside call annotations. Pipeline-managed TMA copies keep their ring parity in the explicit waits the pass inserts, and the explicit-async T.tcgen05_gemm never auto-waits, so neither is annotated.

Drop the matching ring branch from LowerTileOp's fallback: it keyed on the "num_stages" annotation, which InjectSoftwarePipeline strips, so it was dead on pipelined loops and wrong on unpipelined ones -- a compiler-generated per-site barrier flips once per loop iteration regardless of pipeline depth. The fallback is now the zero-based epoch of the nearest enclosing serial loop.

Add transform-level tests asserting the annotated phase and the emitted wait parities on sm_90a, so the fix is covered without sm_100 hardware; the loops start at a nonzero min to pin the epoch normalization. Add a tcgen05 correctness test running the failing kernel at num_stages 2 and 3.

## Summary

- Fixed wait-parity handling for explicit mbarriers in pipelined loops.
- Annotated synchronous `T.gemm` operations and eligible explicit-barrier copies with logical parity `(k - min) % 2`.
- Excluded pipeline-managed TMA copies and explicit-async `T.tcgen05_gemm` operations.
- Updated `LowerTileOp` to derive fallback parity from the nearest serial loop epoch.
- Added transform tests for barrier annotations and emitted wait parity on `sm_90a`.
- Added `tcgen05` GEMM correctness tests for two- and three-stage pipelines.

## C++ style / lint notes

- The PR changes C++ sources.
- No changes to `docs/developer_guide/cpp_style.md` are included.
- The C++ API Style Audit remains warning-only when applicable.
- No correctness or build issues are identified. Advisory style warnings should not block merge unless they indicate a new API, FFI, or maintainability risk.

### DIFF
--- a/src/op/operator.h
+++ b/src/op/operator.h
@@ -117,8 +117,7 @@ struct LowerArgs {
   ffi::Map<tirx::Var, PrimExpr> bind_var_to_expr;
   // Fallback mbarrier parity for ops that do not carry an explicit
   // tl.pipeline_mbar_phase_expr annotation. LowerTileOp derives this from the
-  // nearest enclosing serial loop so non-pipelined TMA loops still alternate
-  // barrier phase correctly.
+  // zero-based epoch of the nearest enclosing serial loop.
   PrimExpr mbar_phase_expr = IntImm(DataType::Int(32), 0);
   // Pointer to the shared.barrier buffer for compiler-generated mbarriers.
   // Points to the LowerTileOpPass member so copy.cc sees the buffer
--- a/src/transform/inject_pipeline.cc
+++ b/src/transform/inject_pipeline.cc
@@ -334,11 +334,24 @@ private:
     return Call(call->dtype, call->op, call->args, annotations, call->span);
   }
 
+  // Ops that auto-emit a paired arrive+wait on an explicit (user-allocated)
+  // mbarrier: sync gemm with an mbar (isTcgen05_ marks the explicit-async
+  // T.tcgen05_gemm op, which never waits) and copies with an explicit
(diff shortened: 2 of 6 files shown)


LeiWang1999:
@regression-perf

LeiWang1999 (approved):

Issue #2948, opened 2026-08-12
[BUG][Fuzzer][wrong-code] `buf[(i+1)%N]` fragment read in `T.Parallel` silently returns the local element instead of the indexed value


Hughshine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and found no similar report.

### What version of TileLang are you using?

0.1.13 (latest release)

### System information

TileLang 0.1.13; CUDA 12.8; NVIDIA L40S (sm_89), single GPU.

### Problem description

A `local.fragment` READ whose index is a non-identity affine function of the `T.Parallel`
loop variable — for example a rotation `buf[(i+1)%N]` or a shifted read `buf[(i+OFF)%N]` —
is silently miscompiled: the index expression is discarded and each thread reads *its own*
local element instead. The kernel compiles with no error and runs to completion; the result
is deterministically wrong.

Concretely, for `B[i] = buf[(i+1) % N]` (a rotate-by-1) over an `int32` fragment of `N=128`
with `threads=128`, the emitted CUDA is:

```
buf[0] = A[((int)threadIdx.x)];
B[((int)threadIdx.x)] = buf[0];      // the (i+1)%N index is gone — no cross-thread read
```

so `B == A` (identity) instead of the rotation. All 128 elements are wrong. The same kernel
over a `shared` buffer emits the real indexed read and is correct (see control below).

<details><summary>Trigger boundary (offset and elements-per-thread)</summary>

The whole index expression is dropped, so the *kind* of wrongness depends only on the
loop→thread partition, not on the offset value:

| config | epr | got | expected | wrong |
|---|---|---|---|---|
| `threads=128`, `OFF=1` | 1 | `[0,1,2,3,4,5,6,7]` | `[1,2,3,4,5,6,7,8]` | 128/128 |
| `threads=128`, `OFF=3` | 1 | `[0,1,2,3,4,5,6,7]` | `[3,4,5,6,7,8,9,10]` | 128/128 |
| `threads=128`, `OFF=5` | 1 | `[0,1,2,3,4,5,6,7]` | `[5,6,7,8,9,10,11,12]` | 128/128 |
| `threads=64`, `OFF=3` | 2 | `[1,0,3,2,5,4,7,6]` | `[3,4,5,6,7,8,9,10]` | 128/128 |
[...]

PR #3090, opened 2026-08-26
[Layout] Reject non-round-tripping inferred inverses

KellyFrog:
This fixes issue #2948.

## Summary

- Reject inferred layout inverses that fail round-trip validation.
- Validate inferred bijective inverses symbolically and, when required, across static domains.
- Report unknown results for symbolic or oversized domains.
- Add layout inverse tests for valid split layouts and invalid cyclic shifts.
- Add a CUDA regression test for shifted modulo fragment indexing in `LowerTileOp`.

## C++ style / lint notes

- The change touches C++ implementation code in `src/layout/layout.cc`.
- No public or exported declarations changed.
- The change does not modify rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) remains advisory. Warning-only findings are not correctness or build blockers.

### DIFF
--- a/src/layout/layout.cc
+++ b/src/layout/layout.cc
@@ -356,7 +356,7 @@ Map<Var, Range> FragmentNode::GetVarMap() const {
 
 namespace {
 
-constexpr int64_t kMaxExactInjectivityDomain = 1 << 18;
+constexpr int64_t kMaxExactLayoutDomain = 1 << 18;
 
 struct IntTupleHash {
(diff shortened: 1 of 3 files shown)


SiriusNEO (approved):
LGTM, this PR adds a check to reject the wrong inverse (like `(i+1)%N`). But maybe we need to fix the TVM side in the future to address the root problem

Issue #2942, opened 2026-08-12
[BUG][Fuzzer][wrong-code] `T.fill` of a row-offset fragment sub-region silently fills the wrong rows instead of the sliced region


Hughshine:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and [Discussions](https://github.com/tile-ai/tilelang/discussions) that this hasn't already been reported. (+1 or comment there if it has.)

### What version of TileLang are you using?

0.1.13 (latest release)

### System information

TileLang 0.1.13; CUDA 12.8, PyTorch 2.8; NVIDIA L40S (sm_89).

### Problem description

`T.fill(f[off:off+h, :], v)` on a `T.alloc_fragment` sub-region whose row offset `off` is **not** a multiple of the fragment's per-thread row atom silently fills the *wrong rows* — the kernel compiles and runs with no diagnostic. For a `(32, 64)` fragment (128 threads, row atom 8), `T.fill(f[3:11, :], 7.0)` writes rows `[0,1,2,3,4,13,14,15]` instead of `[3..10]`. Offsets that are multiples of the atom (`off=0`, `off=8`) fill the intended rows correctly, so the region is computable and the compiler mishandles only the misaligned offset.

The same misalignment can also abort compilation with `InternalError: Loop layout is not injective: Fragment(... forward_thread: (_i + 5) % 8 * 16 ...)` for some `(off, h)` (e.g. `off=5, h=10`); the silent-wrong-rows face and the ICE face are two manifestations of the same layout mis-map. Not a regression — wrong since the feature was added (see Provenance).

### Reproducible example code

```python
import tilelang, tilelang.language as T, torch

M, N = 32, 64

def build(off, h):
    @tilelang.jit(out_idx=[-1])
    def prog():
        @T.prim_func
        def main(A: T.Tensor((M, N), "float32"), C: T.Tensor((M, N), "float32")):
            with T.Kernel(1, threads=128) as _:
                f = T.alloc_fragment((M, N), "float32")
                T.copy(A, f)
                T.fill(f[off:off + h, :], 7.0)     # fill rows off..off+h of the fragment
                T.copy(f, C)
        return main
    return prog

torch.manual_seed(0)
a = torch.randn(M, N, device="cuda", dtype=torch.float32)

def check(off, h):
    ref = a.clone(); ref[off:off + h, :] = 7.0
[...]

PR #3091, opened 2026-08-26
[Layout] Reject unanchored fragment slicing

KellyFrog:
See issue #2942.

The issue of fragment slicing involves some bugs and lack of consideration in infering layouts of `T.Parallel`.

Now fragment slicing is rejected in `LayoutInference` when no explicit layouts are avalible (e.g. layouts infered from `T.gemm`, `T.annotate_layout`, etc.).

Codes like:

```python
  # 128 threads
  f = T.alloc_fragment((32, 64), "float32")
  s = T.alloc_shared((8, 64), "float32")

  T.copy(A, f)
  T.copy(f[3:11, :], s)
  T.copy(s, D)
  T.copy(f, C)
```

will now be rejected because no explicit layouts are given and the compiler have to find a layout for both `f[3:11, :]` and `f`, compiler may fail or find sub-optimal layouts.

Fragment slicing is also observed in tilelang's example codes. In these senarios, they are usually input/output buffers and their layouts are fixed by requirements of `T.gemm`. These codes still work correctly as expected.

After this PR, `f[3:11, :]` will pass when `f`'s layout is fixed (e.g. by `T.gemm`) but may fail with errors involving "incorrect inverse" because `tvm` still struggles with fragment cyclic indexing: codes like `[(i + 3) % 8]` will be generated and is currently not supported.

## Summary

- Rejects unsupported unanchored fragment slicing during `LayoutInference`.
- Keeps slicing enabled when explicit layouts anchor inference, such as `T.gemm` and `T.annotate_layout`.
- Validates fragment shapes, thread ranges, injective mappings, and buffer access coverage.
- Supports replicated stores when fragment and global stores write identical values.
- Adds regression coverage for fragment slicing, parallel access validation, replicated writes, and reducer layouts.
- Updates attention-sink examples to use row-based sink fragments.

## C++ style / lint notes

- The PR changes C++ implementation and API files.
- The style guide and CI warning configuration require review of any reported `TLCPP003` or `TLCPP004` findings.
- Warning-only findings do not block merge unless they indicate a clear API, FFI, or maintainability risk.


examples/attention_sink/example_gqa_sink_bwd_bhsd.py
```diff
@@ -70,7 +70,7 @@ def flash_fwd(
             scores_scale = T.alloc_fragment([block_M], accum_dtype)
             scores_sum = T.alloc_fragment([block_M], accum_dtype)
             logsum = T.alloc_fragment([block_M], accum_dtype)
-            sinks = T.alloc_fragment([heads], dtype)
```
SiriusNEO, line 73: Why this PR introduces some changes in the examples? Like change the sinks fragment's shape

SiriusNEO (commented):
It seems that if we write like the following:
```python3
for i, j in T.Parallel(8, N):
    f[i + 3, j] = patch[i, j]
```
It will not be rejected by current checking? Should we reject this case as well as we do to fragment slicing?

KellyFrog:
Will fix this

PR #3093, opened 2026-08-26
[Refactor] Give reducers a first-class PartialFragment layout solved by layout inference

LeiWang1999:
## Summary

Reducer v2's physical plan (narrow vs wide partials) used to be decided **after** layout inference by `ReducerPlanAndMaterialize`, with dst-steering emulating the missing propagation through side channels (wake tables, `LayoutInferArgs.reducer_update_sites` hint plumbing, ownership gates). The same induced layout was computed twice and kept consistent by convention only, and the reducer split the layout use-def graph into disconnected components.

This PR makes the reducer's layout a first-class solver value that **fully determines the physical plan**:

- **New layout kind `tl.PartialFragment`** (`PartialFragmentNode : FragmentNode`): Fragment algebra whose replication coordinate carries an explicit **combine decomposition** — under a low-bits convention enforced at construction, `_rep % combine_size` enumerates *addend lanes* the finalize collective must reduce, `_rep / combine_size` enumerates *equal-value copy groups*. `(storage algebra, combine width)` is the complete plan: collective steps are derivable from the node alone (`CombineSteps`), the FullParticipant object (`combine == replicate == participants`, `thread == _rep`) is the wide fallback, and the combine width is part of layout identity — same-storage/different-decomposition proposals **widen at the commit point** instead of unifying silently. A boundary-preserving condensation keeps the convention true through projection; kind-aware `IsEqual` in both dispatch directions prevents silent unification with plain Fragments.
- **Update nests propose, the engine arbitrates**: `ParallelOp` records its `reducer_update` sites and proposes the induced partial (per-site `AnalyzeReducerUpdateSite` proofs) or the participant-wide plan; multi-site disagreement resolves by a monotone **widen-on-conflict** rule; zero-update epochs are seeded wide up front; `tl.reducer_force_baseline` pins the proposal side.
- **`finalize_reducer` reads the solved partial** and steers an unconstrained destination to its post-collective reading. The wake table, hint plumbing and `reducer_use_list_` component glue are deleted — reducers travel ordinary `use_list_` buffer edges.
- **The materializer reads the plan off the node**: storage = `AsPostCollective()`, steps = `CombineSteps()`, wide lowering iff the node is the FullParticipant object. `TryNarrowPlan` is retired; the remaining `DeriveNarrowLowering` derives only what lives outside the layout — the packed-lane heuristic and the destination-containment gate (which still demotes gracefully to the wide plan, preserving the unsatisfiable-steering fallback contract). The destination override chains are kept as a **verified-no-op fallback** with retirement telemetry.
- **`T.annotate_layout` accepts `PartialFragment` on reducers**, pinning plans the automatic solution cannot reach (e.g. a middle plan between narrow and wide: an annotated 8-lane plan reshapes the input fragment's load layout via back-propagation). Annotated partials are strict; a disagreeing update site raises with the narrow-rejection reason.
- **Layout cost models price register-array spills**: a fragment access whose physical slot index depends on the thread variable (shared `FragmentThreadIndexProbe`) demotes the whole array to local memory; it is charged as byte-denominated traffic in both models' `mem` channel so it *competes with* — rather than vetoes — global-bandwidth estimates. `ChooseBestCandidate`'s anti-spill preference now applies regardless of the loop's store target. With generation and pricing combined, **inference produces the wide publication chain's replicated layouts on its own** — the reducer_v2 suite and the mhc_post_bwd production kernels pass with the destination override fully disabled.

Structurally, the `36-vs-576` bug class dies at the root: the finalize collective width is read off the reducer's own layout, and the once-ambiguous case (fully replicated narrow storage vs true FullParticipant — e.g. `Parallel(8)` into a scalar over 128 threads, `rep 128 = 8 addends x 16 copy groups`) is now representable and structurally distinct.

## Testing

All with the kernel cache disabled (`TILELANG_DISABLE_CACHE=1`):

- `test_tilelang_language_reducer_v2.py`: **70 passed** — new coverage includes pinned middle plans (`AllReduce<SumOp, 8>` where auto solves 4), annotation steering the input fragment, conflict diagnostics carrying the narrow-rejection reason, combine-step roundtrip (stride-1 vs stride-16 lanes), and same-storage/different-combine widening at the commit point.
- `testing/python/language`: 1261 passed; layout-inference cost-model tests and `maint/layout_inference` goldens unchanged.
- Generated CUDA for existing reducer kernels is byte-identical to the previous behavior.
- `mhc_post_bwd` (legacy v1 reducer through `CanonicalizeLegacyReducer`): correctness passes; six benchmark sizes show no latency change and zero STL/LDL in SASS; also verified green with the destination override disabled (the retirement experiment).

Note: this branch stacks on the `reducer-dst-steering` commits (`1275ef6f..32e9dba1`), which appear in this PR until they land separately.

## Summary

- Introduces `tl.PartialFragment` as a first-class reducer layout.
- Refactors reducer layout inference to solve narrow and wide partial plans through update-site proposals and monotone widening.
- Includes combine decomposition in partial-layout identity.
- Uses the solved partial layout during reducer finalization and materialization.
- Adds reducer annotation validation and input-fragment back-propagation.
- Removes reducer-specific wake-table, hint-plumbing, and use-list component infrastructure.
- Adds register-array spill traffic to layout cost models.
- Adds reducer v2 coverage for annotations, conflicts, zero-update epochs, combine widths, and layout widening.
- Updates layout visualization and Python exports for `PartialFragment`.

## Validation

- 70 reducer v2 tests pass.
- 1,261 language tests pass.
- Existing layout-inference goldens remain unchanged.
- Existing reducer kernels produce byte-identical generated CUDA.
- Legacy reducer correctness and benchmark checks pass.
- No latency change or STL/LDL instructions were observed.

## C++ style / lint notes

- The PR changes C++ implementation and public API code.
- The changes touch TVM Object and FFI types covered by `docs/developer_guide/cpp_style.md`.
- No correctness, build, or test issue is identified in the provided results.
- TLCPP003/TLCPP004 findings are advisory unless they indicate a clear API, FFI, or maintainability risk.


LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

PR #3109, opened 2026-08-28
[CPU] Support OpenMP parallel of grid loops

penguin-wwy:
## Summary

- Added opt-in CPU OpenMP parallelization for grid loops.
- Added `MaterializeCPUParallelGrid` to convert annotated CPU grid loops into parallel loops.
- Added loop collapsing, thread-count configuration, minimum-trip gating, and allocation sinking.
- Added OpenMP compile-flag handling for Linux, macOS, and Windows.
- Added CPU pass configuration keys and Python helpers.
- Added tests for correctness, pragmas, allocation sinking, gating, unit dimensions, thread counts, and disabled behavior.

## C++ style / lint notes

- The PR changes C++ code and touches APIs and FFI registration.
- The relevant guidance is documented in `docs/developer_guide/cpp_style.md`.
- Review new advisory TLCPP003/TLCPP004 warnings. These warnings are not merge blockers unless they indicate a clear API, FFI, or maintainability risk.
- Correctness, build, and test failures remain separate from warning-only style findings.


src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,419 @@
...
+    Touch(op->buffer);
+    StmtExprVisitor::VisitExpr_(op);
+  }
+  void Touch(const Buffer &buffer) {
+    AccessInfo &entry = info_[buffer->data];
+    if (grid_depth_ == 0) {
+      entry.min_depth = 0;
+    } else {
+      // First inside access records its depth; later ones take the minimum.
+      entry.min_depth =
```
LeiWang1999, line 164: **[bug — confirmed locally] Alloc sinking breaks with two sequential `T.Kernel` launches in one PrimFunc**

`grid_depth_` accumulates across *all* annotated nests in the function, so `min_depth` can't tell which nest an access belongs to. When a function contains two sequential `T.Kernel` launches and the second kernel's scratch alloc sits above the first nest, `CollectSinkableAllocs` sees `min_depth >= sink_depth` and sinks it into the *first* nest, leaving the second nest's uses undeclared.

Repro (two `T.Kernel` blocks, each with a `local` scratch buffer, `tl.cpu_parallel=True`, target `c`) — generated C:

```c
#pragma omp parallel for collapse(2)
for (int32_t bx = 0; bx < 4; bx += 1) {
  for (int32_t by = 0; by < 4; by += 1) {
    float buf1[64];
    float buf2[64];   // <- second kernel's buffer, sunk into the first nest
    ...
  }
}
for (int32_t bx_1 = 0; bx_1 < 4; ++bx_1) {
  ...
  buf2[i_2] = ...;    // error: 'buf2' was not declared in this scope
```

g++ then fails with `'buf2' was not declared in this scope`.

Suggested fix: record the nest root (outermost annotated For) per access in `GridAccessAnalysis`, and only sink allocs whose uses all lie inside the nest currently being converted. Relatedly, `converted_` limiting conversion to the first nest looks more conservative than necessary — sequential sibling nests are safe to parallelize independently (the llvm backend only rejects *nested* parallel loops, not siblings).

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,419 @@
...
+        return SeqStmt(std::move(elements));
+      }
+      return body;
+    };
+
+    // Total trip count must be statically known and above the threshold.
+    int64_t trip = 1;
+    for (const ForNode *op : loops) {
+      const auto *extent = op->extent.as<IntImmNode>();
+      if (!extent || extent->value <= 0)
```
LeiWang1999: **[behavior gap — confirmed locally] Dynamic grid extents silently fall back to serial, with no warning**

A symbolic extent fails this IntImm check and takes `rebuild_serial()`, which also strips the grid annotations — so the residual-annotation warning at the end of the pass can never fire for this case. Verified: `T.Kernel(T.ceildiv(m, BM))` with symbolic `m` compiles fine under `tl.cpu_parallel=True` but emits no pragma and no warning.

On the `c` target `#pragma omp parallel for` handles runtime bounds fine; the static-extent requirement is only needed by the min-trip gate. Suggest allowing dynamic extents when `min_trip == 0` (at least on `c`), or failing that, logging a warning so users know why nothing got faster. Dynamic shapes are a fairly central tilelang use case.

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,419 @@
...
+      out.Set(kv.first, kv.second);
+    }
+  }
+  return out;
+}
+
+/*! \brief For each buffer (keyed by its data Var): the shallowest grid-nest
+ * depth of any access (0 = accessed outside the nest) and whether it is
+ * accessed inside the nest at all. */
+class GridAccessAnalysis : public StmtExprVisitor {
```
LeiWang1999, line 72: **[soundness gap] Opaque buffer accesses are invisible to the sinking analysis**

`GridAccessAnalysis` only visits `BufferLoad`/`BufferStore`. If a buffer's data var appears only in opaque positions (`T.call_extern` args, `address_of`, `tvm_access_ptr`), the analysis concludes it isn't used inside the nest: the alloc stays function-scope (shared across workers) and neither race warning in `CollectSinkableAllocs` can fire — a silent data race once the grid runs parallel. A mixed case (loads inside the nest, opaque use outside) can instead mis-sink and hit the undeclared-identifier failure.

Suggest treating any occurrence of a tracked data var outside BufferLoad/Store (e.g. in Call args) as a depth-0 access so the buffer is pinned shared with a warning — or refusing to parallelize that nest.

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,419 @@
...
+    // collapse(n)); exactly one — the first non-unit dim — on ``llvm``.
+    size_t parallel_idx = SinkIndex(loops, collapse_all_dims_);
+
+    if (!collapse_all_dims_) {
+      // TVM's llvm codegen requires min=0 / step=1 on the parallel loop.
+      // A missing step is the implicit default of 1 and is normalized to a
+      // literal during the rebuild below.
+      const ForNode *marked = loops[parallel_idx];
+      const auto *min = marked->min.as<IntImmNode>();
+      ICHECK(min && min->value == 0)
```
LeiWang1999: **[minor] Hard crash vs. graceful fallback** — this ICHECK (plus the step check and the nested-parallel check below) aborts compilation. Since the pass already has a clean `rebuild_serial()` escape hatch, falling back to serial with a warning would be friendlier for anything user IR can trigger.

LeiWang1999 (commented):
Reviewed the full diff and tested locally on Linux: the PR's 8 tests pass, the full `testing/python/cpu/` suite passes (96 tests), plus two edge-case probes.

Overall the design is clean — the opt-in contract is airtight (config off → no annotation, no extra pass, no extra flags, bit-identical output, with a test guarding it), the codegen's identity-based collapse-chain tracking handles nested kParallel through wrappers correctly (nice regression test), and the alloc-sinking warnings cover the common unsafe cases.

Two confirmed bugs worth fixing before merge, both verified with local repros (details inline):

1. **Two sequential `T.Kernel` launches in one PrimFunc** → the second kernel's scratch buffer gets sunk into the first parallel nest → C compile error (`'buf2' was not declared in this scope`).
2. **Dynamic (symbolic) grid extents silently fall back to serial** with no warning — `rebuild_serial()` strips the annotations, which also disarms the residual-annotation warning.

The remaining inline comments are smaller: an analysis blind spot for opaque buffer uses (potential silent race), no llvm-target test coverage, `cpu_num_threads` silently ignored on llvm, and the `-O2` flag bundling.

testing/python/cpu/test_tilelang_cpu_parallel.py
```diff
@@ -0,0 +1,189 @@
+"""OpenMP CPU parallelization (``tl.cpu_parallel``) tests.
```
LeiWang1999, line 1: **[test coverage] No `llvm`-target test** — all eight cases here use `target="c"` (or `build_c`). The entire `collapse_all_dims_ = false` branch — single-dim marking, the min/step ICHECKs, `ParallelLoopFinder`, the `TVMBackendParallelLaunch` lowering — is untested, though the PR advertises llvm support. One correctness test on `llvm` would cover it.

tilelang/language/kernel.py
```diff
@@ -298,6 +299,12 @@ def Kernel(
     prelude : str
         The import c code of the kernel,
         will be injected before the generated kernel code.
+    cpu_num_threads : int
+        OpenMP thread count for CPU targets: emitted as the
+        ``num_threads(n)`` clause on the grid parallel region when the
+        ``tl.cpu_parallel`` pass config is enabled. ``None`` (default)
+        omits the clause and lets the OpenMP runtime pick the thread
+        count (e.g. from ``OMP_NUM_THREADS``). Ignored on non-CPU targets.
```
LeiWang1999: **[docs] `cpu_num_threads` is silently ignored on the `llvm` target** — only the C codegen consumes the annotation; llvm's `TVMBackendParallelLaunch` uses its own thread pool. "Ignored on non-CPU targets" suggests it works on both CPU targets. Worth documenting here (or warning at compile time when the target is llvm and `cpu_num_threads` is set).

tilelang/contrib/openmp.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def get_openmp_compile_flags() -> list[str]:
+    """Compile flags for OpenMP CPU kernels: ``-O2`` plus OpenMP flags.
+
+    ``-O2`` is unconditional: the default CPU library build has no
+    optimization level at all, which alone costs ~8.6x on dense kernels and
+    would swamp the parallel speedup.
+    """
+    flags = ["-O2"]
```
LeiWang1999, line 85: **[design note] `-O2` is bundled with the parallel switch** — toggling `tl.cpu_parallel` also toggles the optimization level, so a serial-vs-parallel comparison conflates the parallel speedup with the ~8.6x `-O2` effect the docstring mentions. Longer term `-O2` probably belongs to the CPU build itself, independent of parallelization.

Also, win32 behavior differs between the two injection paths: `libgen.cpu_openmp_flags` still injects `-O2` on Windows, while `tvm_ffi._cpu_openmp_flags` early-outs to `[]`.

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,557 @@
...
+                                    : "it is used across grid levels";
+        LOG(WARNING) << "tl.cpu_parallel: buffer `" << data->name_hint
+                     << "` is mutated inside the grid nest but cannot be "
+                        "privatized ("
+                     << reason << "); the nest stays serial";
+        return rebuild_serial();
+      }
+    }
+
+    Stmt body = loops[parallel_idx]->body;
```
LeiWang1999: **[bug — blocking] llvm path: the rebuild duplicates every grid loop below `parallel_idx`**

`body` starts from `loops[parallel_idx]->body`, but the loop below rebuilds **all** of `loops[size-1..0]` around it. On the `c` target `parallel_idx == size-1`, so this is correct. On llvm, `parallel_idx` is the *first* non-unit dim — for the 4×4 grid in `test_llvm_cpu_parallel_gemm_correctness`, `parallel_idx = 0` and `loops[0]->body` already *contains* the original `by` loop. The rebuild wraps a fresh serial `by` around it:

```
parallel bx { serial by (rebuilt) { allocs; serial by (original, still annotated) { body } } }
```

Consequences:
- The same loop `Var` is bound by two nested `For`s. `CodeGenLLVM::CreateSerialFor` has `TVM_FFI_ICHECK(!var_map_.count(loop_var.get()))`, so compilation crashes outright on any multi-dim llvm grid whose first non-unit dim isn't the last. This wasn't caught because both llvm tests were SKIPPED on every CI runner (no `USE_LLVM` build in CI).
- Even absent the ICHECK, the body would execute `extent(inner dims)`× redundantly — silently wrong for non-idempotent kernels (e.g. `T.atomic_add` accumulation).
- The original inner loops keep `kCPUGridDim`, so the residual-annotation warning fires spuriously on every multi-dim llvm compile.

Suggested fix — always rebuild from the true innermost body and splice the sunk allocs in when the wrap reaches the parallel dim (behavior-identical on `c`, where `parallel_idx == size-1`):

```cpp
Stmt body = loops.back()->body;
for (int i = static_cast<int>(loops.size()) - 1; i >= 0; --i) {
  if (static_cast<size_t>(i) == parallel_idx && !sunk.empty()) {
    Array<Stmt> elements(sunk.begin(), sunk.end());
    elements.push_back(body);
    body = SeqStmt(std::move(elements));
  }
  // ... wrap loops[i] as today ...
}
```

On llvm the allocs then land inside the parallel loop but outside the inner serial dims — per-worker privatization preserved, and there's no collapse clause to keep perfectly nested there.

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,557 @@
...
+private:
+  void VisitStmt_(const ForNode *op) override {
+    if (op->kind == ForKind::kParallel)
+      found = true;
+    StmtVisitor::VisitStmt_(op);
+  }
+};
+
+//! True if the subtree still carries a grid annotation (unconverted nest).
+class GridAnnotationFinder : public StmtVisitor {
```
LeiWang1999: **[design — blocking] Replace the rewrite-then-rescan residual detection**

This post-hoc scan uses the IR annotation as a side channel: `Rewrite`'s fallthrough comment says it *deliberately* keeps the annotation "so the residual-annotation warning fires". That costs us on every axis:

- The warning can't say **which** loop or **why** — it has to conflate two unrelated causes ("nested inside another parallelized nest, or wrapped in an unrecognized construct") and fires once per function no matter how many nests were left behind.
- Unconverted nests leak `tl.cpu_grid_dim` into the **final IR** — inconsistent with `rebuild_serial()`, which strips it. Harmless today, a trap the day anything downstream starts reading that key.
- The finder only exists because `GridRewriter`'s hand-rolled traversal is incomplete: it doesn't recurse into `LetStmt` or non-annotated `For` bodies, so a nest hidden there is invisible to the rewriter and needs a second full-tree pass to detect.

The hand-rolled `SeqStmt`/`AttrStmt`/`IfThenElse` copy-on-write recursion in `GridRewriter` is exactly what `StmtMutator` provides for free. Please restructure: make `GridRewriter` a `StmtMutator` (default recursion reaches every annotated `For` wherever it hides; `in_parallel` becomes member state; the sibling-alloc collection moves into an overridden `VisitStmt_(const SeqStmtNode*)`), and at each nest that can't convert, warn **in place** with `loop_var->name_hint` and the actual reason, then strip the annotation. That establishes a clean invariant — no `tl.cpu_grid_dim` survives this pass — and `GridAnnotationFinder` can be deleted (or reduced to a debug-only `PostOrderVisit` ICHECK). `ParallelLoopFinder` above can become a `PostOrderVisit` lambda for the same reason.

I'd like this reworked before merge — the side-channel pattern is the kind of thing that gets copied once it's in-tree.

src/cpu/transform/materialize_cpu_parallel_grid.cc
```diff
@@ -0,0 +1,557 @@
...
+      info_[op->buffer->data].store_inside = true;
+    }
+    Touch(op->buffer);
+    StmtExprVisitor::VisitStmt_(op);
+  }
+  void VisitExpr_(const BufferLoadNode *op) override {
+    Touch(op->buffer);
+    StmtExprVisitor::VisitExpr_(op);
+  }
+  void VisitExpr_(const VarNode *op) override {
```
LeiWang1999, line 131: **[soundness — should fix before merge] `T.address_of` is still invisible to the opaque-use detection**

The refusal warning below claims coverage of "call_extern / address_of / access_ptr", but `address_of` wraps a **BufferLoad** (`tirx/op.py::address_of`), and `ExprVisitor::VisitExpr_(BufferLoadNode)` visits only the indices — the data var never surfaces as a bare `VarNode`, so this hook never sees it. The analysis records `T.address_of(buf[0])` as a plain *load*.

Concrete race that slips through: `buf` used inside the nest only via `T.call_extern("writer", T.address_of(buf[0]))` (callee mutates it) and also read after the nest. `outside = true` blocks sinking, but `store_inside`/`opaque_inside` are both false → no refusal → the nest parallelizes with the mutated buffer shared across workers.

Fix is small: override `VisitExpr_(const CallNode*)` here, and when `op->op.same_as(builtin::address_of())`, treat the argument BufferLoad's buffer as opaque (same handling as the bare-var path). Alternatively, if you'd rather defer, the warning text must stop claiming `address_of` coverage.

src/transform/materialize_kernel_launch.cc
```diff
@@ -87,16 +91,29 @@ class KernelLaunchMaterializer : public StmtMutator {
...
+    bool is_block = IsBlockBinding(op);
+    PrimExpr extent =
+        is_block ? op->extent : PrimExpr(IntImm(op->extent.dtype(), 1));
+    Map<ffi::String, ffi::Any> annotations = op->annotations;
+    if (annotate_grid_ && is_block && !annotations.count(attr::kCPUGridDim)) {
+      // Annotate every grid dim, including unit-extent ones: the consumer
+      // treats the annotated nest as one contiguous chain, and skipping a
+      // middle dim would cut the deeper dims off from parallelization (the
+      // llvm side skips unit dims itself when picking the loop to mark).
+      annotations.Set(attr::kCPUGridDim, IntImm(DataType::Int(32), block_dim_));
```
LeiWang1999, line 104: **[minor]** `block_dim_` keeps counting across sequential kernels in the same PrimFunc, so the second launch's grid dims get annotated with 2, 3 — contradicting `attr.h`'s "valued by its grid dimension index". Harmless today (the consumer only checks `count()`), but either reset the counter per launch or fix the `attr.h` comment so the two don't drift.

testing/python/llvm/test_tilelang_llvm_parallel.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+
+@tilelang.testing.requires_llvm
+def test_llvm_cpu_parallel_gemm_correctness():
+    torch.manual_seed(0)
+    kernel = _compile({PassConfigKey.TL_CPU_PARALLEL: True})
+    A = torch.randn(M, K, dtype=torch.float32)
+    B = torch.randn(K, N, dtype=torch.float32)
+    torch.testing.assert_close(kernel(A, B), A @ B, rtol=1e-3, atol=1e-3)
```
LeiWang1999, line 71: **[test coverage — blocking] These tests have never executed — and this one should crash once it does**

Both tests were `SKIPPED` on every runner of the PR's CI run (`requires_llvm`; the CI builds have `USE_LLVM OFF`), so the llvm path this PR advertises has zero executed coverage anywhere. With the loop-duplication bug in `ConvertGridNest` (see comment there), this GEMM test should fail with an ICHECK in `CodeGenLLVM::CreateSerialFor` on any LLVM-enabled build.

Two asks before merge:
1. Run these tests on an LLVM-enabled build at least once and post the result (CI can't cover them today).
2. Assert that parallel lowering actually happened — as written, the test also passes if `TL_CPU_PARALLEL` silently no-ops and the kernel runs serially. Checking the lowered IR for a `kParallel` loop (or the module for `TVMBackendParallelLaunch`) before the numerical assert closes that hole.

If an LLVM-enabled verification isn't practical right now, I'd rather descope llvm from this PR and land the well-tested `c`-target support alone.

tilelang/language/kernel.py
LeiWang1999: **[docs]** "Ignored on non-CPU targets" undersells the caveat: `cpu_num_threads` is also silently ignored on the **llvm** CPU target — only the C codegen consumes the annotation; `TVMBackendParallelLaunch` uses TVM's own thread pool. One sentence here ("no effect on the llvm target, which uses TVM's thread pool") prevents the natural misreading that it works on both CPU targets.

LeiWang1999 (changes requested):
Re-reviewed the head after the fix round — thanks for the quick turnaround. The fixes for the previous round are solid: the per-nest attribution genuinely resolves the two-sequential-kernels mis-sinking, dynamic extents now parallelize (the right call), the opaque-use analysis + refusal warnings cover the cases I flagged, and the ICHECKs are now graceful serial fallbacks. All 14 `c`-target tests passed in CI.

However, this round found three blocking issues, two of them on the llvm path — which, it turns out, has **zero executed coverage**: both new llvm tests were `SKIPPED` on every CI runner (`requires_llvm`; the CI builds have `USE_LLVM OFF`). Details inline:

1. **[blocking]** The llvm rebuild duplicates the inner grid loops (`materialize_cpu_parallel_grid.cc:475`). Traced through the PR's own 4×4 llvm GEMM test, the result binds the same loop `Var` twice, which `CodeGenLLVM::CreateSerialFor`'s `TVM_FFI_ICHECK(!var_map_.count(loop_var.get()))` hard-rejects — the test should crash on any LLVM-enabled build. Masked only by the CI skip.
2. **[blocking]** The rewrite-then-rescan residual detection (`GridAnnotationFinder` + "annotation kept so the warning fires") needs to be reworked — it loses all diagnostic context, leaks `tl.cpu_grid_dim` into the final IR inconsistently, and exists only to compensate for `GridRewriter`'s incomplete hand-rolled traversal.
3. **[blocking]** The llvm tests must actually run (on an LLVM-enabled build) and assert that parallel lowering occurred before this PR can claim llvm support — or descope llvm from this PR and keep it `c`-target only, which I'd also be fine with.

Smaller items inline: the `address_of` blind spot in the opaque-use analysis (the warning text claims coverage it doesn't have), the `block_dim_` counter across sequential kernels, and the `cpu_num_threads` llvm doc gap. From the previous round, the `-O2` bundling / win32 asymmetry note and coderabbit's trip-count saturation point are still open — non-blocking, but please address or reply.

testing/python/llvm/test_tilelang_llvm_parallel.py
penguin-wwy: LLVM Test success
```
testing/python/llvm/test_tilelang_llvm_parallel.py::test_llvm_cpu_parallel_gemm_correctness PASSED
testing/python/llvm/test_tilelang_llvm_parallel.py::test_llvm_cpu_parallel_disabled_by_default PASSED
```

PR #3114, opened 2026-08-30
[BugFix][CUDA] Support non-constant int4/uint4 broadcast

### LINKED ISSUE #2997: [BUG][Fuzzer][ice-on-valid-code] `T.fill`/`T.clear` of an `int4` buffer crashes the compiler with `broadcast_var is not a constant` instead of emitting the fill
### Required prerequisites

0.1.13 (latest release)

### System information

- TileLang 0.1.13, Python 3.13, Linux.
- GPU: NVIDIA L40S (compute capability 8.9, sm_89). The defect is in the CUDA `Broadcast` codegen (`src/cuda/codegen/codegen_cuda.cc`), which runs at code-generation time before any arch-specific lowering, so it is architecture-independent.

### Problem description

`T.fill` (and `T.clear`, which lowers to `T.fill(..., 0)`) on an `int4` buffer aborts at compile time with

```
InternalError: Check failed: (p) is false: BroadcastNode ... value: broadcast_var is not a constant
```

`int4` is a supported element dtype (it is the operand dtype of the shipped `examples/gemm_int4` kernel) and `T.fill`/`T.clear` place no dtype restriction on the buffer, so the kernel is legal. The fill lowering broadcasts the fill value across the vectorized store, but binds that value to a temporary (`broadcast_var`) first, so the `Broadcast` node's value is a `Var`, not a constant. The `int4` branch of the CUDA `Broadcast` visitor unconditionally requires a constant value (`ICHECK(p)` on `as_const_int(op->value)`), so it fatal-aborts. The identical kernel on the sibling `int8` dtype compiles, because the `int8` branch only special-cases the constant sub-case and otherwise falls through to the generic (non-constant) path.

The crash does not depend on `T.copy` or on any GEMM machinery — a bare `T.fill`/`T.clear` of an `int4` shared buffer reproduces it. Not a regression — see Provenance.

### Reproducible example code

```python

M, N = 16, 64

def make(dtype):
    @T.prim_func
    def func(B: T.Tensor((M, N), dtype)):
            sh = T.alloc_shared((M, N), dtype)
            T.fill(sh, 1)          # T.clear(sh) fails identically
            T.copy(sh, B)
    return func

def try_compile(dtype):
    try:
        tilelang.compile(make(dtype)).get_kernel_source()
        return "compiled"
    except Exception as e:
        return "CRASH: " + str(e).splitlines()[-1]

print("int4:", try_compile("int4"))   # -> CRASH ("broadcast_var is not a constant")
print("int8:", try_compile("int8"))   # -> compiled (control)
```

Output:

```
int4: CRASH: Check failed: (p) is false: BroadcastNode 0x... value: broadcast_var is not a constant
int8: compiled
```

### Traceback

```
tvm.error.InternalError: Check failed: (p) is false: BroadcastNode 0x... value: broadcast_var is not a constant
  ...
  CodeGenTileLangCUDA::VisitStmt_(BufferStoreNode const*)
  CodeGenC::PrintExpr(PrimExpr const&)
  CodeGenTileLangCUDA::VisitExpr_(BroadcastNode const*, std::ostream&)
  tvm::runtime::detail::LogFatalImpl(...)
```

### Expected behavior

Filling an `int4` buffer with a runtime (non-constant) broadcast value should emit code rather than aborting, matching what the sibling `int8` path already does: the `int8` branch handles the constant case as an optimization and otherwise falls through to the generic `Broadcast` lowering, so a variable fill value is fine there. The `int4` branch could take the same shape — keep the constant fast-path, and fall through for a non-constant value instead of asserting.

### Additional context

**Root cause.** The CUDA `Broadcast` codegen does not handle a non-constant value in its `int4` branch. In [`CodeGenTileLangCUDA::VisitExpr_(BroadcastNode)`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L5614-L5618) the `int4` (`bits() == 4`) branch reads `p = as_const_int(op->value)` and immediately `ICHECK(p) << "... is not a constant"`, so any broadcast of a `Var` fatal-aborts.

The `broadcast_var` in the message is *not* something `T.fill` wrote — it is created by the device-lowering pass [`HoistBroadcastValues`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/transform/hoist_broadcast_values.py), which runs unconditionally in `engine/lower.py` (`device_mod = tilelang.transform.HoistBroadcastValues()(device_mod)`). That pass deliberately takes a `Broadcast` of an *immediate constant* (`IntImm`/`FloatImm`) and rebinds the constant to a let-bound `Var` named `broadcast_var`, leaving a `Broadcast(broadcast_var, lanes)` behind. So a fill value that was a constant in the frontend is a `Var` by the time codegen runs, and `as_const_int` returns null. Verified this session: the same crash fires on a *genuinely* runtime fill value (a kernel scalar argument, no hoist pass involved) with the message `value: T.Cast("int4", s) is not a constant` — the `int4` branch cannot emit any non-constant broadcast, hoisted or not.

<details><summary>Why the <code>int8</code> control compiles</summary>

The [`int8` branch of the same visitor](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L5503-L5525) guards its constant packing with `if (p) { ... }` and returns only inside it; when `p` is null it does not `return`, so control reaches the generic `Broadcast` path lower in the function. The `int4` branch instead asserts on `p` before doing anything, so it has no non-constant fall-through.

</details>

**Suggested fix.** Give the `int4` branch the same structure as the `int8` one: gate the constant packing on `if (p)` and, when the value is not a constant, fall through to the generic `Broadcast` lowering rather than `ICHECK`-aborting.

**Provenance.** The `int4` `Broadcast` branch (with the unconditional `ICHECK(p)`) was introduced in [#2073](https://github.com/tile-ai/tilelang/pull/2073) ("[CUDA] Improve int4 GEMM lowering and packed codegen support", 2026-04-22); a later namespace refactor ([#2297](https://github.com/tile-ai/tilelang/pull/2297)) relocated but did not change it. When it was written, a constant fill value still reached codegen as an `IntImm`, so the `ICHECK(p)` held for the fill/clear path. The `HoistBroadcastValues` pass that rebinds those constants to a `Var` was added three weeks later in [#2166](https://github.com/tile-ai/tilelang/pull/2166) (first commit 2026-05-12) and wired into the default device-lowering pipeline — this is what turns the fill value non-constant before it reaches the branch. So the fill/clear crash is best described as a **regression by pass interaction**: neither PR is wrong alone, but the hoist pass now feeds the `int4` branch exactly the non-constant input its `ICHECK` forbids. (The branch was independently already broken for a *truly* runtime broadcast value, which no shipped kernel produces.)

**Dedup.** I searched the open and closed tracker and found no existing report of this defect.

**Generalization (root + 4-axis sweep, all cells RUN this session on sm_89 / v0.1.13).** Two-level root:
- **Source-level root:** `CodeGenTileLangCUDA::VisitExpr_(BroadcastNode)`, the `(is_int()||is_uint()) && bits()==4` branch (codegen_cuda.cc:5614–5618). Fragile idiom = *assert-constant before doing anything*, with no non-constant fall-through — versus every other width's *gate on `if(p)` then fall through*. It is the only `as_const_int(op->value)`-gated branch in the whole visitor that `ICHECK`s instead of guarding (the int8 site at 5504 is the safe twin).
- **Operator-level root:** any store into an `int4` buffer whose stored value lowers to a *vectorized `Broadcast`*. `T.fill`/`T.clear` are the frontend triggers, but so is a raw vectorized assignment of a scalar. Two independent ways to make the value non-constant both hit it: the `HoistBroadcastValues` pass (constant → `Var`) and a genuinely runtime scalar.

|---|---|---|---|
| related-type | `uint4` fill | CRASH `broadcast_var is not a constant` | **same** (shares the 4-bit branch; not signedness-gated) |
| related-type | `int1` fill | CRASH `Cannot convert type int1x4 to CUDA type!` | **distinct** — falls through to the generic `make_` path, dies in `PrintType` (int1 codegen gap at :2321), not the `ICHECK` |
| related-type | `float4_e2m1fn` (fp4) fill | compiled | distinct path — fp4 is excluded by `!is_float4()`, routed through the fp8/fp4 branch (:5776); confirms the predicate boundary |
| related-type | `int8` / `int16` / `int32` fill | compiled | control — 8-bit gates on `if(p)`, wider ints take generic path |
| related-type | `float16` / `bfloat16` / `float8_e4m3` / `float8_e5m2` / `float32` fill | compiled | control — those branches `PrintExpr` the (Var) value |
| related-operator | `T.clear(int4)` | CRASH (same msg) | **same** — `clear` = `fill(...,0)` |
| related-operator | `T.fill(int4, 0)` / `(int4, 7)` | CRASH (same msg) | **same** — value magnitude irrelevant |
| related-operator | raw vectorized store of a runtime scalar into int4 | CRASH `value: T.Cast("int4", s) is not a constant` | **same** — direct non-constant, no hoist pass; proves the branch is broken for real runtime values too |
| similar-logic | int8 branch fed a genuinely runtime scalar (the safe twin) | compiled | control — same idiom done with `if(p)` fall-through |
| related-source | other `as_const_int(op->value)` sites in the visitor | only two exist (int8 :5504 safe, int4 :5616 buggy) | single buggy instance — no other sibling to fix |

**Boundary is narrow and stays specific.** The defect is exactly the one 4-bit integer branch; `int4` and `uint4` are the *same* branch (one fix covers both), so this is not a per-dtype family but a single missing fall-through. Adjacent widths, fp4, and the int8 twin are all clean. Title/Problem kept specific to the 4-bit `Broadcast` branch.

**Distinct-adjacent bug found (not filed).** `T.fill` of an `int1` buffer also aborts, but with a *different* root: `Cannot convert type int1x4 to CUDA type!` in `PrintType` (the int1 codegen at codegen_cuda.cc:2321 area), reached via the generic `make_` fall-through — not the `ICHECK(p)`. Different mechanism, different fix; recorded here, not filed.

**Impact.** The trigger is narrow: it needs a broadcast store into an `int4` buffer, and no shipped kernel does this (the one int4-buffer example, `gemm_int4`, only clears its `int32` accumulator — verified by running it). Any `T.fill`/`T.clear` of an `int4` buffer fires it, because `HoistBroadcastValues` makes even a constant fill value non-constant before codegen; a genuinely runtime broadcast store fires it too. When it fires it is a compile-time `ICHECK` abort — loud and caught immediately when building that kernel; nothing is silently miscompiled and no wrong value can reach a run. Fixing it closes a dtype-parity gap so a legal `T.fill`/`T.clear` on `int4` compiles the way it already does on `int8`, rather than refusing to build.

**Reach.** (Example run this session, v0.1.13 / sm_89.) The only shipped example that allocates real `T.int4` buffers is `examples/gemm_int4/example_tilelang_gemm_int4.py` — `A_shared`/`B_shared` are `T.alloc_shared(..., T.int4)`. I ran that example verbatim (its default `check_int4_gemm_correctness()` config): it **compiles and prints `Correctness check passed.`** It dodges this bug because it only `T.clear`s the `int32` accumulator `C_local` (line 29) and loads the `int4` operands via `T.copy`, never filling the `int4` buffers themselves — so no shipped kernel fills an `int4` buffer, which is why CI is green. Adding a single `T.clear(A_shared)` to that example-shaped kernel reproduces the crash (ran this session → `broadcast_var is not a constant`). Correction to an earlier claim in this draft: the `examples/dequantize_gemm/` kernels do **not** use `int4` buffers — they carry the 4-bit weights in a packed `storage_dtype` (int8/int32) and dequantize into an `in_dtype` (fp16/bf16) buffer, so none of them allocate or fill a `T.int4` buffer or exercise this branch. `gemm_int4` is the sole shipped int4-buffer site.

jjppp:
The `int4` branch of `CodeGenTileLangCUDA::VisitExpr_(BroadcastNode)` unconditionally `ICHECK`ed that the broadcast value was a compile-time constant, yet `HoistBroadcastValues` rebinds a constant fill value to a `Var` before codegen, so the value can be non-constant by the time the the assert fires, even for plain `T.fill(sh, 1)`.

This patch strengthens the `int4`/`uint4` branch to handle both constant and non-constant 4-bit packing.
For constants, we fold the packed 4-bit values into a literal as before; for non-constant, we emit the packing process as a runtime expression.

Note that a trivial fallback to the generic `make_<Type>(v, ...)` path does not work: `PrintType` maps e.g. `int4x4` to `int16_t`, emitting an undefined `make_int16_t(...)` and fail nvcc compilation.
This is why we handle 4-bit packing in the dedicated branch.

Fixes #2997.

## Summary

- Updated CUDA code generation for `int4` and `uint4` broadcasts.
- Preserved packed-literal generation for constant values.
- Added runtime masking and shifting for non-constant values.
- Prevented `ICHECK` failures after `HoistBroadcastValues`.
- Added vectorization tests for widths 4, 8, 16, and 32.
- Added `T.fill` tests for `int4` and `uint4` buffers.

## C++ style / lint notes

- The change follows the TVM-style invariant-check guidance in `docs/developer_guide/cpp_style.md`.
- It does not change a public API, reflected field, namespace rule, or header interface.
- The C++ API Style Audit is warning-only.
- No correctness or build issue is indicated by the style guidance.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -5791,49 +5791,52 @@ void CodeGenTileLangCUDA::VisitExpr_(const BroadcastNode *op,
     return;
   }
 
+  // Note that packed 4-bit broadcast cannot trivially fallback to the
+  // generic make_<Type>(v, ...) path, as it can emit malformed make_int16_t()
+  // calls that cause nvcc compilation errors when lanes==4.
   if ((op->dtype.is_int() || op->dtype.is_uint()) && op->dtype.bits() == 4) {
-    bool fail = false;
     const int64_t *p = as_const_int(op->value);
-    ICHECK(p) << "BroadcastNode " << op << " value: " << op->value
-              << " is not a constant";
-    int64_t v = *p & 0xF;
 
-    if (lanes == 4) {
-      v = (v << 12) | (v << 8) | (v << 4) | v;
-      if (op->dtype.is_uint()) {
-        os << "(uint16_t)" << v;
-      } else {
-        os << "(int16_t)" << v;
-      }
-    } else {
-      v = (v << 28) | (v << 24) | (v << 20) | (v << 16) | (v << 12) | (v << 8) |
-          (v << 4) | v;
-      if (lanes == 8) {
-        if (op->dtype.is_uint()) {
-          os << "(uint)" << v;
-        } else {
-          os << "(int)" << v;
+    auto emit_packed_field = [&](int nibbles_per_field) {
+      if (p) {
+        int64_t v = *p & 0xF;
+        int64_t packed = 0;
+        for (int i = 0; i < nibbles_per_field; ++i) {
+          packed |= v << (i * 4);
         }
-      } else if (lanes == 16 || lanes == 32) {
-        os << "make_";
-        PrintType(op->dtype, os);
+        os << packed;
+      } else {
+        std::string v = PrintExpr(op->value);
         os << '(';
-        for (int i = 0; i < lanes / 8; ++i) {
+        for (int i = 0; i < nibbles_per_field; ++i) {
           if (i != 0)
-            os << ", ";
-          if (op->dtype.is_uint()) {
-            os << "(uint)" << v;
-          } else {
-            os << "(int)" << v;
-          }
+            os << " | ";
+          os << "((static_cast<unsigned int>(" << v << ") & 0x0fu) << "
+             << (i * 4) << ")";
         }
         os << ')';
-      } else {
-        fail = true;
       }
-    }
+    };
 
-    if (!fail) {
+    if (lanes == 4) {
+      os << (op->dtype.is_uint() ? "(uint16_t)" : "(int16_t)");
+      emit_packed_field(4);
+      return;
+    } else if (lanes == 8) {
+      os << (op->dtype.is_uint() ? "(uint)" : "(int)");
+      emit_packed_field(8);
+      return;
+    } else if (lanes == 16 || lanes == 32) {
+      os << "make_";
+      PrintType(op->dtype, os);
+      os << '(';
+      for (int i = 0; i < lanes / 8; ++i) {
+        if (i != 0)
+          os << ", ";
+        os << (op->dtype.is_uint() ? "(uint)" : "(int)");
+        emit_packed_field(8);
+      }
+      os << ')';
       return;
     }
   }
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -5797,6 +5797,12 @@ void CodeGenTileLangCUDA::VisitExpr_(const BroadcastNode *op,
   if ((op->dtype.is_int() || op->dtype.is_uint()) && op->dtype.bits() == 4) {
     const int64_t *p = as_const_int(op->value);
 
+    // Materialize for reuse to avoid side-effects.
+    std::string sval;
+    if (!p) {
+      sval = SSAGetID(PrintExpr(op->value), op->value.dtype());
+    }
+
     auto emit_packed_field = [&](int nibbles_per_field) {
       if (p) {
         int64_t v = *p & 0xF;
@@ -5806,12 +5812,11 @@ void CodeGenTileLangCUDA::VisitExpr_(const BroadcastNode *op,
         }
         os << packed;
       } else {
-        std::string v = PrintExpr(op->value);
         os << '(';
         for (int i = 0; i < nibbles_per_field; ++i) {
           if (i != 0)
             os << " | ";
-          os << "((static_cast<unsigned int>(" << v << ") & 0x0fu) << "
+          os << "((static_cast<unsigned int>(" << sval << ") & 0x0fu) << "
              << (i * 4) << ")";
         }
         os << ')';
--- a/testing/python/language/test_tilelang_language_vectorize.py
+++ b/testing/python/language/test_tilelang_language_vectorize.py
@@ -216,6 +216,64 @@ def test_vectorize_broadcast_int8(vec_num):
     vectorize_broadcast_int8.compile(vec_num=vec_num)
 
 
+def vectorize_broadcast_int4(dtype, vec_num):
+    @tilelang.jit(pass_configs={tilelang.PassConfigKey.TL_DISABLE_BUFFER_INIT_CHECK: True})
+    def kernel():
+        with T.Kernel(1, threads=128):
+            a = T.alloc_local((64,), dtype)
+            b = T.alloc_var(dtype)
+
+            for i in T.vectorized(vec_num):
+                a[i] = b
+
+    return kernel
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("dtype", ["int4", "uint4"])
+@pytest.mark.parametrize("vec_num", [4, 8, 16, 32])
+def test_vectorize_broadcast_int4(dtype, vec_num):
+    """Broadcasting a non-constant 4-bit value to a vectorized store must emit
+    the packed-nibble carrier rather than aborting (see issue #2997). The int4
+    branch used to ICHECK on a non-constant value; the int8 branch already
+    handled it. Every packed lane count (16/32-bit carriers and make_(u)int{N})
+    is exercised here."""
+    vectorize_broadcast_int4(dtype, vec_num).compile()
+
+
+def fill_int4_buffer(dtype, fill_value):
+    M, N = 16, 64
+
+    @T.prim_func
+    def func(B: T.Tensor((M, N), dtype)):
+        with T.Kernel(1, threads=128):
+            sh = T.alloc_shared((M, N), dtype)
+            T.fill(sh, fill_value)
+            T.copy(sh, B)
+
+    return func
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("dtype", ["int4", "uint4"])
+@pytest.mark.parametrize("fill_value", [0, 1, 7])
+def test_fill_int4_buffer(dtype, fill_value):
+    """End-to-end reproduction of issue #2997: T.fill/T.clear of an int4 buffer.
+    HoistBroadcastValues rebinds the constant fill value to a Var, so the CUDA
+    Broadcast codegen sees a non-constant value. Verify it both compiles and
+    produces the correct packed nibbles."""
+    M, N = 16, 64
+    kernel = tilelang.compile(fill_int4_buffer(dtype, fill_value))
+    # int4 buffer packs two nibbles per byte, so N columns -> N // 2 bytes.
+    out = torch.empty(M, N // 2, dtype=torch.int8, device="cuda")
+    kernel(out)
+    expected = fill_value & 0x0F
+    low = out & 0x0F
+    high = (out >> 4) & 0x0F
+    assert bool((low == expected).all()), f"low nibble mismatch for {dtype}={fill_value}"
+    assert bool((high == expected).all()), f"high nibble mismatch for {dtype}={fill_value}"
+
+
 @tilelang.jit
 def vectorize_test_call_infinity():
     A = T.empty((4,), dtype=T.float32)


src/cuda/codegen/codegen_cuda.cc
```diff
@@ -5791,49 +5791,57 @@ void CodeGenTileLangCUDA::VisitExpr_(const BroadcastNode *op,
...
+          packed |= v << (i * 4);
         }
-      } else if (lanes == 16 || lanes == 32) {
-        os << "make_";
-        PrintType(op->dtype, os);
+        os << packed;
+      } else {
         os << '(';
-        for (int i = 0; i < lanes / 8; ++i) {
+        for (int i = 0; i < nibbles_per_field; ++i) {
```
LeiWang1999, line 5816: Verified the sign-handling here beyond the PR's own tests, since this is the subtle part: for **signed** int4 the carrier value can be sign-extended (e.g. a `signed char` holding -1 = `0xFF`), and `static_cast<unsigned int>` + `& 0x0fu` reduces it to the correct nibble regardless. Probed on an H100:

- constant `T.fill(sh_int4, -1)` → all nibbles `0xF` ✓
- truly-runtime fill from a kernel scalar arg (`T.fill(sh, T.Cast(dtype, v))`, so the value never was a constant): int4/uint4 × {0, 5, 7} ✓, and `v = -2` → `0xE` ✓

Also confirmed `SSAGetID` is a no-op reuse when the value is already a `Var` (the `HoistBroadcastValues` case that triggered #2997) — no redundant temp in the emitted source.

One optional micro-cleanup: for lanes 16/32 the 8-term packing expression is re-emitted once per 32-bit `make_` field (4× for lanes=32). nvcc CSEs it, so purely cosmetic, but SSA-ing the packed field once would shrink the generated source.

LeiWang1999 (commented):
Reviewed with local verification on H100. **Correct fix at the right layer, and the test coverage is genuinely good — LGTM.**

What I checked:

- **Premise reproduces**: with the codegen change reverted, `T.fill` of an int4 shared buffer dies with the `ICHECK` (`value ... is not a constant`) exactly as #2997 describes — `HoistBroadcastValues` rebinding the constant to a `Var` is what pushes the old branch over.
- **Constant path preserved**: `emit_packed_field` folds the same packed literals the old code produced for lanes 4/8/16/32, and lane counts outside that set still fall through to the generic path exactly as the old `fail` flag did (no behavior change there). The lanes==2 case isn't handled in this branch but lands on `PrintVecConstructor`'s `tl_pack_int4x2` from #3036, so there's no remaining gap.
- **Runtime path**: PR's 14 new tests pass on H100, plus adversarial probes for signed negatives and a genuinely runtime (kernel-argument) fill value — details in the inline comment.
- **Regressions**: full `test_tilelang_language_vectorize.py` (50 passed) and the int4x2 suite `test_tilelang_issue_2994.py` (97 passed) are clean with the patch.

Two small suggestions inline (a `-1` test case to pin the sign masking, and an optional codegen cosmetic) — neither blocks merge.

testing/python/language/test_tilelang_language_vectorize.py
```diff
@@ -216,6 +216,64 @@ def test_vectorize_broadcast_int8(vec_num):
...
+            T.copy(sh, B)
+
+    return func
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("dtype", ["int4", "uint4"])
+@pytest.mark.parametrize("fill_value", [0, 1, 7])
+def test_fill_int4_buffer(dtype, fill_value):
+    """End-to-end reproduction of issue #2997: T.fill/T.clear of an int4 buffer.
```
LeiWang1999, line 261: Suggest adding a negative value for the signed case, e.g.:

```python
@pytest.mark.parametrize("fill_value", [0, 1, 7, -1])
```

with `-1` skipped (or expected as `0xF`) for `uint4`. The runtime path's sign masking (`static_cast<unsigned int> ... & 0x0fu`) is exactly the part a future refactor could silently break, and `[0, 1, 7]` never exercises a sign-extended carrier. I probed `-1` and `-2` locally and they produce the right nibbles — pinning that in the suite is cheap.

LeiWang1999 (approved):
LGTM

PR #3117, opened 2026-08-30
[BugFix][CUDA][HIP] Handle logical not for vectorized bool

### LINKED ISSUE #3021: [BUG][Fuzzer][ice-on-valid-code] Logical `T.Not` on a `bool` buffer in an auto-vectorized loop emits `!ushort4`, crashing nvcc instead of negating lane-by-lane
### Required prerequisites

0.1.13

### System information

TileLang 0.1.13; L40S (sm_89), CUDA 12.8, PyTorch 2.8.0, Python 3.13.

### Problem description

A logical NOT (`T.Not` or Python `not`) applied elementwise inside a `T.Parallel` loop fails to compile whenever the loop auto-vectorizes (more than one element per thread). The emitted CUDA applies the scalar operator `!` directly to a vector operand, and nvcc rejects it:

(Note: `~` on a `bool` buffer routes through a *different* visitor and emits `~ushortN` — see "Class" below; that is the bitwise sibling, filed separately. This issue is specifically the `!` / `NotNode` site.)

```
tvm_kernels.cu(26): error: no operator "!" matches these operands
            operand types are: ! ushort4
    ushort4 v__1 = !__2;
```

The same kernel compiles and returns the correct result when the loop is not vectorized (one element per thread) — so this is triggered purely by the vector width, not by the `bool` dtype itself.

<details><summary>Emitted CUDA for the vectorized case (width 4) — the <code>!__2</code> is the crash</summary>

```cuda
extern "C" __global__ void __launch_bounds__(64, 1) main_kernel(const signed char* __restrict__ A, signed char* __restrict__ C) {
  int __1;
  ushort4 __2;
  int v_ = *(int*)(A + (((int)threadIdx.x) * 4));
  __2.x = (bool)(((char)(v_ >> 0)));
  __2.y = (bool)(((char)(v_ >> 8)));
  __2.z = (bool)(((char)(v_ >> 16)));
  __2.w = (bool)(((char)(v_ >> 24)));
  ushort4 v__1 = !__2;                 // <-- nvcc: no operator "!" matches ushort4
  __1=((signed char)(v__1.x) << 0);
  ...
}
```

The `bool` vector is correctly represented as a `ushort4`; only the `!` operator is un-scalarized.

</details>

### Reproducible example code

```python

N = 256
def build(threads):
    @T.prim_func
    def main(A: T.Tensor((N,), "bool"), C: T.Tensor((N,), "bool")):
        with T.Kernel(1, threads=threads) as bx:
            for i in T.Parallel(N):          # N/threads elems per thread
                C[i] = T.Not(A[i])
    return main

a = (torch.rand(N, device="cuda") > 0.5)
ref = ~a

# CONTROL: threads == N -> 1 elem/thread -> scalar `!x`
c = tilelang.compile(build(256), out_idx=[1])(a)
print("threads=256 (scalar) :", "PASS" if torch.equal(c, ref) else "FAIL")   # -> PASS

# REPRO: threads < N -> 4 elems/thread -> vectorized `!ushort4`
c = tilelang.compile(build(64), out_idx=[1])(a)                              # -> nvcc compile error
print("threads=64  (vector) :", "PASS" if torch.equal(c, ref) else "FAIL")
```

### Traceback

```
tvm_kernels.cu(26): error: no operator "!" matches these operands
            operand types are: ! ushort4
    ushort4 v__1 = !__2;
                   ^
tvm_kernels.cu(26): note #3328-D: built-in operator!(<bool>) does not match because argument #1 does not match parameter
1 error detected in the compilation of "tvm_kernels.cu".
```

### Expected behavior

The negation of a `bool` vector should compile and compute lane-by-lane, matching the scalar path (which already returns the correct result over the same inputs). Every other elementwise op in this codegen — `&&`, `||`, `+`, comparisons — already scalarizes over a vector via `PrintBinaryExpr`'s `op->dtype.lanes() == 1` split; `NotNode` is the one unary op that skips that check.

### Additional context

**Root cause.** The C codegen's `NotNode` visitor does not account for a vector `dtype`, so it emits `!` against a vector operand that CUDA has no `operator!` for. In [`CodeGenC::VisitExpr_(const NotNode*)`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L617) it unconditionally writes `os << '!'; PrintExpr(op->a)`, with no `lanes()` branch. Contrast [`PrintBinaryExpr`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L514), which every arithmetic/logical/comparison node routes through: for `lanes() > 1` it defers to `PrintVecBinaryOp` instead of emitting a scalar operator on a vector. The CUDA codegen does not override `NotNode`, so the base-class scalar `!` reaches nvcc verbatim.

**Suggested fix.** One direction is to give `NotNode` the same lane-aware handling the binary nodes have — emit a per-lane expansion (e.g. `make_ushortN(!a.x, !a.y, ...)`) when `op->dtype.lanes() > 1`, mirroring `PrintVecBinaryOp`. Whether the cleaner place is the visitor or an earlier scalarize/legalize pass is a call for the maintainers.

**Class / generalization (two-level root + 4-axis sweep, all cells run on 0.1.13 / sm_89).**

- **Source-level root:** `CodeGenC::VisitExpr_(const NotNode*)` at [codegen_c.cc L617](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L617) unconditionally emits `os << '!'; PrintExpr(op->a)` with no `op->dtype.lanes()` branch. Both binary helpers ([`PrintBinaryExpr` L514](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L514), `PrintBinaryIntrinsic` L536) split on `lanes()==1` and defer wide operands to `PrintVecBinaryOp`; the unary visitors do not.
- **Operator-level root:** the *only* two visitor sites in `CodeGenC` that emit a C operator on a raw operand with no lane split are `NotNode` (`!`) and the `bitwise_not` intrinsic branch (`os << "(~"`, [L677](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L677)). Every other unary (`Cast`, and negation which is lowered to `0 - x` = a binary `SubNode`) already handles vectors. So the class boundary is exactly these two sites.

| axis | cell tested (vectorized, `threads<N`) | observed result | same-root? |
|---|---|---|---|
| repro | `T.Not(A[i])`, bool, threads=64 (width 4) | `ushort4 v__1 = !__2;` → `no operator "!" matches ... ushort4` | — (the bug) |
| similar-logic | Python `not A[i]`, bool, width 4 | `ushort4 v__1 = !__2;` → same `operator "!"` crash | SAME (NotNode) |
| related-type | `T.Not(A[i])`, bool, threads=128 (width 2) | `ushort2 v__1 = !__2;` → same `operator "!"` crash | SAME (NotNode, any legal width) |
| related-operator | `T.Not(A[i] > 0)`, int32→bool, width 4 | **PASS** — simplifier folds `!(x>0)` to `(x<=0)`, NotNode eliminated pre-codegen | dodges (NotNode never reaches codegen) |
| related-operator | `~A[i]`, bool buffer, width 4 | `ushort4 v__1 = ~__2;` → `no operator "~" matches ... ushort4` | SIBLING site (`bitwise_not`, not `NotNode`) |
| related-operator / related-type | `~A[i]`, int32, width 4 | `*(int4*)... = (~*(int4*)...)` → `no operator "~" matches ... int4` | SIBLING site (`bitwise_not`) |

Takeaways: (1) the NotNode crash is width-agnostic (widths 2 and 4 both fire) and trigger-agnostic between `T.Not` and Python `not`. (2) The bug only survives when the NotNode reaches codegen — `T.Not` of a *comparison* is constant-folded by the simplifier (`!(x>0)`→`(x<=0)`) and never crashes; the defect needs a NotNode over a value the simplifier can't rewrite, e.g. a plain `bool` buffer load. (3) `~` on any buffer (bool or int) is a **distinct-but-same-root sibling** at the `bitwise_not` site, filed separately as the "vectorized `~`" issue; kept separate because they are two independent visitor sites fixable independently. Both are the single missing-`lanes()`-split pattern in `CodeGenC`.

**Provenance.** The `NotNode` visitor has carried this scalar-only form in the vendored TVM `CodeGenC` (`3rdparty/tvm`, pinned at `8df8ebd6` for the 0.1.13 tag). Introduced-vs-relocated before the fork is unverified.

**Dedup.** I searched the open and closed tracker and found no report of a logical-NOT crash on a vectorized `bool`. The closest are distinct mechanisms: #2206 is a `PrintType` `LogFatal` on `boolx8` (type printing, not `operator!`), and #2911/#2382 concern vector `Select`/`atomic_add` value handling.

**Reach.** The trigger is first-class (`T.Not` and Python `not` over a `bool`). Auto-vectorization is the default: any `T.Parallel(N)` with `threads < N` (and `N/threads` a legal vector width) takes the vector path, so the crash fires on the common case; only `threads == extent` (one element per thread) dodges it.

No shipped example negates a `bool` on the device side inside a `T.Parallel`, so CI is green. The one example that *contains* a boolean negation, `examples/block_causal_attention/block_causal_attention.py`, writes `~q_clean & ~k_clean` (line 429) only in `_dllm_mask` — a **pure-torch host-side reference** operating on `torch.Tensor`s, not device-side TileLang. Its device kernels use `int` masks with `T.if_then_else`, not `T.Not`. I ran it verbatim on 0.1.13: it compiles without the `operator!` crash (it fails only a numeric backward-gradient tolerance check, unrelated to this bug), confirming the example does **not** exercise this defect. So the "attention/masking" reach is a *plausible idiom* argument, not a shipped-example one.

**Impact.** The trigger is narrow-looking (a `bool` negation whose loop happens to auto-vectorize) but sits on the default code path for the common `threads < N` case. When it fires it is a nvcc compile-time error, so it is loud and caught immediately: the kernel simply fails to build, and it cannot silently corrupt any output or slip into a running workload. Fixing it lets a legal, common idiom (elementwise negation of a boolean mask) compile in the vectorized path instead of only when one element maps to each thread; it closes a compile-time crash class, not a silent-data hazard.

jjppp:
Previously, both CUDA and HIP codegen delegate to `CodeGenC`, which cannot emit logical not for vectorized bools, causing device compilation to fail.
This PR handles vectorized `NotNode` expressions lane by lane in CUDA and HIP.

Fixes #3021

## Summary

- Added lane-by-lane `NotNode` code generation for vectorized CUDA and HIP booleans.
- Preserved scalar handling through `CodeGenC`.
- Added CUDA and HIP codegen tests for vector widths 2, 3, and 4.
- Added a regression test for auto-vectorized `T.Parallel` loops.

## C++ style / lint notes

- The PR does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- No correctness or build issues are indicated by the changes.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -5554,6 +5554,31 @@ void CodeGenTileLangCUDA::VisitExpr_(const SelectNode *op, std::ostream &os) {
   os << r_var;
 }
 
+void CodeGenTileLangCUDA::VisitExpr_(const NotNode *op, std::ostream &os) {
+  if (!op->dtype.is_fixed_length_vector()) {
+    CodeGenC::VisitExpr_(op, os);
+    return;
+  }
+
+  std::string result = name_supply_->FreshName("_");
+  this->PrintIndent();
+  this->PrintType(op->dtype, stream);
+  stream << ' ' << result << ";\n";
+  int ssa_scope = BeginScope();
+  {
+    std::string value = SSAGetID(PrintExpr(op->a), op->a.dtype());
+    for (int i = 0; i < op->dtype.lanes(); ++i) {
+      std::ostringstream lane;
+      lane << "!bool(";
+      PrintVecElemLoad(value, op->a.dtype(), i, lane);
+      lane << ')';
+      PrintVecElemStore(result, op->dtype, i, lane.str());
+    }
+  }
+  EndScope(ssa_scope);
+  os << result;
+}
+
 void CodeGenTileLangCUDA::VisitExpr_(const ShuffleNode *op,
                                      std::ostream &os) { // NOLINT(*)
   // For bfloat16x2 / float16x2 construction from two scalar lanes, emit a
--- a/src/cuda/codegen/codegen_cuda.h
+++ b/src/cuda/codegen/codegen_cuda.h
@@ -58,6 +58,7 @@ public:
   void VisitExpr_(const ShuffleNode *op, std::ostream &os) final;
   void VisitExpr_(const MinNode *op, std::ostream &os) final;
   void VisitExpr_(const MaxNode *op, std::ostream &os) final;
+  void VisitExpr_(const NotNode *op, std::ostream &os) final;
   void VisitStmt_(const EvaluateNode *op) final;
   void VisitStmt_(const AllocBufferNode *op) final;
   void VisitStmt_(const AttrStmtNode *op) final;
--- a/src/rocm/codegen/codegen_hip.cc
+++ b/src/rocm/codegen/codegen_hip.cc
@@ -806,6 +806,31 @@ void CodeGenTileLangHIP::VisitExpr_(const SelectNode *op, std::ostream &os) {
   os << result;
 }
 
+void CodeGenTileLangHIP::VisitExpr_(const NotNode *op, std::ostream &os) {
+  if (!op->dtype.is_fixed_length_vector()) {
+    CodeGenC::VisitExpr_(op, os);
+    return;
+  }
+
+  std::string result = name_supply_->FreshName("_");
+  this->PrintIndent();
+  this->PrintType(op->dtype, stream);
+  stream << ' ' << result << ";\n";
+  int ssa_scope = BeginScope();
+  {
+    std::string value = SSAGetID(PrintExpr(op->a), op->a.dtype());
+    for (int i = 0; i < op->dtype.lanes(); ++i) {
+      std::ostringstream lane;
+      lane << "!bool(";
+      PrintVecElemLoad(value, op->a.dtype(), i, lane);
+      lane << ')';
+      PrintVecElemStore(result, op->dtype, i, lane.str());
+    }
+  }
+  EndScope(ssa_scope);
+  os << result;
+}
+
 void CodeGenTileLangHIP::PrintVecElemLoad(const std::string &vec, DataType t,
                                           int i,
                                           std::ostream &os) { // NOLINT(*)
--- a/src/rocm/codegen/codegen_hip.h
+++ b/src/rocm/codegen/codegen_hip.h
@@ -50,6 +50,7 @@ public:
   void VisitExpr_(const FloatImmNode *op, std::ostream &os) final;
   void VisitExpr_(const CallNode *op, std::ostream &os) final;
   void VisitExpr_(const CastNode *op, std::ostream &os) final;
+  void VisitExpr_(const NotNode *op, std::ostream &os) final;
   void VisitExpr_(const SelectNode *op, std::ostream &os) final;
   void VisitExpr_(const ShuffleNode *op, std::ostream &os) final; // NOLINT(*)
   void VisitStmt_(const AllocBufferNode *op) final;
--- a/testing/python/amd/test_tilelang_hip_vector_not_codegen.py
+++ b/testing/python/amd/test_tilelang_hip_vector_not_codegen.py
@@ -1,0 +1,40 @@
+import pytest
+import tilelang
+import tilelang.testing
+
+from tilelang import tvm
+from tvm import tirx
+
+
+def _make_vector_not_module(lanes):
+    value = tirx.Var("value", f"boolx{lanes}")
+    func = tirx.PrimFunc([value], tirx.Evaluate(tirx.Not(value)))
+    func = func.with_attr("global_symbol", "vector_not")
+    func = func.with_attr("calling_conv", tvm.ir.CallingConv.DEVICE_KERNEL_LAUNCH)
+    return tvm.IRModule({"vector_not": func})
+
+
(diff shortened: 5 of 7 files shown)


jjppp:
HIP and CUDA now deviate on how they support boolx8: HIP uses per-bit encoding while CUDA simply aborts. I'm still concerned about HIP's implementation choice that every write to vectorized bools now become a read-modify-write, as we cannot update a single bit.

That said, this PR is orthogonal to boolx8 support, and should also work for boolx8 if such types are later implemented.

SiriusNEO (approved):

PR #3121, opened 2026-08-31
[BugFix] Always vectorize `T.Parallel` loops

Yongqi-Zhuo:
Previously `T.Parallel` loops involving only local buffers (including fragment buffers) are not allowed to be vectorized. After careful inspection we found that this does not make sense. So we lift the restriction.

This allows us to vectorize non-casting `T.Parallel` loops on registers to utilize e.g. `fp32x2`.

## Summary

- Removes the restriction on vectorizing `T.Parallel` loops that use only local or fragment buffers.
- Enables register-only loops to use vector types such as `fp32x2`.
- Simplifies `LowerParallelLoop` by removing its `should_vectorize` parameter and restoring always-vectorize behavior with access analysis selecting the vector width.
- Updates backend call sites to use the shared vectorization behavior.
- Updates tests to verify register-register vectorization and generated loop ordering.

## C++ style / lint notes

- The PR changes C++ APIs and implementation code, but it does not modify rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) may report advisory findings. These warnings are separate from correctness, build, and test results and are not merge blockers unless they indicate a clear API, FFI, or maintainability risk.

### DIFF
--- a/src/backend/common/op/atomic_reduce.h
+++ b/src/backend/common/op/atomic_reduce.h
@@ -183,10 +183,11 @@ struct AtomicReduce {
                           level);
     }
     auto loop_layout = par_op->GetLoopLayout();
-    return LowerParallelLoop(
-        fused_loop, loop_layout, lower_args.thread_index, analyzer,
-        lower_args.layout_map, par_op->GetPredicate(lower_args.thread_index),
-        /*parallel_loop=*/true, par_op->LoopLayoutRequiresPaddingGuard());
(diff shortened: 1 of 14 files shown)


Yongqi-Zhuo:
@regression-perf

Yongqi-Zhuo:
@regression-perf

Yongqi-Zhuo:
@regression-perf

PR #3125, opened 2026-09-01
[BugFix] Reject non-bijective sm8x sparse metadata shapes

### LINKED ISSUE #2606: [BUG][Fuzzer][wrong-code] `make_cutlass_metadata_layout` (sm8x) silently emits a non-bijective, out-of-range E-metadata layout when M is not a multiple of the interleave group
### Required prerequisites

0.1.11

### System information

- Installed: pip (dev build reporting `0.1.11`; the source I cite matches commit `cd37ed5f` exactly and is unchanged on `origin/main`).
- GPU: NVIDIA L40S (sm_89) — but the bug is an **architecture-independent integer permutation** in a Python layout helper; it manifests for any sm_80..sm_89 target that reaches the `sm8x` branch.
- Python 3.13, `torch 2.8`, CUDA 12.8.

### Problem description

`tilelang.layout.make_cutlass_metadata_layout(buffer, mma_dtype, arch=...)` builds the layout used to place the 2:4-structured-sparse **E metadata** buffer in the CUTLASS sparse-GEMM convention. For an `sm_8x` target it dispatches to `make_cutlass_metadata_layout_sm8x`, whose inner `ColumnMajorInterleaved(i, j)` map is intended to be a **bijection** of the `(M, K)` metadata buffer onto itself (a pure permutation — every logical metadata element must land on exactly one distinct physical slot inside `[0, M*K)`).

It is a bijection **only when `M` is a multiple of the interleave group** (`group = 32` for 16-bit metadata, `group = 16` for 32-bit metadata). Whenever `M % group != 0`, the map:

1. **collides** — two distinct logical elements map to the *same* physical slot (one silently overwrites the other), and
2. **goes out of range** — some logical elements map *past the end* of the metadata buffer (an out-of-bounds write/read).

There is no exception and no warning: the helper just returns a `T.Layout` that silently mis-routes and corrupts metadata. The result is deterministic. Because the metadata E selects which columns of A are the non-zeros, a corrupted E silently produces a wrong GEMM result (or an OOB access).

The upstream admission check for sparse M is only `M % 16 == 0` ([`gemm_sp.cc:115-118`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/op/gemm_sp.cc#L115-L118), `kMPerWarp = 16`), which does **not** imply `M % 32 == 0`. So an fp16 metadata buffer with `M ∈ {16, 48, 80, 112, …}` passes every check yet breaks this layout.

### Reproducible example code

```python
# Pure-integer bijection oracle for make_cutlass_metadata_layout_sm8x.
# Transcribed EXACTLY from tilelang/layout/gemm_sp.py:138-145. No FP, no GPU needed.
def column_major_interleaved(i, j, m, k, group, interweave):
    i = i // group * group + (i % 8) * interweave + (i % group) // 8
    topright   = (1 - (i % 2)) & (j % 2)
    bottomleft = (i % 2) & (1 - (j % 2))
    i += topright - bottomleft
    j -= topright - bottomleft
    offset = (j // 2) * m * 2 + i * 2 + (j % 2)
    return offset // k, offset % k

def is_bijection(m, k, group, interweave):
    size, seen, coll, oob = m * k, {}, [], []
    for i in range(m):
        for j in range(k):
            pr, pc = column_major_interleaved(i, j, m, k, group, interweave)
            flat = pr * k + pc
            if not (0 <= pr < m and 0 <= pc < k and 0 <= flat < size):
                oob.append(((i, j), (pr, pc), flat))
            if flat in seen: coll.append(((i, j), seen[flat], flat))
            else: seen[flat] = (i, j)
    return len(seen) == size and not coll and not oob, coll, oob

# fp16 / bf16 -> 16-bit metadata: group=32, interweave=4, k=4
for m in (32, 48):            # 32 % 32 == 0 (control) ; 48 % 32 == 16 (broken)
    ok, coll, oob = is_bijection(m, 4, 32, 4)
    print(f"fp16 m={m}: bijection={ok}  collisions={len(coll)}  oob={len(oob)}")
    if coll: print("   collide:", coll[0])   # two logical -> same slot
    if oob:  print("   oob    :", oob[0])     # logical -> past end of buffer
```

Output:

```
fp16 m=32: bijection=True  collisions=0  oob=0
fp16 m=48: bijection=False collisions=16 oob=16
   collide: ((36, 0), (0, 2), 96)     # logical (36,0) and (0,2) both -> physical flat 96
   oob    : ((36, 2), (48, 0), 192)   # logical (36,2) -> physical row 48 (>= M=48), flat 192 >= size 192
```

The **same** collisions/OOB come out of the real `Layout` object (not just the transcription above), via `Layout.map_forward_index`, including through the public dispatcher:

```python
import tilelang, tvm
import tilelang.language as T
from tilelang.layout.gemm_sp import make_cutlass_metadata_layout

# fp16 metadata, M=48 (48 % 32 = 16), K=4 -> sm8x path via the public API
buf = tvm.tirx.decl_buffer((48, 4), T.int16)
layout = make_cutlass_metadata_layout(buf, mma_dtype=T.float16, arch="8.6")
print(layout.map_forward_index([36, 0]))   # -> [24, 0]
print(layout.map_forward_index([0, 2]))    # -> [24, 0]   <-- SAME slot: collision
print(layout.map_forward_index([36, 2]))   # -> [48, 0]   <-- row 48 >= M=48: OUT OF RANGE
```

Sweeping `M ∈ [8, 128]`, the set of `M` that break is **exactly** `{M : M % group != 0}`:

- 16-bit meta (fp16/bf16, group 32): `{8, 16, 24, 40, 48, 56, 72, 80, 88, 104, 112, 120}`
- 32-bit meta (int8/fp8, group 16): `{8, 24, 40, 56, 72, 88, 104, 120}`

Every `M % group == 0` is a perfect bijection (control). The `sm90` sibling `make_cutlass_metadata_layout_sm90` is **not** affected (its col-major-decompose + ceil-repeat construction is injective even for `M % group != 0`).

### Traceback

```pytb
No traceback — `make_cutlass_metadata_layout(...)` returns a `T.Layout` normally, with no exception or warning; the defect is that the returned permutation is non-bijective (collisions) and maps out of range. The corruption surfaces only later as a silently wrong sparse-GEMM result or an out-of-bounds metadata access, not as an error at layout-construction time.
```

### Expected behavior

`make_cutlass_metadata_layout(...)` should either return a genuine bijection for every `M` it accepts, or reject an `M` it cannot lay out correctly with a clear error. It should never silently return a layout that maps two metadata elements onto one slot or writes past the buffer.

### Additional context

**Root cause.** The defect is in [`tilelang/layout/gemm_sp.py:138-145`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/layout/gemm_sp.py#L138-L145):

```python
def ColumnMajorInterleaved(i: int, j: int) -> int:
    i = i // group * group + (i % 8) * interweave + (i % group) // 8   # <-- (A)
    topright   = (1 - (i % 2)) & (j % 2)
    bottomleft = (i % 2) & (1 - (j % 2))
    i += topright - bottomleft
    j -= topright - bottomleft
    offset = (j // 2) * m * 2 + i * 2 + (j % 2)
    return offset // k, offset % k
```

Line (A) is the interleave that assumes each `group`-row block is complete: `(i % 8) * interweave` spreads 8 consecutive rows by `interweave`, and `(i % group) // 8` picks the sub-block within the group. That row-permutation is a bijection **only within a full group of `group` rows**. When `M % group != 0` the final partial block has fewer than `group` rows, so `(i % 8) * interweave` scatters those partial rows to permuted positions that (i) coincide with positions already used by earlier full blocks (collision) and (ii) exceed `M` (OOB), because the `* m * 2` / `offset // k` recombination still assumes the group-complete stride.

Constructed via [`make_cutlass_metadata_layout_sm8x` (lines 121-147)](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/layout/gemm_sp.py#L121-L147) and reached publicly via the [dispatcher (lines 150-161)](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/layout/gemm_sp.py#L150-L161) for `8.0 <= compute < 9.0`. `group`/`interweave` are set at [lines 135-136](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/layout/gemm_sp.py#L134-L136).

**Suggested fix** (either is sufficient to remove the silent corruption, proposed not built):

- **Reject** at construction: `assert m % group == 0, f"metadata rows M={m} must be a multiple of {group} for the sm8x interleave"` in `make_cutlass_metadata_layout_sm8x`. Cheapest, no behavior change for the valid path.
- **Correct** the interleave to be a bijection for any `M` (handle the partial trailing group), if partial-M metadata is intended to be supported. Which is appropriate depends on whether a non-`group`-multiple M is meant to be laid out here — a call for the maintainers.

**Provenance.** The buggy `ColumnMajorInterleaved` interleave was introduced by PR [#1056](https://github.com/tile-ai/tilelang/pull/1056) (commit `283a9a00`, "support `T.gemm_sp_v2` on sm80 and sm89", 2025-12-01, @botbw) — `git blame` on the interleave line points there. The later `tirx` refactor ([#2216](https://github.com/tile-ai/tilelang/pull/2216), commit `b939fa01`) only relocated the code, and the `sm90` sibling `make_cutlass_metadata_layout_sm90` (unaffected) was added separately. Not a regression — the sm8x interleave has been non-bijective for `M % group != 0` since it first shipped. The project's own sparse tests / `examples/gemm_sp` use `M` that is a multiple of the group, so the partial-M case is never exercised.

I searched the open and closed tracker (`make_cutlass_metadata_layout`, `metadata layout bijection`, `gemm_sp metadata`, `ColumnMajorInterleaved`) and found no existing report. Distinct from the arch-gate issues #2503/#2504 (those are arch-version-legality bugs; this is a pure integer permutation in a layout helper) and from the sm_90 WGMMA `T.gemm_sp` compute-atom bugs (this is the sm_8x 2:4 E-metadata *layout*, a different file section, arch tier, and failure mode).

saiyambharara:
## Summary
- `make_cutlass_metadata_layout_sm8x`'s interleave is a bijection only when M is a multiple of the group size (32/16) and K is even; otherwise it silently collides or writes out of range.
- Reject both cases with a `ValueError`.

## Test plan
- [x] `pytest testing/python/layout/test_tilelang_layout_gemm_sp_metadata.py`

Fixes #2606

## Summary

- Validate SM8x sparse metadata shapes before layout construction.
- Raise `ValueError` when `M` is not a multiple of the required interleave group or when `K` is odd.
- Add tests for rejected and accepted metadata shapes and dtypes.

## Impact

This prevents non-bijective mappings, collisions, and out-of-range metadata accesses for unsupported shapes.

### DIFF
--- a/testing/python/tilelibrary/test_tilelang_tilelibrary_gemm_sp.py
+++ b/testing/python/tilelibrary/test_tilelang_tilelibrary_gemm_sp.py
@@ -5,6 +5,8 @@ from tilelang.utils.tensor import torch_assert_close
 import tilelang.testing
 import torch
 import tilelang.language as T
+from tilelang import tvm
+from tilelang.layout.gemm_sp import make_cutlass_metadata_layout_sm8x
 
 
 def matmul(
@@ -755,6 +757,22 @@ def test_gemm_rr(
 )
 def test_compress_dtype_combinations(in_dtype, out_dtype, dtypeAccum, meta_dtype):
     run_gemm_ss(128, 128, 128, False, True, in_dtype, out_dtype, dtypeAccum, 128, 128, 64, 2, 128, meta_dtype=meta_dtype)
+
+
+def test_make_cutlass_metadata_layout_sm8x_rejects_bad_shapes():
+    with pytest.raises(ValueError, match="multiple of"):
+        make_cutlass_metadata_layout_sm8x(tvm.tirx.decl_buffer((48, 4), T.int16), mma_dtype=T.float16)  # 48 % 32 != 0
+
+    with pytest.raises(ValueError, match="even"):
+        make_cutlass_metadata_layout_sm8x(tvm.tirx.decl_buffer((64, 3), T.int16), mma_dtype=T.float16)  # odd K
+
+    with pytest.raises(ValueError, match="multiple of"):
+        make_cutlass_metadata_layout_sm8x(tvm.tirx.decl_buffer((40, 4), T.int32), mma_dtype=T.int8)  # 40 % 16 != 0
+
+
+def test_make_cutlass_metadata_layout_sm8x_accepts_valid_shapes():
+    make_cutlass_metadata_layout_sm8x(tvm.tirx.decl_buffer((64, 4), T.int16), mma_dtype=T.float16)
+    make_cutlass_metadata_layout_sm8x(tvm.tirx.decl_buffer((48, 4), T.int32), mma_dtype=T.int8)
 
 
 if __name__ == "__main__":
--- a/tilelang/layout/gemm_sp.py
+++ b/tilelang/layout/gemm_sp.py
@@ -135,6 +135,12 @@ def make_cutlass_metadata_layout_sm8x(buffer: tvm.tirx.Buffer, mma_dtype: str):
     group = 32 if buffer.dtype.bits == 16 else 16
     interweave = 4 if buffer.dtype.bits == 16 else 2
 
+    # ref: https://github.com/pytorch/pytorch/blob/18acb4f62d1557fb068bfeec4622967885484da6/torch/sparse/_semi_structured_conversions.py#L72-L81
+    if m % group != 0:
+        raise ValueError(f"metadata M={m} must be a multiple of {group}")
+    if k % 2 != 0:
+        raise ValueError(f"metadata K={k} must be even")
+
     def ColumnMajorInterleaved(i: int, j: int) -> int:
         i = i // group * group + (i % 8) * interweave + (i % group) // 8
         topright = (1 - (i % 2)) & (j % 2)


saiyambharara:
@botbw @LeiWang1999 tagging you both since this touches code you've worked on directly. Would appreciate a review when you have a chance.

botbw:
Sorry, I totally forgot this. Just did a quick check, pytorch performed the same shape check here (and it might be better if we add a comment for reference):

https://github.com/pytorch/pytorch/blob/18acb4f62d1557fb068bfeec4622967885484da6/torch/sparse/_semi_structured_conversions.py#L72-L81

Though I don't think a separate test file is necessary.

saiyambharara:
Good call, added the reference. Moved the tests into the existing gemm_sp test file instead of a new one.

LeiWang1999 (approved):
LGTM

saiyambharara:
@LeiWang1999, the CUDA-auto job is failing during venv setup while importing flash_attn_2_cuda due to an undefined c10::cuda_check_implementation symbol. The actual CUDA tests aren't reached. Could you rerun the failed job? It looks like a runner/dependency issue.

saiyambharara:
Hi @LeiWang1999, thanks for the approval. Merge seems blocked by the 

> Test for Python 3.12 with CUDA-auto

 job, but the failure looks unrelated to this PR: it's an `ImportError: undefined symbol ... c10_cuda_check_implementation` from `flash_attn_2_cuda.so`, which looks like a stale/ABI-mismatched build artifact on the runner rather than anything from this diff (this PR only touches tilelang/layout/gemm_sp.py and its test). The run is also from Sep 2, so the runner environment has likely moved on since.

Could you re-run that job when you get a chance? Happy to help debug further if it fails again. Thanks

PR #3127, opened 2026-09-01
[BugFix] Coalesce sub-32-bit elements in parallel loop partition   

zhouyangye1076:
### Problem                                                                                                                                                   
                                                                                                                                                                 
   When a non-fragment `T.Parallel` loop touches global memory with a sub-32-bit                                                                                 
   dtype (fp8/int8/uint16/fp16), the loop partitioner may assign each thread fewer                                                                               
   than 4 bytes. For example, copying 128 fp8 elements with 128 threads gives                                                                                    
   1 byte per thread.                                                                                                                                            
                                                                                                                                                                 
   cp.async requires a transfer width of 4/8/16 bytes, so with a sub-4B per-thread                                                                               
   chunk `InjectPTXAsyncCopy` finds no eligible vectorized global→shared store and                                                                               
   the copy **silently falls back to a synchronous scalar global load** — even when                                                                              
   `prefer_instruction="cp_async"` is given explicitly. These sub-4B per-thread                                                                                  
   global accesses are also memory-inefficient in general.                                                                                                       
                                                                                                                                                                 
   ### Reproducer                                                                                                                                                
                                                                                                                                                                 
   ```python                                                                                                                                                     
   import tilelang                                                                                                                                               
   import tilelang.language as T                                                                                                                                 
   from tilelang.tools.compile_only import compile_kernel_source                                                                                                 
                                                                                                                                                                 
   N, THREADS = 128, 128                                                                                                                                         
                                                                                                                                                                 
                                                                                                                                                                 
   @T.prim_func                                                                                                                                                  
   def kernel(                                                                                                                                                   
       A: T.Tensor((N,), T.float8_e4m3fn),                                                                                                                       
       B: T.Tensor((N,), T.float8_e4m3fn),                                                                                                                       
   ):                                                                                                                                                            
       with T.Kernel(1, threads=THREADS) as bx:                                                                                                                  
           a_sh = T.alloc_shared((N,), T.float8_e4m3fn)                                                                                                          
           T.copy(A, a_sh, prefer_instruction="cp_async")                                                                                                        
           T.copy(a_sh, B)                                                                                                                                       
                                                                                                                                                                 
                                                                                                                                                                 
   src = compile_kernel_source(kernel, "cuda -arch=sm_80")                                                                                                       
   print(src)                                                                                                                                                    
   assert "tl::cp_async" in src          # used to silently fall back to a sync load                                                                             
   assert "threadIdx.x) < 32" in src     # idled threads are predicated off                                                                                      
   ```                                                                                                                                                           
                                                                                                                                                                 
   ### Generated code (kernel body)                                                                                                                              
                                                                                                                                                                 
   **Before** — 1 byte per thread, cp.async rewrite misses, silent sync fallback:                                                                                
                                                                                                                                                                 
   ```cuda                                                                                                                                                       
   extern "C" __global__ void __launch_bounds__(128, 1) kernel_kernel(const fp8_e4_t* __restrict__ A, fp8_e4_t* __restrict__ B) {                                
     extern __shared__ __align__(1024) fp8_e4_t a_sh[];                                                                                                          
     a_sh[((int)threadIdx.x)] = A[((int)threadIdx.x)];                                                                                                           
     B[((int)threadIdx.x)] = a_sh[((int)threadIdx.x)];                                                                                                           
   }                                                                                                                                                             
   ```                                                                                                                                                           
                                                                                                                                                                 
   **After** — per-thread chunk coalesced to 4 bytes, lowered to cp.async, idle                                                                                  
   threads covered by the partition predicate:                                                                                                                   
                                                                                                                                                                 
   ```cuda                                                                                                                                                       
   extern "C" __global__ void __launch_bounds__(128, 1) kernel_kernel(const fp8_e4_t* __restrict__ A, fp8_e4_t* __restrict__ B) {                                
     extern __shared__ __align__(1024) fp8_e4_t a_sh[];                                                                                                          
     if (((int)threadIdx.x) < 32) {                                                                                                                              
       tl::cp_async_gs<4>((&(a_sh[(((int)threadIdx.x) * 4)])), (&(A[(((int)threadIdx.x) * 4)])));                                                                
     }                                                                                                                                                           
     tl::cp_async_commit();                                                                                                                                      
     tl::cp_async_wait<0>();                                                                                                                                     
     __syncthreads();                                                                                                                                            
     if (((int)threadIdx.x) < 32) {                                                                                                                              
       *(fp8_e4_4_t*)(B + (((int)threadIdx.x) * 4)) = *(fp8_e4_4_t*)(a_sh + (((int)threadIdx.x) * 4));                                                           
     }                                                                                                                                                           
   }                                                                                                                                                             
   ```                                                                                                                                                           
                                                                                                                                                                 
   ### Fix                                                                                                                                                       
                                                                                                                                                                 
   In `ParallelOpNode::ComputePlanCandidate`, after `SelectMinPaddingVectorSize`                                                                                 
   picks the smallest-padding vector width, scan the loop body for the smallest                                                                                  
   dtype bit-width among the accessed global-memory buffers                                                                                                      
   (`MinGlobalAccessDtypeBits`). If it is below 32 bits (and divides 32), scale the                                                                              
   per-thread chunk up to `32 / dtype_bits` elements — but only when the original                                                                                
   (maximum legal) vector width allows it, so we never exceed what vectorization                                                                                 
   would have used anyway.                                                                                                                                       
                                                                                                                                                                 
   Threads left idle by the wider chunk are covered by the partition predicate                                                                                   
   (and by predicated cp.async after the rewrite pass), so there is no                                                                                           
   out-of-bounds access, as shown by the `threadIdx.x < 32` guard in the generated                                                                               
   code above. Loops that do not touch global memory and loops with explicit                                                                                     
   `coalesced_width` annotations are unaffected.                                                                                                                 
  

## Summary

- Fixes partitioning for non-fragment `T.Parallel` loops that access global memory with sub-32-bit dtypes.
- Coalesces each thread’s work to at least 32 bits when the legal vector width allows it.
- Enables supported-width `cp.async` transfers instead of scalar synchronous loads.
- Guards threads that become inactive after chunk widening.
- Leaves loops without global-memory accesses and loops with explicit `coalesced_width` unchanged.

## C++ style / lint notes

- The PR changes C++ code in `src/op/parallel.cc`.
- It does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step remains advisory.
- No correctness, build, or test issue is identified from the described changes.

### DIFF
--- a/src/op/parallel.cc
+++ b/src/op/parallel.cc
@@ -10,6 +10,7 @@
 #include <tvm/runtime/logging.h>
 
 #include <algorithm>
+#include <limits>
 #include <tvm/tirx/analysis.h>
 #include <tvm/tirx/op.h>
 #include <unordered_set>
@@ -120,7 +121,85 @@ int SelectMinPaddingVectorSize(int max_vector_size, PrimExpr loop_total_size,
   return best_vector_size;
 }
 
+/**
+ * @brief Whether the statement contains a call with non-pure effects.
+ *
+ * Atomics and stateful intrinsics rule out treating the loop as a pure copy.
+ * Pure calls (e.g. shift/logical builtins inside index arithmetic) do not
+ * block copy coalescing.
+ */
+bool ContainsImpureCall(const Stmt &stmt) {
+  bool impure = false;
+  PostOrderVisit(stmt, [&](const ObjectRef &node) {
+    if (impure) {
+      return;
+    }
+    if (const auto *call = node.as<CallNode>()) {
+      impure = SideEffect(GetRef<PrimExpr>(call)) > CallEffectKind::kPure;
+    }
+  });
+  return impure;
+}
+
+/**
+ * @brief Whether every buffer store is a raw global->shared copy.
+ *
+ * A raw copy is a shared-memory store whose value is a load from a global
+ * buffer with a matching dtype — the only shape that the cp.async injector
+ * can rewrite. cp.async only applies to global->shared transfers, so loops
+ * with other side effects (global stores, mixed compute) must keep their
+ * original partition: re-coalescing them changes their codegen (or even
+ * their semantics, e.g. vectorized atomics) without enabling cp.async.
+ */
+bool IsPureGlobalToSharedCopyLoop(const Stmt &stmt) {
+  bool pure = true;
+  bool saw_store = false;
+  PostOrderVisit(stmt, [&](const ObjectRef &node) {
+    const auto *store = node.as<BufferStoreNode>();
+    if (store == nullptr || !pure) {
+      return;
+    }
+    saw_store = true;
+    const auto *load = store->value.as<BufferLoadNode>();
+    if (load == nullptr || !IsGlobalBuffer(load->buffer) ||
+        !IsSharedBuffer(store->buffer) ||
+        load->buffer->dtype != store->buffer->dtype) {
+      pure = false;
+    }
+  });
+  return saw_store && pure;
+}
+
+/**
+ * @brief Smallest element bit-width among the loop's global->shared copies.
+ *
+ * Measures the full element width (dtype bits times lanes) so vector-typed
+ * buffers are not underreported and over-widened. Callers must only use this
+ * after IsPureGlobalToSharedCopyLoop() accepted the loop.
+ */
+int MinGlobalToSharedCopyElementBits(const Stmt &stmt) {
+  int min_bits = std::numeric_limits<int>::max();
+  PostOrderVisit(stmt, [&](const ObjectRef &node) {
+    if (const auto *store = node.as<BufferStoreNode>()) {
+      if (const auto *load = store->value.as<BufferLoadNode>()) {
+        int bits = load->buffer->dtype.bits() * load->buffer->dtype.lanes();
+        min_bits = std::min(min_bits, bits);
+      }
+    }
+  });
+  return min_bits;
+}
+
 } // anonymous namespace
+
+std::optional<int>
+ParallelOpNode::GlobalToSharedCopyMinElementBits(const For &root) const {
+  if (ContainsImpureCall(root->body) ||
+      !IsPureGlobalToSharedCopyLoop(root->body)) {
+    return std::nullopt;
+  }
+  return MinGlobalToSharedCopyElementBits(root->body);
+}
 
 /**
  * @brief Handle a parallel For node during traversal, collecting loop metadata.
@@ -996,6 +1075,16 @@ Fragment ParallelOpNode::ComputeLoopLayoutFromBuffer(
   return result;
 }
 
+/**
+ * @brief Compute the plan-based loop layout candidate.
+ *
+ * Determines a vectorized loop layout candidate from the maximum legal
+ * vectorize size, then adjusts it to the smallest-padding width for
+ * non-fragment loops. Pure global->shared copies with sub-32-bit dtypes are
+ * additionally coalesced to keep each active thread's chunk at or above
+ * 4 bytes, so the copy stays compatible with the cp.async minimum transfer
+ * width instead of silently falling back to a synchronous load.
+ */
 Fragment
 ParallelOpNode::ComputePlanCandidate(const LayoutInferArgs &layout_args) const {
   // Vectorize Size must be aware of the buffer_remap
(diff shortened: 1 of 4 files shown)


LeiWang1999:
ci failed cause of torch upgrade, please rebase into main to pick the pr https://github.com/tile-ai/tilelang/pull/3136 to fix the ci :)

LeiWang1999 (changes requested):
Thanks for the fix — the problem is real and the change verifiably works. I reproduced both sides locally on sm_90 hardware: on current main the reproducer silently falls back to a synchronous scalar copy even with an explicit `prefer_instruction="cp_async"`, and with this patch it emits predicated `tl::cp_async` (32 active threads x 4B) with bitwise-correct results. Copy / parallel / layout-inference suites pass locally as well.

Two things before this can merge:

**1. Please add regression tests.**
The reproducer in the PR description is already 90% of a host-side test (`compile_kernel_source` + the two assertions). Please add it under `testing/python/` (e.g. next to the copy or layout-inference tests), plus a negative case asserting that a loop containing an atomic call or a non-copy store keeps its original partition (no coalescing). Without tests nothing pins this behavior.

**2. Please restructure `MinGlobalToSharedCopyDtypeBits`.**
Functionally it checks out, but the shape doesn't meet the bar for this file:

- It fuses a predicate ("is this loop a pure g2s copy?") with an aggregation (min element bits) by threading three mutable flags through a `PostOrderVisit` lambda, with `0` overloaded to mean "no stores", "not a pure copy", and "has side effects". Split the two questions and return `std::optional<int>` instead of a sentinel.
- `32` appears three times (function name, `< 32`, `32 / bits`) but its real origin is cp.async's 4-byte minimum transfer width. Give it a named constant that lives with the cp.async machinery and reference it from the partitioner, so the two can't silently drift apart.
- `ParallelOpNode` already models the loop's buffer accesses (`indice_map_`). Consider making this a private method of the op that reuses that model, with the stmt walk reduced to the side-effect check that `indice_map_` can't answer.
- The blanket `CallNode` bail-out also trips on pure intrinsics inside index expressions (e.g. a shift builtin in address math), silently disabling coalescing for loops that are still pure copies. Conservative is fine, but this deserves either a narrower side-effect check or at least a comment acknowledging the limitation.

Happy to re-review once these land.

zhouyangye1076:
Thanks for the detailed review — restructured per your suggestions and added tests.          
                                                                                                  
   **Refactor of the dtype-bits helper**                                                        
   - Split the fused predicate+aggregation into three pieces: `ContainsImpureCall`              
     (side-effect check), `IsPureGlobalToSharedCopyLoop` (structural predicate:                 
     every store is a same-dtype global->shared copy), and                                      
     `MinGlobalToSharedCopyElementBits` (pure aggregation). They are composed by                
     a private `ParallelOpNode::GlobalToSharedCopyMinElementBits` which returns                 
     `std::optional<int>` — the sentinel 0 is gone.                                             
   - The bare 32 is replaced by named constants `kCPAsyncMinTransferBytes` /                    
     `kCPAsyncMinTransferBits`, defined in `src/op/utils.h` next to                             
     `IsValidCPAsyncTransferBytes` and referenced from the partitioner.                         
   - The blanket `CallNode` bail-out is replaced by a real purity check                         
     (`tirx::SideEffect > kPure`), so pure intrinsics in index arithmetic no                    
     longer disable coalescing. (Note: shift/floormod-style address math is                     
     already modeled as PrimExpr binary nodes, not CallNodes, so it was never                   
     affected; the narrowed check covers genuinely pure Calls.)                                 
   - On reusing `indice_map_`: I looked into it and it only models **fragment**                 
     buffers — `RecordBufferAccess` is called exclusively for                                   
     `IsFragmentBuffer`, so global/shared copies aren't in the map at all.                      
     Extending it would entangle fragment layout inference with an unrelated                    
     query, which felt out of proportion here. The logic still lives on the op                  
     as a private method as you suggested; the stmt walk is what answers both                   
     the pairing shape (store-value-is-load) and dtype equality, which the map                  
     can't answer even for fragments.                                                           
                                                                                                
   **Regression tests** (`testing/python/language/test_tilelang_language_copy.py`)              
   - Positive: the PR reproducer as a compile-only test — 128 threads copying 128               
     fp8 elements with `prefer_instruction="cp_async"` asserts `tl::cp_async` and               
     the `threadIdx.x < 32` predicate.                                                          
   - Negative: a bf16 atomic-add loop and a fp16 compute/store loop both keep                   
     their original one-element-per-thread partition (no coalescing guard).                     
   - All three are compile-only and guarded with the repo-standard                              
     `cuda_codegen_available()` skip, so they pass on builds without CUDA codegen               
     (e.g. ROCm-only).                                                                          
                                                                                                
   Verified locally: new tests pass, the previously broken CI cases (packed                     
   int4 copy, `float x` let-binding, bf16 atomic numerics) are unaffected,                      
   transform-level pipeline suites pass, and `pre-commit run --all-files` is                    
   clean.

PR #3132, opened 2026-09-02
[CUDA] Complete 16-bit bridges for CUDA-lowered unary math

Chennesxu:
## Summary

`T.sinh`, `T.cosh`, `T.tanh`, `T.atan`, `T.erf`, and `T.nearbyint` fail to compile on `float16` and `bfloat16`.

CUDA lowering emits half-style names for these operations, but no CUDA overload of those names accepts the `cutlass::half_t` and `cutlass::bfloat16_t` values TileLang emits. Most toolkits do not declare the names at all, and the overloads that do exist take native `__half` or `__nv_bfloat16`, which the CUTLASS wrappers have no implicit conversion to. These operations therefore fail on every toolkit:

```text
error: identifier "hsinh" is undefined
```
## Fix

Add `half_t` and `bfloat16_t` overloads for the six names, each evaluating in float32 and converting back — the same approach as the existing `htan` bridge (#2894).

This completes the set: every unary floating-point op that CUDA lowering already handles now has a 16-bit bridge. Verified by compiling all 22 of them in both `float16` and `bfloat16`: 44/44 emit the expected `h*` name and compile.

## Tests

Extended the existing 16-bit intrinsics kernel with the six ops.

Verified locally on sm_75 / CUDA 12.4:

* Touched file: 5 passed.
* Negative control (bridges reverted): fails only with the six undefined identifiers.
* Fast-math and sm_75 GEMM suites alongside the touched file: 141 passed.

## Summary

- Added CUDA 16-bit bridges for `sinh`, `cosh`, `tanh`, `atan`, `erf`, and `nearbyint`.
- Added `half_t` and `bfloat16_t` overloads for the corresponding `h*` CUDA names.
- Evaluated each operation in `float32`, then converted the result back to the source type.
- Expanded the CUDA intrinsic test kernel to cover all six operations.
- Validation compiled 22 unary floating-point operations for both 16-bit types. All 44 cases passed.
- Additional CUDA 12.4 `sm_75` test suites passed.

## C++ style / lint notes

- The PR changes C++ code in `src/tl_templates/cuda/common.h`.
- The “C++ API Style Audit (warning only)” CI step is relevant to the changed helper overloads.

### DIFF
--- a/src/tl_templates/cuda/common.h
+++ b/src/tl_templates/cuda/common.h
@@ -213,15 +213,63 @@ TL_PATCH TL_DEVICE bfloat16_t __hfma(const bfloat16_t x, const bfloat16_t y,
 #endif
 }
 
-// CUDA has no half-precision tangent intrinsic, but tangent lowering uses the
-// half-style `htan` name for 16-bit inputs. Evaluate in float32 and convert the
-// result back to the source type.
+// 16-bit lowering emits these h* names, but CUDA has no overload of them
+// that accepts the CUTLASS wrapper types. Evaluate in float32 and convert
+// the result back to the source type.
 TL_PATCH TL_DEVICE half_t htan(const half_t x) {
   return half_t(tanf(float(x)));
 }
 
 TL_PATCH TL_DEVICE bfloat16_t htan(const bfloat16_t x) {
   return bfloat16_t(tanf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t hsinh(const half_t x) {
+  return half_t(sinhf(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t hsinh(const bfloat16_t x) {
+  return bfloat16_t(sinhf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t hcosh(const half_t x) {
+  return half_t(coshf(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t hcosh(const bfloat16_t x) {
+  return bfloat16_t(coshf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t htanh(const half_t x) {
+  return half_t(tanhf(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t htanh(const bfloat16_t x) {
+  return bfloat16_t(tanhf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t hatan(const half_t x) {
+  return half_t(atanf(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t hatan(const bfloat16_t x) {
+  return bfloat16_t(atanf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t herf(const half_t x) {
+  return half_t(erff(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t herf(const bfloat16_t x) {
+  return bfloat16_t(erff(float(x)));
+}
+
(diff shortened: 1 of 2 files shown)


LeiWang1999:
ci failed cause of torch upgrade, please rebase into main to pick the pr https://github.com/tile-ai/tilelang/pull/3136 to fix the ci :)

src/tl_templates/cuda/common.h
```diff
@@ -224,6 +224,54 @@ TL_PATCH TL_DEVICE bfloat16_t htan(const bfloat16_t x) {
...
+TL_PATCH TL_DEVICE half_t hcosh(const half_t x) {
+  return half_t(coshf(float(x)));
+}
+
+TL_PATCH TL_DEVICE bfloat16_t hcosh(const bfloat16_t x) {
+  return bfloat16_t(coshf(float(x)));
+}
+
+TL_PATCH TL_DEVICE half_t htanh(const half_t x) {
+  return half_t(tanhf(float(x)));
```
LeiWang1999, line 244: Empirical note on toolkit collisions, since this is the one name in the set that CUDA *does* now declare natively: CUDA 12.8+ ships `htanh(__half)` / `htanh(__nv_bfloat16)`. Verified on CUDA 13.1 (headers declare it) that both overload sets coexist fine — a call with `half_t` picks this bridge by exact match over the user-defined conversion to `__half`, no ambiguity, and the test passes.

Two side effects worth knowing (neither blocking):
- On toolkits with the native `htanh`, this bridge still routes through float32 rather than delegating like the `hsin`/`hcos` bridges above do. That's the *more accurate* choice (fp32 tanh then one rounding), just not the fastest; a `CUDART_VERSION`-guarded delegation could be a later micro-optimization.
- The float32 fallback is universally correct on every arch/toolkit, which the native path isn't — so as written this is the right default.

LeiWang1999 (commented):
Reviewed with local verification on H100 + CUDA 13.1. **Correct, complete for its stated scope, and well-tested — LGTM.**

What I checked:

- **Premise**: `CUDAMath` in the CUDA intrinsic lowering unconditionally h-prefixes any fp16/bf16 math name, so the six names are exactly what codegen emits. Reproduced the failure with the header change reverted: `identifier \"hsinh\" is undefined` — on CUDA **13.1**, i.e. even the newest toolkit doesn't declare these (only `htanh` gained a native overload in 12.8+; see inline comment on the coexistence check).
- **Completeness claim**: cross-checked every op registered with the `CUDAMath` dispatcher against the bridges in `common.h` — after this PR all registered *unary* float ops (floor/ceil/trunc/fabs/round/nearbyint/exp/exp2/exp10/erf/log/log2/log10/tan/cos/cosh/sin/sinh/atan/tanh/sqrt/rsqrt) have 16-bit bridges. ✓
- **Tests**: the touched file passes 5/5 on H100 (fp16 + bf16, with numerics). Nice touch that `B`'s shape derives from `len(HALF_MATH_OPS)`, so the new rows can't go out of bounds, and `nearbyint → torch.round` is consistent with the existing `round → hrint` convention (both round-half-to-even; the inputs avoid ties).

One follow-up worth a tracking issue (out of scope here since the PR scopes itself to unary ops): the same failure class exists for **binary** ops — probed locally, `T.pow` on `float16` dies with `identifier \"hpow\" is undefined` via the same `CUDAMath` h-prefixing (`fmod` is likely in the same boat). A sibling PR adding `hpow`/`hfmod` bridges (or teaching `CUDAMath` to fall back to fp32 for names with no native/bridged overload) would close the class entirely.

LeiWang1999 (approved):

Issue #2267, opened 2026-05-25
[CI] Windows wheel build is much slower as ccache is not effectively used


LeiWang1999:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) that this hasn't already been reported. (comment there if it has.)

## Problem

The Windows dist/wheel build in the `Dist` workflow takes significantly longer  than the other platforms.

Example from run:
https://github.com/tile-ai/tilelang/actions/runs/26380636948

Build wheel step duration:

- macOS Metal: ~2m13s
- Linux CUDA x86_64: ~5m02s
- Linux CUDA arm64: ~5m01s
- Windows CUDA 13.0: ~39m54s

LeiWang1999:
would you mind taking a look? @sepcnt

sepcnt:
The Windows job appears to be ccache-thrashing because embedded MSVC PDB debug info makes the .obj files much larger, so I suggest increasing the Windows ccache size from 200MB to 1GB, which should keep enough objects cached and likely resolve the slowdown.

PR #3133, opened 2026-09-02
[CI] Fix slow Windows wheel builds by making ccache effective

LeiWang1999:
## Summary

The `Dist` workflow's Windows wheel job takes ~48 min while Linux takes ~9 min (see e.g. [run 33572214431](https://github.com/tile-ai/tilelang/actions/runs/33572214431)). Almost all of it is cibuildwheel recompiling all ~700 TVM + TileLang TUs with clang-cl on every run: the ccache hit rate was **8/698 (1.15%)** on Windows vs **678/714 (95%)** on Linux.

## Root causes

1. **PEP 517 build-isolation tempdir names leak into ccache's hash.** cibuildwheel's `build` frontend materializes each build into a randomly-named `Temp\build-env-XXXXXXXX` dir, which hosts the pip-provided CUDA cu13 / z3 headers. That path is preserved in the preprocessor output as `#line` directives, so ccache's preprocessed-mode result key changes on every run. The repo already ships the fix — `cmake/ccache_strip_pep517.py`, plugged in via ccache `prefix_command_cpp` — but:
   - CMakeLists.txt only activates it when a stable interpreter exists at `.venv/Scripts/python.exe`, and CI never created one (the configure log shows `ccache: no stable .venv python found`);
   - the wrapper's regex only matched uv's `.tmpXXX` naming, not the `build-env-XXX` naming pypa/build actually uses (nor random suffixes containing `_`, e.g. `build-env-gna_6ikk`).
2. **`max-size: 200MB` is smaller than one full build's output.** ccache evicted entries mid-build (522 cleanups, cache 99.96% full), so the cache could never converge to a full set of objects.

The command-line side is already safe: ccache 4.9 natively treats `-imsvc` / `-external:I` as preprocessor path options excluded from the cpp-mode hash, so the `#line` text was the only leak.

## Changes

- `cmake/ccache_strip_pep517.py`: widen normalization to `(?:\.tmp|build-env-|cibw-run-)[A-Za-z0-9_]+`. The rewritten text is only hashed, never compiled, so this cannot miscompile.
- `.github/workflows/dist.yml`: move `setup-uv` before the build step and create a repo-root `.venv` on Windows so the wrapper activates; raise ccache `max-size` to 2GB for the wheel jobs.
- `CMakeLists.txt`: suppress clang-cl's `-Wdelayed-template-parsing-in-cxx20` deprecation warning (feature-checked so older LLVM doesn't emit unknown-warning-option noise instead); it fired on every C++20 TU and flooded the logs.

## Expected effect / how to verify

The first Windows run after this lands still compiles cold (~45 min) and populates the cache. Subsequent runs should hit ~95% and drop "Build wheels" from ~44 min to well under 10 min. Check the `Post Setup ccache` stats and the `ccache: stripping PEP 517 .tmpXXX paths via ...` configure line (instead of `no stable .venv python found`).

Since this PR touches `dist.yml` / `CMakeLists.txt`, the Dist workflow runs on the PR itself — the second datapoint needs a re-run or a follow-up push.

## Summary

- Improved Windows wheel build caching by normalizing more PEP 517 temporary-directory patterns.
- Created a stable repository-root `.venv` before the Windows cibuildwheel step.
- Increased the wheel-job ccache limit from 200 MB to 2 GB.
- Added conditional suppression for the clang-cl C++20 delayed-template-parsing deprecation warning.
- Expected subsequent Windows builds to reach about 95% cache hits and complete in under 10 minutes.

## C++ style / lint notes

- The PR changes a C++ compiler warning in `CMakeLists.txt`.
- The change does not affect the “C++ API Style Audit (warning only)” CI step.
- The clang-cl warning suppression is conditional and avoids unknown-warning-option noise on unsupported toolchains.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -141,10 +141,35 @@ jobs:
           fetch-depth: 1
           submodules: recursive
 
+      - name: Setup Python and uv with caching
+        id: setup-uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: ${{ matrix.target.python_version }}
+          activate-environment: true
+
+      # CMakeLists.txt only plugs cmake/ccache_strip_pep517.py (normalizes the
+      # random PEP 517 build-env paths before ccache hashes the preprocessed
+      # output) in front of the preprocessor when a stable interpreter exists
+      # at .venv/Scripts/python.exe. Without it ccache misses on ~99% of
+      # lookups and the Windows wheel build recompiles all of TVM + TileLang
+      # from scratch (~40 min).
+      - name: Create stable .venv for ccache wrapper (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          if (-not (Test-Path ".venv\Scripts\python.exe")) {
+            uv venv .venv
+          }
+          & ".venv\Scripts\python.exe" --version
+
       - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1
         with:
-          max-size: "200MB"
+          # One full Windows build writes >200MB of compressed objects; a
+          # smaller cap makes ccache evict entries mid-build (522 cleanups
+          # observed at 200MB) so the cache never converges.
+          max-size: "2GB"
           create-symlink: ${{ runner.os != 'Windows' }}
           evict-old-files: "7d"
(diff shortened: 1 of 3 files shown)


LeiWang1999:
related issue: #2267

PR #2444, opened 2026-06-23
[Pipeline] Simplify async copy lowering

LeiWang1999:
## Summary

- Remove the obsolete full-function `LowerPTXAsyncCopy` pass and make async copy injection explicit at tile-op lowering call sites.
- Drop software-pipeline-only `async_scope` and `pipeline_mvb_*` attr plumbing now that tile-op pipeline rewriting handles multi-version state directly.
- Fix vectorization analysis for predicated tail copies so guarded residual tiles do not become invalid 2-byte `cp.async` operations.

## Changes

- Simplified CUDA and ROCm async copy injectors by removing `enable_auto_async_copy` and `async_scope` handling; call sites now decide whether to invoke injection.
- Removed `pipeline_context_num_stages` and `pipeline_mvb_*` emission and late consumption from `PipelineRewriter` and `MultiVersionBufferRewriter`.
- Added condition-aware vectorization sizing for `if_then_else` predicates and an issue 2365 regression case.
- Improved `cp.async` byte-width diagnostics to include the offending call.

## Validation

- `pre-commit run --all-files`
- `cmake --build build -j$(nproc)`
- `python -m pytest testing/python/transform/test_tilelang_transform_Inject_software_pipeline.py -q`
- `python -m pytest testing/python/transform/test_tilelang_transform_lower_tile_op.py -q`
- `python testing/python/issue/test_tilelang_issue_2365.py`

## Notes

- `tl.pipeline_mbar_phase_expr` remains in place for tile-op lowering; only the obsolete MVB/context attrs are removed.

## Summary

This pull request simplifies the async copy lowering pipeline in the tilelang compiler by removing the global `LowerPTXAsyncCopy` pass and making async copy injection explicit at call sites. The main changes include:

### Core Changes

**Removed global pass**: The `tl.cuda.transform.LowerPTXAsyncCopy` pass is eliminated (source file removed), moving async copy decision-making to lower-level call sites via `LowerTileOpPass`.

**Simplified async copy injector APIs**: Both `InjectPTXAsyncCopy` (CUDA) and `InjectROCmAsyncCopy` (ROCm) have their signatures simplified by removing the `enable_auto_async_copy` parameter. These helpers now only take `body` and `async_without_async_commit_wait` as control inputs, with documentation updated to clarify that callers decide whether to invoke injection before calling the helper.

**Removed software-pipeline-specific attributes**: Obsolete attributes are removed as pipeline stage “context” plumbing is dropped, including:
- `async_scope`
[...]

PR #3137, opened 2026-09-02
[BugFix] Allow async copies in conditionally executed pipeline stages 

zhouyangye1076:
PipelineStageAnalyzer rejected any copy stage whose statement may be conditionally executed, so a pipelined global->shared copy guarded by a runtime condition (e.g. 'if (cond) { T.copy; T.gemm }') silently stayed a synchronous load and the whole pipeline fell back to __syncthreads.

The rejection is unnecessary: the injected cp.async carries the branch predicate, and the pipeline lowering already places commit_group outside the branch and the wait before the conditional consumer, with the gemm re-guarded by the original predicate, so skipped iterations stay correct. Drop the conditional_execution check from IsAsyncProducerCandidate and let conditional copy stages use cp.async.

### Problem                                                                                                                           
                                                                                                                                           
   When a pipelined loop's copy is guarded by a runtime condition:                                                                       
                                                                                                                                         
   ```python                                                                                                                             
   for k in T.Pipelined(T.ceildiv(K, BK), num_stages=2):                                                                                 
       if k < limit:                          # dynamic condition                                                                        
           T.copy(A[k * BK : (k + 1) * BK], a_sh)   # global -> shared                                                                   
           T.gemm(a_sh, b_sh, c_loc)                                                                                                     
   ```                                                                                                                                   
                                                                                                                                         
   `PipelineStageAnalyzer::IsAsyncProducerCandidate` rejects any copy stage whose                                                        
   statement `MayBeConditionallyExecuted`, so the copy silently stays a                                                                  
   synchronous global load and the whole pipeline degenerates to                                                                         
   `__syncthreads`-based staging. The condition is legitimate for bounds-style                                                           
   tiling, and the ideal lowering is well-defined:                                                                                       
   `if (cond) { cp.async } commit; if (cond) { wait, gemm }`.                                                                            
                                                                                                                                         
   ### Reproducer (compile-only, no GPU needed)                                                                                          
                                                                                                                                         
   ```python                                                                                                                             
   import tilelang                                                                                                                       
   import tilelang.language as T                                                                                                         
   from tilelang.tools.compile_only import compile_kernel_source                                                                         
                                                                                                                                         
   M = N = K = 256                                                                                                                       
   BM, BN, BK = 64, 64, 32                                                                                                               
                                                                                                                                         
                                                                                                                                         
   @T.prim_func                                                                                                                          
   def kernel(                                                                                                                           
       A: T.Tensor((M, K), T.float16),                                                                                                   
       B: T.Tensor((K, N), T.float16),                                                                                                   
       C: T.Tensor((M, N), T.float16),                                                                                                   
       limit: T.int32,                                                                                                                   
   ):                                                                                                                                    
           a_sh = T.alloc_shared((BM, BK), T.float16)                                                                                    
           b_sh = T.alloc_shared((BK, BN), T.float16)                                                                                    
           c_loc = T.alloc_fragment((BM, BN), T.float32)                                                                                 
           T.clear(c_loc)                                                                                                                
           for k in T.Pipelined(T.ceildiv(K, BK), num_stages=2):                                                                         
               if k < limit:                                                                                                             
                   T.copy(A[by * BM:(by + 1) * BM, k * BK:(k + 1) * BK], a_sh)                                                           
                   T.copy(B[k * BK:(k + 1) * BK, bx * BN:(bx + 1) * BN], b_sh)                                                           
                   T.gemm(a_sh, b_sh, c_loc)                                                                                             
           T.copy(c_loc, C[by * BM:(by + 1) * BM, bx * BN:(bx + 1) * BN])                                                                
                                                                                                                                         
                                                                                                                                         
   src = compile_kernel_source(kernel, "cuda -arch=sm_80")                                                                               
   assert "tl::cp_async" in src  # used to silently stay a synchronous copy                                                              
   ```                                                                                                                                   
                                                                                                                                         
   ### Generated code, steady-state loop (index arithmetic abridged)                                                                     
                                                                                                                                         
   **Before** — conditional stage excluded from async, everything synchronous:                                                           
                                                                                                                                         
   ```cuda                                                                                                                               
   for (int k = 0; k < 7; ++k) {                                                                                                         
     __syncthreads();                                                                                                                    
     if ((k + 1) < limit) {                                                                                                              
       *(uint4*)(a_sh_smem + ... ) = *(uint4*)(A + ...);   // sync vector copy                                                           
       *(uint4*)(b_sh_smem + ... ) = *(uint4*)(B + ...);                                                                                 
     }                                                                                                                                   
     __syncthreads();                                                                                                                    
     if (k < limit) { /* ldmatrix + mma */ }                                                                                             
   }                                                                                                                                     
   ```                                                                                                                                   
                                                                                                                                         
   **After** — cp.async keeps the branch predicate, commit hoisted out of the                                                            
   branch, wait before the conditional consumer:                                                                                         
                                                                                                                                         
   ```cuda                                                                                                                               
   if (0 < limit) { tl::cp_async_gs<16>(...); }  // prologue                                                                             
   tl::cp_async_commit();                                                                                                                
   for (int k = 0; k < 7; ++k) {                                                                                                         
     __syncthreads();                                                                                                                    
     if ((k + 1) < limit) {                                                                                                              
       tl::cp_async_gs<16>((&a_sh[...]), (&A[...]));                                                                                     
       tl::cp_async_gs<16>((&b_sh[...]), (&B[...]));                                                                                     
     }                                                                                                                                   
     tl::cp_async_commit();                       // outside the if: always runs                                                         
     tl::cp_async_wait<1>();                                                                                                             
     __syncthreads();                                                                                                                    
     if (k < limit) { /* ldmatrix + mma */ }                                                                                             
   }                                                                                                                                     
   tl::cp_async_wait<0>();                        // drain                                                                               
   __syncthreads();                                                                                                                      
   if (7 < limit) { /* epilogue gemm */ }                                                                                                
   ```                                                                                                                                   
                                                                                                                                         
   ### Fix                                                                                                                               
                                                                                                                                         
   Drop the `conditional_execution` early-return in                                                                                      
   `IsAsyncProducerCandidate` (`src/transform/pipeline_planning.cc`). No new pass                                                        
   is needed: the injected cp.async carries the branch predicate, and the                                                                
   existing pipeline lowering already places `commit_group` *outside* the branch                                                         
   (so skipped iterations still commit a consistent, possibly empty group) and                                                           
   the wait before the consumer guarded by the original predicate. TMA copies                                                            
   and non-copy stages remain excluded.                                                                                                  
                                                                                                                                         
   ### Validation                                                                                                                        
                                                                                                                                         
   - Reproducer above emits `tl::cp_async_gs<16>` with correct commit/wait                                                               
     placement; checked for `num_stages=2` and `num_stages=3`                                                                            
     (prologue fills stages, steady state `wait<2>`, drain `wait<1>`/`wait<0>`).                                                         
   - Transform-level pipeline suites:                                                                                                    
     `test_tilelang_transform_Inject_software_pipeline.py`,                                                                              
     `test_tilelang_transform_pipeline_planning.py`,                                                                                     
     `test_tilelang_transform_pipeline_barrier_ownership.py`                                                                             
     — 34 passed, 3 skipped.                                                                                                             
   - `pre-commit run --all-files` is clean.                                                                                              
   - Note: device-run numerics should be validated by CI on an SM80+ GPU;                                                                
     the verification here is at the codegen/TIR level.

## Summary

- Allow conditionally executed global-to-shared copy stages to use `cp.async`.
- Keep conditional TMA copies and non-copy stages excluded.
- Emit commit groups outside conditional branches.
- Wait before conditional consumers while preserving consumer predicates.
- Avoid unnecessary `__syncthreads` fallback for runtime-guarded copies.

## Validation

- Compile-only checks passed for two-stage and three-stage pipeline reproducers.
- 34 pipeline tests passed.
- 3 pipeline tests were skipped.
- Pre-commit checks passed.
- GPU numerical validation remains deferred to CI on SM80+ hardware.

## C++ style / lint notes

- The change touches C++ implementation code but does not alter rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) remains relevant to CI.
- This change introduces no public API changes or known warning-only style issues.
- Advisory TLCPP003/TLCPP004 findings do not indicate correctness or build issues.

### DIFF
--- a/src/transform/pipeline_planning.cc
+++ b/src/transform/pipeline_planning.cc
@@ -494,9 +494,12 @@ public:
   }
 
   bool IsAsyncProducerCandidate(const PipelineStageInfo &pinfo) const {
-    if (pinfo.conditional_execution) {
-      return false;
-    }
+    // Conditionally executed copies are still eligible for cp.async: commit is
+    // emitted outside the user branch (a skipped iteration commits an empty
+    // group, which completes immediately), so group counting stays consistent.
+    // Conditional TMA remains excluded (tma_copy is never set for conditional
+    // stages): mbarrier arrive/wait must pair up, and a skipped arrive with a
+    // non-skipped wait deadlocks (see #2002, blocksparse_gemm launch failures).
     if (pinfo.IsTmaCopy()) {
       return false;
     }
@@ -602,6 +605,10 @@ public:
   void ClassifyCopyLikeStage(const Stmt &stmt, PipelineStageInfo *pinfo) const {
     ICHECK(pinfo != nullptr);
     if (pinfo->conditional_execution) {
+      // Intentional exclusion (safe, not policy): conditional plain copies
+      // are already classified by the pure-copy path in MakePipelineStageInfo,
+      // so this fallback only gates conditional im2col, whose pipeline-managed
+      // cp.async path has no test coverage yet.
       return;
     }
 
@@ -942,6 +949,10 @@ public:
       ClassifyCopyLikeStage(pipeline_stmts[i], &pinfo);
       pinfo.order = static_cast<int>(order_array[i]->value);
       pinfo.stage = static_cast<int>(stage_array[i]->value);
+      // The conditional check here is an intentional conservative exclusion:
+      // this legacy heuristic promotes non-pure (mixed compute+copy)
+      // statements to copy stages, and extending that guess to conditionally
+      // executed statements is not covered by any use case.
       if (!pinfo.IsCopyStage() && !pinfo.conditional_execution &&
           pinfo.stage == 0) {
         bool reads_global = false;
--- a/testing/python/transform/test_tilelang_transform_pipeline_planning.py
+++ b/testing/python/transform/test_tilelang_transform_pipeline_planning.py
@@ -1,7 +1,10 @@
 from tilelang import tvm as tvm
+import pytest
 import tilelang as tl
+import tilelang.testing
 from tilelang.backend.target import determine_target
 import tilelang.language as T
+from tilelang.tools.compile_only import compile_kernel_source, cuda_codegen_available
 from tvm.tirx.stmt_functor import post_order_visit
 
 auto_target = tvm.target.Target(determine_target("auto"))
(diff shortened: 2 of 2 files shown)


LeiWang1999:
ci failed cause of torch upgrade, please rebase into main to pick the pr #3136 to fix the ci :)

src/transform/pipeline_planning.cc
```diff
@@ -494,9 +494,9 @@ class PipelineStageAnalyzer {
   }
 
   bool IsAsyncProducerCandidate(const PipelineStageInfo &pinfo) const {
-    if (pinfo.conditional_execution) {
-      return false;
-    }
+    // Conditionally executed copies are still eligible for async: the injected
+    // cp.async carries the branch predicate, and the pipeline commit/wait
+    // placement keeps the synchronization correct for the skipped iterations.
```
LeiWang1999: **Verified this locally — the claim in the PR description holds, and the reasoning can be strengthened: the commit placement is *structural*, not incidental.**

`AsyncCommitWaitAttrLowerer` lowers `async_commit_queue_scope` as `SeqStmt{body, commit}` (inject_pipeline.cc:369-377), and the user's `IfThenElse` lives *inside* the group body, so the commit can never end up under the user's branch. Every iteration therefore commits exactly one (possibly empty) group regardless of the condition, which is what keeps `wait<N>` group counting consistent. The wait and both `__syncthreads()` are likewise emitted outside the conditional, so there's no divergent-barrier risk either.

Also confirmed this check was the single gate: `copy_stage` is already set for conditional stages via the `pure_copy_stage` path in `MakePipelineStageInfo` (`IsPureCopyStmt` uses `PostOrderVisit` and sees through the `If`), and `tma_copy` stays `false` for conditional stages (the `!pinfo.conditional_execution` term at line 990), so TMA remains excluded as the description says.

Verification on an H100 (details in the review body): sm80 codegen inspection for ns=2/3, device numerics on the cp.async path (`tl.disable_warp_specialized=True`) sweeping `limit` ∈ {0, 1, 3, full, past-the-end} for ns=2/3 plus a mixed conditional/unconditional-copy pipeline — all exact against a truncated-K torch reference — and 129 transform tests (the three pipeline suites + the three WS suites) pass with the change.

LeiWang1999 (commented):
Reviewed with local verification on H100 (sm90) and sm80 codegen inspection. **The change is correct and well-scoped** — details in the inline comment; summary of what was checked:

- **sm80 compile-only, ns=2/3**: conditional copies become `tl::cp_async_gs`, the commit is unconditional (outside the user branch), steady-state/drain waits are `{1,0}` and `{2,1,0}`, and no synchronous `uint4` fallback copies remain.
- **H100 device numerics** on the cp.async path (`tl.disable_warp_specialized=True`): the PR's kernel shape with `limit` swept over {0, 1, 3, full, past-the-end} at runtime, `num_stages` ∈ {2, 3}, plus a mixed pipeline (conditional A-copy + unconditional B-copy feeding one gemm) — all bit-exact against a truncated-K torch reference. The default Hopper WS/TMA path was run as a control and is unaffected (WS runs before planning and does its own copy classification).
- **Regression**: the three pipeline transform suites plus the three warp-specialization suites — 129 passed.
- **Wait-relaxation safety**: `DeterministicNoWaitCommitAnalyzer::Analyze` returns `Unknown` for `IfThenElse`, so commits from conditional stages simply don't count toward the relaxation lower bound — conservative direction only, never under-waiting.

Two non-blocking follow-ups:

1. **Please add the description's reproducer as a regression test** — a compile-only test asserting `tl::cp_async` presence (plus commit/wait placement) in e.g. `test_tilelang_transform_pipeline_planning.py` would keep this from regressing; a device-side numeric test with a runtime `limit` sweep would be even better. Right now the PR ships no test, and this exact check has been re-introduced once before (#2002 → #2444 kept it), so a test pinning the new behavior is worth it.
2. **Leftover asymmetries**: `ClassifyCopyLikeStage` still early-returns on `conditional_execution` (line 604), and the legacy annotation path checks it too (line 945) — so a conditional `im2col` and legacy-annotated loops still fall back to sync. That's safe, just inconsistent with the new policy; fine to leave, but a short comment there (or a follow-up) would help the next reader understand the remaining exclusions are intentional.

LeiWang1999:
@regression-perf

LeiWang1999:
**Some archaeology on why this check existed in the first place** (follow-up to my review, for the record):

The `conditional_execution` gate was added in #2002 (April), a large stabilization PR that was fighting real launch failures at the time. The motivating bug is documented in that PR's commit history:

> PipelinePlanning and InjectSoftwarePipeline would then generate non-WS TMA pipeline code with broken barrier phase tracking for conditional pipeline bodies (**barrier waits outside conditionals cause deadlocks when the condition is false**). Fixes CUDA_ERROR_LAUNCH_FAILED in blocksparse_gemm and related TMA kernels with conditional loop bodies.

So the fire was **TMA/mbarrier-specific**: mbarrier `arrive`/`wait` must pair up, and with the producer's arrive inside the branch and the consumer's wait outside, a false condition means nobody arrives and the wait deadlocks. The fix at the time was layered (strip pipeline annotations on WS fallback, guarded phase counters in the WS pass) plus the blanket "conditional ⇒ fully synchronous" in planning — which swept cp.async along even though its commit/wait semantics are group-counting, not paired: a skipped iteration still commits an (empty) group, and empty groups complete immediately. Judging by the #2002-era `inject_pipeline.cc`, the commit placement was already structural back then too (`SeqStmt{body, commit}` with the user's branch inside the body), so the cp.async half of the blanket was likely conservatism under fire rather than a known miscompile. #2444 later carried the check forward unchanged — there was never a test pinning down why it existed.

This PR removes exactly the safe half: conditional stages still get `tma_copy = false` in `MakePipelineStageInfo`, so the conditional+TMA combination that actually deadlocked remains excluded, and conditional copies only ever take the group-counting cp.async path.

Given this history, I'd suggest extending the new comment slightly to record the distinction, e.g.:

```cpp
// Conditionally executed copies are still eligible for cp.async: commit is
// emitted outside the user branch (a skipped iteration commits an empty
// group, which completes immediately), so group counting stays consistent.
// Conditional TMA remains excluded (tma_copy is never set for conditional
// stages): mbarrier arrive/wait must pair up, and a skipped arrive with a
// non-skipped wait deadlocks (see #2002, blocksparse_gemm launch failures).
```

That way the next person auditing this code knows which half of the old blanket was load-bearing.

zhouyangye1076:
Thanks for the thorough review and the archaeology — that history also explains                                                                                                                     
   why no test ever pinned the check down. All feedback addressed:                                                                                                                                     
                                                                                                                                                                                                       
   **Comment updated with the load-bearing distinction**                                                                                                                                               
                                                                                                                                                                                                       
   Adopted your suggested wording in `IsAsyncProducerCandidate`, recording that                                                                                                                        
   cp.async is group-counting (a skipped iteration commits an empty group, which                                                                                                                       
   completes immediately) while conditional TMA remains excluded because mbarrier                                                                                                                      
   arrive/wait must pair up — the #2002 blocksparse_gemm deadlock:                                                                                                                                     
                                                                                                                                                                                                       
       // Conditionally executed copies are still eligible for cp.async: commit is                                                                                                                     
       // emitted outside the user branch (a skipped iteration commits an empty                                                                                                                        
       // group, which completes immediately), so group counting stays consistent.                                                                                                                     
       // Conditional TMA remains excluded (tma_copy is never set for conditional                                                                                                                      
       // stages): mbarrier arrive/wait must pair up, and a skipped arrive with a                                                                                                                      
       // non-skipped wait deadlocks (see #2002, blocksparse_gemm launch failures).                                                                                                                    
                                                                                                                                                                                                       
   **Regression tests** (`testing/python/transform/test_tilelang_transform_pipeline_planning.py`)                                                                                                      
                                                                                                                                                                                                       
   - Compile-only codegen test for `num_stages` ∈ {2, 3}: asserts `tl::cp_async`,                                                                                                                      
     unconditional `tl::cp_async_commit()`, steady-state `wait<1>`/`wait<2>`, and                                                                                                                      
     drain `wait<0>`. Following your structural argument, it also brace-matches                                                                                                                        
     the producer `if ((k + n) < limit) { ... }` block and asserts the statement                                                                                                                       
     immediately after its closing brace is exactly `tl::cp_async_commit();` —                                                                                                                         
     pinning the "commit outside the user branch" invariant itself, not just the                                                                                                                       
     presence of cp.async.                                                                                                                                                                             
   - Device numerics test sweeping `limit` over {0, 1, 3, full, past-the-end}                                                                                                                          
     with warp specialization disabled, checked against a truncated-K reference.                                                                                                                       
     (You verified the same sweep on H100; the test now lives in-tree so the                                                                                                                           
     behavior can't regress, and the numerics pass on our local device as well.)                                                                                                                       
                                                                                                                                                                                                       
   **Leftover asymmetries** — kept the gates, documented intent at both sites:                                                                                                                         
                                                                                                                                                                                                       
   - `ClassifyCopyLikeStage`: the early return now notes that conditional plain                                                                                                                        
     copies are already classified by the pure-copy path, so this fallback only                                                                                                                        
     gates conditional im2col, whose pipeline-managed cp.async path has no test                                                                                                                        
     coverage yet (possible follow-up).                                                                                                                                                                
   - Legacy annotation path: noted that promoting *non-pure* conditional                                                                                                                               
     statements to copy stages is a heuristic guess, and extending it to                                                                                                                               
     conditionally executed statements has no use case — a deliberate                                                                                                                                  
     conservative exclusion.                                                                                                                                                                           
                                                                                                                                                                                                       
   **Drive-by fix:** the new codegen test reproduces on platforms without CUDA                                                                                                                         
   codegen FFI (the ROCm job failure you may have seen on the previous push) —                                                                                                                         
   it now carries the repo-standard `cuda_codegen_available()` skip, same as the                                                                                                                       
   `compile_only` tool tests.                                                                                                                                                                          
                                                                                                                                                                                                       
   Local verification: 37 passed / 3 skipped across the three pipeline                                                                                                                                 
   transform suites, numerics sweep passes on-device, `pre-commit` clean.

PR #3139, opened 2026-09-02
[CUDA] Add async-proxy fences for sparse MMA intrinsics

### LINKED ISSUE #2634: [BUG][Fuzzer][wrong-code] Hopper `T.gemm_sp` emits code that omits the ISA-required `fence.proxy.async` before the sparse WGMMA — silently reads stale operand data
### Required prerequisites

0.1.11 (source at `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`). Also live on `main` (`1e075149`).

### System information

NVIDIA H100 80GB HBM3 (sm_90), CUDA 13.0, tilelang 0.1.11. `fence.proxy.async` is a Hopper construct (`InjectFenceProxy` is gated on `TargetHasBulkCopy`, i.e. sm_90+), so the defect is sm_90-specific. Blackwell (sm_100) is affected by inspection — the blockscaled sparse builtin `ptx_tcgen05_mma_blockscaled_ss` has the same omission — but I did not run it.

### Problem description

**`T.gemm_sp` emits code that omits the `fence.proxy.async` the PTX proxy model requires** between a generic-proxy write to a shared WGMMA operand tile and the async-proxy `wgmma.mma_async.sp` read of that same tile. On the identical kernel, the dense `T.gemm` emits the fence and the sparse `T.gemm_sp` emits **zero**. Per PTX ISA §9.7.16.7.1 — *"An async proxy fence must be used to establish an ordering between prior writes to shared memory matrices and subsequent reads of the same matrices in a `wgmma.mma_async` instruction"* — the generated sparse code is **incorrect by the documented contract**: the WGMMA is permitted to read stale/torn operand data, a silent wrong result, with nothing ordering the two.

This is a wrong-code defect in the emitted PTX/SASS, established at the code level (counts + adjacency below) and grounded in the ISA. I did not force a divergent value out of `T.gemm_sp` itself on the tested H100 — an incidental `bar.sync` (the pipeline `__syncthreads()`) happens to drain the write on that schedule — but that barrier is **not** the ISA-required ordering (§8.7: it is a different proxy, so it establishes no happens-before edge), and a minimal control carrying the exact "`bar.sync` present, proxy fence absent" fingerprint **does** race the silicon (see Additional context).

The two kernels below are identical except `T.gemm` vs `T.gemm_sp`; both put a `B_s[i,j] += 3` generic store right before the matmul. `T.gemm` emits one `fence.proxy.async`; `T.gemm_sp` emits zero.

<details><summary>Emitted PTX and SASS — the required fence is present for dense, absent for sparse (H100, sm_90, CUDA 13.0)</summary>

Counts (grep of the emitted code): CUDA `fence_proxy_async` dense **1** / sparse **0**; PTX `fence.proxy.async` dense **1** / sparse **0**; SASS main-loop proxy fence (`MEMBAR.ALL.CTA` + `FENCE.VIEW.ASYNC.S` right after the generic operand store) dense **present** / sparse **absent** (both kernels share one *prologue* `FENCE.VIEW.ASYNC.S`; the load-bearing difference is the main-loop one).

```
=== PTX, dense (proxy fence PRESENT) ===
    wgmma.fence.sync.aligned;                 // register/accumulator fence — NOT the proxy fence
    fence.proxy.async.shared::cta;            // <-- the generic->async-proxy fence
    wgmma.mma_async.sync.aligned.m64n128k16.f32.f16.f16 {...};

=== PTX, sparse (proxy fence ABSENT) ===
    wgmma.fence.sync.aligned;                 // register fence only
    wgmma.mma_async.sp.sync.aligned.m64n128k32.f32.f16.f16 {...};   // async read, unfenced
    // the generic operand store st.shared.v4.u32 precedes this with nothing ordering it

=== SASS, dense: STS.128 -> MEMBAR.ALL.CTA ; FENCE.VIEW.ASYNC.S ; WARPGROUP.ARRIVE ; HGMMA.64x128x16.F32
=== SASS, sparse: STS.128 -> BAR.SYNC.DEFER_BLOCKING ; WARPGROUP.ARRIVE ; HGMMA.SP.64x128x32.F32
       (full scan of the sparse store->WGMMA.SP window: zero MEMBAR / FENCE.VIEW / DEPBAR)
```

The `wgmma.fence.sync.aligned` present in **both** is the accumulator/register fence, not the generic→async-proxy fence — only dense additionally emits `fence.proxy.async`. In SASS, where dense has `FENCE.VIEW.ASYNC.S`, sparse has only a `BAR.SYNC` — a barrier, which the PTX memory model (§8.7) does not accept as a substitute because it does not change an access's proxy.

</details>

### Reproducible example code

```python
from tilelang.utils.sparse import get_e_factor

M = N = K = 256; bM = bN = 128; bK = 64
e_dtype = "uint8"; e_factor = get_e_factor(T.float16, e_dtype)

def dense():                                              # CONTROL: gets the fence
    @T.prim_func
    def main(A: T.Tensor((M, K), T.float16), B: T.Tensor((K, N), T.float16),
        with T.Kernel(T.ceildiv(N, bN), T.ceildiv(M, bM), threads=128) as (bx, by):
            A_s = T.alloc_shared((bM, bK), T.float16); B_s = T.alloc_shared((bK, bN), T.float16)
            C_l = T.alloc_fragment((bM, bN), "float32"); T.clear(C_l)
            for k in T.Pipelined(T.ceildiv(K, bK), num_stages=2):
                T.copy(A[by * bM, k * bK], A_s); T.copy(B[k * bK, bx * bN], B_s)
                for i, j in T.Parallel(bK, bN):
                    B_s[i, j] = B_s[i, j] + T.float16(3)   # generic-proxy write to operand tile
                T.gemm(A_s, B_s, C_l)
            T.copy(C_l, C[by * bM, bx * bN])
    return main

def sparse():                                             # BUG: no fence
    @T.prim_func
    def main(A_sp: T.Tensor((M, K // 2), T.float16), E: T.Tensor((M, K // e_factor), e_dtype),
             B: T.Tensor((K, N), T.float16), C: T.Tensor((M, N), "float32")):
        with T.Kernel(T.ceildiv(N, bN), T.ceildiv(M, bM), threads=128) as (bx, by):
            A_s = T.alloc_shared((bM, bK // 2), T.float16); E_s = T.alloc_shared((bM, bK // e_factor), e_dtype)
            B_s = T.alloc_shared((bK, bN), T.float16); C_l = T.alloc_fragment((bM, bN), "float32"); T.clear(C_l)
            for k in T.Pipelined(T.ceildiv(K, bK), num_stages=2):
                T.copy(A_sp[by * bM, k * bK // 2], A_s); T.copy(E[by * bM, k * bK // e_factor], E_s)
                T.copy(B[k * bK, bx * bN], B_s)
                for i, j in T.Parallel(bK, bN):
                    B_s[i, j] = B_s[i, j] + T.float16(3)   # same generic write
                T.gemm_sp(A_s, E_s, B_s, C_l, transpose_A=False, transpose_E=False, transpose_B=False)
            T.copy(C_l, C[by * bM, bx * bN])
    return main

for name, fn, idx in [("DENSE T.gemm", dense, 2), ("SPARSE T.gemm_sp", sparse, 3)]:
    k = tilelang.compile(fn(), out_idx=[idx])
    print(f"{name}: fence.proxy.async count = {k.get_kernel_source().count('fence_proxy_async')}")
# DENSE  T.gemm:    fence.proxy.async count = 1
# SPARSE T.gemm_sp: fence.proxy.async count = 0   <-- missing
```

### Traceback

No traceback — both kernels compile and run to completion. The defect is missing memory-ordering in the emitted code: the generated sparse PTX/SASS omits the `fence.proxy.async` the ISA requires, so the sparse WGMMA is permitted to read a stale/torn operand — a silent wrong result whenever the (absent) ordering actually matters.

### Expected behavior

`T.gemm_sp` should emit the same `fence.proxy.async` that `T.gemm` emits when a generic-proxy shared write precedes it — i.e. the sparse WGMMA should be treated as an async-proxy op by `InjectFenceProxy`, exactly like the dense WGMMA already is. cutlass emits this fence for the same generic→async-proxy transition: [`cutlass/arch/barrier.h#L715-L721`](https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/include/cutlass/arch/barrier.h#L715-L721) (`fence_view_async_shared()` = `fence.proxy.async.shared::cta`) and [`cute/arch/copy_sm90_tma.hpp#L1214-L1217`](https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/include/cute/arch/copy_sm90_tma.hpp#L1214-L1217) (`tma_store_fence()`). The tilelang pass already encodes this rule for the dense path, so the sparse path being unfenced looks like an omission in the intrinsic list rather than an intentional difference.

### Additional context

**Root cause.** `InjectFenceProxy` does not recognize the sparse WGMMA builtins as async-proxy operations, so it omits the `fence.proxy.async` before a generic write that precedes a sparse matmul.

<details><summary>Mechanism — the sparse variants are missing from the async-intrinsic list</summary>

The pass classifies calls via `IsAsyncIntrinsic` ([`inject_fence_proxy.cc#L99-L106`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/inject_fence_proxy.cc#L99-L106)). That list contains the dense WGMMA builtins `ptx_wgmma_ss`/`ptx_wgmma_rs` and the dense tcgen05 ops, but **omits the sparse variants** `ptx_wgmma_sp_ss`, `ptx_wgmma_sp_rs`, and `ptx_tcgen05_mma_blockscaled_ss` (all public builtins). `T.gemm`/`T.gemm_sp` are lowered by `LowerTileOp` ([`pipeline.py#L117`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/pipeline.py#L117)) **before** `InjectFenceProxy` runs ([`pipeline.py#L229`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/cuda/pipeline.py#L229)): dense lowers to `ptx_wgmma_ss` (in the list → fenced), sparse lowers to `ptx_wgmma_sp_ss` (not in the list → not fenced). The default for unlisted intrinsics is now fail-open (`return ProxyEvent::kNone`), so the sparse op slips through with no fence. `T.gemm_sp` is a public op exported into the `T` namespace.

</details>

**Suggested fix (one line).** Add `ptx_wgmma_sp_ss()`, `ptx_wgmma_sp_rs()`, and `ptx_tcgen05_mma_blockscaled_ss()` to the `IsAsyncIntrinsic` disjunction, alongside their dense siblings. All three C++ symbols already exist. After adding them, the sparse repro emits `fence.proxy.async count = 1`, matching the dense control — restoring the ISA-required ordering the emitted code was missing.

**Wrong-code confirmed in the emitted code; runtime symptom masked on the tested silicon.** The defect is confirmed at the level a `[wrong-code]` classification requires — the generated PTX/SASS omits the `fence.proxy.async` that PTX ISA §9.7.16.7.1 says *must* order a generic write to a shared WGMMA operand ahead of the async `wgmma.mma_async.sp` read; dense emits it, sparse does not (counts + adjacency in Problem description). I did **not** force a divergent numerical result out of `T.gemm_sp` itself (~29,000 iterations matched a manual-`T.fence_proxy_async()` control bit-exact), because an incidental partial `bar.sync` (the pipeline `__syncthreads()`) drains the generic write on this schedule. That `bar.sync` is not the ISA-required ordering — per PTX §8.7, two operations are morally strong only if performed via the *same* proxy, so a generic-proxy barrier establishes no ordering against an async-proxy read — so it protects neither other Hopper silicon/schedules, nor other tile shapes, nor Blackwell, nor any refactor that moves or removes it. The code is wrong; the current schedule merely masks the symptom.

<details><summary>Positive control — the silicon <b>does</b> race when the drain is truly absent</summary>

On a minimally-derived single-warpgroup WGMMA kernel, three variants differ only in the ordering primitive between the generic operand write and the async read: **A** = nothing, **B** = `fence.proxy.async` (correct golden), **C** = `bar.sync` only, fence removed (the exact `T.gemm_sp` fingerprint). Racing C against the fenced golden B on identical fp16 inputs under a concurrent spoiler kernel, on H100 sm_90:

- 100k iterations, seed 0x1001: **2** divergences, worst |diff| **30.9**;
- 1M iterations, seed 0xBEEF (different GPU): **30** divergences, worst |diff| **31.4**; a captured launch had 512/8192 output elements wrong (a torn/partial operand read);
- fence-vs-fence control (B vs B, same input): **0** divergences — the golden is deterministic, so every C divergence is a genuine ordering effect.

This demonstrates that a `bar.sync` without the proxy fence is not a valid substitute — the missing fence is a real, silicon-observable ordering hazard, not a theoretical one. (The divergence rate is low and timing-sensitive — inserting a per-launch sync often suppresses it — consistent with "the silicon usually drains fast enough, but the ISA ordering guarantee is genuinely absent and *is* occasionally violated under scheduling pressure.")

</details>

**Provenance.** Regression, introduced by [#2048](https://github.com/tile-ai/tilelang/pull/2048) (`f11954cb`, "[Backend] Refactor gemm_sp", 2026-05-19), first shipped in **0.1.10**. Before #2048, the sparse sm90 path emitted the `call_intrin` `tl::tl_gemm_sp()`, which **is** in the `IsAsyncIntrinsic` list ([`inject_fence_proxy.cc:114`](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/transform/inject_fence_proxy.cc#L114)) — so sparse gemm **was** fenced at 0.1.9 (verified: `f11954cb^:src/backend/cuda/op/gemm_sp.cc:128` emits `tl::tl_gemm_sp()`). #2048 replaced that template path with direct `ptx_wgmma_sp_ss`/`ptx_wgmma_sp_rs` lowering and did **not** touch `inject_fence_proxy.cc` (git-verified via `--stat`), orphaning the fence and leaving the `tl_gemm_sp` list entry dead for the WGMMA path. The dense `ptx_wgmma_ss`/`ptx_wgmma_rs` entries were added earlier by #977; a later change (#1068) flipped the unlisted-intrinsic default from fail-safe (`kAsync`) toward fail-open, so an unlisted async intrinsic is no longer caught by default. On `main`, the `tl_gemm`/`tl_gemm_sp` fallbacks are removed entirely and the sparse variants are still absent — the defect persists.

**Dedup.** I searched the open and closed tracker (`fence proxy`, `wgmma_sp fence`, `InjectFenceProxy`) and found no existing report. Distinct from the `cp.async.bulk`/cluster 16-byte-size and offset-alignment crash reports — those are alignment UB on the copy path; this is a missing proxy fence on the sparse WGMMA path.

ZenAlexa:
## Summary

- classify sparse WGMMA `ss` and `rs` calls as async-proxy operations
- classify block-scaled TCGen05 `ss` calls as an async-proxy operation
- add a transform-level regression covering fence placement after generic shared-memory writes

Fixes #2634.

## Rationale

`LowerTileOp` emits these intrinsics for sparse Hopper and block-scaled Blackwell MMA. `InjectFenceProxy` handles their dense siblings at the same compiler boundary. Classifying the three sparse calls restores `fence.proxy.async` between a generic shared-memory write and the following async-proxy MMA.

## Validation

- `bash format.sh --files src/cuda/transform/inject_fence_proxy.cc testing/python/transform/test_tilelang_transform_inject_fence_proxy.py`
- `python -m py_compile testing/python/transform/test_tilelang_transform_inject_fence_proxy.py`

Repository CI will execute the native transform coverage on its self-hosted NVIDIA runner.

## Summary

- Classifies sparse WGMMA `ss`/`rs` and block-scaled TCGen05 `ss` intrinsics as async-proxy operations.
- Inserts `fence.proxy.async` after generic shared-memory writes and before sparse MMA reads.
- Adds regression coverage for all three intrinsics.

## C++ style / lint notes

- The PR changes C++ code but does not modify rules in `docs/developer_guide/cpp_style.md`.
- CI runs the **C++ API Style Audit (warning only)** step.
- Any audit findings are advisory and separate from correctness, build, and test results.

### DIFF
--- a/src/cuda/transform/inject_fence_proxy.cc
+++ b/src/cuda/transform/inject_fence_proxy.cc
@@ -102,7 +102,10 @@ bool IsAsyncIntrinsic(const CallNode *call) {
       call->op.same_as(tma_load_gather4()) ||
       call->op.same_as(tma_store_scatter4()) ||
       call->op.same_as(ptx_wgmma_ss()) || call->op.same_as(ptx_wgmma_rs()) ||
+      call->op.same_as(ptx_wgmma_sp_ss()) ||
+      call->op.same_as(ptx_wgmma_sp_rs()) ||
       call->op.same_as(ptx_tcgen05_mma_ss()) ||
+      call->op.same_as(ptx_tcgen05_mma_blockscaled_ss()) ||
       call->op.same_as(ptx_tcgen05_mma_ts())) {
     return true;
   }
--- a/testing/python/transform/test_tilelang_transform_inject_fence_proxy.py
+++ b/testing/python/transform/test_tilelang_transform_inject_fence_proxy.py
@@ -383,6 +383,40 @@ def test_wgmma_marked_async():
 
 
 @tilelang.testing.requires_cuda_compute_version_ge(9, 0)
+def test_sparse_mma_intrinsics_marked_async():
+    for op_name in (
+        "tl.ptx_wgmma_sp_ss",
+        "tl.ptx_wgmma_sp_rs",
+        "tl.ptx_tcgen05_mma_blockscaled_ss",
+    ):
+        target_op = tirx.op.Op.get(op_name)
+
+        @T.prim_func
+        def before():
+            with T.Kernel(1):
+                smem = T.decl_buffer((1,), T.float16, scope="shared")
+                smem[0] = T.float16(0)
+                T.evaluate(T.call_intrin("handle", target_op))
+
+        mod = tvm.IRModule.from_expr(before.with_attr("global_symbol", "main"))
+        mod = tvm.tirx.transform.BindTarget(auto_target)(mod)
+        mod = tl.cuda.transform.InjectFenceProxy()(mod)
+        order = []
+
+        def visit(node):
+            if isinstance(node, tirx.Evaluate):
+                call = node.value
+                if isinstance(call, tirx.Call):
+                    order.append(getattr(call.op, "name", ""))
+
+        tirx.stmt_functor.post_order_visit(mod["main"].body, visit)
+
+        assert op_name in order
+        assert "tl.fence_proxy_async" in order
+        assert order.index("tl.fence_proxy_async") < order.index(op_name)
+
+
+@tilelang.testing.requires_cuda_compute_version_ge(9, 0)
 def test_shared_barrier_ops_do_not_trigger_fence_proxy():
     @T.prim_func
     def before(A_desc: T.handle("uint8x128", "grid_constant")):
--- a/testing/python/transform/test_tilelang_transform_inject_fence_proxy.py
+++ b/testing/python/transform/test_tilelang_transform_inject_fence_proxy.py
@@ -404,15 +404,19 @@ def test_sparse_mma_intrinsics_marked_async():
         order = []
 
         def visit(node):
-            if isinstance(node, tirx.Evaluate):
+            if isinstance(node, tirx.BufferStore):
+                order.append("shared_store")
+            elif isinstance(node, tirx.Evaluate):
                 call = node.value
                 if isinstance(call, tirx.Call):
                     order.append(getattr(call.op, "name", ""))
 
         tirx.stmt_functor.post_order_visit(mod["main"].body, visit)
 
+        assert order.count("shared_store") == 1
+        assert order.count("tl.fence_proxy_async") == 1
+        assert order.index("shared_store") < order.index("tl.fence_proxy_async")
         assert op_name in order
-        assert "tl.fence_proxy_async" in order
         assert order.index("tl.fence_proxy_async") < order.index(op_name)
 
 


ZenAlexa:
The regression now records the shared-memory `BufferStore`, requires exactly one proxy fence, and asserts store → fence → sparse MMA ordering for all three intrinsics.

ZenAlexa:
CUDA CI completed on the self-hosted 8× H100 80 GB HBM3 runner with CUDA 13.1. The targeted `test_sparse_mma_intrinsics_marked_async` passed, and the CUDA Python suite reported 3,360 passed / 444 skipped. Quick Lint, ROCm 7.2, Metal, pre-commit.ci, and CodeRabbit also passed; CuTeDSL Examples is queued for the NVIDIA runner.

ZenAlexa:
`CuTeDSL Examples` has completed successfully. The current head is green across all five GitHub CI jobs, pre-commit.ci, and CodeRabbit; the 8× H100 CUDA 13.1 run also completed with 3,360 passed / 444 skipped.

LeiWang1999 (approved):

PR #3140, opened 2026-09-02
[Bugfix][Quantize] Use plain int masks in interleave_weight special branches

### LINKED ISSUE #3018: [BUG][Fuzzer][ice-on-valid-code] `interleave_weight` raises `TypeError: 'torch.dtype' object is not callable` on its `nbits=1`/`nbits=2` branches
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and no existing report covers this.

0.1.13 (latest release; the function lives at `examples/dequantize_gemm/quantize/utils.py`, relocated from `tilelang/quantize/utils.py`).

### System information

TileLang 0.1.13, PyTorch 2.13.0. This is a host-side (CPU/`torch`) tensor utility; no GPU is involved and it is architecture-independent.

### Problem description

`tilelang.quantize.interleave_weight(qweight, nbits, target_dtype)` crashes with `TypeError: 'torch.dtype' object is not callable` for three of its documented input combinations — `(nbits=1, "int8")`, `(nbits=2, "float16")`, and `(nbits=1, "float16")` — and never returns. The three affected branches each build a bit mask by calling `torch.int32(0x....)`, but `torch.int32` is a dtype object, not a constructor, so the first such call aborts the function.

`target_dtype="float16"` is one of only two values the function's own `assert` admits (utils.py:88), and `nbits=1`/`nbits=2` each have dedicated, hand-written interleave logic in the function, so these are inputs the function is written to handle. The generic `nbits=4` path (which never reaches these lines) works, which is why the default-config example and CI don't catch it. Not a regression — see Provenance.

<details><summary>All input combinations (run on 0.1.13)</summary>

```
nbits=1 int8     -> CRASH TypeError: 'torch.dtype' object is not callable
nbits=2 float16  -> CRASH TypeError: 'torch.dtype' object is not callable
nbits=1 float16  -> CRASH TypeError: 'torch.dtype' object is not callable
nbits=4 float16  -> PASS shape=(4, 4) dtype=torch.int8    (generic path)
nbits=4 int8     -> PASS shape=(4, 4) dtype=torch.int8    (generic path)
```
</details>

### Reproducible example code

```python
import sys, torch

# 0.1.13: interleave_weight moved from `tilelang.quantize` into the
# dequantize-gemm example. The pip wheel does not ship examples/, so point
# at a source checkout of the tag (adjust the path to your clone).
sys.path.insert(0, "/path/to/tilelang/examples/dequantize_gemm/quantize")
from utils import interleave_weight

qw = torch.randint(0, 15, (4, 4), dtype=torch.int8)

# Fails: takes the nbits=2 / float16 special branch (utils.py:110-114).
interleave_weight(qw, nbits=2, target_dtype="float16")   # -> TypeError

# Control (same inputs, generic nbits=4 path) returns normally:
out = interleave_weight(qw, nbits=4, target_dtype="float16")
print(out.shape, out.dtype)   # -> torch.Size([4, 4]) torch.int8
```

### Traceback

```
  File ".../_tb.py", line 4, in <module>
    interleave_weight(qw, nbits=2, target_dtype="float16")
  File ".../examples/dequantize_gemm/quantize/utils.py", line 111, in interleave_weight
    n8_weight = new_qweight & torch.int32(0xFF0000FF)
                              ~~~~~~~~~~~^^^^^^^^^^^^
TypeError: 'torch.dtype' object is not callable
```

### Expected behavior

The call should return the interleaved `int8` tensor, as the `nbits=4` path already does. The three special branches are computable — replacing each `torch.int32(0xMASK)` with a plain Python integer mask (`0xMASK`), which `torch`'s bitwise `&`/`|` accept directly, produces a result of the expected shape and dtype with no other change.

<details><summary>Fix verified (nbits=2 / float16 branch, plain-int masks)</summary>

Running the function's own core loop followed by the `nbits=2/float16` branch rewritten with plain-int masks returns a `torch.Size([4, 4]) torch.int8` tensor instead of raising:

```python
n8_weight  =  new_qweight & 0xFF0000FF
n8_weight |= ((new_qweight & 0x0000FF00) >> 8)  << 16
n8_weight |= ((new_qweight & 0x00FF0000) >> 16) << 8
return n8_weight.view(torch.int8)
```
</details>

### Additional context

**Root cause.** `interleave_weight` uses the dtype object `torch.int32` as if it were an integer constructor when building its bit masks, so any code path that reaches one of those masks aborts. The masks live only in the `nbits`/`target_dtype` special branches: [`nbits=1, "int8"` at utils.py:104-108](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/examples/dequantize_gemm/quantize/utils.py#L104), [`nbits=2, "float16"` at :111-113](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/examples/dequantize_gemm/quantize/utils.py#L111), and [`nbits=1, "float16"` at :116-122](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/examples/dequantize_gemm/quantize/utils.py#L116). The generic tail (`return new_qweight.view(torch.int8)`) never touches them, so `nbits=4` is unaffected.

**Suggested fix.** In the three special branches (`utils.py:104-122`), drop the `torch.int32(...)` wrapper on each mask and pass the hex integer directly (`new_qweight & 0xFF0000FF`, …); `torch` widens a Python int to the tensor dtype for `&`/`|`. This matches the working `nbits=4` path, which already uses a plain-int mask (`mask = (1 << nbits) - 1`).

**Provenance.** Introduced with the quantize components in [#494](https://github.com/tile-ai/tilelang/pull/494) (2025-05-14), which added `tilelang/quantize/utils.py` with these lines already in their current form; the branches have never returned. The function was relocated verbatim into `examples/dequantize_gemm/quantize/utils.py` (it was `tilelang.quantize.interleave_weight` up to 0.1.9; on 0.1.13 it is re-exported from the example's `quantize/__init__.py`).

**Adjacent inputs (tested this session).** Ran all six `(nbits, target_dtype)` combinations the `assert` admits (0.1.13, v0.1.13 source extracted from the checkout — the wheel does not ship `examples/`); the three that reach a `torch.int32(...)` mask crash, the three that take the generic tail return normally:

| nbits | target_dtype | result |
|---|---|---|
| 1 | int8 | CRASH `TypeError: 'torch.dtype' object is not callable` (utils.py:104) |
| 2 | float16 | CRASH (utils.py:111) |
| 1 | float16 | CRASH (utils.py:116) |
| 2 | int8 | PASS `shape=(4,4) dtype=torch.int8` (generic tail) |
| 4 | float16 | PASS `shape=(4,4) dtype=torch.int8` (generic tail) |
| 4 | int8 | PASS `shape=(4,4) dtype=torch.int8` (generic tail) |

**Generalization (roots + 4-axis sweep, run this session).**

*Root, two levels.* (a) SOURCE-level: a `torch.dtype` object is used as an integer constructor — `torch.int32(0xMASK)` — in the three special branches of `interleave_weight` (`utils.py:104–122`); `torch.dtype.__call__` does not exist, so the first mask aborts before any tensor op. (b) OPERATOR-level: the "dtype-name-as-mask-constructor" idiom for building a bit mask. The correct spellings in the same function (`.view(torch.int32)`, `mask = (1<<nbits)-1`) pass the dtype as an *argument* / use a plain int, and work.

| axis | cell tested (run) | observed result | same-root? |
|---|---|---|---|
| similar-logic | plain-int fix on all three branches (`& 0xFF0000FF` … , `torch.int32(…)`→hex) | all three now PASS `(4,4) int8` — one fix clears the whole branch-family | confirms root |
| related-source | `general_compress` (sibling fn, same `utils.py`) | PASS `(4,4) int8` — it uses `dtype` as an argument (`torch.zeros(dtype=…)`, `.to(torch.int8)`), never calls it | distinct (unaffected) |
| related-operator | bitnet-1.58b local `interleave_weight` copies (`np.int32(0x…)` masks), `nbits=2`/`float16` branch | CRASH `OverflowError: Python integer 4278190335 out of bounds for int32` (numpy 2.x) — same idiom, different failure | distinct-adjacent, not filed |
| related-type | `torch.<dt>(0x0F)` for dt ∈ {int8,int16,int32,int64,uint8,float16,bfloat16,float32,bool} | every one `TypeError: 'torch.dtype' object is not callable` | root is dtype-agnostic (not int32-specific) |

Class: **one root, three affected branches** — every special branch spells its mask `torch.int32(0x…)` (dtype-object-as-constructor); the generic tail uses a plain-int mask and is unaffected. Title/Problem already framed as the branch-family, kept. Distinct-adjacent (not filed): (1) the bitnet `np.int32(0x…)` copies break the same way on numpy 2.x via `OverflowError`; (2) the docstring's own `(10,10)` example crashes at the `.view(torch.int32)` on a non-divisible shape (a shape defect in the example, unrelated to the mask). Neither is this bug.

**Dedup.** I searched the open and closed tracker and found no existing report of this defect (#2477 concerns `decode_i4s_to_i8s` int4 decode, a different function and mechanism).

**Reach.** `target_dtype` is typed and `assert`-restricted to `{"float16","int8"}` at [utils.py:88](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/examples/dequantize_gemm/quantize/utils.py#L88), and `nbits=1`/`nbits=2` each have their own dedicated branch — the function is written to accept exactly these. The only shipped callers of *this* copy are the dequantize-gemm gemv example ([`example_dequant_gemv_fp16xint4.py:204,266`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/examples/dequantize_gemm/example_dequant_gemv_fp16xint4.py#L204), which calls `interleave_weight(qB, num_bits, in_dtype)` with `num_bits=4, in_dtype="float16"` → generic tail, no mask) — verified this session by reading the example's `main()`. The docstring's own example `interleave_weight(qweight, 4, "float16")` reaches the generic tail, but **run verbatim it does not even reach a mask — it raises earlier** (`RuntimeError: self.size(-1) must be divisible by 4 to view Char as Int … got 10`, because the docstring's `(10,10)` int8 shape is not `.view(torch.int32)`-able); with a divisible last dim (e.g. `(10,12)`) the generic tail returns normally. So no shipped caller or test of this copy exercises the three special branches — that is why CI is green. The `nbits=2` interleave is itself a real idiom — the bitnet-1.58b examples carry their own local `interleave_weight` and call it — but both bitnet `__main__`s pass `in_dtype=T.int8` → `interleave_weight(qw, 2, "int8")`, which is the `nbits=2`/`int8` generic tail (no mask), so they dodge the mask branches regardless of copy. Those copies also spell their masks as a dtype-name-as-constructor (`np.int32(0x…)`); on modern numpy (2.x) that path is *also* broken but with a different failure (`OverflowError: Python integer 4278190335 out of bounds for int32`, since `np.int32(0xFF0000FF)` overflows) — a distinct-adjacent defect, not this one. The defect reported here is specific to the `torch.int32(...)` spelling in `examples/dequantize_gemm/quantize/utils.py`.

**Impact.** The trigger is narrow: only the three special branches `(nbits=1,"int8")`, `(nbits=2,"float16")`, `(nbits=1,"float16")` reach a mask, so it needs a non-default `nbits`/`target_dtype` combination no shipped caller uses. When it fires it is a loud, deterministic `TypeError` at call time — the function aborts before returning, so nothing is silently miscomputed and no output tensor is corrupted; the cost is that these three otherwise-legal input combinations cannot run at all. Fixing it closes that boundary class (makes the documented `nbits=1`/`nbits=2` branches usable), rather than correcting any wrong result in a running workload.

rishabhsinha17:
## Summary

`interleave_weight` in `examples/dequantize_gemm/quantize/utils.py` raises `TypeError: 'torch.dtype' object is not callable` on its `(nbits=1, "int8")`, `(nbits=2, "float16")` and `(nbits=1, "float16")` branches, so those input combinations never return.

## Root cause

The three special branches build their bit masks with `torch.int32(0x...)`. `torch.int32` is a dtype object, not a constructor, so the first mask expression aborts the call. The generic `nbits=4` tail uses a plain int mask and is unaffected, which is why the shipped example and CI never reach the failing lines.

## Changes

- Replace each `torch.int32(0x...)` mask with the plain int literal. torch widens a Python int to the tensor dtype for `&` and `|`, matching the existing `mask = (1 << nbits) - 1` path.
- Add `testing/python/issue/test_tilelang_issue_3018.py`, which checks all six `(nbits, target_dtype)` combinations bit-exactly against a pure-Python reference of the interleave.

## Validation

- `python -m pytest -q testing/python/issue/test_tilelang_issue_3018.py`: 6 passed. On main, the three special-branch cases fail with the `TypeError`.
- `bash format.sh --files ...`: clean.
- Host-side, no GPU required to verify.

Fixes #3018

## Summary

- Replaced invalid `torch.int32(...)` mask construction with Python integer masks in the affected `interleave_weight` branches.
- Preserved generic `nbits=4` behavior.
- Added regression coverage for six `(nbits, target_dtype)` combinations against a pure-Python reference.
- Verified output dtype, shape, and bit-exact values.

## Testing

- Six regression tests pass.
- Formatting checks pass.
- Tests run on the host without GPU requirements.

### DIFF
--- a/examples/dequantize_gemm/quantize/utils.py
+++ b/examples/dequantize_gemm/quantize/utils.py
@@ -101,25 +101,25 @@ def interleave_weight(qweight, nbits=4, target_dtype="float16"):
 
     if nbits == 1 and target_dtype == "int8":
         # special handling for 1b interleave
-        n16_weight = new_qweight & torch.int32(0xF0F00F0F)
-        n16_weight |= ((new_qweight & torch.int32(0x000000F0)) >> 4) << 16
-        n16_weight |= ((new_qweight & torch.int32(0x0000F000)) >> 12) << 24
-        n16_weight |= ((new_qweight & torch.int32(0x000F0000)) >> 16) << 4
-        n16_weight |= ((new_qweight & torch.int32(0x0F000000)) >> 24) << 12
+        n16_weight = new_qweight & 0xF0F00F0F
+        n16_weight |= ((new_qweight & 0x000000F0) >> 4) << 16
+        n16_weight |= ((new_qweight & 0x0000F000) >> 12) << 24
+        n16_weight |= ((new_qweight & 0x000F0000) >> 16) << 4
+        n16_weight |= ((new_qweight & 0x0F000000) >> 24) << 12
         return n16_weight.view(torch.int8)
     elif nbits == 2 and target_dtype == "float16":
-        n8_weight = new_qweight & torch.int32(0xFF0000FF)
-        n8_weight |= ((new_qweight & torch.int32(0x0000FF00)) >> 8) << 16
-        n8_weight |= ((new_qweight & torch.int32(0x00FF0000)) >> 16) << 8
+        n8_weight = new_qweight & 0xFF0000FF
+        n8_weight |= ((new_qweight & 0x0000FF00) >> 8) << 16
+        n8_weight |= ((new_qweight & 0x00FF0000) >> 16) << 8
         return n8_weight.view(torch.int8)
     elif nbits == 1 and target_dtype == "float16":
-        n8_weight = new_qweight & torch.int32(0xF000000F)
-        n8_weight |= ((new_qweight & torch.int32(0x000000F0)) >> 4) << 8
-        n8_weight |= ((new_qweight & torch.int32(0x00000F00)) >> 8) << 16
-        n8_weight |= ((new_qweight & torch.int32(0x0000F000)) >> 12) << 24
-        n8_weight |= ((new_qweight & torch.int32(0x000F0000)) >> 16) << 4
-        n8_weight |= ((new_qweight & torch.int32(0x00F00000)) >> 20) << 12
-        n8_weight |= ((new_qweight & torch.int32(0x0F000000)) >> 24) << 20
+        n8_weight = new_qweight & 0xF000000F
+        n8_weight |= ((new_qweight & 0x000000F0) >> 4) << 8
+        n8_weight |= ((new_qweight & 0x00000F00) >> 8) << 16
+        n8_weight |= ((new_qweight & 0x0000F000) >> 12) << 24
+        n8_weight |= ((new_qweight & 0x000F0000) >> 16) << 4
+        n8_weight |= ((new_qweight & 0x00F00000) >> 20) << 12
+        n8_weight |= ((new_qweight & 0x0F000000) >> 24) << 20
         return n8_weight.view(torch.int8)
 
     return new_qweight.view(torch.int8)
--- a/testing/python/issue/test_tilelang_issue_3018.py
+++ b/testing/python/issue/test_tilelang_issue_3018.py
@@ -1,0 +1,83 @@
+# Regression test for https://github.com/tile-ai/tilelang/issues/3018
+# interleave_weight used ``torch.int32(0x...)`` to build bit masks on its
+# nbits=1 / nbits=2 branches; ``torch.int32`` is a dtype object, not a
+# constructor, so those branches raised ``TypeError`` instead of returning.
+# Host-side only: the utility is a plain CPU torch tensor transform.
+import pytest
+import torch
+
+import tilelang.testing
+from examples.dequantize_gemm.quantize.utils import interleave_weight
+
+_MASK32 = 0xFFFFFFFF
+
+
+def _reference_interleave(word: int, nbits: int, target_dtype: str) -> int:
+    """Pure-Python reference for interleave_weight on a single uint32 word."""
+    bits_stride = 8 if target_dtype == "int8" else 16
+    mask = (1 << nbits) - 1
+    num_groups = 32 // bits_stride
+    elems_per_group = bits_stride // nbits
+    new = 0
+    for i in range(num_groups):
+        for j in range(elems_per_group):
+            offset = i * elems_per_group + j
+            shift = (offset % num_groups) * bits_stride + (offset // num_groups) * nbits
+            new |= ((word >> (nbits * offset)) & mask) << shift
+    new &= _MASK32
+
+    if nbits == 1 and target_dtype == "int8":
+        out = new & 0xF0F00F0F
+        out |= ((new & 0x000000F0) >> 4) << 16
+        out |= ((new & 0x0000F000) >> 12) << 24
+        out |= ((new & 0x000F0000) >> 16) << 4
+        out |= ((new & 0x0F000000) >> 24) << 12
+        return out & _MASK32
+    if nbits == 2 and target_dtype == "float16":
+        out = new & 0xFF0000FF
+        out |= ((new & 0x0000FF00) >> 8) << 16
+        out |= ((new & 0x00FF0000) >> 16) << 8
+        return out & _MASK32
+    if nbits == 1 and target_dtype == "float16":
+        out = new & 0xF000000F
+        out |= ((new & 0x000000F0) >> 4) << 8
+        out |= ((new & 0x00000F00) >> 8) << 16
+        out |= ((new & 0x0000F000) >> 12) << 24
+        out |= ((new & 0x000F0000) >> 16) << 4
+        out |= ((new & 0x00F00000) >> 20) << 12
+        out |= ((new & 0x0F000000) >> 24) << 20
+        return out & _MASK32
+    return new
+
+
+def _as_uint32_words(tensor: torch.Tensor) -> list[int]:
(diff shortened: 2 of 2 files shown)


rishabhsinha17:
@LeiWang1999 could you take a look at this one when you have time, or point me to the right reviewer? It is the last open PR from the same batch as #3141, #3142 and #3143, which have all merged. Pre-commit and CodeRabbit are green.

PR #3149, opened 2026-09-03
[ROCm] Select feasible flash-decoding tiles

### LINKED ISSUE #2912: [ROCm] examples/flash_decoding has no working tile configuration on gfx942
`examples/flash_decoding/test_example_flash_decoding.py::test_example_example_mha_inference` fails on gfx942 (MI300X). The test is ungated, so it runs on any ROCm machine. Reproduced on current `main` (`c7fabc4c`), ROCm 7.2, torch 2.13.0+rocm7.2.

```
tvm.error.InternalError: ROCM HIP Error: hipModuleLaunchKernel(...) failed
    with error: invalid argument
3rdparty/tvm/src/runtime/rocm/rocm_module.cc:183
```

There are **two independent causes**, both stemming from `BLOCK_M = 128` / `BLOCK_N = 64` being hardcoded in `main()`. I initially assumed only the first.

### 1. Dynamic LDS exceeds the 64 KB cap

The generated HIP source lays the buffers out past the device limit:

```
O_shared    = buf_dyn_shmem + 0
Q_shared    = buf_dyn_shmem + 0        (aliased with O_shared)
K_shared    = buf_dyn_shmem + 32768
V_shared    = buf_dyn_shmem + 49152
workspace   = buf_dyn_shmem + 65536    <- at the cap
workspace_1 = buf_dyn_shmem + 66048    <- past it
```

Q/K/V alone consume exactly 65536 bytes; the reduction workspaces push the requirement over. `torch.cuda.get_device_properties(0).shared_memory_per_block` on gfx942 is **65536**.

### 2. Register exhaustion, independent of LDS

Halving `BLOCK_N` brings LDS well under the cap but the kernel still fails, with a different error. From `kernel.resource_usage`:

| BLOCK_M | BLOCK_N | max shmem offset | VGPRs | AGPRs | Occupancy | result |
| --- | --- | --- | --- | --- | --- | --- |
| 128 | 64 | 66048 | **256** | 228 | 1 wave/SIMD | FAIL — `hipModuleLaunchKernel` |
| 128 | 32 | 49664 | **256** | 188 | 1 wave/SIMD | FAIL — `invalid argument` |
| 64 | 64 | 49664 | 154 | 96 | 2 waves/SIMD | **PASS** |

Both failing configs sit at the 256-VGPR architectural ceiling. `BLOCK_M` drives `acc_o[block_M, dim]`, so `BLOCK_N` barely moves it — which is why reducing `BLOCK_N` alone does not help.

### The passing configuration is genuinely correct

`BLOCK_M=64, BLOCK_N=64` passes 10/10 across five shapes at both the example's tolerance and a 5x tighter one:

| shape | rtol/atol 0.01 | rtol/atol 0.002 |
| --- | --- | --- |
| B1 H32 Q128 KV2048 D128 | PASS | PASS |
| B1 H32 Q128 KV4096 D128 | PASS | PASS |
| B2 H16 Q128 KV2048 D128 | PASS | PASS |
| B1 H32 Q256 KV2048 D128 | PASS | PASS |
| B1 H32 Q128 KV2048 D64 | PASS | PASS |

### Probably gfx942-specific

`tilelang/carver/arch/cdna.py` records `_GFX950_LDS_SIZE = 160 * 1024`, so CDNA4 has 160 KB of LDS against CDNA3's 64 KB. The 66048-byte requirement would therefore fit on gfx950, and cause 1 would not reproduce there. I have not verified this on hardware — we currently only have gfx942 in CI.

### Out of scope: `causal=True`

While testing I found `causal=True` also fails, at every tile size, with a numerical mismatch rather than a launch error. That looks like a pre-existing, backend-independent gap rather than a ROCm issue — the example carries two `# TODO: Handle causal split case` comments, and with `num_split = 4` the K/V reads use the split offset `(seqlen_kv // num_split) * sid + k * block_N` while the causal mask compares against `k * block_N` with no split offset. The test only exercises `causal=False`. Noting it here for completeness; happy to split it out if you'd prefer a separate issue.

### Suggested direction

I did not send a patch because the obvious one — hardcoding `BLOCK_M = 64` — seems wrong on two counts: it would leave performance on the table on gfx950, where 128 fits, and it does not address cause 2 in any principled way.

TileLang already ships what is needed: `get_arch(target).smem_cap` gives the correct per-arch LDS budget, and there is an autotuner that other examples use (`example_convolution_autotune.py`, `example_dequant_gemm_bf16_mxfp4_cdna4.py`). So the two options look like arch-aware tile defaults derived from `smem_cap`, or autotuning the example.

Worth noting that only autotuning handles cause 2, since register usage is not known until after compilation.

Happy to implement whichever you prefer.

### Note on coverage

`examples/` is not part of the ROCm CI invocation, so none of this is currently caught automatically.

andyluo7:
## Summary

- preserve the existing `128x64` flash-decoding MHA tile as the first choice and unchanged CUDA path
- inspect generated HIP LDS and VGPR usage before launch
- fall back to the existing correct `64x64` tile only when the larger tile exceeds the current AMD device's resource budget
- add a gfx942 runtime regression against the PyTorch reference

## Why

`examples/flash_decoding/example_mha_inference.py` currently hardcodes `BLOCK_M=128`, `BLOCK_N=64`. On gfx942 the generated split kernel requires 66,048 bytes of dynamic LDS against a 65,536-byte device limit and reaches the 256-VGPR ceiling, so `hipModuleLaunchKernel` fails with `invalid argument`.

A device-name conditional would unnecessarily freeze smaller tiles onto future AMD GPUs. Instead, this change compiles candidates in descending order and accepts the first whose generated module fits `get_arch(target).smem_cap` and the available register-file budget. CUDA returns the original `128x64` candidate without entering the HIP resource-selection path.

## Test coverage

- batch 1, two heads, query length 128, KV length 512, head dimension 128
- non-causal split MHA path
- assert gfx942 selects `64x64`
- compare output against the existing PyTorch reference

## Validation

- exact head: `33e0223a`
- changed-file pre-commit hooks: passed
- Python syntax compilation: passed
- gfx942 job [`100527389226`](https://github.com/tile-ai/tilelang/actions/runs/33716628301/job/100527389226): passed
  - focused flash-decoding test: passed in 8.73s
  - full suite: 2236 passed, 1574 skipped
- CUDA and Metal: passed
- CodeRabbit: passed with no unresolved review threads
- CuTeDSL CUDA examples: pending

The existing causal split TODO is unchanged and remains out of scope.

Closes #2912

### DIFF
--- a/testing/python/amd/test_tilelang_hip_flash_decoding.py
+++ b/testing/python/amd/test_tilelang_hip_flash_decoding.py
@@ -1,0 +1,32 @@
+import sys
+from functools import partial
+from pathlib import Path
+
+import tilelang
+import tilelang.testing
+
+
+_EXAMPLE_DIR = Path(__file__).resolve().parents[3] / "examples" / "flash_decoding"
+sys.path.insert(0, str(_EXAMPLE_DIR))
+
+import example_mha_inference as mha  # noqa: E402
+
+
+@tilelang.testing.requires_rocm
+def test_mha_flash_decoding_selects_launchable_tile():
+    """Run the resource-selected MHA kernel against its PyTorch reference."""
+    batch, heads, query_length, kv_length, head_dim = 1, 2, 128, 512, 128
+    kernel, config = mha.compile_flashattn_for_current_device(
+        batch,
+        heads,
+        query_length,
+        kv_length,
+        head_dim,
+        False,
+    )
+
+    assert config in ({"block_M": 128, "block_N": 64}, {"block_M": 64, "block_N": 64})
+    if "gfx942" in str(kernel.target.attrs.get("mcpu", "")):
+        assert config == {"block_M": 64, "block_N": 64}
+    profiler = kernel.get_profiler(tensor_supply_type=tilelang.TensorSupplyType.Normal)
+    profiler.assert_allclose(partial(mha.ref_program, causal=False), rtol=1e-2, atol=1e-2)
--- a/examples/flash_decoding/example_mha_inference.py
+++ b/examples/flash_decoding/example_mha_inference.py
@@ -2,10 +2,13 @@ import torch
 import torch.nn.functional as F
 import tilelang
 from tilelang.autotuner import *
+from tilelang.carver.arch import get_arch
 import tilelang.language as T
 from functools import partial
 
 num_split = 4
+_KERNEL_THREADS = 128
+_TILE_CANDIDATES = ((128, 64), (64, 64))
 
 
 @tilelang.jit(out_idx=[3])
@@ -27,7 +30,7 @@ def flashattn(batch, heads, seqlen_q, seqlen_kv, dim, is_causal, block_M, block_
         glse = T.alloc_global([batch, heads, num_split, seqlen_q], dtype)
         Output_partial = T.alloc_global(part_shape, dtype)  # [batch, seqlen_q, heads, num_split, dim]
         # split
-        with T.Kernel(T.ceildiv(seqlen_q, block_M), heads * batch, num_split, threads=128) as (bx, by, bz):
+        with T.Kernel(T.ceildiv(seqlen_q, block_M), heads * batch, num_split, threads=_KERNEL_THREADS) as (bx, by, bz):
             Q_shared = T.alloc_shared([block_M, dim], dtype)
             K_shared = T.alloc_shared([block_N, dim], dtype)
             V_shared = T.alloc_shared([block_N, dim], dtype)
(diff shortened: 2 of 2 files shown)


LeiWang1999 (approved):

testing/python/amd/test_tilelang_hip_flash_decoding.py
LeiWang1999, line 1: This test can likely be moved to `test_example_flash_decoding.py`.

andyluo7: I kept this regression under `testing/python/amd` because the ROCm CI job collects only `testing/python`; `examples/test_example_flash_decoding.py` is collected by the CUDA examples step, not the ROCm step. Keeping the AMD test here ensures the gfx942 runtime path and tile-selection assertion are exercised on every ROCm run.

PR #3150, opened 2026-09-03
[ROCm] Validate Seer block-sparse attention

andyluo7:
## Summary

- add gfx942 runtime coverage for the existing Seer block-sparse attention kernel
- cover causal self-attention and a query-shorter-than-KV decode shape
- compare the TileLang output against a direct PyTorch masked-attention reference

## Why

The Seer TileLang kernel uses backend-neutral copy, GEMM, reduction, and masking operations, but its example test is CUDA-gated and the ROCm CI job runs only `testing/python`. This PR adds bounded AMD runtime coverage and leaves the production kernel unchanged because the exact-head gfx942 run passed without a backend-specific fix.

## Test coverage

- FP16, batch 1, head dimension 64, block size 64
- self-attention: two heads, query/KV length 128, top-2 block mask
- decode-style shape: one head, query length 128, KV length 256, top-1 block mask
- causal masking and output comparison against PyTorch

## Validation

- exact head: `c47a2e94`
- changed-file pre-commit hooks: passed
- Python syntax compilation: passed
- gfx942 job [`100527419589`](https://github.com/tile-ai/tilelang/actions/runs/33716658170/job/100527419589): passed
  - focused Seer cases passed in 4.08s and 3.83s
  - full suite: 2237 passed, 1574 skipped
- CUDA and Metal: passed
- CodeRabbit: passed with no unresolved review threads
- CuTeDSL CUDA examples: pending

Local pytest collection was unavailable because the macOS Python environment does not have PyTorch installed. No production kernel implementation is changed.

### DIFF
--- a/testing/python/amd/test_tilelang_hip_seer_attention.py
+++ b/testing/python/amd/test_tilelang_hip_seer_attention.py
@@ -1,0 +1,77 @@
+import math
+import sys
+from pathlib import Path
+
+import pytest
+import torch
+
+import tilelang.testing
+
+
+_EXAMPLE_DIR = Path(__file__).resolve().parents[3] / "examples" / "seer_attention"
+sys.path.insert(0, str(_EXAMPLE_DIR))
+
+import block_sparse_attn_tilelang as seer  # noqa: E402
+
+
+def _reference(query, key, value, block_mask, block_size):
+    query_length = query.shape[-2]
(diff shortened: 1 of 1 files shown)


andyluo7:
Closing in favor of #3165, which keeps this validation with the existing example and runs the selected portable example directly in ROCm CI, following maintainer feedback.

PR #3153, opened 2026-09-03
[BugFix] Fix 256-bit vectorized codegen for fp4/int4/fp16

LeiWang1999:
## Summary

256-bit vectorization (sm_100+) failed to compile for several packed dtypes. Minimal repro — an fp4 predicated copy with a loop-invariant predicate:

```python
for row, col in T.Parallel(128, 128):
    dst[row, col] = src[row, col] if predicate[0] != 0 else 0
```

At 64 lanes this materializes a typed vector temporary and a zero broadcast, hitting two gaps: `condval = tl::load_global_256(...)` has no matching `operator=` (the generic overload returned `ulonglong4`, and `fp4_e2_64_t` never got the hand-written assignment operator — it was copy-pasted onto the 128-bit `fp4_e2_32_t` with a broken body), and the broadcast fallback emits `make_fp4_e2_64_t` with 64 args while the constructors stop at 32 lanes.

A compile sweep over dtype × op × width found the same class of bug in two more places:

- **int4/uint4 × 64 lanes**: `PrintType` emitted the nonexistent CUDA type `(u)int8`, and the packed-nibble broadcast only handled 16/32 lanes.
- **fp16 × 16 lanes**: broadcast calls `tl::pack_float16x4(const half ...)`, but codegen prints fp16 scalars as `cutlass::half_t` (the bf16 variant already took `bfloat16_t`, which is why bf16 worked).

## Approach

Rather than adding one more hand-written case per cell, make the contracts generic so new dtypes/widths don't need per-type glue:

- `tl::load_global_256` / `_conditional` generic overloads now return the pointee type (`static_assert(sizeof(T) == 32)`) and bit-cast internally. Assignment then type-checks for any 32-byte vector struct; the `operator=(ulonglong4)` hacks on `fp8_e{4,5,8}_32_t` become dead and are removed, along with the misplaced/broken `fp4_e2_32_t` one. The `T.ldg256` builtin keeps returning `uint32x8` via an explicit `ulonglong4` pointer cast.
- New `tl::broadcast<V>(unit)` replicates a packed unit across any vector width. fp4 broadcast emits one packed fp4x2 byte + `tl::broadcast`, covering every lane count uniformly (and shrinking the generated source).
- `int4x64` maps to `(u)longlong4` (following the `int8x32` precedent); its broadcast packs two 8-nibble 32-bit fields per 64-bit component.
- `pack_float16x4` takes `half_t`, consistent with `__pack_half2` / `pack_bfloat16x4`.

## Testing

- New `testing/python/language/test_tilelang_language_vectorize_matrix.py`: compile matrix of 12 dtypes × {copy, predicated copy, non-zero fill} × {128b, 256b} (72 cells, all previously-failing ones covered), plus bitwise numeric checks for fp4/int4 predicated copies and int4/fp16 fills. On < sm_100 targets the 256b params compile at 128 bits, so the test stays portable.
- Verified on sm_103a: matrix all green; fp4/fp8(e4m3, e5m2)/int4/fp16 predicated copies bitwise-correct for both predicate values; `test_tilelang_language_ldg.py`, `test_tilelang_language_ldst.py`, `test_tilelang_language_vectorize.py`, `test_tilelang_language_vectorized_cast.py` all pass (109 tests) with `TILELANG_DISABLE_CACHE=1`.

## Summary

- Fixes 256-bit vectorized code generation for fp4, int4/uint4, and fp16 on sm_100+.
- Makes `load_global_256` generic for 32-byte vector types.
- Adds width-independent packed-value broadcasting.
- Updates int4/uint4 type mapping and broadcasting.
- Aligns `pack_float16x4` with generated fp16 scalar types.
- Removes obsolete vector assignment operators.

## Testing

- Adds a compile matrix for 12 dtypes, three operations, and 128-/256-bit widths.
- Adds numeric checks for fp4, int4, and fp16.
- Tests pass on sm_103a.

## C++ style / lint notes

- The PR changes C++ code and CUDA headers.
- No evidence indicates changes to rules documented in `docs/developer_guide/cpp_style.md`.
- No C++ API Style Audit warning is reported.
- The changes remove obsolete operators and add template checks. No correctness or build issue is identified.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -985,7 +985,9 @@ void CodeGenTileLangCUDA::PrintType(DataType t, std::ostream &os) { // NOLINT(*)
         os << "int4";
         return;
       } else if (t.lanes() == 64) {
-        os << "int8";
+        // 256 bits; CUDA has no (u)int8 vector type, use (u)longlong4 like
+        // int8x32.
+        os << "longlong4";
         return;
       } else {
         LOG(FATAL) << "Cannot convert type " << t << " to CUDA type!";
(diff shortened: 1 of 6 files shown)


LeiWang1999:
@regression-perf

LeiWang1999:
Closing: this PR's entire content already landed on main via #3158 (e07bd1d0), which was stacked on top of this branch and squash-merged including both commits. The reported merge conflict was this branch conflicting with its own already-merged changes — verified that the file trees are identical (empty diff against main for all touched files).

Issue #3120, opened 2026-08-31
[Perf] Lower T.any_of / T.all_of via warp vote intrinsics (__any_sync / __all_sync) where safe


LeiWang1999:
## Current behavior

`T.any_of(buffer)` / `T.all_of(buffer)` (`tilelang/language/logical.py`) lower through the `tl.any_of` / `tl.all_of` builtins (`src/op/logical.cc`) to `call_extern` of the device helpers `tl::Any` / `tl::All` (`src/tl_templates/cuda/common.h:846-863`, HIP equivalents at `src/tl_templates/hip/common.h:304-322`):

```cpp
template <typename T> TL_DEVICE bool Any(T *a, int size) {
  for (int i = 0; i < size; i++)
    if (a[i]) return true;
  return false;
}
```

Every thread scans the buffer serially and independently. Two costs:

- **Shared buffers:** all 32 lanes of a warp redundantly read all `size` elements — 32x redundant shared-memory traffic per warp, and the loop is O(size) per thread.
- **Fragments:** each thread reduces only its own local slice, so callers that want a tile-wide / warp-uniform answer still need a manual cross-thread combine (typically a shared-memory round-trip).

## Proposal

Use the warp vote intrinsics that #1858 already wired end-to-end (`tl.any_sync` / `tl.all_sync` → `__any_sync` / `__all_sync` on CUDA at `src/cuda/codegen/codegen_cuda.cc:4492`, `__any` / `__all` on HIP at `src/rocm/codegen/codegen_hip.cc:1589`):

**1. Shared/global buffers — cooperative warp-strided scan + one vote:**

```cpp
template <typename T> TL_DEVICE bool AnyWarp(T *a, int size, uint32_t mask = 0xffffffff) {
  bool local = false;
  for (int i = (threadIdx.x % 32); i < size; i += 32)
    local |= static_cast<bool>(a[i]);
  return __any_sync(mask, local);   // AllWarp: &= + __all_sync
}
```

Per-thread reads drop by the warp size (32x on CUDA, 64x on CDNA wavefronts), and the result is warp-uniform.

**2. Fragments — fold the warp with one vote:** keep the per-thread local scan, then `__any_sync` combines the 32 partials in a single instruction, replacing the shared-memory staging a tile-wide reduction needs today.

## Why this cannot be a blanket replacement

[...]

PR #3170, opened 2026-09-04
warp reduction for any_of and all_of

KareemMusleh:
#3120

tested locally on H100.

Will implemented automatic warp divergence detection next

## Summary

- Added automatic warp-or-thread scope resolution for `T.any_of` and `T.all_of`.
- Removed the public `scope` argument while preserving raw two-argument TIR compatibility.
- Added CUDA and HIP warp reduction helpers: `tl::AnyWarp` and `tl::AllWarp`.
- Added conservative warp-uniformity analysis through `ResolveLogicalScope`.
- Integrated scope resolution into the CUDA and ROCm lowering pipelines.
- Added tests for control-flow uniformity, partial warps, shared and global buffers, and target-specific warp sizes.
- Verified native C++ builds, resolver tests, H100 tests, expected ROCm skips, and pre-commit checks.

## C++ style / lint notes

- The PR changes C++ source and headers but does not change rules documented in `docs/developer_guide/cpp_style.md`.
- TLCPP003/TLCPP004 findings are advisory unless they indicate a clear API, FFI, or maintainability risk.
- No correctness, build, or test issue is identified.

### DIFF
--- a/src/op/logical.cc
+++ b/src/op/logical.cc
@@ -17,22 +17,30 @@ PrimExpr any_of_op(PrimExpr args) {
   const CallNode *call = args.as<CallNode>();
   ICHECK(call != nullptr);
   const ffi::Array<PrimExpr> &arg = call->args;
-  ICHECK_EQ(arg.size(), 2);
+  ICHECK_EQ(arg.size(), 3);
   PrimExpr buffer_address = arg[0];
   PrimExpr elems = arg[1];
+  const auto scope_imm = arg[2].as<StringImmNode>();
+  ICHECK(scope_imm != nullptr);
+  const auto scope = scope_imm->value;
   return tirx::Call(DataType::Bool(), tirx::builtin::call_extern(),
-                    {StringImm("tl::Any"), buffer_address, elems});
+                    {StringImm((scope == "warp") ? "tl::AnyWarp" : "tl::Any"),
+                     buffer_address, elems});
 }
 
 PrimExpr all_of_op(PrimExpr args) {
   const CallNode *call = args.as<CallNode>();
   ICHECK(call != nullptr);
   const ffi::Array<PrimExpr> &arg = call->args;
-  ICHECK_EQ(arg.size(), 2);
+  ICHECK_EQ(arg.size(), 3);
   PrimExpr buffer_address = arg[0];
   PrimExpr elems = arg[1];
+  const auto scope_imm = arg[2].as<StringImmNode>();
+  ICHECK(scope_imm != nullptr);
+  const auto scope = scope_imm->value;
   return tirx::Call(DataType::Bool(), tirx::builtin::call_extern(),
-                    {StringImm("tl::All"), buffer_address, elems});
+                    {StringImm((scope == "warp") ? "tl::AllWarp" : "tl::All"),
+                     buffer_address, elems});
 }
 
 TVM_REGISTER_OP("tl.any_of")
--- a/src/op/logical.cc
+++ b/src/op/logical.cc
@@ -23,9 +23,18 @@ PrimExpr any_of_op(PrimExpr args) {
   const auto scope_imm = arg[2].as<StringImmNode>();
   ICHECK(scope_imm != nullptr);
   const auto scope = scope_imm->value;
+  ffi::String fn_name;
+  if (scope == "warp") {
+    fn_name = "tl::AnyWarp";
+  } else if (scope == "thread" || scope == "auto") {
+    // By default auto uses tl::Any, if it can prove no warp divergence it'll
+    // use tl::AnyWarp instead
+    fn_name = "tl::Any";
+  } else {
+    ICHECK(false) << "Invalid scope: " << scope;
+  }
   return tirx::Call(DataType::Bool(), tirx::builtin::call_extern(),
-                    {StringImm((scope == "warp") ? "tl::AnyWarp" : "tl::Any"),
-                     buffer_address, elems});
+                    {StringImm(fn_name), buffer_address, elems});
 }
 
 PrimExpr all_of_op(PrimExpr args) {
@@ -38,9 +47,18 @@ PrimExpr all_of_op(PrimExpr args) {
   const auto scope_imm = arg[2].as<StringImmNode>();
   ICHECK(scope_imm != nullptr);
   const auto scope = scope_imm->value;
+  ffi::String fn_name;
+  if (scope == "warp") {
+    fn_name = "tl::AllWarp";
+  } else if (scope == "thread" || scope == "auto") {
+    // By default auto uses tl::Any, if it can prove no warp divergence it'll
+    // use tl::AnyWarp instead
+    fn_name = "tl::All";
+  } else {
+    ICHECK(false) << "Invalid scope: " << scope;
+  }
   return tirx::Call(DataType::Bool(), tirx::builtin::call_extern(),
-                    {StringImm((scope == "warp") ? "tl::AllWarp" : "tl::All"),
-                     buffer_address, elems});
+                    {StringImm(fn_name), buffer_address, elems});
 }
 
 TVM_REGISTER_OP("tl.any_of")
--- a/src/op/logical.cc
+++ b/src/op/logical.cc
@@ -13,52 +13,41 @@ namespace tvm {
 namespace tl {
 using namespace tirx;
 
-PrimExpr any_of_op(PrimExpr args) {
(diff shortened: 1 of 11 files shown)


LeiWang1999:
Thanks for your contribution! @KareemMusleh Would love toe see some benchmark results.

KareemMusleh:
@LeiWang
image
scope = "thread"
image

scope = "warp"
image

KareemMusleh:
@LeiWang1999 Now that I've added automatic warp divergence detection, should I remove `scope="thread"` and `scope="warp"` or should I keep them just in case? Is the PR good as is or should I add something else?

BTW thanks for the great description in your issue! I had no prior compiler background

LeiWang1999 (changes requested):
Thanks — the direction is right and the relational uniformity proof in `IsWarpUniformCondition` (two symbolic thread instances + `CanProve(!same_warp || agree)`) is a genuinely nice formulation. I built this branch merged with current main and validated on sm_90 hardware; the PR's own tests pass (27 passed), but three probes each confirmed a correctness-grade problem, and stepping back from them I think the API itself needs one more design iteration.

**1. [Blocking] `ResolveLogicalScope` is not wired into any lowering pipeline — `auto` is dead code.**
There are zero references to the pass outside its registration and the unit tests. Verified empirically: a fully uniform kernel with the default scope still lowers to thread-level `tl::Any`; your own language tests assert `tl::All` for `scope=None/auto`, which codifies that the default never benefits. As written, the perf win of #3120 is only reachable by explicitly writing `scope="warp"`.

**2. [Blocking] Wrong results for partial warps (blockDim % warp_size != 0).**
`AnyWarp` strides by `lane_id + 32k`, so with `threads=48` warp 1 only has lanes 0–15 and the residue classes 16–31 (mod 32) are scanned by nobody. Reproduced: 70 elements, single true at index 20, `scope="warp"` → threads 0–31 return true, threads 32–47 return **false**. Additionally `__any_sync(0xffffffff)` with 16 live lanes is UB on Volta+. Neither the resolver nor the runtime guards this; the resolver should refuse `warp` when the launch extent is not a multiple of `thread_warp_size`, and the explicit-scope contract needs the same validation or documentation.

**3. [Blocking] The uniformity analysis only models `IfThenElse`; thread-dependent loops are unsound.**
`for i in T.serial(tx + 1): ... T.any_of(...)` resolves to `warp` (verified), but lanes exit the loop at different trip counts, so the full-mask vote executes under divergence — UB. Data-dependent loops are common in decode kernels, so this is reachable in practice. `For`/`While` with thread-dependent extents/conditions must mark the context divergent; `LetStmt` also needs handling (`Substitute` does not see through let-bound thread-dependent vars, so a condition on a let var is misjudged as uniform).

**4. [Blocking] No buffer-scope validation — `scope="warp"` on a fragment buffer computes garbage.**
Nothing in `logical.py` or `logical.cc` checks the buffer's memory scope. For a fragment buffer, `AnyWarp` receives a per-thread private pointer, so lane k scans elements k, k+32, … of *its own slice* and then votes — the result is neither "any of my slice" nor "any across the warp's slices". Silent wrong answers.

**Design suggestion — derive the scope instead of asking the user for it.**
For shared/global buffers the result is thread-invariant, so thread-vs-warp is purely an implementation choice whose legality is exactly "call site is warp-converged and the block is warp-aligned" — precisely what your analysis computes. There is no information the user needs to supply; this belongs in lowering, which also dissolves finding 1 (no user-facing `auto` left to wire). For fragment buffers, per-thread and cross-warp reductions are two different *operations*, not two scopes of one op — the cross-warp variant needs fragment-layout awareness and should be a separate (future) op; until then fragment + warp should be rejected outright. If an escape hatch is still desired for cases the analyzer can't prove, make it an explicit assertion (e.g. `assume_warp_uniform=True`) rather than a three-valued implementation selector.

**Smaller items:**
- HIP helpers hardcode a stride of 64; RDNA (gfx11xx) is wave32. Use `warpSize`/`__AMDGCN_WAVEFRONT_SIZE`, and share the constant with the resolver's `thread_warp_size` so they can't drift.
- `any_of_op`/`all_of_op` scope dispatch is copy-pasted and the comment in `all_of_op` still says `tl::Any`/`AnyWarp`; factor a small scope→helper mapping.
- `LogicalScopeResolver::Visit` is dead code; include style mixes `<tvm/...>` and `"tvm/..."`.
- The C++ interface dropping 2-arg calls is a breaking change for raw TIR constructors — worth an explicit note in the PR description.

Happy to re-review after the redesign — the analysis machinery you built is the hard part and it carries over directly.

KareemMusleh:
• Thanks for the thorough review and especially for the concrete reproducers. They exposed real gaps in both the legality analysis and the API design. I’ve reworked the implementation around your
  suggestion and addressed each item.

  1. ResolveLogicalScope is now part of normal lowering

  The pass now runs in both the CUDA and ROCm pipelines immediately before LowerAccessPtr, while buffer scope, access ranges, and structured control flow are still available.

  The language tests now verify that a legal T.any_of/T.all_of call selects AnyWarp/AllWarp through the normal compilation pipeline rather than invoking the resolver manually.

  2. Partial warps now fall back to the thread implementation

  Automatic warp selection additionally requires the total block thread extent to be provably divisible by the target’s thread_warp_size. If the extent is unknown or leaves a partial final warp, the
  resolver selects the serial helper.

  I added the reported 48-thread, 70-element regression with the decisive value at index 20 for both any_of and all_of. The kernels now return the correct result for all 48 threads and generate tl::Any/
  3. The analysis now covers loops, expression control flow, and aliases

  The resolver now handles:

  - ForNode: min, extent, and step must all be warp-uniform.
  - WhileNode: currently treated conservatively as divergent.
  - BindNode: TileLang scalar bindings are expanded transitively.
  - LetNode: scoped TIRX let bindings are expanded within the body and restored afterward.
  - tirx.if_then_else: its lazily evaluated branches inherit the condition’s uniformity.
  - Conditions that load thread-private storage: conservatively treated as non-uniform.

  The original T.serial(tx + 1) reproducer now selects the thread implementation. Tests also distinguish lane-divergent loop bounds from warp-uniform bounds such as tx // 32 + 1.

  4. The public scope selector has been removed

  T.any_of and T.all_of no longer accept scope. The frontend keeps the original two-argument TIR form, and ResolveLogicalScope appends a compiler-owned internal "thread" or "warp" decision.

  Fragment and other thread-private buffers therefore retain their existing per-thread semantics and cannot automatically select the cooperative helper. I agree that a cross-lane fragment reduction is a
  different operation and should be introduced separately in the future with fragment-layout-aware semantics.

  For shared and global buffers, the resolver now additionally proves that every lane in a warp reduces the same range. In particular, it checks:

  - the buffer is shared or global;
  - every base BufferLoad index is warp-uniform;
  - the access-pointer extent is warp-uniform;
  - the reduction size is warp-uniform;
  - the call site is warp-converged;
  - the block contains only complete warps.

  This avoids incorrectly combining lane-dependent regions such as shared[tx:tx+8]. Uniform global-buffer reductions are now optimized, while lane-dependent ranges fall back to the thread
  implementation.

  The smaller items are also addressed:

  - CUDA and HIP helpers use the compiler-provided warpSize; no hardcoded 32/64 stride remains.
  - ROCm target normalization derives thread_warp_size from the same mcpu: gfx9 uses wave64, while gfx10/11/12 use wave32. Compiler tests cover gfx90a and gfx1100.
  - The duplicated any_of_op/all_of_op dispatch is factored into a shared lowering helper, removing the stale copied comment.
  - The unused LogicalScopeResolver::Visit wrapper is removed.
  - TVM and local include styles are normalized.
  - Raw two-argument TIR compatibility is preserved. If a two-argument call reaches intrinsic lowering without scope resolution, it conservatively uses the serial helper.

  Validation after the redesign:

  - Native C++ build passed.
  - Resolver suite: 27 passed.
  - Complete any_of/all_of tests on H100: 12 passed, with the two ROCm runtime tests skipped as expected.
  - The H100 coverage includes shared and global cooperative reductions, divergent fallback, and the partial-warp reproducer.
  - All pre-commit hooks pass for the PR files.

  Thanks again—the relational proof remains the core of the implementation, but these cases helped establish the complete legality boundary around it. Happy to iterate further if you see anything else.

KareemMusleh:
@LeiWang1999  I’ve addressed the four blocking correctness issues and the API redesign from your review in the latest commits. CI and CodeRabbit are green. When you have time, could you please take another look?

PR #3171, opened 2026-09-06
[Layout] Consider scalar reducer plans in register-count search

LeiWang1999:
## Summary

Let the default `register-count` policy consider one additional scalar loop-partition candidate at each eligible reducer-update root. The existing spill/register-slot score is unchanged. Native plans remain available and win same-root ties; explicit layouts and widths remain authoritative. The opt-in `io-aware` policy is unchanged.

This fixes a gap in **candidate generation**, not a missing communication term in the score: a layout with fewer accumulator slots cannot win if the search never constructs it.

## Why this PR is needed

In free-mode inference, `ComputePlanCandidate` starts with the width returned by `GetVectorizeSize` and passes the resulting width to `PlanLoopPartition`. Trying different inference roots does not, by itself, try different planning widths at the same root. Buffer-derived candidates still exist, but the scalar alternative described below is not necessarily among them.

For a column reduction of an `8 x 128` tile using 128 threads, consider the following two valid mappings of element `(i, j)`:

```cpp
// Native width-4 loop-partition candidate:
thread = (i % 4) * 32 + j / 4;
slot   = (i / 4) * 4 + j % 4;

// Scalar loop-partition candidate:
thread = j;
slot   = i;
```

In the width-4 mapping, each thread owns four neighboring columns, but each column is split across four warps. In the scalar mapping, thread `j` owns all eight elements of column `j` and can finish that column's reduction locally.

| Per-thread storage / reduction property | Native width 4 | Scalar width 1 |
| --- | ---: | ---: |
| Input fragment slots | 8 | 8 |
| Accumulator slots | 4 | 1 |
| Result fragment slots | 4 | 1 |
| Threads contributing to each output column | 4 | 1 |
| Cross-thread combine for this column reduction | Required | Not required |

The existing register-count score can already prefer the scalar mapping because the accumulator/result use fewer slots, without increasing the input fragment size. Previously, its scalar planning width was not explicitly explored at that root. The new regression case records both the layout and the absence of `AllReduce` in the selected scalar column-reduction kernel.

A planning width of four is a logical distribution choice, not a guarantee of a single `float4` load or arithmetic instruction in the final code. Conversely, scalar per-thread access can still be coalesced across neighboring lanes. Wider access and packed arithmetic can be valuable, but their benefit must not be assumed to outweigh cross-thread reduction costs in every kernel.

## What changes, and what stays unchanged

1. Try the normal inference attempt first.
2. Under `register-count`, try one additional scalar-root attempt only when the root is a parallel loop with reducer updates, has no resolved or annotated loop layout, and has no explicit `coalesced_width`.
3. Propagate each attempt to a complete component layout and compare it with the **same existing** spill/register-slot score. A scalar attempt must strictly improve the score to replace the native attempt at the same root. Existing root-order tie-breaking remains deterministic.

Existing buffer-layout constraints and normal compatibility checks still apply. This is not an instruction to make every reduction scalar: the packed bf16 full-reduction control still needs all 128 participants and retains its native layout and generated CUDA.

There are no new bandwidth/issue terms, communication weights, register budgets, liveness estimates, or user-facing configuration flags. A communication penalty could make different tradeoffs, including accepting larger fragments to avoid a collective; this PR does not introduce that policy. It makes a previously missing candidate available to the current score and does not claim that register count universally predicts execution time.

Search growth is bounded to at most one additional **complete inference attempt per eligible root**. It is not a Cartesian search over widths or independently chosen scalar/vector layouts at every operator. This bound is not a measured compilation-time overhead guarantee.

## Meaning and lifetime of `LayoutInferArgs::candidate_vector_size_limit`

The parameter is a **candidate-generation vector-width cap**, scoped to one inference call. It affects the width used to construct a new free-mode loop-partition candidate:

```cpp
int vector_size = GetVectorizeSize(...);
if (layout_args.candidate_vector_size_limit > 0) {
  vector_size = std::min(vector_size, layout_args.candidate_vector_size_limit);
}
// Existing legality/divisibility adjustments still apply before:
auto candidate = PlanLoopPartition(root, vector_size, thread_bounds);
```

| Value | Meaning |
| --- | --- |
| `0` (default) | No additional cap; preserve the normal width-selection path. It does not bypass the vectorizer's existing limits. |
| `1` | Generate the scalar loop-partition alternative for this inference call. |
| Positive value in general | Clamp the automatically derived planning width; this PR only supplies `0` or `1` and does not enumerate other widths. |

For example, if `GetVectorizeSize` returns four, repeating the inference without an override would just regenerate the width-4 plan. Passing one is the mechanism that actually creates the alternative, rather than merely naming a second attempt "scalar."

The cap is passed only to the selected root's **first free-mode inference call** in the scalar attempt. Once solved, that root layout is retained for propagation. Queue propagation, other operators, later visits, and lowering-time re-inference receive the default arguments. The inferred layout can naturally influence connected buffers and loops, but the cap itself is not propagated as a constraint.

In particular, this parameter:

- Does **not** disable vectorization for the entire kernel or impose a maximum width on final generated instructions. Later vectorization follows the selected layout and its own legality checks.
- Does **not** cap buffer-derived/annotated layouts or override explicit `coalesced_width`/`loop_layout` choices.
- Does **not** belong to persistent `ParallelOpNode` state or require a special operator clone to carry it. Keeping it in `LayoutInferArgs` makes its per-call lifetime explicit and prevents this search option from leaking into another attempt or later inference.

The name's `candidate` refers specifically to constructing a loop-partition alternative, not a global compilation plan or a final-code vectorization setting. The same integer is passed through `RunOneAttempt` and `RunInferStep` into `LayoutInferArgs`, without an intermediate scalar boolean.

Each complete attempt also owns its propagation queue and membership flags. A discarded attempt can leave pending work; keeping this state local prevents it from affecting the next root, scalar candidate, or wide-fallback retry.

### Latest cleanup validation

For `492684f1`, after rebuilding the changed C++ sources and removing temporary diagnostic instrumentation:

- H100 layout, transform layout-inference, reducer-v2/reduce/reduce-maxmin-NaN, and scalar-candidate tests: **514 passed, 1 skipped**. This includes all 14 scalar-candidate tests, including the two new conflict/retry cases.
- B300 scalar-candidate tests: **14 passed**.
- H100 layout golden verification passed with no golden changes; CuTe/oracle parity: **112/112**. The harness uses an auto-selected target: B300's 26 differences from the H100 goldens are identical before and after this cleanup.
- Six compute-control CUDA sources are byte-identical to the previous PR head (`068d0573`). B300 SwiGLU/MHC controls: **52 correctness tests passed**, with all **50 generated sources** unchanged (42 primary controls plus eight additional correctness-only compilations).
- Scoped `format.sh` checks and `git diff --check` passed. The worktree is clean after committing.

The new retry cases exercise actual conflicts with pending propagation work, but also pass on the previous PR head; they are path coverage, not a reproduced old wrong-result case. No full performance-matrix rerun was performed for this cleanup; the earlier performance results below are retained as earlier validation.

### Earlier candidate-search validation

Fresh CUDA builds of the base and patched source:

- New CUDA regression tests: 12 passed on H100 and 12 passed on B300. Covers scalar column ownership, packed bf16 full reduction, annotations, dynamic shapes, multiple update sites, and runtime correctness.
- Layout tests, transform layout-inference tests, and language reducer-v2/reduce/reduce-maxmin-NaN tests on H100: 500 passed, 1 skipped.
- Layout golden harness passed; existing goldens unchanged. CuTe/oracle parity: 112/112.
- Additional SwiGLU/MHC controls: 42/42 generated CUDA sources identical to the base and 52 correctness tests passed on each GPU.
- `./format.sh` and `git diff --check` passed.

An automatic-layout, compute-only compress control on H100 (median of seven randomized CUDA-graph rounds):

| Configuration | Base | Patched | Speedup |
| --- | ---: | ---: | ---: |
| 1 block, 48 tiles | 68.04 us | 28.33 us | 2.40x |
| 132 blocks, 1 tile | 2.46 us | 1.64 us | 1.50x |

Its three AllReduce sites disappear. The bf16 full-reduction control retains byte-identical CUDA, and all 12 paired compute controls pass correctness.

## Summary

- Added one scalar reducer-plan attempt per eligible root in `register-count`.
- Preserved native plans on ties and explicit layout or width constraints.
- Scoped `candidate_vector_size_limit` to the first root inference.
- Kept `io-aware` behavior unchanged.
- Deferred intermediate-width exploration.

## Tests

- Added reducer scalar-candidate coverage for reductions, constraints, retries, dynamic shapes, and correctness.
- Validation covered CUDA, layout, reducer, parity, generated-source, formatting, and diff checks.
- Compute-only controls showed 2.40× and 1.50× speedups with three fewer `AllReduce` sites.
- The bf16 full-reduction control remained byte-identical in CUDA.

## C++ style / lint notes

- The PR changes C++ API declarations but does not modify the documented rules in `docs/developer_guide/cpp_style.md`.
- TLCPP003/TLCPP004 findings are advisory and are not merge blockers unless they indicate a clear API, FFI, or maintainability risk.

### DIFF
--- a/maint/layout_inference/README.md
+++ b/maint/layout_inference/README.md
@@ -18,6 +18,13 @@ Why this exists:
 
 ## Usage
 
+The default `register-count` model also tries one scalar plan at each
+unannotated reducer-update root. Both native and scalar plans use the unchanged
+spill/register-slot score; native plans win same-root ties. This can remove
+replicated column accumulators without forcing scalar layouts for full
+reductions. Explicit widths and layouts remain authoritative. The opt-in
+`io-aware` model retains its existing candidate search and scoring.
+
 ```bash
 python run.py                # verify all cases against expected/
 python run.py --case NAME    # substring filter
@@ -59,7 +66,7 @@ io-aware scorer on the in-tree CuTe layout algebra
   exactly. `CUTE_STATEMENTS` in a case supplies real enclosing-buffer
   shapes where they differ from the fragment shape (offset_region_copy).
 
-Current status: 88/88 statements match with a 100% conversion hit rate.
+Current status: 112/112 statements match with a 100% conversion hit rate.
 The production scorer in `layout_cost_model.cc` uses this formulation; its
 mode arithmetic was additionally audited once, in-tree, against a full
 exact-enumeration oracle across the layout-relevant test corpus (~264
@@ -111,3 +118,4 @@ nest's `parallel_loop_layout` annotation (see `common.py`).
 | `reduce_broadcast` | Softmax-shaped row-reduce + broadcast consume: the most common real-kernel component. Both models agree, including the reduced fragment's canonical partial replication. |
 | `offset_region_copy` | Multi-block tiled copies whose region mins carry block indices (the model's "foreign vars"): offset regions must rank exactly like zero-offset ones. |
 | `shared_staging` | global→shared→fragment→global chain: the shared-side copy is outside the io model, so the fragment is decided by the copy-out alone; goldens would surface any change to that boundary. |
+| `reducer_scalar_candidates` | Register-count can choose scalar column ownership while preserving native packed layouts for a full reduction; io-aware keeps its existing search. |
--- a/maint/layout_inference/cases/reducer_scalar_candidates.py
+++ b/maint/layout_inference/cases/reducer_scalar_candidates.py
@@ -1,0 +1,58 @@
+"""Scalar column ownership versus a necessary full-participant reduction."""
+
+import tilelang.language as T
+
+
+def make_reducer(total=False, width=None, pinned=False, pin_loop=False, updates=1):
+    width = None if width is None else T.int32(width)
+    rows, cols = (4, 256) if total else (8, 128)
+    outputs = 1 if total else cols
+    dtype = T.bfloat16 if total else T.float32
+    if total:
+        layout = T.Fragment(
+            (rows, cols),
+            forward_thread_fn=lambda i, j: (i * cols + j) // 8,
+            forward_index_fn=lambda i, j: j % 8,
+        )
+    else:
+        layout = T.Fragment(
+            (rows, cols),
+            forward_thread_fn=lambda i, j: (i % 4) * 32 + j // 4,
+            forward_index_fn=lambda i, j: (i // 4) * 4 + j % 4,
+        )
+
+    @T.prim_func
+    def main(A: T.Tensor((rows, cols), dtype), B: T.Tensor((outputs,), T.float32)):
+        with T.Kernel(1, threads=128):
+            shared = T.alloc_shared((rows, cols), dtype)
+            values = T.alloc_fragment((rows, cols), T.float32)
+            if pinned:
+                T.annotate_layout({values: layout})
+            T.copy(A, shared)
+            T.copy(shared, values)
+            acc = T.alloc_reducer((outputs,), T.float32)
+            result = T.alloc_fragment((outputs,), T.float32)
+            T.reducer_init(acc)
+            if updates > 0:
+                for i, j in T.Parallel(rows, cols, coalesced_width=width, loop_layout=layout if pin_loop else None):
+                    T.reducer_update(acc[0 if total else j], values[i, j])
+            if updates > 1:
+                for i, j in T.Parallel(rows, cols):
+                    T.reducer_update(acc[0 if total else j], values[i, j])
+            T.finalize_reducer(acc, result)
+            T.copy(result, B)
+
+    return main
+
+
+VARIANTS = {
+    "columns": lambda: make_reducer(),
+    "full_bf16": lambda: make_reducer(total=True),
+}
+
+
+def check(variant, model, result):
+    if variant == "columns" and model == "register-count":
+        assert result["buffers"]["acc"]["replicate"] == 1
+    if variant == "full_bf16":
+        assert result["buffers"]["acc"]["replicate"] == 128
--- a/maint/layout_inference/expected/reducer_scalar_candidates.json
+++ b/maint/layout_inference/expected/reducer_scalar_candidates.json
@@ -1,0 +1,342 @@
+{
+  "columns": {
+    "io-aware": {
+      "buffers": {
+        "acc": {
+          "forward_index": [
+            "_i % 4"
+          ],
+          "forward_thread": "_rep * 32 + _i // 4",
+          "input_shape": [
+            128
+          ],
+          "kind": "PartialFragment",
+          "output_shape": [
+            4
+          ],
+          "replicate": 4,
+          "thread_range": [
+            0,
+            128
+          ],
+          "threads": 128
+        },
+        "result": {
+          "forward_index": [
+            "_i % 4"
+          ],
+          "forward_thread": "_rep * 32 + _i // 4",
+          "input_shape": [
+            128
+          ],
+          "kind": "Fragment",
+          "output_shape": [
+            4
+          ],
+          "replicate": 4,
+          "thread_range": [
+            0,
+            128
+          ],
+          "threads": 128
+        },
+        "values": {
+          "forward_index": [
+            "_i // 4 * 4 + _j % 4"
+          ],
+          "forward_thread": "_i % 4 * 32 + _j // 4",
+          "input_shape": [
+            8,
+            128
+          ],
+          "kind": "Fragment",
+          "output_shape": [
+            8
+          ],
+          "replicate": 1,
+          "thread_range": [
+            0,
+            128
+          ],
+          "threads": 128
+        }
+      },
+      "loops": {
+        "i,j[8x128]": {
+          "forward_index": [
+            "_i // 4 * 4 + _j % 4"
+          ],
+          "forward_thread": "_i % 4 * 32 + _j // 4",
+          "input_shape": [
+            8,
+            128
+          ],
+          "kind": "Fragment",
+          "output_shape": [
+            8
+          ],
+          "replicate": 1,
+          "thread_range": [
+            0,
+            128
+          ],
+          "threads": 128
+        }
+      }
+    },
+    "register-count": {
+      "buffers": {
+        "acc": {
+          "forward_index": [
+            "0"
+          ],
+          "forward_thread": "_i",
+          "input_shape": [
+            128
+          ],
+          "kind": "PartialFragment",
+          "output_shape": [
+            1
+          ],
+          "replicate": 1,
+          "thread_range": [
+            0,
+            128
+          ],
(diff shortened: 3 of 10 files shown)


LeiWang1999:
### Why this PR only adds the scalar candidate

`plan_vector_size_limit` is an integer cap, not inherently a boolean switch. The search could be extended to try other **legal divisors of the automatically selected vector width**. For example, for a native width of 8, candidates could use widths 8, 4, 2, and 1. Here `0` is only the sentinel for "no additional cap"; it is not a vector width. Existing layout, annotation, and vectorization legality checks would still apply.

Such an extension would increase the search space: each additional width can require another complete inference/propagation/scoring attempt for the component. Enumerating widths per root would not require a Cartesian product across all operators, but it would still cost more than the current native-plus-scalar search.

So far, we have not established a performance-beneficial case in the evaluated **unmodified kernels** that requires an intermediate width instead of the native/scalar candidates. The existing validation was not an exhaustive per-kernel width sweep, so this is not evidence that intermediate widths can never help.

For this PR, we therefore keep the scope to the demonstrated missing scalar candidate: preserve the native attempt (`0`) and add only `vec_size=1` at eligible reducer-update roots, using the unchanged score and keeping native plans on same-root ties. This does **not** force all selected layouts to be scalar. Exploring intermediate divisors remains a possible follow-up once there is a concrete kernel benefit to weigh against the additional search cost.

LeiWang1999:
@regression-perf

LeiWang1999:
### Design follow-up: keep the missing-candidate fix small

Implemented in 492684f138066c406d9b09f19134a6f0b7314512:

- Renamed `plan_vector_size_limit` to `candidate_vector_size_limit`. The same integer now passes through `RunOneAttempt` -> `RunInferStep` -> `LayoutInferArgs`; there is no intermediate scalar boolean. `0` means no additional cap, `1` requests the scalar planning candidate. Only the root's first inference call receives the cap; propagation and later visits use the default. The search still supplies only `0` and `1`.
- Made the propagation queue and its membership flags local to `RunOneAttempt`. A discarded attempt can leave pending work, so restoring operators/layouts alone is insufficient isolation. Local ownership also covers native/scalar retries, different roots, and wide-fallback retries without adding a separate cleanup protocol. The shared-queue lifetime already existed before this PR.
- Added two same-inference retry cases with a pinned downstream consumer: a conflicting native candidate followed by a successful scalar candidate, and an explicit-width case that must fall back to a wide result. Temporary diagnostic instrumentation confirmed failed attempts with one or two pending queue entries, including a native failure with two entries immediately followed by scalar success. No diagnostic instrumentation is included in the patch. These cases also produce correct results on the previous PR head: they cover the failure/retry path, not a claim of a previously reproduced wrong result.

### Alternatives intentionally deferred

1. **Choose native/scalar locally inside `ParallelOp`.** A local comparison does not see the complete propagated component: other fragments, reducer accumulators, finalize results, and downstream consumers can change the total spill/register score or invalidate a choice. Keep candidate construction local, but compare complete attempts in the inference engine.
2. **Derive candidates from reducer output ownership.** This is an attractive follow-up and may naturally expose intermediate widths such as two. A general solution must handle arbitrary update indices, several reducers or update sites, dynamic shapes, and explicit layout constraints. It is a larger candidate-construction change than making the demonstrated scalar alternative available to the existing score.
3. **Enumerate all legal width factors.** The integer cap leaves room for that, but each extra factor can require another complete inference/propagation/scoring attempt. We have not established an intermediate-width performance win in the evaluated unmodified kernels; the evaluation was not an exhaustive width sweep. Keep native plus scalar until a concrete benefit justifies the extra search.
4. **Add communication weights, bandwidth/issue terms, or liveness-aware register budgets.** These change which tradeoffs the policy accepts and need their own calibration. This PR still changes candidate availability, not the register-count scoring formula.

This preserves the small, bounded search and existing constraints while making candidate lifetime and attempt isolation explicit.

LeiWang1999:
@regression-perf

LeiWang1999:
@regression-perf

cklxx:
Evidence that the scalar candidate search this PR added does not currently produce the layout its own invariant asserts. No code change proposed here — reporting so the gap is not mistaken for stale goldens.

`maint/layout_inference/cases/reducer_scalar_candidates.py` asserts `columns/register-count` yields `acc` with `replicate == 1` (`combine_size == 1`, the LocalComplete plan described in `src/layout/layout.h`). Every target gives `replicate == combine_size == 128`:

```
[auto=metal ] columns/register-count  check=FAIL acc[shape=[128] rep=128 fthread='_rep']
[cuda-sm90  ] columns/register-count  check=FAIL acc[shape=[128] rep=128 fthread='_rep']
[cuda-sm100 ] columns/register-count  check=FAIL acc[shape=[128] rep=128 fthread='_rep']
[cuda-sm103 ] columns/register-count  check=FAIL acc[shape=[128] rep=128 fthread='_rep']
```

Instrumenting `InferInFreeMode` (`layout_inference.cc`) shows the mechanism does engage:

```
[TLSDBG] root=3 is_parallel=1 explore=1 has_reducer=1 loop_layout_defined=0 annotated_unbound=0 coalesced=0 model=register-count => try_scalar=1
[TLVDBG] cap=1 vector_size 4 -> 1
[TLSDBG] attempt root=3 cap=0 mem=0 regs=264 acc{replicate=128 combine=128} has_best=1
[TLSDBG] attempt root=3 cap=1 mem=0 regs=264 acc{replicate=128 combine=128} has_best=1
```

So: the gate opens, the scalar attempt runs, the cap reaches `ComputePlanCandidate` and does cut `vector_size` to 1 — but the attempt still returns the replicated plan, scores identically (`mem=0`, `regs=264`), and by the documented tie rule (`Ties keep the earliest root and its native plan`) the native plan is kept.

It reproduces on a minimal reducer-only kernel with no `T.copy` at all, so it is not a copy/pipeline interaction:

```python
    acc = T.alloc_reducer((128,), "float32")
    result = T.alloc_fragment((128,), "float32")
    T.reducer_init(acc)
    for i, j in T.Parallel(8, 128):
        T.reducer_update(acc[j], T.float32(i * 128 + j))
    T.finalize_reducer(acc, result)
    for j in T.Parallel(128):
        B[j] = result[j]
```

That gives `acc shape=[128] replicate=128` too. The PR's own unit test (`testing/python/transform/test_tilelang_transform_reducer_scalar_candidates.py::test_register_count_considers_scalar_column_ownership`) asserts `combine_size == 1` for `infer_reducer()` and is marked `@tilelang.testing.requires_cuda`; that is the assertion to run on the CUDA box the PR was validated on, since the layout inference path here cannot reach it.

#3192 (harness target pinning) marks these variants `KNOWN_DIVERGENCE` so the harness reports `KNOWN-GAP` instead of drift, and writes no golden for them; removing that mapping is the signal the behavior landed.

PR #3176, opened 2026-09-07
[Layout] Make default vector size selection reduction-aware

SiriusNEO:
# Reduction-aware layout cost model

The default `register-count` policy now uses reduction-aware scoring for CUDA reducer-v2 components. This makes reduction communication an explicit selection criterion rather than relying on register replication as an indirect proxy.
- Search native, intermediate power-of-two, and scalar vector-width caps at eligible reducer-update roots, respecting explicit layout and coalescing constraints.
- Reuse the physical reducer planner to account for local work, packing, shuffles, and shared-memory collectives.
- Prefer measurable candidates, then minimize the combined score. Ordinary shared-memory traffic and bank conflicts are outside this model's scope.

```text
reducer_issue_cost = local_issues + shared_issues + combine_issues + 4 * shuffle_issues + 32 * barriers
execution_cost = global_memory_cost + sum_reducers(threads * max_vector_bytes * reducer_issue_cost)
register_penalty = 4 * max_threads * register_slots
total_cost = spill_bytes + execution_cost + register_penalty
```

Here:
- `threads` is each reducer's participant count; `max_threads` is their maximum.
- `max_vector_bytes = MaxVectorLoadBits(target, false) / 8`.
- `global_memory_cost` covers ordinary global accesses, including the vector-width issue estimate.
- `shared_issues` counts only shared-memory communication in the physical reducer plan. There is no ordinary shared-memory access cost or bank-conflict priority.
- Spill, memory, and reducer issue estimates include execution counts.
- `register_slots` sums per-thread physical fragment slots, including packing temporaries, without liveness analysis.

**These are heuristic weights, not calibrated latency ratios.** The byte-like scaling aligns thread-group scope but does not make memory traffic, instruction issues, and register capacity physically equivalent. The register term is a static resource penalty, not a peak-live register or occupancy estimate. For example, with a 16-byte vector cap and equal thread counts, one local issue scores like four additional per-thread register slots, and one barrier like 128 slots.

Non-CUDA targets, non-reducer components, and functions with unknown execution counts retain the register-count fallback. No new pass configuration is introduced; the existing `io-aware` policy remains unchanged.

## Summary

- Make the existing default `register-count` policy automatically reduction-aware for eligible CUDA reducers. **No new pass-config key, policy value, or enable/disable switch.** Leaving the policy unset and explicitly naming `register-count` take the same path.
- Build on #3171's isolated inference attempts to search native, intermediate, and scalar vector widths, deduplicating equivalent root layouts before propagation. Share physical reducer-plan selection with materialization, including wide fallback, arithmetic packing, repetition, batching, and eliminated local-complete finalization.
- Scope execution scoring to global accesses and physical reducer work. Remove ordinary shared-memory issue estimation, bank-address analysis, and the bank-conflict-free ordering field. Explicit layout constraints remain authoritative; shared swizzles no longer make an otherwise measurable reducer attempt unknown.
- Among measurable candidates, compare one combined score rather than lexicographically prioritizing spills, execution, or registers:

  ```text
  register_penalty = 4 * max_threads * register_slots
  total_cost = spill_bytes + execution_cost + register_penalty
  ```

  Execution retains local/shuffle/barrier weights of **1/4/32**, global-memory estimates, and reducer collective shared-memory issues. The register term is a static four-byte-per-slot normalization, charged once per attempt. These are ranking heuristics, not measured latency, register allocation, occupancy, or a performance guarantee. Equal totals retain the earlier root/native candidate.
- Keep the experimental `io-aware` policy and existing unsupported/non-reducer/non-CUDA fallback behavior unchanged. Add candidate diagnostics and regression coverage for communication-aware choices, combined scores, explicit constraints, ordinary shared-access score neutrality, swizzled shared layouts, and numerical correctness.

## Validation

- Native rebuild passed.
- **196 tests passed** on H100 across reduction-aware layout, reducer scalar candidates, reducer v2, loop vectorization, and layout inference; five existing deprecation warnings.
- The changed shared-layout and weighted-pooling tests also ran with no visible GPU: **6 passed, 9 skipped**. Runtime cases skip while compile-only checks remain available with the CUDA backend built.
- All layout goldens passed without changing snapshots for the scope reduction. CuTe/oracle agreement: **176/176**. All vectorization anchors passed.
- The original mixed-dtype shared-staging regression retains its `tid * 4` source assertions; they were not relaxed or removed. Swizzled shared copies have both score and GPU correctness coverage.
- Scoped `./format.sh --files` checks across the complete PR and staged whitespace checks passed.
- CUDA-specific tests have backend requirements. No AMD or Metal hardware validation was run for this update.

## Performance

H100, FP32 weighted pooling with 256 groups, warm-cache CUDA Graph timing, median of five rounds. The legacy control uses the register-count/two-attempt selector; the complete and narrowed models use the same default reduction-aware search.

| Queries | Legacy selector | Before scope reduction | Narrowed model | Speedup vs legacy |
| --- | ---: | ---: | ---: | ---: |
| 1 | 1.862 us | 1.599 us | 1.598 us | 1.17x |
| 32 | 10.550 us | 3.069 us | 3.057 us | 3.45x |
| 128 | 52.116 us | 7.475 us | 7.441 us | 7.00x |

The weighted-pooling CUDA sources are identical before and after narrowing. The existing distilled shared-staging regression measures 2.917 us before and 2.918 us after narrowing, with its original access distribution preserved. These measurements are for the regression reproducer, not the full production kernel. Three ordinary column-reduction shapes and the mixed-dtype shared-copy case also pass numerical checks and show no meaningful timing change in this comparison.

## C++ style / lint notes

- The PR does not change rules documented in `docs/developer_guide/cpp_style.md`.
- The C++ API Style Audit (warning only) is relevant to the new planning APIs. No audit result is reported.
- No correctness or build issue is reported. Any audit findings remain advisory unless they indicate an API, FFI, or maintainability risk.


SiriusNEO:
@regression-perf

SiriusNEO:
@regression-perf

LeiWang1999 (dismissed):
Reviewed with a local build (branch merged with current main), full validation on our side. LGTM — this is careful work: the maint harness contract is fully honored (new cases + goldens + bench reproducer + docs), the saturating cost arithmetic and the gated `[ReducerVectorPlan]` diagnostics are exactly right, and `tl.analysis.ReducerCost` makes the heuristic auditable.

**Validation performed**
- `test_tilelang_transform_reduction_aware_layout.py` + `test_tilelang_transform_layout_inference.py`: 87 passed, 1 skipped.
- `maint/layout_inference/run.py --cute`: all symbolic-vs-oracle checks pass — the io-aware path is verifiably untouched.
- `run.py` golden check reports 26 `layout drift` failures, **but a control run on a pristine main build shows the identical 26 failures** — the copy-case goldens (elementwise_copy / fp8_copy / broadcast_read / mixed_dtype_chain) are stale on main and the harness has no CI hook. Not this PR's problem; I'll open a separate issue for the golden refresh + CI hook. Your new reduction_aware goldens pass.
- `bench_reduction_aware.py`: could not reproduce locally — this box has been swapped to sm_103 hardware and the script's sm90 guard correctly refuses. Your SM90 numbers stand unverified by me, no concerns.

**Compile-time A/B** (same tree, only this PR's src changes reverted for baseline; medians):

| case | LayoutInference base → PR | full compile base → PR |
|---|---|---|
| reducer_basic | 66.5 → 153.3 ms (×2.31) | 1062 → 1168 ms (+10.0%) |
| reducer_comm | 28.9 → 64.0 ms (×2.21) | 1004 → 1050 ms (+4.6%) |
| reducer_shared | 137.4 → 255.3 ms (×1.86) | 1289 → 1387 ms (+7.6%) |
| weighted_pooling | 90.5 → 177.1 ms (×1.96) | 1189 → 1202 ms (+1.1%) |
| control_copy (no reducer) | 11.5 → 12.3 ms (noise) | within noise |

LayoutInference roughly doubles on reducer kernels (expected from the width enumeration), end-to-end +1–10% since nvcc dominates, and the non-reducer path is empirically untouched. Fine for JIT; it will be felt in autotuning sweeps (hundreds of configs × ~100 ms). Worth a line in the PR description, and there's an obvious follow-up: candidate attempts share most of their propagation, so caching across width caps should reclaim much of the 2×.

**Two non-blocking notes**
- `bank_conflict_free` lexicographically dominating `total_cost` admits an unbounded inversion in theory (a huge-spill conflict-free candidate beats a cheap one-conflict candidate). It's documented and pinned by `test_bank_conflict_free_precedes_combined_cost`, so clearly intentional — just confirm the bench evidence covers this trade; if a counterexample shows up later, a weighted penalty is the fallback.
- Heads-up on sequencing: #3127 (in changes-requested) touches the same `ComputePlanCandidate` tail your `GetPlanVectorSize` split refactors — whichever lands second needs a rebase; semantically they compose.

cklxx:
Heads-up while you are reviewing this: the harness numbers quoted in the description are target-dependent, because `run.py` resolved its target with `determine_target("auto")` (i.e. the host GPU's compute capability). On one build of this branch:

| target | `run.py` result |
|---|---|
| `cuda-sm90` | 4 failures, all in `reducer_scalar_candidates` |
| `cuda-sm100` | 30 failures |
| `cuda-sm103` | 30 failures |

So "26 golden layout drift failures" is the count for the architecture you recorded on, and the same tree reports a different number elsewhere — including on the new cases this PR adds.

Re-running `cases/reduction_aware.py` under an explicit `cuda-sm90` target (this is a CPU-only build — layout inference only needs `tl.copy` registered, which is what #3192 contributes — so treat it as a cross-check rather than a substitute for the CUDA box):

```
columns         /register-count  check=FAIL acc[shape=[128] rep=128]
columns         /io-aware        check=PASS  ...
intermediate    /register-count  check=FAIL acc[shape=[256] rep=128]
intermediate    /io-aware        check=PASS  ...
full            /register-count  check=PASS  acc[shape=[1] rep=128]
full            /io-aware        check=PASS  ...
warp_collective /register-count  check=FAIL acc[shape=[128] rep=128]
warp_collective /io-aware        check=FAIL acc[shape=[128] rep=128]
```

`check()` for this case asserts `replicate == 1` for `columns`/`intermediate` under `register-count` and `replicate == 4` for `warp_collective`; none of those hold here, while `full` (which asserts `replicate == 128`) passes. The committed `expected/reduction_aware.json` has `columns/register-count` as `output_shape=[1] replicate=1` — the same value the `reducer_scalar_candidates` invariant wants and that no target I can reach produces (reported on #3171).

If the scalar column path in this PR is meant to close that gap, `columns`/`intermediate`/`warp_collective` under `register-count` are the cases to check on the CUDA box. If not, `check()` and the goldens should state the intended answer explicitly, so the two cases stop disagreeing about the same contract.

Two verifiable notes, not blockers:

- `check()` is enforced in `--record` mode, so re-recording cannot bless these — the disagreement has to be resolved by a decision, not by a recording.
- #3192 pins the harness target, stores goldens per suite, and makes `--record` unable to leave stale files behind. Before it, the `cuda-sm90` suite (which the checked-in goldens belong to) cannot run at all on a build without CUDA codegen.

Correction to my previous comment: I had the `columns/io-aware` and `intermediate/io-aware` rows attributed to the wrong model, and I listed `columns/io-aware` as failing. The table above is the measured output; my apologies for the noise.

SiriusNEO:
We have a more simplified method to modeling communication in reduction workload. But I think we should postpone this PR for now for some reasons :)

PR #3186, opened 2026-09-09
[Refactor][Language] Make T.Kernel target-neutral with dialect-owned launch annotations

LeiWang1999:
`T.Kernel` is traced before the target is known, yet it emitted `threadIdx.*` thread-binding loops with a hard-coded default of 128 threads. That bakes a SIMT launch into the frontend: CPU lowering silently pinned the thread index to 0 (per-thread code computed 1/128 of the work), and every backend inherited CUDA's launch vocabulary.

This PR splits the launch into the part every target shares and the parts a backend interprets.

**Target-neutral encoding.** The frontend records the grid as `blockIdx.*` thread-binding loops (the program-index space all targets share) and reserves thread identities as `tx/ty/tz = tl.launch_thread_idx(axis)` `Bind` placeholders. `threads=` becomes the `tl.launch_threads` annotation on the `tilelang_root` block; the frontend no longer picks a default.

**Backend interpretation.** `MaterializeKernelLaunch(lower_thread_binding, default_threads, unsupported_annotations)` decides the target-dependent meaning right after `BindTarget`:
- SIMT backends rebind each placeholder as a `threadIdx.*` `thread_extent` over the same `Var` (extent from the annotation, else the backend default; CUDA/ROCm/Metal/WebGPU pass 128).
- Backends without SIMT drop the placeholders and reject a body that references a thread index instead of miscomputing.
- Launch annotations a backend cannot honor (`cluster_dims` on cpu/rocm/metal/webgpu) are rejected here rather than silently ignored downstream.

**Dialect-owned launch API.** Each language dialect defines its own `Kernel` with explicit keyword parameters, which is what editors show on hover and what type checkers verify: cuda offers `threads`/`prelude`/`cluster_dims`, rocm and metal `threads`/`prelude`, webgpu `threads`, cpu `prelude`, and the common `Kernel(*grid)` none. All funnel into `tilelang.language.kernel.launch_kernel`; unknown keywords are rejected by Python itself. `tilelang.language` remains the CUDA facade. `ClusterKernel` keeps its interface.

**Cluster identity.** `T.get_cluster_id/ids/dims/size/extent(s)` are program-space arithmetic (`bx // cluster_dims`) computed at trace time and therefore target-neutral; `T.block_rank_in_cluster()` stays the CUDA hardware intrinsic. The two agree under the cluster-aware threadblock swizzle, which permutes the grid at cluster granularity (covered by a runtime test).

**Eager JIT.** Phase 1 recognizes the launch by a `kernel_launch_factory` mark on every dialect's `Kernel` (and `ClusterKernel`) instead of object identity against the common implementation, so the phase-1 body guard is injected for every dialect and for aliases such as `K = T.Kernel`.

**Frontend.** `KernelLaunchFrame` exposes `grid_vars/grid_extents/thread_vars/thread_extents` explicitly instead of relying on fixed frame offsets; `get_thread_extent()` raises at trace time when `threads=` was omitted. A failed trace now unwinds the launch frame stack instead of leaving a stale `KernelLaunchFrame.Current()`.

**Tests.** Tests that applied passes directly to a traced function now materialize the launch first (thread indices otherwise carry no extent); fixtures that wrote per-thread code and compiled it for the `c` target are rewritten at the tile level; the span-preservation check asserts that no statement loses its span rather than that the spanned count never decreases. New coverage in `test_tilelang_transform_materialize_kernel_launch.py` (encoding shape, annotation recording, SIMT default/explicit threads, non-SIMT rejection, Var identity, per-dialect keyword sets, cluster id arithmetic, frame unwinding), eager-rewriter coverage over all dialects plus an eager gemm whose tile extent depends on a `T.const`, and a CUDA runtime check that `get_cluster_id() * cluster_size + block_rank_in_cluster() == blockIdx.x`.

Validation (CUDA build, `./format.sh` and `git diff --check` clean):
- `testing/python/{language,transform,cpu,debug,metal,webgpu}` on the final head: **2229 passed, 69 skipped** (3 pre-existing fp16 `test_reduce` cases on this sm_103 machine deselected; `test_grid_sync` is device-dependent here: passes on one GPU, fails on others with identical code before and after this change).
- `testing/python/{language,jit,kernel,issue,ir,backend,components,primitives,carver,autotune}`: **1853 passed, 95 skipped**, 4 failures unrelated to this change (3 pre-existing fp16 `test_reduce` cases on this sm_103 machine; 1 `jit_diagnostics` test that depends on kernel-cache state).

## Summary

- Move `Kernel` launch ownership into backend dialects.
- Record grid and thread placeholders in `KernelLaunchFrame`.
- Materialize thread placeholders according to backend requirements.
- Add backend-specific launch APIs for CUDA, ROCm, Metal, WebGPU, and CPU.
- Add cluster launch helpers and target-neutral cluster arithmetic.
- Improve eager JIT launch-factory detection and frame cleanup after trace failures.
- Update backend pipelines, documentation, and launch-contract tests.
- Keep the CUDA launch API available through `tilelang.language`.

## Validation

Validation covered formatting, diff checks, native builds, and affected tests. Reported results included 947 passed and 30 skipped tests in one group, plus 1,853 passed and 95 skipped in another. Four unrelated failures remained: three pre-existing fp16 reduction failures and one kernel-cache-dependent JIT diagnostics failure.

## C++ style / lint notes

The PR changes C++ reflection and launch infrastructure. The reflected `KernelLaunchFrameNode` fields touch guidance in `docs/developer_guide/cpp_style.md` for public `ObjectNode` fields.

The `C++ API Style Audit (warning only)` CI step applies. Any TLCPP003/TLCPP004 findings are advisory and separate from correctness, build, and test issues. They are not merge blockers unless they introduce a clear API, FFI, or maintainability risk.

### DIFF
--- a/docs/programming_guides/language_basics.md
+++ b/docs/programming_guides/language_basics.md
@@ -85,16 +85,40 @@ Notes
 
 ## 2. Launching Work with `T.Kernel`
 
-`with T.Kernel(...)` declares a launch context and creates block/thread
-bindings. For GPU backends, specify a grid and threads per block.
+`with T.Kernel(...)` declares a grid of tile programs. The positional
+arguments give the grid extent along each axis and the returned variables are
+the program indices along those axes. This is the part of a launch every
+target shares: on CUDA a program is a thread block and `bx`/`by` are
+`blockIdx.x`/`blockIdx.y`; on CPU the grid becomes the outer loop.
 
 ```python
 with T.Kernel(grid_x, grid_y, threads=128) as (bx, by):
-    ...  # bx/by are blockIdx.x/y
+    ...  # bx/by are the program indices (blockIdx.x/y on CUDA)
 ```
 
-You rarely need raw thread indices; most kernels use structured loops
-(`T.serial`, `T.unroll`, `T.Parallel`, `T.Pipelined`) inside a `T.Kernel`.
+Keyword arguments are launch annotations that the backend interprets once the
+target is known. Each language dialect's `Kernel` declares the annotations its
+backend understands as explicit keyword parameters, so hovering or
+autocompleting `T.Kernel` shows exactly those and anything else is rejected:
+`tilelang.language` (the CUDA dialect) offers `threads`, `prelude` and
+`cluster_dims`; `tilelang.rocm.language` / `tilelang.metal.language` offer
+`threads` and `prelude`; `tilelang.cpu.language` offers only `prelude`.
+`threads` is the SIMT one: how many threads run each tile program on
+GPU-style backends. Those backends pick a default (128) when it is omitted;
+a kernel written with the CUDA dialect still compiles for CPU, which ignores
+the thread count. Code inside `T.Kernel` operates at the tile-program level,
+so you rarely need raw thread indices; most kernels use structured loops
+(`T.serial`, `T.unroll`, `T.Parallel`, `T.Pipelined`) that the compiler maps
+onto threads. `T.get_thread_binding()` exposes the thread index for
+thread-level code on SIMT targets; a kernel that uses it is rejected when
+compiled for a target without SIMT threads.
+
+`T.ClusterKernel(..., cluster_dims=...)` adds the CUDA thread-block-cluster
+annotation (SM90+). A cluster is a `cluster_dims`-shaped tile of the grid, so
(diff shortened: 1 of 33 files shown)


LeiWang1999:
@perf-regression

LeiWang1999:
@regression-perf

PR #3187, opened 2026-09-09
Fix crash when applying bitwise_not to vec integers

### LINKED ISSUE #2993: [BUG][Fuzzer][ice-on-valid-code] Vectorized `~x` (bitwise-NOT) on 16/32-bit ints fails to compile instead of computing the complement
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and found no existing report of this defect.

0.1.13 (latest release)

### System information

Python 3.13, linux
tilelang 0.1.13
torch 2.8.0+cu128
GPU: NVIDIA L40S (sm_89), CUDA 12.8

### Problem description

An elementwise bitwise-NOT `B[i] = ~A[i]` (equivalently `T.bitwise_not(A[i])`) on a 16- or 32-bit integer tensor fails to compile whenever the `T.Parallel` loop is auto-vectorized — i.e. whenever a thread handles more than one element (`threads < N`), which is the default for elementwise kernels. The generated CUDA is:

```cuda
*(int4*)(B + ...) = (~*(int4*)(A + ...));   // int32, 4 elems/thread
```

and `nvcc` rejects it: `no operator "~" matches these operands / operand types are: ~ int4`. The same kernel with one element per thread (`threads == N`, no vectorization) compiles and returns the correct values, and the binary bitwise operators (`&`, `|`, `^`), the shifts, and `int8` all vectorize correctly. So the elementwise operation the user wrote is well-defined and computable on this hardware; only the vector codegen for the unary `~` is broken.

<details>
<summary>nvcc error and emitted line (int32 and int16)</summary>

```
tvm_kernels.cu(19): error: no operator "~" matches these operands
            operand types are: ~ int4
    *(int4*)(B + (((int)threadIdx.x) * 4)) = (~*(int4*)(A + (((int)threadIdx.x) * 4)));

# int16 takes the same path -> short4:
    *(short4*)(B + (((int)threadIdx.x) * 4)) = (~*(short4*)(A + (((int)threadIdx.x) * 4)));
```
</details>

### Reproducible example code

```python
import numpy as np, torch

N, TH = 128, 32          # N/TH = 4 elements per thread -> auto-vectorized to int4

@T.prim_func
def bitnot(A: T.Tensor((N,), "int32"), B: T.Tensor((N,), "int32")):
    with T.Kernel(1, threads=TH):
        for i in T.Parallel(N):
            B[i] = ~A[i]                 # <-- vectorized unary ~  => nvcc reject

@T.prim_func
def control(A: T.Tensor((N,), "int32"), B: T.Tensor((N,), "int32")):
    with T.Kernel(1, threads=TH):
        for i in T.Parallel(N):
            B[i] = A[i] ^ T.int32(-1)    # same complement via binary ^, SAME vector config

a   = np.arange(-64, 64, dtype=np.int32)
exp = ~a

# control: binary bitwise op scalarizes per lane -> compiles, correct
kc  = tilelang.compile(control, out_idx=[1])
oc  = kc(torch.tensor(a, device="cuda")).cpu().numpy()
print("control (^ -1):", "PASS" if np.array_equal(oc, exp) else "FAIL")   # -> PASS

# bug: vectorized unary ~ -> nvcc compile crash
try:
    kb = tilelang.compile(bitnot, out_idx=[1])
    ob = kb(torch.tensor(a, device="cuda")).cpu().numpy()
    print("bitnot  (~):", "PASS" if np.array_equal(ob, exp) else "FAIL")
except Exception as e:
    print("bitnot  (~): COMPILE CRASH ->", str(e).splitlines()[-1][:80])  # no operator "~" ... int4
```

### Traceback

No traceback in the Python sense — the failure is an `nvcc` compile error surfaced as a `RuntimeError` from `tilelang.compile`; the key line is:

```
error: no operator "~" matches these operands
            operand types are: ~ int4
```

### Expected behavior

The vectorized `~` should compute the elementwise complement and compile, matching the scalar path and the `^ -1` binary path, both of which return the correct result over the same inputs on this hardware. The binary bitwise operators already handle the vector case by scalarizing per lane; the natural expectation is that unary `~` does the same rather than emitting `~` on a packed vector type CUDA has no operator for.

### Additional context

**Root cause.** The C codegen's `bitwise_not` handler does not scalarize a vector operand, so for a vectorized lane it emits `~` applied to a packed vector type (`int4`/`short4`) that has no `operator~`. The binary bitwise/shift operators go through [`PrintBinaryIntrinsic`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L536), which branches on `op->dtype.lanes()` and, for `lanes() > 1`, calls `PrintVecBinaryOp` to emit the op lane-by-lane. The [`bitwise_not` branch](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L677) has no such lanes check — it unconditionally emits `os << "(~"; PrintExpr(arg); os << ")"`, which becomes `(~*(int4*)...)` for a vectorized store. `int8` dodges it because 4×int8 is vectorized as a single scalar `int` reinterpret (`~*(int*)...`), on which `~` is legal C. The CUDA codegen (`codegen_cuda.cc` at `8df8ebd6`) overrides neither the `bitwise_not` branch nor `NotNode`, so the base-class scalar `~`/`!` reaches nvcc verbatim.

**Suggested fix.** One direction is to give `bitwise_not` the same lane handling the binary intrinsics have — when `op->dtype.lanes() > 1`, emit the complement per lane (the analogue of `PrintVecBinaryOp`) rather than `~` on the whole vector. Whether that is preferable to declining to vectorize the unary `~` in the vectorizer is a call for the maintainers; either way the current outcome (a hard nvcc reject on a legal, computable elementwise op) seems to be the problem.

**Generalization (roots + 4-axis sweep, all cells RUN on 0.1.13 / sm_89).**

*Two-level root.*
- **SOURCE-level:** `CodeGenC::VisitExpr_(CallNode)` `bitwise_not` branch at [`codegen_c.cc:677`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L677) emits `os << "(~"; PrintExpr(arg); os << ")"` with **no** `op->dtype.lanes() > 1` guard — unlike `PrintBinaryIntrinsic` ([`codegen_c.cc:536`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L536)) which routes `lanes>1` to `PrintVecBinaryOp`. The class is: **the unary-operator visitors in `CodeGenC` that print their operator token directly onto a possibly-vectorized operand, lacking the per-lane scalarize the binary path has.** There are exactly two such sites: the `bitwise_not` branch and `NotNode` ([`codegen_c.cc:617`](https://github.com/tile-ai/tvm/blob/8df8ebd61659505dd7005f4820b030e983e2c005/src/target/source/codegen_c.cc#L617), `os << '!'`). `codegen_cuda.cc` overrides neither.
- **OPERATOR-level:** the TIR unary builtins/nodes that go through those two sites — `bitwise_not` (`~`) and logical `Not` (`!`). Adjacent-looking unary spellings (`-x`) do **not** route here.

*4-axis sweep.*

| axis | cell tested (vectorized `T.Parallel`, N=128 TH=32 → 4/thread) | observed result | same-root? |
|------|--------------------------------------------------------------|-----------------|------------|
| related-operator | logical `not A[i]` on `bool` (`NotNode`) | nvcc `no operator "!" matches these operands / operand types are: ! ushort4` | **same** (sibling site, line 617) |
| related-operator | unary neg `-A[i]` on int32 | COMPILED OK + values PASS; emitted per-lane `__1.x = (v_.x*v__1.x)` (lowered to `Sub`/mul-by-−1, routes through binary `PrintVecBinaryOp`) | **distinct** (not a unary visitor) |
| related-type | int16 `~A[i]` | nvcc `operand types are: ~ short4` | same |
| related-type | uint32 `~A[i]` | nvcc `operand types are: ~ uint4` | same |
| related-type | int64 `~A[i]` (N=128 → longlong2) | nvcc `operand types are: ~ longlong2` | same |
| related-type | int8 `~A[i]` | COMPILED OK + values PASS; emitted `(~*(int*)...)` — 4 bytes reinterpreted as scalar `int`, `~` legal & bit-correct | **dodges** (scalar-int path, not packed vector) |
| similar-logic | binary `A[i] ^ -1` (draft control), same vector config | COMPILED OK + PASS (routes through `PrintBinaryIntrinsic` → per-lane) | n/a — the correctly-scalarized comparand |
| related-source | `NotNode` (same file, sibling unary visitor) | see logical-`not` row above — crashes identically | same |

**Reframe decision.** The defect is a **two-site class**: `bitwise_not` (`~`, `no operator "~" ... {int4,short4,uint4,longlong2}`) and `NotNode` (`!`, `no operator "!" ... ushort4`). Both are the same missing-per-lane-scalarize pattern in the same file, and both were run and crash identically across integer widths. Kept as **two separate issues** (this one = the bitwise `~` half; the logical `!` half is filed separately) because they are two distinct visitor sites and can be fixed independently — but they should reference each other as one root class.

**Distinct-adjacent (not filed):** unary neg `-x` looks like a sibling but is lowered to a binary `Sub`/multiply and scalarizes correctly (PASS) — a *distinct, non-bug* path, recorded so the class is not over-claimed to "all unary ops". No NEW bug was found while sweeping.

**Example-run (PART 1):** `examples/deepseek_v32/topk_selector.py` — RAN verbatim on 0.1.13, compiles and passes; its `~bits_uint` is scalar-inside-`if_then_else`, so it **dodges** this bug (confirms the draft's Reach claim, corrected in the Reach section).

**Provenance.** Present in the vendored TVM C codegen (`3rdparty/tvm`, pinned at `8df8ebd6` for the 0.1.13 tag). The `bitwise_not` branch has carried this scalar-only form since it was vendored; introduced-vs-relocated before the TileLang fork is unverified. Not introduced by a recent change.

**Dedup.** I searched the open and closed tracker and found no existing report of a vectorized `bitwise_not` codegen failure.

**Reach.** `~` in a kernel maps to the documented TIR `bitwise_not` builtin and is used in-kernel in a shipped example (`examples/deepseek_v32/topk_selector.py` uses `~bits_uint`), so the operator is in real use. I ran that shipped example verbatim on 0.1.13 (`git show v0.1.13:examples/deepseek_v32/topk_selector.py`, default `test_topk_selector` config): it compiles and passes (`selected/all: 2048 / 2048 = 1.0` for every batch). It dodges this bug because its two `~` uses are on a scalar `bits_uint` inside `T.if_then_else`, not in a vectorized `T.Parallel` store, so the `~` never lands on a packed vector type. Triggering the crash requires the elementwise loop to auto-vectorize — a thread covering >1 element (`threads < N`), the common default for a plain elementwise `T.Parallel` kernel — on a 16-, 32-, or 64-bit integer. `int8` dodges for a different reason: 4×int8 vectorizes by reinterpreting the four bytes as a single scalar `int` (`*(int*)`), so the emitted `~` lands on a scalar C `int`, which is legal and — because bitwise-NOT distributes over byte boundaries — also numerically correct (verified PASS). No test exercises a vectorized elementwise `~`, which is why CI is green.

**Impact.** The trigger is narrow: a vectorized elementwise `~` on a 16/32-bit int, i.e. a thread covering >1 element (`threads < N`) — the common default, but int8 and the scalar/`^ -1` paths all dodge it. When it fires it is a hard `nvcc` compile error, so it fails loudly at build time on that exact kernel and corrupts no data and cannot slip into a running workload silently. Fixing it closes a vector-codegen gap for one unary builtin, letting a legal, computable elementwise op build the same way its binary-bitwise siblings already do.

KareemMusleh:
implement `PrintVecUnaryOp_` to fix #2993. Doing it through XOR seems more annoying because we'd have to find the equivalent of -1 to other unsigned dtypes. This seems more generally useful

Tested 8 dtypes:
- Signed: int8, int16, int32, int64
- Unsigned: uint8, uint16, uint32, uint64

## Summary

- Added `PrintVecUnaryOp_` to generate lane-by-lane vector unary operations in CUDA code.
- Fixed vectorized integer bitwise NOT for signed and unsigned `int8`, `int16`, `int32`, and `int64` types.
- Masked 8-bit results to prevent integer-promotion bits from affecting packed stores.
- Added code-generation and runtime regression tests for 1, 2, and 4 elements per thread.

## C++ style / lint notes

- The PR changes C++ code but does not alter rules documented in `docs/developer_guide/cpp_style.md`.
- The “C++ API Style Audit (warning only)” CI step may report advisory findings. These warnings do not affect correctness or build behavior.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -1377,6 +1377,34 @@ void CodeGenTileLangCUDA::PrintVecBinaryOp(const std::string &op, DataType t,
   os << sret;
 }
 
+void CodeGenTileLangCUDA::PrintVecUnaryOp_(const std::string &op,
+                                           DataType dtype,
+                                           const PrimExpr &input,
+                                           std::ostream &os) {
+  std::string result = name_supply_->FreshName("_");
+  PrintIndent();
+  PrintType(dtype, stream);
+  stream << ' ' << result << ";\n";
+  int ssa_scope = BeginScope();
+  // Evaluate the vector once before extracting its lanes.
+  std::string value = SSAGetID(PrintExpr(input), input.dtype());
+  for (int i = 0; i < dtype.lanes(); ++i) {
+    std::ostringstream lane;
+    lane << '(' << op;
+    PrintVecElemLoad(value, input.dtype(), i, lane);
+    lane << ')';
+    // The packed byte store combines lanes with OR. Integer promotion must
+    // not leave upper bits set and overwrite neighboring bytes.
+    std::string lane_value = lane.str();
+    if (dtype.bits() == 8 && (dtype.is_int() || dtype.is_uint())) {
+      lane_value = "(" + lane_value + " & 0xffu)";
+    }
+    PrintVecElemStore(result, dtype, i, lane_value);
+  }
+  EndScope(ssa_scope);
+  os << result;
+}
+
 void CodeGenTileLangCUDA::PrintVecElemLoad(const std::string &vec, DataType t,
                                            int i,
                                            std::ostream &os) { // NOLINT(*)
@@ -4692,6 +4720,11 @@ void CodeGenTileLangCUDA::VisitExpr_(const CallNode *op, std::ostream &os) {
        << ", " << PrintExpr(offset) << ")";
   } else if (HandleLateIntrinsicCall(op, os)) {
     // Handled by a helper to keep MSVC's parser away from the giant tail chain.
+  } else if (op->op.same_as(builtin::bitwise_not()) &&
+             op->dtype.is_fixed_length_vector() && !op->dtype.is_bool() &&
+             (op->dtype.is_int() || op->dtype.is_uint())) {
+    ICHECK_EQ(op->args.size(), 1U);
+    PrintVecUnaryOp_("~", op->dtype, op->args[0], os);
   } else {
     CodeGenC::VisitExpr_(op, os);
   }
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -1377,6 +1377,15 @@ void CodeGenTileLangCUDA::PrintVecBinaryOp(const std::string &op, DataType t,
   os << sret;
 }
 
+/*!
+ * \brief Emit a prefix unary operator, such as `~`, on each vector lane.
+ *
+ * Materializes the input once and uses the lane load/store helpers to
+ * assemble the result, including packed integer representations. Masks
+ * 8-bit integer results so promoted upper bits cannot affect adjacent lanes.
+ * Supporting statements go to `stream`; the result variable name goes to
+ * `os`. The input and result must have the same fixed-length vector type.
+ */
 void CodeGenTileLangCUDA::PrintVecUnaryOp_(const std::string &op,
                                            DataType dtype,
                                            const PrimExpr &input,
--- a/src/cuda/codegen/codegen_cuda.h
+++ b/src/cuda/codegen/codegen_cuda.h
@@ -80,6 +80,8 @@ protected:
                        std::ostream &os) final; // NOLINT(*)
 
 private:
+  void PrintVecUnaryOp_(const std::string &op, DataType dtype,
+                        const PrimExpr &input, std::ostream &os);
   // Handle volatile loads
   void HandleVolatileLoads(const std::string &value, const BufferLoadNode *op,
                            std::ostream &os) final;
--- a/testing/python/language/test_tilelang_language_bitwise_not.py
+++ b/testing/python/language/test_tilelang_language_bitwise_not.py
@@ -1,0 +1,60 @@
+"""Regression coverage for vectorized integer bitwise NOT (#2993)."""
+
+import pytest
+import torch
+
+import tilelang
+from tilelang import tvm
+from tilelang.cuda import language as T
+
+
+DTYPES = [f"{sign}{bits}" for sign in ("int", "uint") for bits in (8, 16, 32, 64)]
+
+
+def make_bitwise_not(dtype, elements_per_thread):
+    n = 32 * elements_per_thread
+
+    @T.prim_func
+    def kernel(A: T.Tensor((n,), dtype), B: T.Tensor((n,), dtype)):
+        with T.Kernel(1, threads=32):
+            for i in T.Parallel(n):
+                B[i] = ~A[i]
+
+    return kernel
+
+
+@pytest.mark.parametrize("dtype", DTYPES)
+@pytest.mark.parametrize("elements_per_thread", [1, 2, 4])
+def test_bitwise_not_codegen(dtype, elements_per_thread):
+    target = tvm.target.Target({"kind": "cuda", "arch": "sm_90"})
+    with target:
+        source = tilelang.lower(make_bitwise_not(dtype, elements_per_thread), target=target).kernel_source
+    assert "~" in source
+    if elements_per_thread > 1:
+        # A direct complement of a packed vector load is invalid for CUDA structs.
+        assert "(~*(" not in source
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA device required")
+@pytest.mark.parametrize("dtype", DTYPES)
+@pytest.mark.parametrize("elements_per_thread", [1, 2, 4])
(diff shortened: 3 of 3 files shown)


sepcnt:
@KareemMusleh I guess we can just reuse the existing code path. https://github.com/tile-ai/tilelang/pull/3228

PR #3198, opened 2026-09-11
[Metal] Implement fragment reductions through the shared planner

anerli:
## Problem

Metal has no lowering for `T.reduce_sum`, `T.reduce_max`, `T.reduce_min` and the supported bitwise reductions. Lowering a kernel with any of them on the Metal target fails with an unregistered-implementation error, so softmax, normalization and routing kernels cannot be written portably for Apple GPUs; each one has to hand-encode lane shuffles.

## Change

1. `src/backend/common/op/reduce.h`: the shared `ReduceLowerer` gains one hook, `ThreadReduceArgs`, returning extra arguments for the per-thread reduce intrinsic. The default returns nothing, so CUDA and ROCm lowering is unchanged.
2. `src/metal/op/reduce.cc`: a Metal lowerer on top of the shared planner. It passes the thread index through the hook and rejects cross-simdgroup reductions whose participating threads are not the whole threadgroup, because those need a threadgroup barrier that every thread must reach.
3. `src/metal/codegen/reduce.h`, `codegen_metal.cc/.h`: emit SIMD lane shuffles for intra-simdgroup steps and a threadgroup workspace with barriers for cross-simdgroup steps; accumulation and batched reductions are supported.
4. `docs/compiler_internals/metal_tilelang_development.md`: implemented-features row.

## Tests

`testing/python/metal/test_metal_reduce.py` executes sum/max/min and bitwise reductions on MPS against Torch references, in FP32 and BF16, with and without accumulation, batched over rows, across simdgroup boundaries, and checks that a partial-threadgroup cross-simdgroup reduction is rejected at lowering time.

## Validation

Apple M4 Max, macOS 15.5 (24F74), Command Line Tools 16.4 / macOS SDK 15.5, Python 3.12.11, PyTorch 2.14.0, `TILELANG_DISABLE_CACHE=1`, native build of this branch on top of `main` (85fd8fc2).

- `python -m pytest testing/python/metal -q`: 100 passed, 3 skipped (Metal 4 hardware).
- The CUDA and ROCm reduction translation units compile with the hook in place; those GPUs were not available here, so their execution coverage relies on CI.
- `bash format.sh --files <changed>`, `pre-commit run --files <changed>`, `git diff --check`: clean.

## Related

#2968 proposes a separate single-simdgroup butterfly `ReduceImpl` for Metal and moves helpers out of the common header. This change instead reuses the existing shared planner through one hook, and supports cross-simdgroup reductions. Happy to reconcile with the author of #2968; the two touch the same common header.

## Summary

- Adds Metal lowering for fragment reductions through the shared reduction planner.
- Supports sum, max, min, absolute, and bitwise reductions.
- Uses SIMD shuffles for intra-simdgroup reductions.
- Uses threadgroup workspace and barriers for cross-simdgroup reductions.
- Supports accumulation and batched reductions.
- Rejects unsafe partial-threadgroup reductions.
- Adds Metal reduction code generation and documentation.
- Adds MPS coverage for supported data types, batching, accumulation, SIMD boundaries, and invalid cases.

## C++ style / lint notes

- The PR changes C++ backend code and C++ headers.
- Review `docs/developer_guide/cpp_style.md` for applicable rules.
- The C++ API Style Audit (warning only) CI step is relevant.
- No correctness, build, or test issue is identified from the provided changes.

## Validation

- 100 Metal tests passed on Apple M4 Max.
- 3 tests were skipped for Metal 4 hardware.
- Formatting, pre-commit checks, and whitespace validation passed.

### DIFF
--- a/docs/compiler_internals/metal_tilelang_development.md
+++ b/docs/compiler_internals/metal_tilelang_development.md
@@ -640,3 +640,14 @@ Default benchmark:
 TILELANG_DISABLE_CACHE=1 python benchmark/matmul_metal/benchmark_matmul_metal.py \
   --m 4096 --n 4096 --k 4096 --warmup 10 --repeats 100
 ```
+
+## 6. Fragment Reductions
+
+Metal `T.reduce` uses the shared GPU reduction planner for logical fragment
+ownership, local accumulation, and thread reduction steps. The Metal backend
+emits SIMD-group shuffles, with threadgroup workspace and barriers for steps
+that cross SIMD groups. Such steps require the entire threadgroup to participate;
+partial threadgroup participation is rejected during lowering. Sum, max, min,
+absolute sum/max, bitwise reductions, and batched outputs use this path.
+
+Regression coverage is in `testing/python/metal/test_metal_reduce.py`.
--- a/src/backend/common/op/reduce.h
+++ b/src/backend/common/op/reduce.h
@@ -633,6 +633,12 @@ inline PrimExpr MakeUpdate(const ReduceOpNode &op, PrimExpr dst_val,
 } // namespace reduce
 
 template <typename Impl> struct ReduceLowerer {
+  // Some device languages require the thread index as an explicit argument.
+  static Array<PrimExpr> ThreadReduceArgs(const LowerArgs &, const Fragment &,
+                                          int) {
+    return {};
+  }
+
   static Stmt LowerLocal(const ReduceOpNode &op, const Buffer &src_buffer,
                          const Buffer &dst_buffer,
                          const LowerArgs &lower_args) {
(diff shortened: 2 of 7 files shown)


anerli:
Both flagged items are in code this PR does not introduce.

- `TargetMetalGetWarpSize` returning 32 is the backend-wide assumption the existing Metal copy and GEMM lowerings already build on (`src/metal/op/copy.cc`, `src/metal/op/gemm.cc`). Every Apple-silicon GPU reports a 32-wide SIMD group, and the torch adapter exposes `thread_execution_width` so a launch can check the compiled pipeline against it. Making width a target parameter is a backend change, not something this lowering should do on its own.
- The `count = end - base + 1` approximation in `ResolveAllReduceThreadRange`, and the TODO about restoring `CountSatisfyingValues`, are existing common-header code that the CUDA and ROCm reductions already go through; this PR only reads its result. On Metal the cross-simdgroup case is additionally required to span the whole threadgroup, which is stricter than the CUDA path.

PR #3199, opened 2026-09-11
[BugFix][Metal] Lower fragment accumulators and register operands for simdgroup GEMM

anerli:
## Problem

A valid portable kernel that touches a GEMM accumulator after `T.gemm` compiles on CUDA and crashes the Metal lowering:

```python
@T.prim_func
def main(A: T.Tensor((32, 32), "float16"), B: T.Tensor((32, 32), "float16"), C: T.Tensor((32, 32), "float32")):
    with T.Kernel(1, threads=32):
        As = T.alloc_shared((32, 32), "float16")
        Bs = T.alloc_shared((32, 32), "float16")
        Cl = T.alloc_fragment((32, 32), "float32")
        T.copy(A, As)
        T.copy(B, Bs)
        T.clear(Cl)
        T.gemm(As, Bs, Cl)
        for i, j in T.Parallel(32, 32):
            Cl[i, j] = Cl[i, j] * 2.0
        T.copy(Cl, C)

tilelang.compile(main, target="metal", execution_backend="torch")
# InternalError: Check failed: (dtype == Float(16) || dtype == Float(32) || dtype == BFloat(16)) is false:
# Only float16, float32, and bfloat16 are supported, but got float32x4
```

Root cause:

1. `local.fragment` GEMM accumulators are rewritten into opaque `metal.simdgroup` matrices before layout inference (`metal_fragment_to_simdgroup.py`). After that point the accumulator has no thread-level layout, so element access (`C_local[i, j]`), elementwise operations and `T.reduce_*` on it reach the simdgroup codegen with vector types it cannot represent; a kernel cannot do GEMM -> softmax -> GEMM while keeping the intermediate in registers.
2. The same rewrite means GEMM operands can only come from shared memory. A fragment produced by one GEMM (or by elementwise code) cannot feed the next GEMM.

Neither limitation exists on the `metal.cooperative_tensor` path, where fragments keep their regular scope through layout inference. That path is gated on Metal 4, so today it covers M5. The simdgroup path is what every M1 through M4 GPU runs, and the docs mark it as the legacy fragment treatment. That seems like the wrong split: the pre-M5 installed base is the Apple hardware that exists in volume, and a kernel written once against `T.alloc_fragment` and `T.gemm` should lower on both paths with the same semantics. This change brings the simdgroup path up to the same fragment contract rather than leaving it behind.

## Change

1. Keep logical fragments through layout inference and supply native simdgroup fragment layouts, so element access lowers to native matrix element reads and writes.
2. `tilelang/metal/transform/legalize_simdgroup.py` (replacing `metal_fragment_to_simdgroup.py`): legalize vector access widths on simdgroup-backed fragments after layout inference.
3. `src/metal/op/gemm.cc`, `tilelang/metal/op/gemm/gemm_metal.py`, `metal_macro_generator.py`, `codegen_metal.cc/.h`: fragment/shared operand combinations for A and B, transposes, warp partition policies and tails, with the emitter handling register operands.
4. `tilelang/metal/pipeline.py`: run the new legalization at the right point.
5. `docs/compiler_internals/metal_tilelang_development.md`: feature and implementation-map rows.

Fragment inputs currently require full buffer regions; that limitation is enforced at lowering time.

## Tests

`testing/python/metal/test_metal_fragment_access.py` and `test_metal_gemm_fragments.py`: elementwise operations on accumulators, fragment copies and initialization, repeated GEMMs, all operand combinations and transposes, warp policies, and tail shapes, executed on MPS against Torch. `test_metal_gemm_v2_linux.py` and `test_metal_simdgroup_store.py` are updated for the new lowering.

## Validation

Apple M4 Max, macOS 15.5 (24F74), Command Line Tools 16.4 / macOS SDK 15.5, Python 3.12.11, PyTorch 2.14.0, `TILELANG_DISABLE_CACHE=1`, native build of this branch on top of `main` (85fd8fc2).

- `python -m pytest testing/python/metal -q`: 111 passed, 3 skipped (Metal 4 hardware).
- `bash format.sh --files <changed>`, `pre-commit run --files <changed>`, `git diff --check`: clean.

## Related

Independent of #2970 (simdgroup store synchronization); no shared files beyond `codegen_metal.cc`.

## Summary

- Reworked Metal SIMD-group GEMM lowering to keep fragment accumulators logical through layout inference.
- Added post-layout vector-access legalization with native simdgroup fragment layouts.
- Added fragment element access, elementwise operations, reductions, copies, initialization, and repeated GEMM support.
- Updated GEMM lowering and Metal code generation for fragment/shared operands, transposes, warp policies, register operands, and partial tiles.
- Enforced full buffer regions for fragment inputs during lowering.
- Replaced the legacy `MetalFragmentToSimdgroup` pass with `LegalizeSimdgroupVectorization`.
- Updated Metal documentation and tests for fragment operations and GEMM combinations.
- Validation passed 111 Metal tests, with 3 skipped for Metal 4 hardware. Formatting, pre-commit checks, and whitespace validation also passed.

## C++ style / lint notes

- The PR changes C++ code in the Metal code generator.
- No C++ API Style Audit warnings are reported.
- No correctness or build issues are reported. Any warning-only style findings are advisory.


anerli:
On the `bfloat16` fragment-load type: correct for this branch in isolation, and the reason is a dependency rather than a gap here. `PrintType` on `main` prints every BF16 vector as the scalar `bfloat`, for fragment loads and for ordinary vectorized loads alike. #3194 fixes that (`bfloat2`, `packed_bfloat3`, `bfloat4`), and this branch is cut from `main` without it. With #3194 in, the vector constructor emitted here prints the right type. Duplicating the fix in this PR would conflict with #3194; a BF16 fragment-access regression test can be added once both are merged.

PR #3204, opened 2026-09-11
[CUDA] Fuse exact FP4 to FP8 conversion through FP32

ZenAlexa:
## Motivation

The [DeepSeek V4.1 expert GEMM](https://huggingface.co/deepseek-ai/DeepSeek-V4.1-Flash/blob/dba1be0a40aa45a94ad051997016db3960a90277/inference/kernel.py#L477-L559) converts E2M1 weights through FP32 into E4M3 on every K tile. Every E2M1 value, including signed zero, has an exact E4M3 encoding.

## Implementation

I fuse this unannotated CUDA cast chain into two `__byte_perm` operations per four elements. The code handles scalars and vector widths 2, 4, 8, 16, and 32. Explicit cast annotations retain their existing lowering. The conversion uses registers and adds no global or shared-memory allocation.

## Validation

- Native source-generation tests cover both E4M3 dtype spellings, all supported widths, and explicit annotations.
- The [upstream H100 job](https://github.com/tile-ai/tilelang/actions/runs/34607356388/job/103289438636) passed the source-generation cases and all four exhaustive PyTorch runtime cases. Each runtime case checks every packed four-nibble word against an independent encoding table. The subsequent helper rename preserves the generated CUDA operations; the renamed source-generation cases and CUDA compilation pass.
- The extracted GEMM produces identical BF16 outputs in every configuration below. An independent sampled CPU reference also passes.
- Project hooks pass for all three contribution files. Current-head CUDA, ROCm, Metal, and CuTeDSL CI checks pass.

## H100 performance

The benchmark uses the linked GEMM with `T.symbolic` adapted to `T.dynamic`: 128 threads, tiles of 32 by 128 by 32, two pipeline stages, 10,240 bytes of shared memory, FP32 accumulation, and BF16 output. The consumer pins TileLang 0.1.8; the benchmark uses compiler `c629fe5e` and this patch. Baseline CUDA source matches the unmodified compiler's output.

Both variants run on the same H100 80 GB HBM3 with CUDA 13.0, targeting `sm_90a`, and use identical inputs and power-of-two scales. Each graph contains 100 calls. Values are per-call medians from ten retained alternating-order CUDA-event measurements after warmup.

| M | N | K | Baseline (µs) | This PR (µs) |
|---:|---:|---:|---:|---:|
| 1 | 2,304 | 5,120 | 547.94 | 108.65 |
| 8 | 2,304 | 5,120 | 549.32 | 108.59 |
| 32 | 2,304 | 5,120 | 550.15 | 111.02 |
| 128 | 2,304 | 5,120 | 551.82 | 107.71 |
| 512 | 2,304 | 5,120 | 717.30 | 224.21 |
| 2,048 | 2,304 | 5,120 | 2,069.77 | 641.31 |
| 1 | 5,120 | 2,304 | 245.17 | 45.85 |
| 8 | 5,120 | 2,304 | 245.16 | 45.60 |
| 32 | 5,120 | 2,304 | 244.54 | 46.31 |
| 128 | 5,120 | 2,304 | 292.08 | 70.83 |
| 512 | 5,120 | 2,304 | 586.63 | 157.61 |
| 2,048 | 5,120 | 2,304 | 1,760.22 | 591.92 |

Register usage is 128 for the baseline and 156 for this PR; both have zero spills. These measurements cover the complete extracted GEMM under graph replay. Blackwell execution and complete-model serving remain unverified.

## Summary

- Fuses the default CUDA `FP4 E2M1 → FP32 → FP8 E4M3` conversion into direct `__tl_cvt` intrinsics.
- Supports scalar values and vector widths 2, 4, 8, 16, and 32.
- Preserves explicitly annotated cast behavior.
- Registers the conversion as CUDA-vectorizable.
- Adds source-generation and exhaustive packed-value correctness tests.
- Uses registers without global or shared-memory allocations.

## Performance

- H100 expert GEMM measurements show 2.97x–5.38x speedups across twelve configurations.
- The fused path produces no spills.
- Register usage increases from 128 to 156 registers.

## C++ style / lint notes

- The PR changes C++ code and a CUDA kernel template.
- The C++ API Style Audit (warning only) remains relevant.
- No correctness or build issue is identified in the supplied evidence.
- TLCPP003/TLCPP004 findings are advisory and should not block merging without a clear API, FFI, or maintainability risk.

### DIFF
--- a/src/cuda/codegen/codegen_cuda.cc
+++ b/src/cuda/codegen/codegen_cuda.cc
@@ -1706,6 +1706,51 @@ void CodeGenTileLangCUDA::VisitExpr_(const CastNode *op, std::ostream &os) {
   bool cast_sat = get_bool_anno("sat", true);
   Optional<PrimExpr> cast_rbits = get_expr_anno("rbits");
 
+  // The FP32 intermediate preserves every E2M1 value, including signed zero.
+  // Emit packed E4M3 encodings directly for this exact conversion chain.
+  const auto *inner_cast = op->value.as<CastNode>();
+  if (inner_cast && op->annotations.empty() &&
+      inner_cast->annotations.empty() && from_ty.is_float() &&
+      from_ty.bits() == 32 && inner_cast->value.dtype().is_float4_e2m1fn() &&
+      (target_ty.is_float8_e4m3() || target_ty.is_float8_e4m3fn())) {
+    DataType packed_ty = inner_cast->value.dtype();
+    int lanes = packed_ty.lanes();
+    if (lanes == 1) {
+      this->PrintType(target_ty, os);
+      os << "::bitcast(static_cast<uint8_t>(__tl_cvt_e2m1x4_to_e4m3x4(("
+         << PrintExpr(inner_cast->value) << ").__x)))";
+      return;
+    }
+    if (lanes == 2 || lanes == 4 || lanes == 8 || lanes == 16 || lanes == 32) {
+      std::string packed = SSAGetID(PrintExpr(inner_cast->value), packed_ty);
+      std::string result = name_supply_->FreshName("fp8_cast");
+      PrintIndent();
+      PrintType(target_ty, stream);
+      stream << " " << result << ";\n";
+      for (int first_lane = 0; first_lane < lanes; first_lane += 4) {
+        std::string converted = name_supply_->FreshName("fp8_bits");
+        PrintIndent();
+        stream << "uint32_t " << converted << " = __tl_cvt_e2m1x4_to_e4m3x4("
+               << "reinterpret_cast<const uint8_t*>(&" << packed << ")["
+               << first_lane / 2 << "]";
+        if (first_lane + 2 < lanes) {
+          stream << " | (uint16_t(reinterpret_cast<const uint8_t*>(&" << packed
+                 << ")[" << first_lane / 2 + 1 << "]) << 8)";
+        }
+        stream << ");\n";
+        for (int lane = first_lane; lane < first_lane + 4 && lane < lanes;
+             ++lane) {
+          PrintIndent();
+          stream << "reinterpret_cast<uint8_t*>(&" << result << ")[" << lane
+                 << "] = static_cast<uint8_t>(" << converted << " >> "
+                 << (lane - first_lane) * 8 << ");\n";
+        }
+      }
+      os << result;
+      return;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


ZenAlexa:
I renamed the helper to `ConvertE2M1x4ToE4M3x4` and updated its code-generation call sites and source assertions.

LeiWang1999 (commented):
Thanks for the PR — the numerics are right (I checked both `__byte_perm` tables and the sign selector by hand; the exhaustive test is a nice touch) and the speedup is real.

Rather than leave line comments I pushed a follow-up commit (885476d4) to this branch that restructures the codegen side. What changed and why:

- **Split the fold from the intrinsic.** `VisitExpr_(CastNode)` now just rewrites `Cast(e4m3, Cast(f32, e2m1))` with no annotations into `Cast(e4m3, e2m1)` and lets the normal per-dtype-pair table handle it. The exactness argument lives in one 8-line block at the top.
- **Byte plumbing moved into `cuda_fp4.h`.** Codegen shouldn't index bytes out of `fp4_e2_N_t` structs or scatter result bytes; every other cast maps a dtype pair to a `__tl_cvt_*` helper and lets `PrintVectorizedCast` chunk. `ConvertE2M1x4ToE4M3x4` became `__tl_cvt_e2m1x4_to_e4m3x4(__nv_fp4x4_storage_t) -> __nv_fp8x4_storage_t` with `x2` and scalar wrappers, following the header's existing naming (the PascalCase lint hit is advisory; consistency with the neighbouring helpers matters more).
- **`PrintVectorizedCast` gained a `chunk_lanes` parameter** (default 2, existing call sites unchanged) so this path consumes four lanes per call.
- **Direct `T.Cast("float8_e4m3fn", fp4)` takes the same fast path** instead of the elementwise float fallback; tests are parametrized over both the folded chain and the direct cast.

Generated code for 8 lanes is now one line per four lanes:

```cpp
(reinterpret_cast<__nv_fp8x4_storage_t*>(&_1))[0] = __tl_cvt_e2m1x4_to_e4m3x4((reinterpret_cast<__nv_fp4x4_storage_t*>(&_2))[0]);
```

SASS on sm_90a: same instruction count as before, 0-byte stack frame, 12 PRMT instead of 16. All 72 tests in `test_tilelang_language_vectorized_cast.py` and the neighbouring fp4 test files pass locally.

Please take a look; if it looks good to you we can merge once CI is green.

PR #3205, opened 2026-09-12
[BugFix][Language] Make T.clamp propagate NaN

### LINKED ISSUE #3024: [BUG][Fuzzer][wrong-code] `T.clamp` silently drops `NaN` and returns the `min` bound instead of propagating `NaN`
### Required prerequisites

- [x] I have tried the latest version of TileLang.

0.1.13

### System information

NVIDIA L40S (sm_89), CUDA 13.0, PyTorch 2.13.0, Python 3.13. The behavior is architecture-independent — the wrong value comes from the CUDA `fminf`/`fmaxf` NaN semantics, which are identical on every NVIDIA target; only sm_89 was run.

### Problem description

`T.clamp(x, lo, hi)` on a `NaN` input silently returns `lo` (the lower bound) instead of `NaN`. The reference semantics `T.clamp` should match are **`torch.clamp` and `numpy.clip`, both of which propagate `NaN`** (verified this session) — these are the frameworks a TileLang kernel is written against and differentially tested against. The kernel compiles and runs to completion with no error, so the wrong value is silent.

Clamping is used in the shipped fp8-quantization examples (`T.clamp(y / y_s, fp8_min, fp8_max)` in `examples/cast/example_group_per_split_token_cast_to_fp8.py:48`, `example_per_token_cast_to_fp8.py:35`, and the DeepSeek act-quant examples), where a `NaN` in the input would be quantized to `fp8_min` rather than surfacing.

`T.clamp` lowers to `min(max(x, lo), hi)`; the emitted CUDA is `min(max(A[i], -1.0f), 1.0f)`. `NaN` fed to CUDA `max`/`min` (`fmaxf`/`fminf`) returns the non-`NaN` operand, so `max(NaN, -1.0f)` gives `-1.0f` and the `NaN` is gone before the outer `min` runs.

The same `NaN`-suppression is visible in `T.max`/`T.min` on their own (not just via `T.clamp`) — see the neighbor note below.

Not a regression: `clamp` has been `min(max(...))` since it was added.

### Reproducible example code

```python

N = 4

@T.prim_func
def main(A: T.Tensor((N,), "float32"), C: T.Tensor((N,), "float32")):
    with T.Kernel(1, threads=N) as bx:
        i = T.get_thread_binding()
        C[i] = T.clamp(A[i], T.float32(-1.0), T.float32(1.0))

kernel = tilelang.compile(main, out_idx=[1])

a   = torch.tensor([float('nan'), 2.0, 0.5, -5.0], dtype=torch.float32, device="cuda")
got = kernel(a).cpu()
ref = torch.clamp(a.cpu(), -1.0, 1.0)                 # trusted oracle
print("T.clamp :", got.tolist())                      # -> [-1.0, 1.0, 0.5, -1.0]  NaN -> -1.0 (min bound)
print("torch   :", ref.tolist())                      # -> [nan, 1.0, 0.5, -1.0]   NaN preserved
print("NaN preserved:", torch.isnan(got[0]).item())   # -> False  (should be True)

# control: with no NaN present, T.clamp matches torch exactly
b = torch.tensor([2.0, 0.5, -5.0, 0.0], dtype=torch.float32, device="cuda")
print("finite ctrl OK:", torch.equal(kernel(b).cpu(), torch.clamp(b.cpu(), -1.0, 1.0)))  # -> True
```

### Traceback

### Expected behavior

`T.clamp(NaN, lo, hi)` returns `NaN`.

The reference semantics that matter most here are **`torch.clamp` and `numpy.clip`** — both propagate `NaN` (a `NaN` input yields a `NaN` output; verified this session on 0.1.13's PyTorch and NumPy). These are the libraries TileLang kernels are written against and validated against, so `T.clamp` diverging from them on `NaN` is the core defect. The finite-input control already matches `torch` exactly, so only a `NaN` operand disagrees.

This also aligns with the newer IEEE-754 direction: 754**-2019** `minimum`/`maximum` propagate `NaN` (whereas the older 754**-2008** `minNum`/`maxNum` — which is what CUDA's `fmaxf`/`fminf` follow — suppress it). So "propagate" is not a single universal rule for min/max, but it is what both dominant array libraries do and what the current IEEE revision moved to; the concrete ask is that `T.clamp` match `torch`/`numpy`. Whether that is achieved by making `T.clamp` `NaN`-transparent or by exposing an opt-in is the maintainers' call.

### Additional context

**Root cause.** `T.clamp` composes clamping out of `T.max`/`T.min`, whose CUDA lowering does not preserve `NaN`. `clamp` is defined as `T.min(T.max(dst, min_val), max_val)` at [`customize.py#L55-L56`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/customize.py#L55); on CUDA `T.max`/`T.min` on floats emit `max`/`min`, i.e. `fmaxf`/`fminf`, which return the non-`NaN` argument. So `T.clamp(NaN, lo, hi)` computes `min(max(NaN, lo), hi) = min(lo, hi) = lo`.

<details>
<summary>Emitted CUDA + the T.max/T.min neighbor (both run on 0.1.13, L40S)</summary>

Emitted device expression for the repro:

```c++
C[((int)threadIdx.x)] = min(max(A[((int)threadIdx.x)], -0x1p+0f), 0x1p+0f);   // min(max(A[i], -1.0f), 1.0f)
```

The building blocks drop `NaN` on their own, so this is not specific to `clamp`:

```
T.max(nan, 0.0) -> 0.0    (torch.maximum -> nan)
T.min(nan, 0.0) -> 0.0    (torch.minimum -> nan)
```

A fix targeting `clamp`'s composition (e.g. an explicit `dst != dst` re-injection of `NaN`, or a `NaN`-propagating min/max) closes `T.clamp`; the `T.max`/`T.min` primitives keep the CUDA `fmaxf`/`fminf` behavior unless they are separately changed.
</details>

**Suggested fix.** One direction is to make `clamp` `NaN`-transparent — carry the input `NaN` through, or build it from a `NaN`-propagating min/max — so `T.clamp(NaN, …)` returns `NaN` like `torch.clamp`. Whether `T.max`/`T.min` themselves should also propagate `NaN` (vs keep the `fmaxf`/`fminf` reduction semantics used by softmax-style `m_i = T.max(m_i, …)` accumulators) is a separate call for the maintainers. Not built/verified.

**Provenance.** `clamp` was introduced in [#192](https://github.com/tile-ai/tilelang/pull/192) ("[Language] Support clamp in language", merged 2025-03-12) with the `min(max(...))` body it still has; the `NaN`-dropping behavior has been present since then. Not a regression.

**Dedup.** I searched the open and closed tracker for `clamp` / `NaN` / `min max`. Distinct from [#2882](https://github.com/tile-ai/tilelang/issues/2882) (`T.max`/`T.min` on two compile-time float **constants** fold order-dependently on `NaN` via the host `const_fold.h` path); that report treats the **runtime** `max(...)` lowering as the well-defined reference — this issue is that the runtime lowering itself drops `NaN` versus `torch.clamp`/`numpy.clip` (and IEEE-754-2019). No existing report covers `T.clamp`.

**Reach.** `T.clamp` is documented and appears in shipped examples. On 0.1.13 the cited sites are present and verbatim: `T.clamp(y_local[i,j] / y_s_local[i], fp8_min, fp8_max)` at `examples/cast/example_per_token_cast_to_fp8.py:35` and `example_group_per_split_token_cast_to_fp8.py:48`, and `T.clamp(x_local[i,j] / s_local[i], fp8_min/fp8_max)` at `examples/deepseek_v32/inference/kernel.py:72` and `examples/deepseek_v4/act_quant.py:78,156`. The underlying `T.max`/`T.min` appear in ~100 example sites (softmax `m_i = T.max(...)`, `T.max(x, 0)` ReLU).

I **ran** `example_per_token_cast_to_fp8.py`'s shipped kernel verbatim on 0.1.13 (L40S). Two results: (1) with the example's shipped input (`torch.randn`, all finite) the correctness check passes exactly — the shipped example **dodges** this bug because no `NaN` ever reaches the clamp, and no test in the tree injects one. (2) Injecting a single `NaN` into one input row of the same kernel makes that row diverge from the torch reference: torch propagates `NaN` across the whole 128-element group (row0 fp8 = `[nan, nan, …]`) while TileLang emits finite fp8 values (`[-448.0, -60.0, -384.0, …]`). So the bug is reachable through the shipped kernel; the shipped test just never exercises `NaN`. (This particular divergence is compounded by `T.reduce_absmax` also dropping the `NaN` by default — see the generalization table.) The trigger requires a `NaN` reaching the clamp; finite inputs clamp correctly (control passes). CI is green because no test asserts `NaN` propagation through `T.clamp`.

**Generalization.** Root, two levels: (a) SOURCE-level — `clamp` at `tilelang/language/customize.py#L55-L56` composes out of `T.max`/`T.min` (`tilelang/language/tir/ir.py:184-193` → TIR `min`/`max`), which lower to the CUDA NaN-dropping intrinsics (`fmaxf`/`fminf` for fp32, `__hmax`/`__hmin` for fp16/bf16 — never the `_nan` variants). (b) OPERATOR-level — every op built on scalar `T.max`/`T.min` inherits the drop: `T.clamp`, elementwise `T.max`/`T.min` themselves, and (by default) the `reduce_max`/`reduce_min`/`reduce_absmax` family. 4-axis sweep, all cells RUN on 0.1.13 (L40S):

| axis | cell tested (input → observed) | same-root? |
|---|---|---|
| related-operator | `T.max(nan,0.0)→0.0`, `T.min(nan,0.0)→0.0` (torch: `nan`); but `T.max(nan,nan)→nan` (`fmaxf` returns the other operand, so `NaN` survives only when BOTH are `NaN`) | same root (elementwise; the clamp building blocks) |
| related-operator | `T.reduce_absmax` fp32, default: `[…,nan,…]→3.0` (torch amax(abs)=`nan`) | same root, **distinct face** — reduce family has a `nan_propagate=True` opt-in (`reduce_op.py`, GH-#2697) that clamp lacks; but it **defaults to dropping** and is fp16/bf16-only |
| related-type | `T.clamp` fp16: `clamp(nan,-1,1)→-1.0` (device emits `__hmax`/`__hmin`, not `__hmax_nan`) | same root |
| related-type | `T.clamp` bf16: `clamp(nan,-1,1)→-1.0` (device emits `cutlass::bfloat16 __hmax/__hmin`) | same root |
| related-type | `T.clamp` int32/int (no `NaN` representation) | N/A — does not apply (integers have no `NaN`) |
| similar-logic | reduce family got a per-call `nan_propagate` escape hatch; elementwise `T.max`/`T.min`/`T.clamp` have NO such kwarg (`git grep nan_propagate` hits only `reduce_op.py`) | same root, asymmetric coverage — the fix already exists for reductions but not for the elementwise/clamp composition |

Framing kept **specific to `T.clamp`** (title/Problem unchanged): the elementwise `T.max`/`T.min` NaN-drop is arguably intended (softmax `m_i = T.max(...)` accumulators rely on `fmaxf` reduction semantics), and the reduce family already has an opt-in, so the crisp defect is that `T.clamp` — which per `torch.clamp`/IEEE should be `NaN`-transparent — has no way to preserve `NaN`. The reduce-default NaN-drop (`reduce_absmax` etc.) is a **distinct, adjacent** behavior with an existing opt-in and test coverage; not filed here.

**Impact.** Silent-wrong, deterministic, bounded to the clamped value: a `NaN` at the input becomes the `min` bound (`lo`) with no error, so a `NaN` that a workload relies on to detect (or propagate through a quantization step) is silently replaced by a finite in-range value every run. Fixing it aligns `T.clamp` with the `torch.clamp`/`numpy.clip` `NaN` semantics kernels are validated against (and the IEEE-754-2019 direction) and removes a silent frontend/runtime disagreement.

xy200303:
## Summary

Fixes #3024.

`T.clamp` was `T.min(T.max(x, lo), hi)`. On CUDA those lower to the `fmaxf`/`fminf` family, which returns the *non*-NaN operand, so `T.clamp(NaN, lo, hi)` silently returned `lo` — a wrong value with no error — where `torch.clamp` and `numpy.clip` both propagate `NaN`:

```
T.clamp : [-1.0, 1.0, 0.5, -1.0]
torch   : [nan, 1.0, 0.5, -1.0]
```

The composition now re-injects an input `NaN`:

```python
clamped = T.min(T.max(dst, min_val), max_val)
if DataType(dst.dtype).type_code in _NON_FLOAT_TYPE_CODES:
    return clamped
return T.if_then_else(T.isnan(T.cast(dst, "float32")), dst, clamped)
```

Two details worth flagging:

- `tir.isnan` is implemented for only a subset of the float dtypes — it raises `InternalError: Data type bfloat16 not supported for isnan op` for `bfloat16`, `float8_e4m3` and `float8_e5m2`. The predicate is therefore evaluated on an `fp32` cast, which preserves `NaN`, instead of on `dst` directly. Without that, the naive `T.isnan(dst)` version breaks the four existing fp8 clamp tests.
- Dtypes that cannot hold a `NaN` skip the guard, so integer kernels are unchanged and do not grow an int→float cast.

`T.max`/`T.min` themselves are deliberately untouched. As the issue notes, their `fmaxf`/`fminf` behaviour is arguable for softmax-style `m_i = T.max(m_i, ...)` accumulators, and whether they should propagate is a separate call; this PR fixes `clamp`, which has no such excuse because it is specified against `torch.clamp`/`numpy.clip`.

## Test plan

Added `test_clamp_propagates_nan` to `testing/python/language/test_tilelang_language_clamp.py`, parametrized over `float16`, `bfloat16`, `float32`, `float64`, `float8_e4m3` and `float8_e5m2`. Each case puts one `NaN` in the input and asserts it survives, then checks the finite elements exactly against `torch.clamp`.

Hardware: NVIDIA A100-PCIE-40GB (sm80), CUDA 12.8, torch 2.8.0+cu128, tilelang 0.1.14 wheel. The issue reported on sm_89; this reproduces and fixes identically on sm80, which is consistent with the claim that the behaviour comes from the `fmaxf`/`fminf` semantics rather than anything architecture-specific.

A/B on the same machine, same test file:

| | result |
|---|---|
| before the fix | `6 failed, 11 passed` — the new test fails for all six dtypes; the 11 pre-existing tests pass |
| after the fix | `17 passed` |

I also ran a probe outside the test file to confirm the guard does not disturb the integer paths: `T.clamp` on `int32`, `int8` and `uint8` still compiles and matches `torch.clamp` exactly.

`ruff check` and `ruff format --check` are clean on both changed files.

## Notes

- Not a regression: `clamp` has had this body since #192, as the issue reports.
- The five shipped fp8 quantization examples that call `T.clamp` (`examples/cast/example_per_token_cast_to_fp8.py`, `example_group_per_split_token_cast_to_fp8.py`, `examples/deepseek_v32/inference/kernel.py`, `examples/deepseek_v4/act_quant.py`) call it on `fp32` values before casting to `fp8`, so they now surface a `NaN` instead of quantizing it to `fp8_min`. Their shipped tests use finite inputs and are unaffected.

## Summary

- Fixed `T.clamp` to propagate floating-point `NaN` values.
- Preserved finite-value clamping and existing integer behavior.
- Added regression coverage for fp16, bf16, fp32, fp64, and supported FP8 types.
- Verified finite outputs against `torch.clamp`.
- Kept `T.min` and `T.max` unchanged.

### DIFF
--- a/testing/python/language/test_tilelang_language_clamp.py
+++ b/testing/python/language/test_tilelang_language_clamp.py
@@ -115,6 +115,60 @@ def test_clamp():
     run_clamp_value_range(1024, 128, T.float32)
 
 
+CLAMP_NAN_DTYPES = [
+    T.float16,
+    T.bfloat16,
+    T.float32,
+    T.float64,
+    T.float8_e4m3,
+    T.float8_e5m2,
+]
+
+
+def clamp_nan_kernel(N, block_N, dtype):
+    @T.prim_func
+    def main(
+        A: T.Tensor((N,), dtype),
+        C: T.Tensor((N,), dtype),
+    ):
+        with T.Kernel(T.ceildiv(N, block_N), threads=block_N) as bx:
+            for i in T.Parallel(block_N):
+                C[bx * block_N + i] = T.clamp(A[bx * block_N + i], T.cast(-1.0, dtype), T.cast(1.0, dtype))
+
+    return main
+
+
+@tilelang.testing.requires_cuda
+@pytest.mark.parametrize("dtype", CLAMP_NAN_DTYPES)
+def test_clamp_propagates_nan(dtype):
+    """``T.clamp`` must propagate ``NaN`` the way ``torch.clamp`` does.
+
+    ``T.clamp`` composes ``T.min(T.max(...))``, and on CUDA those lower to the
+    ``fmaxf``/``fminf`` family, which returns the non-``NaN`` operand. Without an
+    explicit re-injection a ``NaN`` input silently becomes ``min_val``.
+    """
+    import torch
+
+    N = 32
+    kernel = tilelang.compile(clamp_nan_kernel(N, N, dtype))
+    torch_dtype = dtype.as_torch()
+
+    a = torch.linspace(-2.0, 2.0, N, device="cuda", dtype=torch.float32)
+    a[7] = float("nan")
+    a = a.to(torch_dtype)
+
+    # Caller-allocated output: rebuilding a torch tensor from an fp8 DLPack
+    # tensor is not supported on every torch version.
+    C = torch.empty(N, device="cuda", dtype=torch_dtype)
+    kernel(a, C)
+    got = C.float()
+
+    assert torch.isnan(got[7]).item(), f"T.clamp dropped NaN for {dtype}: got {got[7].item()} instead of NaN"
+    ref = torch.clamp(a.float(), -1.0, 1.0)
+    finite = ~torch.isnan(ref)
+    torch.testing.assert_close(got[finite], ref[finite], atol=0, rtol=0)
+
+
 FP8_OPS = {
     "max": lambda a, b, dtype: T.max(a, b),
     "min": lambda a, b, dtype: T.min(a, b),
--- a/tilelang/language/customize.py
+++ b/tilelang/language/customize.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 from tilelang._typing import ShapeType, DType, BufferLikeType
 import tilelang.language as T
-from tvm import arith
+from tvm import DataType, DataTypeCode, arith
 from tvm.tirx import PrimExpr, Buffer, op
 from tilelang.utils.language import bits_product, prim_expr_equal, retrieve_buffer_and_offset
 from .atomic import atomic_max, atomic_min, atomic_add, atomic_addx2, atomic_addx4, atomic_load, atomic_or, atomic_store  # noqa: F401
@@ -41,8 +41,23 @@ def dp4a(A: BufferLikeType, B: BufferLikeType, C: BufferLikeType) -> PrimExpr:
     )
 
 
+_NON_FLOAT_TYPE_CODES = (
+    DataTypeCode.INT,
+    DataTypeCode.UINT,
+    DataTypeCode.BOOL,
+    DataTypeCode.HANDLE,
+)
+
+
 def clamp(dst: PrimExpr, min_val: PrimExpr, max_val: PrimExpr) -> PrimExpr:
     """Clamps the input value dst between [min_val, max_val]
+
+    A ``NaN`` input yields ``NaN``, matching ``torch.clamp`` and ``numpy.clip``.
+    ``T.max``/``T.min`` lower to the CUDA ``fmaxf``/``fminf`` family, which return
+    the non-``NaN`` operand, so the composition below would otherwise replace a
+    ``NaN`` with ``min_val`` silently. Only floating-point dtypes can hold a
+    ``NaN``, and ``tir.isnan`` is not implemented for every one of them, so the
+    predicate is evaluated on an ``fp32`` cast, which preserves ``NaN``.
 
     Args:
         dst: Input value to be clamped
@@ -52,9 +67,10 @@ def clamp(dst: PrimExpr, min_val: PrimExpr, max_val: PrimExpr) -> PrimExpr:
     Returns:
         Value clamped to the specified range
     """
-    dst = T.max(dst, min_val)  # Ensure value is not less than minimum
-    dst = T.min(dst, max_val)  # Ensure value is not greater than maximum
-    return dst
+    clamped = T.min(T.max(dst, min_val), max_val)
+    if DataType(dst.dtype).type_code in _NON_FLOAT_TYPE_CODES:
+        return clamped
+    return T.if_then_else(T.isnan(T.cast(dst, "float32")), dst, clamped)
 
 
 def reshape(src: Buffer, shape: ShapeType) -> Buffer:


LeiWang1999:
Pushed 9697ef89 to use a dedicated `tl.clamp` intrinsic and device templates, following the NaN-aware reduction helpers.

The Python guard could evaluate an input expression more than once and prevented explicit vector operands from lowering correctly. The new path evaluates each operand once and preserves vectorization: half/bf16 pairs reuse `max2_nan`/`min2_nan`, while float32 keeps vector loads and stores. NaNs in either bound now propagate as well. CPU templates and shared lowering cover the other targets.

Added regression tests for NaN bounds, reversed bounds, type promotion, scalar/vector operands, and returning atomics. The clamp suites pass with **47 passed, 2 skipped**; the related CPU vector, bf16 math, and reduce NaN suites have **16 passed**. The native build and `./format.sh` also pass. Runtime validation was on CUDA and CPU; other backends were checked through source generation or intrinsic lowering.

LeiWang1999:
Fixed the ROCm CI failure in f948e146. The HIP codegen test used `arch="gfx942"`, but HIP targets require `mcpu="gfx942"`; it failed while constructing the target, before code generation.

Reproduced the original `ValueError` locally and verified that all five test target configurations now construct successfully. The target suite has **11 passed, 2 skipped** locally, and `./format.sh` passes. Full HIP validation is left to the new ROCm CI run since this environment has no ROCm backend.

PR #3207, opened 2026-09-12
[BugFix][JIT] Allocate a dynamic-shape output that precedes its sizing input

### LINKED ISSUE #3017: [BUG][Fuzzer][ice-on-valid-code] Dynamic-shape output placed before the input supplying its symbolic dim crashes at call time (`IndexError`) instead of allocating the output
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/tile-ai/tilelang/issues) and [Discussions](https://github.com/tile-ai/tilelang/discussions) that this hasn't already been reported. (comment there if it has.)

0.1.13 (latest release)

### System information

TileLang 0.1.13 / PyTorch 2.8 / CUDA 12.8, single NVIDIA L40S (sm_89). Backend `tvm_ffi` (the default) and `cython` both affected.

### Problem description

A `@tilelang.jit` kernel whose **output** tensor has a symbolic (dynamic) dimension crashes at call time with `IndexError: list index out of range` whenever that output parameter appears in the signature *before* the input parameter that supplies the symbolic size. The kernel compiles fine; the crash is in the host wrapper that allocates the output.

The size resolution assumes every symbolic output dim is backed by an *already-allocated* tensor. The wrapper builds its tensor list in parameter order in a single pass, so when the output comes first it looks up a slot that has not been filled yet.

The precise root is in `_process_dynamic_symbolic`: it records each symbolic dim against the **first parameter (in signature order) whose buffer shape mentions it**, and does **not exclude output buffers**. When the output is the first param to mention the dim, the recorded owner *is the output itself*, and the allocation loop then reads the output's own not-yet-appended slot (`tensor_list[ref_tensor_idx]` where `ref_tensor_idx` is the output's own index). It is a self-reference, not (as one might assume) a reference to a later input. Verified by inspecting the map: for `main(B(N,), A(N,))` with `out_idx=[0]` the map is `{'N': (0, 0, 0, 1)}` — owner index 0 = B, the output.

The same kernel with the output parameter placed *after* the input (the shape it copies from) works correctly, which is why every shipped example — and the dynamic-symbolic test, which uses `out_idx=[-1]` — dodges it.

<details><summary>Console output (repro below): crash vs working control</summary>

```
output-before-input: CRASH IndexError: list index out of range
output-after-input : PASS
```

Traceback for the crashing case:

```
  File ".../tilelang/jit/adapter/tvm_ffi.py", line 255, in func
    shape.append(tensor_list[ref_tensor_idx].shape[ref_shape_idx])
                 ~~~~~~~~~~~^^^^^^^^^^^^^^^^
IndexError: list index out of range
```
</details>

### Reproducible example code

```python
N = T.symbolic("N")

# out_idx=[0]: output B is param 0, input A is param 1.
# B's shape (N,) is resolved from a LATER input param -> crash.
@tilelang.jit(out_idx=[0])
def out_first():
    @T.prim_func
    def main(B: T.Tensor((N,), "float32"), A: T.Tensor((N,), "float32")):
            for i in T.Parallel(N):
                B[i] = A[i] + T.float32(9)
    return main

# Control: same kernel, output B placed AFTER input A.
@tilelang.jit(out_idx=[1])
def out_last():
    @T.prim_func
    def main(A: T.Tensor((N,), "float32"), B: T.Tensor((N,), "float32")):
            for i in T.Parallel(N):
                B[i] = A[i] + T.float32(9)
    return main

a = torch.randn(64, device="cuda")
ref = a.double() + 9

try:
    b = out_first()(a)
    print("output-before-input:", "PASS" if torch.allclose(b.double(), ref, atol=1e-4) else "FAIL")
except Exception as e:
    print(f"output-before-input: CRASH {type(e).__name__}")   # -> IndexError

b = out_last()(a)
print("output-after-input :", "PASS" if torch.allclose(b.double(), ref, atol=1e-4) else "FAIL")  # -> PASS
```

### Traceback

```
  File ".../tilelang/jit/kernel.py", line 204, in __call__
    return self.torch_function(*args, **kwds)
  File ".../tilelang/jit/adapter/tvm_ffi.py", line 255, in func
    shape.append(tensor_list[ref_tensor_idx].shape[ref_shape_idx])
                 ~~~~~~~~~~~^^^^^^^^^^^^^^^^
IndexError: list index out of range
```

### Expected behavior

The kernel should allocate the output and run regardless of where the output parameter sits in the signature — as it already does when the output is placed last. `out_idx` is documented as "Index(es) of the output tensors to return" with no ordering restriction, so an output at index 0 whose dim is shared with a later input is a legal signature that the compiler already compiles; only the runtime output-allocation step fails on it.

### Additional context

**Root cause.** The symbolic-dim owner recording does not exclude output buffers, so when the output is the first parameter to name a shared dim it records itself as that dim's owner, and the output-allocation loop then reads the output's own not-yet-appended slot.

Two-level:

- **Source root (where the wrong index is recorded).** [`_process_dynamic_symbolic`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/tvm_ffi.py#L145-L175) records each symbolic shape dim against the **first parameter, in signature order, whose buffer shape mentions it** — [`dynamic_symbolic_map[shape] = (0, i, j, 1)`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/tvm_ffi.py#L166), gated only by `shape not in dynamic_symbolic_map` (first-wins) — with **no exclusion of `result_idx` (output) parameters**. So if the output is the first param to name the dim, the owner index it stores is the output's own index.
- **Consumption root (where it crashes).** In [`_convert_torch_func.func`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/tvm_ffi.py#L224-L282) the allocation loop walks parameters in order and for an output param resolves each symbolic dim via [`tensor_list[ref_tensor_idx].shape[ref_shape_idx]`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/tvm_ffi.py#L255). `tensor_list` only holds the params processed *before* the current one, so a `ref_tensor_idx` at-or-after the current output — including the **self-reference** where the owner is the output itself — indexes past the end and raises `IndexError`.

The self-reference case is the common one: whenever the output precedes every input that shares the dim, `_process_dynamic_symbolic` points the dim at the output's own slot.
</details>

<details><summary>Why the control passes, and why it hits both backends</summary>

When the output is last (`out_idx=[-1]`, the form used everywhere in `examples/`/`testing/`), an input already names the shared dim, so `_process_dynamic_symbolic` records that input as the owner; by the time the output is allocated the input is already in `tensor_list`, so the lookup succeeds. The `cython` backend carries the identical single-pass resolution at [`cython_wrapper.pyx#L202`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/cython/cython_wrapper.pyx#L202) and crashes the same way (tested: `IndexError` on the repro kernel with `execution_backend="cython"`).
</details>

**Suggested fix.** Two independent points would each close it: (1) in `_process_dynamic_symbolic`, prefer a non-output (`result_idx`) parameter as the recorded owner of a shape dim, so a shared dim is always bound to an input that is populated before any output; and/or (2) in the allocation loop of both adapters, split into two passes (bind all inputs first, then allocate outputs) so `tensor_list[ref_tensor_idx]` is always populated, or resolve output shapes from a position-keyed inputs map rather than the `tensor_list` under construction. Both adapters carry the same allocation-loop shape (not verified end-to-end).

**Provenance.** The single-pass `tensor_list[ref_tensor_idx]` resolution plus the output-inclusive owner recording are present in both adapters at the 0.1.13 release; I did not pin the introducing PR. Not established as a regression of previously-correct behavior.

**Dedup.** Searched the open and closed tracker and found no existing report of this defect.

**Tested boundary.** The trigger is exactly whether the **recorded owner index** of a symbolic dim (the first buffer param, in order, that mentions it — recorded in [`_process_dynamic_symbolic`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/jit/adapter/tvm_ffi.py#L145-L175)) is >= the index of the output being allocated. Because output buffers are not excluded from that recording, the common case is a **self-reference**: the output is the first param to name the dim. Cells run on 0.1.13/L40S (full sweep in the Generalization table below): (a) 2-param output-first (`out_idx=[0]`) — `IndexError`, map `N:(0,0,0)` (owner = the output itself); (b) same kernel `execution_backend="cython"` — identical `IndexError` (`cython_wrapper.pyx:202`); (c) 3-param middle output (`out_idx=[1]`) whose dim is owned by input idx 0 (processed first) — runs correctly, map `N:(0,0,0)`. So (a)/(b) and (c) share one root; the boundary is owner-index-vs-output-index, not the output's absolute position.

**Generalization — two-level root.** (a) *Source-level:* `_process_dynamic_symbolic` (`tvm_ffi.py:161-166`, twin in `cython_wrapper.pyx`) records a symbolic dim against the first buffer param that mentions it, first-wins, **not excluding output (`result_idx`) buffers**. (b) *Operator-level:* every path that returns a symbolic-shaped output through the same allocation loop — single output, multi-output, any output dtype/rank — since the symbol resolution runs before the output tensor is created and is dtype/rank-agnostic. 4-axis sweep, all cells run on 0.1.13 / L40S:

| axis | cell tested (input) | observed result | same-root? |
|---|---|---|---|
| repro (baseline) | `out_idx=[0]`, `main(B(N,),A(N,))`, call `(a[64])` | `IndexError` @ `tvm_ffi.py:255`; map `N:(0,0,0)` (owner=output) | yes |
| related-source | same kernel, `execution_backend="cython"` | `IndexError` @ `cython_wrapper.pyx:202` | yes (identical logic, twin file) |
| related-source | `out_idx=[1]`, `main(X(8,)static, B(M,)out, A(M,)in)` — var first named by output | `IndexError`; map `M:(0,1,0)` (owner=output idx1, **self-ref**, not the later input idx2) | yes — sharpens root to self-reference |
| related-operator | multi-output `out_idx=[0,1]`, `main(B(N,),C(N,),A(N,))` | `IndexError` | yes |
| related-operator | scalar-owner: `out_idx=[0]`, `main(B(N,), n:int32, A(N,))` | `IndexError` @ 255 — N still recorded as buffer-shape (ref_id 0) against output B, NOT the scalar (ref_id 2); scalar `ref_id==2` branch (`inputs[...]`) never reached | yes (same shape-loop path) |
| related-type | 2D int8 output `out_idx=[0]`, `main(B(M,4)int8, A(M,4)int8)` | `IndexError` | yes (resolution precedes tensor creation; dtype/rank orthogonal) |
| control (dodge) | `out_idx=[1]`, `main(A(N,)in, B(N,)out)` — input names dim first | PASS, correct value | n/a (owner idx0 < output idx1) |
| control (dodge) | 3-param `out_idx=[1]`, `main(A(N,)in idx0, B out, C in)` | PASS | n/a (owner idx0 processed first) |

**Reframe.** Kept the specific title/framing but the Problem now names the class: *output is the first param to name a shared symbolic dim* (subsumes both "output before its input" and the self-reference case). The `ref_id==1` stride branch (`tvm_ffi.py:257`, `cython_wrapper.pyx` twin) is code-identical to the crashing shape branch and shares the exact `tensor_list[ref_tensor_idx]` lookup, so it is **same-root by inspection** (symbolic output strides are uncommon and awkward to express via `T.Tensor`, not run here). No distinct-adjacent bug and no new bug surfaced during the sweep. The scalar-owner cell revealed the draft's original "later input" narrative was imprecise — the recorded owner is a *buffer* (here the output itself), and the `ref_id==2` scalar branch that reads the fully-available `inputs` list is not the crashing path.

**Impact.** The trigger is narrow: a dynamic-shape kernel whose symbolic output parameter is the first param to name a shared dim (e.g. declared before the input it borrows the dim from) — an ordering the shipped `out_idx=[-1]` convention never uses. When it fires it is a deterministic call-time `IndexError` in the host output-allocation wrapper: the kernel never runs, nothing is computed, and the failure is loud and immediate rather than a silent wrong value, so it cannot reach a workload's results undetected. Fixing it closes an output-parameter-ordering class that `out_idx` documents as legal but the runtime allocator rejects, aligning the two adapters' allocation path with what the compiler already accepts.

**Reach.** (Run, not asserted.) The trigger is a legal signature (`out_idx` is documented with no ordering constraint) with two ingredients: a **symbolic output dimension** shared with an input (common in dynamic-shape kernels) and the output being the **first parameter to name that dim** (e.g. placed before the input). Placing the output last is the overwhelmingly common convention. I ran the shipped citations on 0.1.13/L40S: the dynamic-symbolic test `testing/python/profiler/test_tilelang_profiler_dynamic_symbolic.py` (`out_idx=[-1]`) — **11 passed** (output C last, dim owned by input A idx 0, dodges); the canonical `examples/dynamic_shape/example_dynamic.py` (eager `C = T.empty((M,N),...)`, `out_idx=[-1]`) — **runs, matches torch** (output constructed after inputs, dodges). A grep of `examples/` at v0.1.13 found no `out_idx=[0]` and no symbolic kernel whose output precedes its owning input; the only symbolic example with a non-trailing out_idx is `deepseek_v32/inference/kernel.py` (`out_idx=[4]`, output after its inputs, dodges). So no shipped example or test exercises the output-first ordering that trips it, which is why CI is green.

xy200303:
## Summary

Fixes #3017.

A `@tilelang.jit` kernel whose **output** has a symbolic dimension and appears *before* the input that supplies that dimension crashed at call time with `IndexError: list index out of range`. The kernel compiled fine; the crash was in the host wrapper that allocates the output.

Two things had to line up, and both are fixed:

**1. The symbolic dimension was owned by the output itself.** `_process_dynamic_symbolic` walked the parameters in signature order and recorded each symbolic dimension against the first parameter whose buffer mentioned it, without excluding outputs. For `main(B(N,), A(N,))` with `out_idx=[0]` the entry came out as `N: (0, 0, 0, 1)` — owner index `0` is `B`, the output being allocated. Inputs are now visited first, so the entry is `N: (0, 1, 0, 1)` — the input `A`.

**2. The caller built its tensor list in a single pass.** Even with the owner pointing at an input, an output at parameter 0 would still have read `tensor_list[1]` before it was filled. Both adapters now place all input tensors first and allocate the outputs afterwards, in parameter order.

`tilelang/jit/adapter/tvm_ffi.py` and `tilelang/jit/adapter/cython/{adapter.py,cython_wrapper.pyx}` get the same two changes; they are duplicated implementations of the same logic.

## Test plan

Added `test_cython_dynamic_shape_output_before_input` to `testing/python/jit/test_tilelang_jit_cython.py`.

Hardware: NVIDIA A100-PCIE-40GB (sm80), CUDA 12.8, tilelang built from source at `66c003c` (the `.pyx` change has to be compiled, so the pip wheel cannot exercise it).

A/B on the same tree, stashing only `tilelang/jit/adapter/` and rebuilding in between:

| backend | shape | before | after |
|---|---|---|---|
| `cython` | output last (control) | OK, map `N: (0, 0, 0, 1)` | OK, map unchanged |
| `cython` | **output first** | **`IndexError`** at `cython_wrapper.pyx:202` | **OK**, results match, map `N: (0, 1, 0, 1)` |
| `tvm_ffi` | output first | OK — see note | OK |

The new test is `1 failed` before and `1 passed` after. `testing/python/jit/test_tilelang_jit_cython.py` is `1 passed, 1 skipped` after (the skip is the PDL test, which requires sm90). `testing/python/profiler/test_tilelang_profiler_dynamic_symbolic.py` is `11 passed`.

Two things worth being explicit about:

- **The `tvm_ffi` adapter's preallocated-output path is not the default on current main.** It is only taken when the host target is not the C host, because the callee-allocated-output ABI otherwise short-circuits `_convert_torch_func` entirely. I fixed it because the defect is identical and the code is a duplicate, but the test above covers the `cython` backend, which is where the bug is reachable today. I exercised the `tvm_ffi` path separately by forcing `_uses_ffi_callee_allocated_output_abi()` to return `False`: before the fix it raises the same `IndexError` with `N: (0, 0, 0, 1)`, after the fix it completes with `N: (0, 1, 0, 1)`. Note that forcing only the Python side leaves the lowered function on the callee-allocated ABI, so the numerical result in that setup is meaningless for both sides of that comparison — it is the disappearance of the `IndexError` and the owner change that it demonstrates.
- **Renamed API.** The reproducer in the issue uses `T.symbolic`, which current main renamed to `T.dynamic` (`tilelang/language/symbolics.py`, `__all__ = ["dynamic"]`). I reproduced and tested against `T.dynamic`; the defect is unchanged.

`ruff check` and `ruff format --check` are clean on the changed Python files, and `clang-format` reports no changes on the `.pyx`.

## Adjacent observation (not changed here)

In the same resolution loop, the scalar-parameter case (`ref_id == 2`) indexes the caller's argument list with the *parameter* index:

```python
if ref_id == 2:
    shape.append(inputs[ref_tensor_idx])
```

That is only correct while no output precedes the scalar parameter. It looks wrong for the same reason the reported bug was, but I have no reproduction for it and did not want to change behaviour I cannot demonstrate, so I left it alone and mention it rather than guess.

## Summary

- Fixed dynamic-shape output allocation when an output precedes the input that supplies its symbolic dimension.
- Updated the `tvm_ffi` and Cython adapters to resolve inputs before outputs while preserving parameter positions.
- Fixed scalar symbolic-dimension lookup in `tvm_ffi` and added a clear error for unset slots.
- Added regression tests for output-before-input ordering in both adapters.
- Preserved existing output-last behavior.
- CUDA test execution remains pending because the environment lacks a GPU and TileLang build artifacts.

### DIFF
--- a/testing/python/jit/test_tilelang_jit_cython.py
+++ b/testing/python/jit/test_tilelang_jit_cython.py
@@ -57,5 +57,34 @@ def test_cython_pdl():
     print("pdl test passed!")
 
 
+@tilelang.testing.requires_cuda
+def test_cython_dynamic_shape_output_before_input():
+    """An output may precede the input that supplies its symbolic dimension.
+
+    The symbolic-dimension map used to be built over every parameter in signature
+    order, outputs included, and the caller assembled its tensor list in a single
+    pass. When the output came first it was therefore recorded as the owner of its
+    own dimension and then read its own not-yet-filled slot.
+    """
+    N = T.dynamic("N")
+
+    @tilelang.jit(out_idx=[0], execution_backend="cython")
+    def kernel():
+        @T.prim_func
+        def main(
+            B: T.Tensor((N,), "float32"),
+            A: T.Tensor((N,), "float32"),
+        ):
+            with T.Kernel(1, threads=128):
+                for i in T.Parallel(N):
+                    B[i] = A[i] + T.float32(9)
+
+        return main
+
+    a = torch.arange(128, dtype=torch.float32).cuda()
+    out = kernel()(a)
+    tilelang.testing.torch_assert_close(out, a + 9, atol=0, rtol=0)
+
+
 if __name__ == "__main__":
     tilelang.testing.main()
--- a/tilelang/jit/adapter/cython/adapter.py
+++ b/tilelang/jit/adapter/cython/adapter.py
@@ -226,13 +226,21 @@ class CythonKernelAdapter(BaseKernelAdapter):
         params = func.params
         buffer_map = func.buffer_map
         dynamic_symbolic_map = {}
-        for i, param in enumerate(params):
+        # Inputs are visited first. An output's shape is resolved from these entries
+        # while that output is being allocated, so a dimension mentioned by both an
+        # input and an output must be owned by the input; owning it on the output
+        # would make the allocation loop read a slot it has not filled yet.
+        ordered = [i for i in range(len(params)) if i not in self.result_idx]
+        ordered += [i for i in range(len(params)) if i in self.result_idx]
+        for i in ordered:
+            param = params[i]
             if param in buffer_map:
                 buffer = buffer_map[param]
                 for j, shape in enumerate(buffer.shape):
                     if isinstance(shape, tirx.Var) and (shape not in dynamic_symbolic_map) and (shape not in params):
                         dynamic_symbolic_map[shape] = (0, i, j, 1)
-        for i, param in enumerate(params):
+        for i in ordered:
+            param = params[i]
             if param in buffer_map:
                 buffer = buffer_map[param]
                 element_bits = buffer.dtype.bits * buffer.dtype.lanes
--- a/tilelang/jit/adapter/cython/cython_wrapper.pyx
+++ b/tilelang/jit/adapter/cython/cython_wrapper.pyx
@@ -185,10 +185,18 @@ cdef class CythonKernelWrapper:
                 stream = 0
 
         cdef int ins_idx = 0
-        cdef list tensor_list = []
+        cdef list tensor_list = [None] * len(self.params)
         device = None
 
-        # Prepare input and output tensors
+        # Inputs are placed first so that a symbolic dimension owned by an input can be
+        # resolved even when the output that needs it comes earlier in the signature;
+        # the outputs are then allocated below, in parameter order.
+        for i in range(len(self.params)):
+            if i not in self.result_idx:
+                tensor_list[i] = inputs[ins_idx]
+                ins_idx += 1
+
+        # Prepare output tensors
         for i in range(len(self.params)):
             if i in self.result_idx:
                 dtype = self.param_dtypes[i]
@@ -212,20 +220,16 @@ cdef class CythonKernelWrapper:
                         f"Cannot create output tensor (name={param_name}) - 0-dimensional tensors are not supported. "
                         f"Expected shape: {shape}"
                     )
-                tensor = torch.empty(*shape, dtype=dtype, device=device)
-            else:
-                tensor = inputs[ins_idx]
-                ins_idx += 1
-            # TODO(chenggang): remove this check or rewrite by ourselves?
-            '''
-            if isinstance(tensor, torch.Tensor) and tensor._base is not None and not tensor.is_contiguous():
-                base_tensor = tensor._base.as_strided(tensor._base.shape, tensor.stride())
-                if torch._debug_has_internal_overlap(base_tensor):
-                    raise ValueError(f"Cannot use an overlapping tensor"
-                                     f"(shape={tensor.shape}, strides={tensor.stride()}, "
-                                     f"overlap={torch._debug_has_internal_overlap(base_tensor)}) as the kernel input")
-            '''
-            tensor_list.append(tensor)
+                tensor_list[i] = torch.empty(*shape, dtype=dtype, device=device)
+                # TODO(chenggang): remove this check or rewrite by ourselves?
+                '''
+                if isinstance(tensor_list[i], torch.Tensor) and tensor_list[i]._base is not None and not tensor_list[i].is_contiguous():
+                    base_tensor = tensor_list[i]._base.as_strided(tensor_list[i]._base.shape, tensor_list[i].stride())
+                    if torch._debug_has_internal_overlap(base_tensor):
+                        raise ValueError(f"Cannot use an overlapping tensor"
+                                         f"(shape={tensor_list[i].shape}, strides={tensor_list[i].stride()}, "
+                                         f"overlap={torch._debug_has_internal_overlap(base_tensor)}) as the kernel input")
+                '''
 
         # Convert tensor pointers to C void pointers for kernel call
         cdef dict dtype_to_ctype = {
--- a/tilelang/jit/adapter/tvm_ffi.py
+++ b/tilelang/jit/adapter/tvm_ffi.py
@@ -172,13 +172,21 @@ class TVMFFIKernelAdapter(BaseKernelAdapter):
         for i, param in enumerate(params):
             if isinstance(param, tirx.Var) and (param not in dynamic_symbolic_map):
                 dynamic_symbolic_map[param] = (2, i, -1, 1)
-        for i, param in enumerate(params):
+        # Inputs are visited first. An output's shape is resolved from these entries
+        # while that output is being allocated, so a dimension mentioned by both an
+        # input and an output must be owned by the input; owning it on the output
+        # would make the allocation loop read a slot it has not filled yet.
+        ordered = [i for i in range(len(params)) if i not in self.result_idx]
+        ordered += [i for i in range(len(params)) if i in self.result_idx]
+        for i in ordered:
+            param = params[i]
             if param in buffer_map:
                 buffer = buffer_map[param]
                 for j, shape in enumerate(buffer.shape):
                     if isinstance(shape, tirx.Var) and (shape not in dynamic_symbolic_map) and (shape not in params):
                         dynamic_symbolic_map[shape] = (0, i, j, 1)
-        for i, param in enumerate(params):
+        for i in ordered:
+            param = params[i]
             if param in buffer_map:
                 buffer = buffer_map[param]
                 element_bits = buffer.dtype.bits * buffer.dtype.lanes
@@ -252,11 +260,18 @@ class TVMFFIKernelAdapter(BaseKernelAdapter):
                 None,
             )
 
-            # Stitch the full positional argument list expected by the TVM executable
+            # Stitch the full positional argument list expected by the TVM executable.
+            # Inputs are placed first so that a symbolic dimension owned by an input can
+            # be resolved even when the output that needs it comes earlier in the
+            # signature; the outputs are then allocated in parameter order.
             ins_idx: int = 0
-            tensor_list: list[torch.Tensor] = []
+            tensor_list: list[torch.Tensor | None] = [None] * len(self.params)
+            for i in range(len(self.params)):
+                if i not in self.result_idx:
+                    tensor_list[i] = inputs[ins_idx]
+                    ins_idx += 1
 
-            # Prepare input and output tensors
+            # Prepare output tensors
             for i in range(len(self.params)):
                 if i in self.result_idx:
                     dtype = param_dtypes[i]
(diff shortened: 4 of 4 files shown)


LeiWang1999:
@regression-perf

xy200303:
Addressed the outstanding `tvm_ffi.py` finding in `420493d`.

## Scalar dimension lookup (`tilelang/jit/adapter/tvm_ffi.py`)

Confirmed still valid, and the root cause is the one described. `_process_dynamic_symbolic` stores **PrimFunc parameter indices**: a scalar variable parameter is recorded as `(2, param_index, -1, 1)`. The output-allocation loop resolved that branch as `inputs[ref_tensor_idx]`, but `inputs` contains only the non-output parameters. For a signature whose output precedes its inputs the index therefore addresses the wrong slot, and for `[B_output, A_input, N_scalar]` it runs past the end of the list.

The scalar branch now resolves against `tensor_list`, which is sized and indexed by parameter position (the same change the non-scalar branch already received). It also reports a clear error if the referenced scalar slot is unset rather than raising `IndexError`. The two branches are now consistent.

## Test coverage

Added `test_tvm_ffi_dynamic_shape_output_before_input`, which exercises the changed `TVMFFIKernelAdapter` path with `out_idx=[0]` and `execution_backend="tvm_ffi"`, placing the output before the input that supplies its symbolic dimension.

## Verification still outstanding — please read

I could not execute the tests: this machine has no GPU (`torch 2.13.0+cpu`, `torch.cuda.is_available() == False`) and tilelang's `build/lib` is not present, so the suite cannot be built or run here. What I did verify:

- `ruff check` — all checks passed on both changed files
- `ruff format --check` — already formatted
- both files parse

So the fix is reviewed by inspection and lint-clean, but it is **not** backed by a GPU run from my side. Before this is merged it should be exercised on CUDA with the dynamic-shape TVM FFI tests, including the new case. If the scalar-parameter signature needs a specific construction to be exercised, that may also need a follow-up test — I'd rather flag that than claim coverage I could not demonstrate.

LeiWang1999 (approved):

PR #3215, opened 2026-09-12
[Metal] Support runtime-valid GEMM rows

anerli:
## Problem

Kernels with a runtime-valid prefix of an otherwise static GEMM tile must either compute padded rows, branch to a different implementation, or abandon the hardware matrix path. Decode and routed-expert workloads need to retain static tile shapes while avoiding invalid M-axis instruction rows.

## Change

- Add an optional target-neutral `valid_m` argument to `T.gemm` while preserving static physical tile dimensions.
- Preserve the existing block-scaled GEMM call protocol and default ordinary GEMMs to the full M extent.
- Require each GEMM implementation to declare support before accepting a partial runtime M extent.
- Implement Metal simdgroup and cooperative-tensor lowering with whole-instruction-row predicates for loads, multiply-accumulate, initialization, and writeback.
- Keep unpredicated fast paths for complete per-warp tiles; the invalid suffix remains unspecified.
- Add construction, generated-source, and Metal execution regressions.

## Tests

- Rebuilt TileLang with `USE_METAL=ON USE_CUDA=OFF USE_ROCM=OFF` from this branch's exact source.
- `pytest -q testing/python/metal/test_metal_gemm_valid_m.py`: 3 passed.
- Remaining Metal suite: 21 passed, 3 skipped.
- `pre-commit run --files ...` and `git diff --check`: passed.

## Validation

Executed a runtime `valid_m=17` GEMM on Apple Metal, verified the valid prefix against PyTorch, and verified wholly invalid instruction rows remain untouched. Two existing upstream codegen-expectation files were excluded from the directory run because they independently fail against upstream `main`.

## Related

Independent of the buffer-region-offset bug fix. This extends the target-neutral GEMM contract and supplies its Metal realization without exposing Metal-specific controls to callers.

## Summary

- Add optional `valid_m` support to `T.gemm`.
- Preserve full-tile behavior by default.
- Add validation and backend capability checks for runtime partial M extents.
- Add Metal predication for loads, accumulation, initialization, and writeback.
- Preserve fast paths for complete warp tiles.
- Add construction, code-generation, and Metal execution tests.

## C++ style / lint notes

- The PR touches FFI-visible GEMM fields and reflection.
- The C++ style guide requires preserving FFI-visible field names.
- The C++ API Style Audit runs in warning-only mode in CI.
- TLCPP003/TLCPP004 findings are advisory unless they indicate an API, FFI, or maintainability risk.
- No correctness, build, or test issue is reported.

## Tests

- 3 dedicated tests pass.
- 21 additional Metal tests pass.
- 3 tests are skipped.
- Metal execution with `valid_m=17` matches PyTorch for valid rows and preserves invalid accumulator rows.

### DIFF
--- a/src/op/gemm.cc
+++ b/src/op/gemm.cc
@@ -76,6 +76,7 @@ void RegisterGemmImpl(GemmImpl impl) {
  *      M (Int), N (Int), K (Int), policy (Int), clear_accum (Bool),
  *      (optional) mbar (BufferLoad or const-0 placeholder),
  *      cCoord_y (PrimExpr), cCoord_x (PrimExpr),
+ *      (optional) valid_m (PrimExpr), or
  *      (optional, blockscaled) SFA, SFB regions, k_start (PrimExpr)]
  *   Backend lowering knobs (k_pack, wg_wait) ride in the annotations map.
  */
@@ -101,6 +102,7 @@ Gemm::Gemm(Array<PrimExpr> args, Map<String, ObjectRef> annotations) {
   node->k_ = args[7].as<IntImm>().value()->value;
   node->policy_ = GemmWarpPolicy(args[8].as<IntImm>().value()->value);
   node->clearAccum_ = args[9].as<PrimExpr>().value();
+  node->validM_ = IntImm(DataType::Int(32), node->m_);
   // k_pack rides in the annotations (a ROCm MFMA/WMMA lowering knob set by
   // the ROCm dialect), not in the positional call protocol.
   if (auto val = annotations.Get("k_pack")) {
@@ -131,13 +133,11 @@ Gemm::Gemm(Array<PrimExpr> args, Map<String, ObjectRef> annotations) {
   }
   node->cCoords_ = Array<PrimExpr>(
       {args[11].as<PrimExpr>().value(), args[12].as<PrimExpr>().value()});
-  if (args.size() > 13) {
+  if (args.size() == 14) {
+    node->validM_ = args[13].as<PrimExpr>().value();
+  } else if (args.size() > 13) {
     node->sfaRegion_ = NormalizeToBufferRegion(args[13]);
-  }
-  if (args.size() > 14) {
     node->sfbRegion_ = NormalizeToBufferRegion(args[14]);
-  }
-  if (args.size() > 15) {
     node->sfKStart_ = args[15].as<PrimExpr>().value();
   }
   node->annotations_ = annotations;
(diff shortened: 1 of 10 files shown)


LeiWang1999:
Thanks for your contribution! but have no idea of this argument, could you explain more about it?

anerli:
The idea is to be able to specify how many rows of a statically shaped T.gemm need to be computed if you don't want to compute all

My use case for this is avoiding multiplying padded token rows

Here is an example where I am using it if that helps: https://github.com/magnitudedev/magnitude/blob/tilelang/inference-v3/src/ops/kernels/packed.py#L152

LeiWang1999:
Thanks for the explanation and for linking the real kernel (`grouped_experts.py` / `packed.py`) — that made the intent clear. Having read it, I don't think `valid_m` belongs on `T.gemm`, so I'm going to close this one. Reasoning:

1. **Correctness is already covered by the existing grouped-GEMM pattern.** Your kernel is structurally the same as `examples/grouped_gemm/example_grouped_gemm_fwd.py`: pad each group to `block_M`, compute the full static tile, mask the epilogue with `if i < actual_rows`. `_group_routes` producing `block_metadata[block] = (expert, valid_rows)`, the `i < valid_m` gather and the `i < valid_m` writeback do exactly that. Dropping `valid_m=` from the `T.gemm` call leaves the kernel correct; the argument is purely a performance knob.

2. **No performance evidence.** The only thing `valid_m` buys is skipping some 8-row MMA instruction rows on Metal. In decode-shaped expert GEMMs the B tile (here the quantized codes + coefficients) is loaded in full regardless of `valid_m`, and the workload is usually weight-bandwidth bound, so it is not obvious the saved MMA work is visible at all. A new `T.gemm` argument needs a benchmark showing the win on a real shape.

3. **Target-neutral API with a single-target implementation.** The kwarg is added to the generic, CUDA and ROCm dialects but only Metal implements it; everywhere else it fails at lowering. We don't want to grow the `T.gemm` contract with a parameter that only one backend honours.

4. **Semantics don't match TileLang's conventions.** `valid_m` means "only the first N rows are guaranteed; the rest (including `clear_accum` init) is unspecified at 8-row granularity". Every other boundary mechanism in TileLang masks or pads to keep the whole buffer well-defined; a partially-undefined accumulator is easy to misuse as a correctness mechanism.

If the padded rows really are the bottleneck in your decode path, the tools that don't change the GEMM contract are: a smaller `block_M` for the decode variant, or the `if valid_m == bm` branch you already have selecting a narrower-tile implementation. Happy to revisit if you can share numbers showing the predicated path beats those on Apple hardware.

(Also noting for the record: the branch currently conflicts with `main` in `src/op/gemm.cc`, `tilelang/language/gemm_op.py` and `metal_macro_generator.py` after #3237 / #3209, and the `args.size()==14` discriminator would need rewriting anyway.)

Issue #3220, opened 2026-09-13
[Transform][Arith] ConstrSet::Populate ordering is load-bearing in two incompatible directions (eager Analyzer::Bind snapshots)


LeiWang1999:
## Summary

`ConstrSet::Populate` (in `src/transform/common/constr_visitor.h`) replays a lexically-collected list of binds and predicates into a fresh `arith::Analyzer`. The replay **order** turns out to be load-bearing in two *incompatible* directions:

- **Program order** (current upstream behavior, rationale documented since #2805) is required so that a bind evaluated under a preceding predicate captures the tightened bound.
- **Binds-first** (what our downstream Ascend fork carries) is required so that a predicate held in a bound boolean var can be expanded back into the condition it represents, even after `Merge` / `RenameFrom` scrambled the definition order.

Both orderings are workarounds for the same root cause: `arith::Analyzer::Bind` **eagerly snapshots** the value's integer bounds at bind time and never re-evaluates them. We would like to converge on one shared implementation instead of carrying a downstream divergence, and we have empirical data plus three candidate fixes below. We are happy to send a PR once a direction is agreed on.

## Background

`ConstrVisitor` (introduced in #1622, used by ThreadSync since #1631, extended with snapshot mode in #3174) collects two kinds of entries while walking the IR:

- **binds** — `v = expr` (flat `Bind`) or `v ∈ range` (`thread_extent`), and
- **predicates** — branch conditions, asserts, assumes.

`ConstrSet::Populate` replays them into an analyzer so passes can ask `CanProve` questions (cross-thread race checks, and in our fork, sync-flag allocation).

The two analyzer entry points have **opposite information-flow requirements**:

- `Analyzer::Bind(v, expr)` eagerly evaluates and caches the value's bounds *at bind time* (vendored TVM, `src/arith/analyzer.cc`):

  ```cpp
  void Analyzer::Bind(const Var& var, const PrimExpr& expr, bool allow_override) {
    ...
    this->const_int_bound.Update(var, this->const_int_bound(new_expr), allow_override);
    this->modular_set.Update(var, this->modular_set(new_expr), allow_override);
    ...
  }
  ```

  So a bind wants every predicate that constrains its value to be entered **before** it.

- `EnterConstraint(pred)` digests the predicate at entry. If `pred` is a bound boolean var (`cond` where `cond = (i < n)` is a bind), the analyzer can only digest it after simplification expands it — which requires the bind to be installed **before** the predicate.

[...]

PR #3221, opened 2026-09-13
[Transform] Restore bind-before-use order when merging constraint sets

LeiWang1999:
Implements option 1 from #3220 (dependency-aware ordering in `ConstrSet::Merge`), plus the `Simplify`-at-entry change it requires.

## Problem

`ConstrSet::Merge` concatenates two lexically-collected sets, so an entry of the first set can read a bind var the second set defines (shared vars after `RenameFrom`, merged per-task sets downstream). Replaying that order digests the predicate before its definition exists — `EnterConstraint` evaluates eagerly, exactly like `Bind` — and the guard fact is silently lost: guard disjointness becomes unprovable on merged sets.

## Change

- `Merge` normalizes the concatenated list (`ConstrSet::Normalized`): entries keep their order, except that a bind is hoisted, together with the binds its own definition reads, directly before the first entry that reads its var. Binds cannot form cycles in SSA form; a defensive cycle guard degrades to collection order with a warning instead of crashing. Within a single lexically-collected set nothing reads a later bind, so this is a no-op there and the #2805 program-order tightening guarantee is untouched (covered by a test).
- `Constr::Populate` simplifies a predicate against the already-installed binds before entering it, so a guard held in a bound boolean var (`cond` where `cond = (i < n)` is a bind) expands into the condition it represents.
- `tl.analysis.ConstrSetsProve` exposes merge + populate + `CanProve` to Python so the merged-set scenarios are testable without staging a full pass pipeline.

Known residue (documented in the code): a hoisted bind can still miss bounds that a first-set predicate would have tightened — that is inherent to eager `Bind` snapshots and only a lazily-evaluated bind (#3220, option 3) removes it. Worth noting the fork's `z3_prover` already stores definitions rather than snapshots, so only the fast sub-analyzers are order-sensitive.

## Validation

- New `testing/python/transform/test_tilelang_transform_constr_set_merge.py`: scrambled predicate-before-bind recovers the guard, transitive definition chains hoist in dependency order, assume predicates expand, #2805 program-order tightening preserved, unbound guard stays unprovable — 5 passed.
- Consumer suites (`test_tilelang_transform_thread_sync.py`, `test_tilelang_transform_verify_parallel_loop.py`, `test_tilelang_transform_auto_schedule.py`, `testing/python/analysis/`): 101 passed, no behavior change.
- `pre-commit` (clang-format, ruff check/format, codespell) on the changed files and `git diff --check`: clean.

Closes nothing on its own — #3220 stays open for the option-3 (lazy bind) follow-up discussion.

## Summary

- Restores bind-before-use ordering when merging `ConstrSet` instances.
- Hoists recursively dependent binds and warns on cyclic definitions.
- Simplifies `kConstr` predicates with installed binds during population.
- Adds `tl.analysis.ConstrSetsProve` for Python-based merge and proof checks.
- Adds regression tests for dependency ordering, predicate expansion, program-order tightening, cycle recovery, and unbound guards.

## C++ style / lint notes

- The PR changes C++ code but does not change the rules documented in `docs/developer_guide/cpp_style.md`.
- CI runs the **C++ API Style Audit (warning only)** step.
- Any TLCPP003/TLCPP004 findings are advisory and should not block merge without a clear API, FFI, or maintainability risk.


LeiWang1999:
@regression-perf

PR #3230, opened 2026-09-15
[Frontend] Support Python iterables and comprehensions

sepcnt:
Separate device loops from compile-time Python iteration instead of globally replacing `range`.
- Preserve serial lowering for `for ... in range(...)`.
- Support Python iterables, `enumerate`, `zip`, and comprehensions.
- Preserve variable scope and distinguish compile-time from device loop control.
This also enables expression-list iteration without allocating local storage:
```python
ks = [k0, k0 + 1, k0 + 8, k0 + 9]
for i, k in enumerate(ks):
    output[i] = X[row, k]
```
Fixes #2451. 
Fixes #2308; indexing Python lists with `T.unroll` variables remains unchanged.

## Summary

Adds compile-time Python iteration support in `@T.prim_func`.

- Supports lists, tuples, comprehensions, generator expressions, `enumerate()`, and `zip()`.
- Keeps `range` and `T.serial` as device-side loops.
- Adds loop and comprehension variable scoping.
- Handles `break` and `continue` in compile-time loops.
- Rejects TIR values and invalid iterables with targeted errors.
- Substitutes expression-list elements without allocating local storage.
- Preserves Python-list indexing with `T.unroll` variables.
- Adds documentation and eager-AST tests.

## Testing

Tests were added, but test execution results were not provided. Review severity counts are unavailable.

PR #3232, opened 2026-09-15
[BugFix] Bind loop targets independently of mutable scalar variables

### LINKED ISSUE #3231: [BUG] Reusing an alloc_var name as a loop target causes an out-of-scope error
### Required prerequisites

 0.1.14 and 0.1.14+cuda.gitd178b4c9

### System information

Reproduced in:

- Linux, Python 3.13, TileLang 0.1.14, PyTorch 2.7.1+cu128, RTX 5090.
- Windows, Python 3.13, TileLang source build at d178b4c9, PyTorch 2.11.0+cu130, RTX 5060 Laptop.

The failure occurs during frontend IR construction, before CUDA compilation.

### Problem description

Reusing a previous T.alloc_var name as the induction variable of a subsequent loop raises an out-of-scope error.

The second loop introduces a new induction variable, but the frontend treats its binding as an assignment to the previous mutable scalar. Renaming only the second loop target avoids the error.

### Reproducible example code

```python
import tilelang.language as T
@T.prim_func
def kernel(A: T.Tensor((2,), "int32")):
    with T.Kernel(1, threads=1):
        for j in T.serial(2):
            d = T.alloc_var("int32")
            d = j
            A[j] = d
        for d in T.Parallel(2):
            A[d] = d
Control case: replace the second loop with:
for k in T.Parallel(2):
    A[k] = k
```

### Traceback

```pytb
Relevant excerpt:
File ".../repro_loop_rebind.py", line 17, in kernel
    A[d] = d
File ".../tilelang/language/eager/builder.py", in rval
    raise RuntimeError(...)
RuntimeError: Immutable variable `d` is used outside its defining region!
The error dump shows the second loop incorrectly storing its induction variable into the old buffer:
d[0] = v
```

### Expected behavior

The second loop should bind d to its own induction variable, without writing to the previous alloc_var. The kernel should produce [0, 1].
Actual reads of expired bindings should remain rejected.

### Additional context

In Builder.bind(), an existing local.var binding takes the mutable-assignment path when the incoming value is a PrimExpr. Loop targets currently use this same path.
Distinguishing loop-target binding from ordinary assignment avoids the unintended buffer store while preserving mutable-scalar updates and scope validation.

sepcnt:
Fixes https://github.com/tile-ai/tilelang/issues/3231.
- Mark loop-target bindings explicitly, including tuple targets.
- Avoid assigning induction variables into existing alloc_var or Ref bindings.
- Preserve ordinary mutable assignments and out-of-scope checks.

## Summary

- Added explicit `loop_target` binding support for loop variables, including tuple targets.
- Created fresh loop bindings instead of rebinding existing `alloc_var` or `Ref` values.
- Preserved ordinary mutable assignments and scope validation.
- Added regression tests for loop forms, execution, tuple swaps, and expired bindings.

## Testing

- Test execution status was not provided.


tilelang/language/eager/builder.py
```diff
@@ -567,7 +567,15 @@ def bind(self, name, value, annot=BaseBuilder.empty):
             else:
                 return orig_value
 
-        orig_value = locals.get(name, self.empty)
+        # An existing Ref/alloc_var is only a store target while the region
+        # that bound the name is still open; once that region has closed the
+        # name is unbound, exactly as `rval` already treats reads of it. Loop
+        # targets always introduce a fresh induction binding, and `_` is the
+        # rewriter's throwaway name for temporaries.
+        if loop_target or name == "_" or self.binding_expired(name):
```
LeiWang1999, line 650: This is the actual root of #3231: an `alloc_var` bound in a region that has already closed should not be a store target any more — that is exactly how `rval` already treats *reads* of the same name, so writes and reads now agree. Your original repro passes with this check alone; `loop_target` is still needed for the case where the old `d` is *live* and gets shadowed by `for d`, which `bind` cannot tell apart from a legitimate `d = j` store by looking at the value (both hand it a `Var`).

tilelang/language/eager/builder.py
```diff
@@ -619,14 +627,19 @@ def bind(self, name, value, annot=BaseBuilder.empty):
         if name != "_":
             frame = self.find_frame_idx(TIR_VAR_SCOPE_FRAME)
             assert frame is not None, f"Variable `{name}` is not defined inside any control flow."
-            if name in self.name_inside_frame and self.name_inside_frame[name] in self.frames:
+            if not loop_target and name in self.name_inside_frame and self.name_inside_frame[name] in self.frames:
```
LeiWang1999: Stepped loops (`T.serial(a, b, step=s)`) yield a `PrimExpr` rather than a `Var`, so a loop target over a live `alloc_var` name landed here and printed "Immutable value `d` is re-bound; use T.alloc_var to create a mutable variable" — the user already did that. Your own `[True-stepped]` test printed it under `-s`.

tilelang/language/eager/builder.py
```diff
@@ -619,14 +627,19 @@ def bind(self, name, value, annot=BaseBuilder.empty):
...
                 logger.warning(
                     f"Immutable value `{name}` is re-bound; use T.alloc_var to create a mutable variable.",
                     stacklevel=2,
                 )
             self.name_inside_frame[name] = self.frames[frame]
         return res
 
+    def binding_expired(self, name: str) -> bool:
```
LeiWang1999: Shared by `bind` (writes) and `rval` (reads). A name with no recorded frame — e.g. a `T.Ref` macro parameter, since each macro gets a fresh `name_inside_frame` — is treated as live, so stores through `Ref` are unaffected (verified with a macro that does `d = d + by`).

testing/python/language/test_tilelang_language_loop_target_binding.py
```diff
@@ -0,0 +1,184 @@
...
+    tvm.tirx.stmt_functor.post_order_visit(
+        kernel.body, lambda node: stores.append(node) if isinstance(node, tvm.tirx.BufferStore) else None
+    )
+    return len([node for node in stores if node.buffer.scope() == "local.var"])
+
+
+def test_stepped_loop_target_over_live_var_does_not_warn(caplog):
+    # The tilelang logger does not propagate to the root logger caplog listens on.
+    builder_logger = logging.getLogger("tilelang.language.eager.builder")
+    builder_logger.addHandler(caplog.handler)
```
LeiWang1999: The `tilelang` logger sets `propagate = False`, so `caplog.at_level(...)` alone never sees these records; attaching the handler directly is what makes this test actually fail on the pre-fix code.

testing/python/language/test_tilelang_language_loop_target_binding.py
```diff
@@ -0,0 +1,184 @@
...
+            for k in T.serial(2):
+                # `d` from the finished loop is unbound here; this is a fresh
+                # binding, not a store into the old variable.
+                d = k + 10
+                A[2 + k] = d
+
+    return kernel
+
+
+def test_assignment_to_expired_var_is_fresh_binding():
```
LeiWang1999, line 76: This is the discriminating test for the liveness check: on the PR head it emits `d[0] = k + 10` into the hoisted buffer (store count 2) even though `d`'s defining loop has ended; the runtime result happens to be identical either way, which is why the structural assertion is the one that matters.

LeiWang1999 (commented):
Thanks for the fix — the diagnosis in #3231 is spot on, and making loop targets fresh bindings is the right call: the compiler needs the TIR loop `Var` (if `d` inside `for d in T.Parallel(...)` became `d[0]`, indices would be buffer loads and layout inference could no longer see the induction variable).

While reviewing I found that the `loop_target` flag patches one syntactic entry point but leaves the underlying asymmetry in place, so I pushed a follow-up commit (300f0bb9) to this branch rather than describe it in prose. What it does:

- **Root cause.** `bind` decided store-vs-fresh from stale Python locals without checking whether the TIR region that bound the name was still open. So a plain `d = k` into an `alloc_var` from a finished loop was silently emitted as a store into the hoisted buffer, while `A[..] = d` on the same name was rejected by `rval`. The new `binding_expired(name)` helper applies `rval`'s rule to writes too; your original repro is fixed by this alone. `loop_target` keeps its remaining job — a *live* `alloc_var` shadowed by `for d` still becomes a fresh binding (your `live=True` structural tests fail without it, so both are needed).
- **Tuple targets.** Phase 1 of the unpack still emitted `__tb.bind("_", tmp)` without the flag; with an `alloc_var` named `_` in scope, `for i, j in T.grid(2, 2)` aliased both loop vars to that buffer and produced `[0, -1, -1, 3]`. `_` is now never resolved against locals.
- **Stepped loops.** `T.serial(a, b, step=s)` yields a `PrimExpr`, not a `Var`, so with `loop_target` it fell into `bind_immutable` and printed "Immutable value `d` is re-bound; use T.alloc_var" — wrong advice, since `d` *is* an `alloc_var`. The shadow warning is skipped for loop targets.
- **Tests.** Added cases for each of the above. Note that the `live=True` variants of `test_loop_target_execution` pass on `main` without any fix (`saved = d` snapshots `d[0]` before the loop), so the structural store-count test is the one carrying that claim; I left them in place.

One thing I deliberately did not touch: `d += 1` after a loop that shadowed `d` still leaks the dead loop `Var` into TIR and fails with a raw TVM `InternalError`, because `visit_AugAssign` targets never pass through `__tb.rval`. That is the deeper "scope is tracked by name, not by value" issue and deserves its own PR.

Local runs: 28/28 in this file, 201 across the other frontend test files. Happy to merge once CI is green — please shout if you disagree with any of the semantics above.

LeiWang1999:
Follow-up: the branch had drifted behind `main` (#3230 rewrote the loop-body generation in `visit_For`, and #3262 removed the "Immutable value is re-bound" warning entirely), so pre-commit.ci was failing on a merge conflict before any hook ran. I merged `main` into the branch (4a0ca760) rather than rebase, to keep the push fast-forward.

Two consequences for the earlier commit:
- The `not loop_target` guard on the re-binding warning, and the `test_stepped_loop_target_over_live_var_does_not_warn` test, are gone — there is no warning left to suppress. My inline comment on that line is now moot.
- `rval` picked up `main`'s `name: str | None` signature; `binding_expired` handles `None` (comprehension locals) by returning `False`.

Everything else is unchanged. 232 tests pass locally across this file and the other frontend test files on the merged tree.

PR #3233, opened 2026-09-15
[BugFix] Restore symbolic loop-layout injectivity proof; reject layouts on symbolic shared tiles

### LINKED ISSUE #2906: [BUG] `makeBufferWithLayout` segfaults on a symbolic shared-buffer extent instead of hitting its own `ICHECK`

sepcnt:
Fixes #2906. Restores `T.Parallel` over mixed static/symbolic iteration spaces, broken by #2719.

## Summary

This PR started as "support symbolic shared layouts" and was rescoped during review (see the review comment below). It now does two independent things:

### 1. Restore the injectivity proof for symbolic padded loop layouts (`src/layout/layout.cc`)

A `T.Parallel` loop over a mixed static/symbolic space such as `(16, n)` is partitioned with a padded tail, so its loop layout `thread=(i*n+j)%128, index=(i*n+j)//128` is injective but not bijective and its domain cannot be enumerated. Before #2719, `DetectInjective` fell back to `NoCheck` for symbolic shapes; #2719 replaced that with a real proof that cannot handle a symbolic stride, so this kernel fails on `main` with `no available layout found` (it worked in v0.1.12):

```python
n = T.dynamic("n")

@T.prim_func
def kernel(A: T.Tensor((16, n), "int32"), B: T.Tensor((16, n), "int32")):
        for i, j in T.Parallel(16, n):
            B[i, j] = A[i, j]
```

`CanProveLeftInverse` builds a candidate inverse from a `NoCheck` IterMap via `InverseAffineIterMap` and then proves `g(f(x)) == x` on the input domain. `NoCheck` alone is not a proof; the round-trip check is. It is a fallback after `CanProveInjective`; the two have disjoint blind spots (documented at the call site), so both are kept.

### 2. Reject layouts on symbolic shared tiles with a clear error (`src/transform/lower_tile_op.cc`)

`makeBufferWithLayout` dereferenced a null `IntImm` for a shared buffer with a symbolic extent and crashed the compiler. TileLang's tiles are meant to be static (replication factor, swizzles, TMA boxes and the shared-memory budget all need constant extents), so instead of deriving a symbolic replication factor this now raises a `ValueError` naming the buffer and the offending extent. The check lives at the remap site so it also covers layouts inferred by ops such as `T.cumsum`, not only `T.annotate_layout`. A symbolic shared buffer *without* a layout keeps working as before.

## Tests

- `testing/python/language/test_tilelang_language_symbolic_loop_layout.py`: pure global-to-global `T.Parallel` over `(n, 16)` and `(16, n)` (the #2719 regression); a non-injective symbolic map is still rejected on injectivity.
- `testing/python/issue/test_tilelang_issue_2906.py`: `ValueError` for both the annotated and the scan-inferred layout on a symbolic shared tile; symbolic shared without a layout still compiles and runs.

Validated locally on H800 with `TILELANG_DISABLE_CACHE=1`: the two files above (6 passed), plus `test_tilelang_language_scan.py`, `test_tilelang_language_let_layout.py`, `test_tilelang_transform_layout_inference.py`, `test_tilelang_transform_lower_tile_op.py`, `test_tilelang_transform_smem_swizzle_alignment.py`, `test_tilelang_transform_producer_consumer_ws.py`, `test_tilelang_language_annotate_safe_value.py` (76 passed). `format.sh` clean.

Thanks to @sepcnt for the left-inverse proof and to @sohampanda000 for the original diagnosis in #2906 / #2909.

## Summary

- Fixes `#2906` by replacing the symbolic shared-buffer extent crash with a clear `ValueError`.
- Rejects symbolic shared-buffer and layout extents when replication cannot be proven.
- Preserves constant replication handling.
- Adds checked left-inverse injectivity proof for layouts where symbolic padding prevents a bijective iter-map proof.
- Converts inverse-construction failures into a declined proof instead of exposing `NoCheck` assertions.

## Tests

- Adds regression tests for annotated and inferred symbolic shared layouts.
- Verifies symbolic shared buffers without layout annotations still compile.
- Adds symbolic loop-layout tests for valid copies and non-injective layouts.
- Test execution results are unavailable.

## C++ style / lint notes

- The PR changes C++ implementation files.
- No changes to `docs/developer_guide/cpp_style.md` are indicated.
- No C++ API Style Audit results or warning findings were supplied.


LeiWang1999:
@sepcnt thanks for the careful work here, and thanks @sohampanda000 for the original diagnosis. I've taken the liberty of rescoping this on the branch directly (maintainer edits are enabled) and updated the title/description; here is why.

The PR bundled two changes that are quite different in nature.

**1. `CanProveLeftInverse` is a regression fix, kept as-is.** The failing injectivity check isn't about symbolic *shared layouts*; it's the *loop layout* of any `T.Parallel` over a mixed static/symbolic space. A pure global-to-global `T.Parallel(16, n)` with no shared buffer fails on `main` with `no available layout found`. It worked in v0.1.12 and broke with #2719 when `DetectInjective` stopped falling back to `NoCheck` for symbolic shapes. Dynamic shapes on global tensors are first-class, so this needed fixing regardless. The regression test is now that pure global loop, and I added a comment at the call site on why both `CanProveInjective` and `CanProveLeftInverse` are kept: they have disjoint blind spots (`(i, i + j)` is only provable by the former, `((i*n+j)%128, (i*n+j)//128)` only by the latter).

**2. Symbolic replication in `makeBufferWithLayout` is replaced by a clear rejection.** TileLang's design center is static tiles: gemm/swizzle layouts, TMA descriptors, vectorization, bank-conflict avoidance and the shared-memory budget all assume compile-time extents. Accepting `T.annotate_layout` / `T.cumsum` on a symbolic shared tile would make that a surface users rely on, and the surrounding machinery isn't ready: the load side of the `(16, n)` copy lowered to per-thread scalar loads, layout extents came out as `Select(n >= 0, 15*n, 0) + (n - 1) + 1`, unprovable replication ended in an `ICHECK`, and `i*(n+8) + j` (a symbolic padded stride) is still rejected by both proofs. Rather than back into partial "dynamic tile" support through a bugfix, the remap site now raises a `ValueError` naming the buffer and extent. It sits in `makeBufferWithLayout` rather than on the annotation path (as #2909 proposed) because scan-inferred layouts reach the same code.

If there's a concrete workload that needs layouts on symbolic shared tiles, let's scope that in a separate issue so it's a deliberate decision. Please shout if you disagree with any of this.

sepcnt:
@LeiWang1999 Looks good to me.

PR #3236, opened 2026-09-15
[CPU] Import the CPU dialect in CPU tests

penguin-wwy:
#3203 follow-up

## Summary

- Updated all nine CPU test files to import `tilelang.cpu.language`.
- Removed the unsupported `threads` argument from the BF16 test.
- Moved the `use_tma` rejection test to cross-dialect operation-hint tests.
- Tightened the expected backend error match.

## Testing

- 92 tests passed.
- 2 tests were skipped.

## Review findings

- Reviewer approved the changes.
- Severity counts were not provided.
- Five LLVM CPU-backend tests still use `tilelang.language` and may require a follow-up update.


LeiWang1999:
**Verdict: ✅ Approve** (suggestions below are non-blocking)

Since #3186 `tilelang.language` is the CUDA-dialect facade, so the `testing/python/cpu/` tests were writing `target="c"` kernels in the CUDA dialect (CUDA-only knobs like `threads=` / `use_tma=` were accepted and silently ignored on CPU). This PR switches all 9 CPU test files to `tilelang.cpu.language`. The two non-mechanical edits are both forced by the dialect switch and handled well:

- `test_tilelang_cpu_bf16_legalize.py:35` drops `threads=1` — `tilelang.cpu.language.Kernel` has no `threads` parameter (`tilelang/cpu/language/kernel.py:13`); keeping it raises `TypeError` at trace time (reproduced locally).
- `test_tilelang_cpu_atomic.py:20,370` builds the `use_tma=True` rejection case via `tilelang.cuda.language.atomic_add`, so it still hits the backend guard (`src/cpu/op/atomic_rmw.h:64`) rather than degrading into a Python signature error. Works, but see the 💡 below for a CPU-dialect-only way to do the same.

**Verified locally** (build has `USE_LLVM=OFF`, so the 2 LLVM numeric cases skip on both sides; everything else is `target="c"` + cython, no GPU needed):

| Test | merge-base `4c9cf5c7` | PR `3eb811b7` |
|---|---|---|
| `pytest testing/python/cpu/` (9 files) | 92 passed, 2 skipped | 92 passed, 2 skipped |

Probes on the PR branch: `cpu.language.Kernel(1, threads=1)` → `TypeError`; `cpu.language.atomic_add(..., use_tma=True)` → `TypeError`; CPU `T.Kernel` + `cuda.language.atomic_add(use_tma=True)` compiled for `c` → `Fatal: CPU atomic_add does not support use_tma=True` (the assertion the test relies on). Not a bug fix, so no red→green is expected; behavior is identical on both sides. The failing `CuTeDSL Examples` CI job was cancelled after ~2h (timeout) and is unrelated to this PR.

**Findings**

- ✅ Coverage is complete: all 9 files under `testing/python/cpu/` are updated. Per-file import is the right mechanism — the dialect is a module-level `import ... as T`, so a conftest/autouse fixture cannot inject it, and making `tilelang` import the CPU dialect by default would contradict the #3186 design.
- ✅ The `use_tma` case is kept and still exercises the backend rejection instead of being deleted or weakened.
- ⚠️ `testing/python/llvm/` has 5 sibling CPU-backend tests (`test_tilelang_llvm_{atomic,gemm,reduce,tgemm,while_repro}.py`) that still `import tilelang.language as T`. None uses `threads=`, so the same one-line change applies; consider including them here or noting a follow-up.
- 💡 `test_tilelang_cpu_atomic.py:20,370`: the cross-dialect `import tilelang.cuda.language as Tcuda` in a CPU-dialect test file is a bit at odds with the point of this PR (CPU tests use only the CPU dialect), and a reader may take it for a leftover. It is not needed: the common `atomic_add` still accepts `annotations` (`tilelang/language/atomic.py:216`), so the same backend guard is reachable from the CPU dialect alone:
  ```python
  with T.Kernel(1):
      T.atomic_add(B, A, annotations={"use_tma": True})
  ```
  Verified on this branch: compiling that for `target="c"` raises the identical `Fatal: CPU atomic_add does not support use_tma=True` from `src/cpu/op/atomic_rmw.h:64`. This also states the test's intent more precisely (the backend rejects the `use_tma` annotation, regardless of which dialect produced it) and lets the `Tcuda` import go. Independently, `:373` `pytest.raises(Exception, match="use_tma")` also matches a `TypeError: ... 'use_tma'`; `match="does not support use_tma"` would pin it to the backend guard.
- 💡 The PR description (auto-generated) has no motivation or test results; one line such as "CPU dialect has no `threads`; `use_tma` case rewritten so it still hits the backend guard" would help future readers of `git log`.

*Automated review by Hermes Agent (tile-ai/tilelang reviewer bot); verified on H800.*

PR #3239, opened 2026-09-17
[BugFix][CUDA][RNG] Harden initialization semantics and default sequence IDs

### LINKED ISSUE #2625: [BUG][Fuzzer][wrong-code] `T.rng_init()` default `seq` is x-dim-only, silently giving y/z-differing threads one RNG stream
### Required prerequisites

0.1.11 (source at `cd37ed5fc35ae7a60a1277c8eb49028174ac51e6`).

### System information

Reproduced on NVIDIA L40S (sm_89), CUDA 13.0, PyTorch 2.11.0, tilelang 0.1.11 (dev wheel, byte-identical to the source tag above). The defect is in the arch-independent Python frontend id-computation (the codegen simply forwards `seq` to `curand_init` on every CUDA target), so it is expected on all arches; I tested sm_89.

### Problem description

`T.rng_init(seed, seq=None, off=0)` initializes a per-thread curand stream for "parallel random number generation". When `seq` is omitted (the default), the per-thread sequence number is auto-derived from **only the x dimension** of the thread and block indices:

```python
# tilelang/language/random.py  (the `seq is None` branch)
if seq is None:
    bx = T.get_block_binding()          # blockIdx.x only  (dim=0)
    ex = T.kernel.get_thread_extent()   # blockDim.x only  (dim=0)
    tx = T.get_thread_binding()         # threadIdx.x only (dim=0)
    id = tx + bx * ex
    seq = tirx.convert(id)
```

`seq` is passed straight through to `curand_init(seed, seq, offset, &state)` — it is the subsequence key that decorrelates parallel streams. Because `threadIdx.y`, `threadIdx.z`, `blockIdx.y`, and `blockIdx.z` are all dropped from `id`, any two threads that differ **only** in one of those dimensions receive the **same `seq`** and therefore the **same random stream**. The kernel compiles and runs with no diagnostic; the outputs are silently identical. This bites any kernel that uses a 2-D/3-D thread block (`threads=(x,y[,z])`) or a multi-dimensional grid with the default `seq`.

### Reproducible example code

```python
tilelang.disable_cache()

def build(explicit_seq):
    @tilelang.jit
    def kf():
        @T.prim_func
        def main(Out: T.Tensor((2, 2), "int32")):     # int32 buffer; uint32 is not torch-feedable
            with T.Kernel(1, threads=(2, 2)):
                tx = T.get_thread_binding(0); ty = T.get_thread_binding(1)
                if explicit_seq:
                    T.rng_init(seed=1234, seq=tx * 2 + ty)   # CONTROL: unique per (tx,ty)
                else:
                    T.rng_init(seed=1234)                    # seq=None default (drops threadIdx.y)
                Out[tx, ty] = T.reinterpret(T.rng_rand(), dtype="int32")
        return main
    return kf()

for label, ex in [("default seq=None", False), ("explicit seq (control)", True)]:
    out = torch.zeros((2, 2), device="cuda", dtype=torch.int32); build(ex)(out)
    a = out.cpu().numpy()
    print(f"{label}: distinct={len(set(a.flatten().tolist()))}/4  Out[tx,0]==Out[tx,1]: "
          f"{bool(a[0,0]==a[0,1])},{bool(a[1,0]==a[1,1])}\n{a}")
# default seq=None      : distinct=2/4  Out[tx,0]==Out[tx,1]: True,True   <-- threadIdx.y dropped
# explicit seq (control): distinct=4/4  Out[tx,0]==Out[tx,1]: False,False
```

Same collision with a 2-D grid `T.Kernel(2, 2, threads=1)` (drops `blockIdx.y`): `Out[bx,0] == Out[bx,1]`.

### Traceback

No traceback — the kernel compiles and runs to completion. The random outputs are silently identical across the dropped dimensions.

### Expected behavior

With the default `seq`, distinct threads/blocks should get distinct curand streams — that is the purpose of the `seq` argument (see the [curand device API](https://docs.nvidia.com/cuda/curand/device-api-overview.html#device-api-overview), cited in the source itself). The control (a `seq` folding in all thread/block dimensions) gives four distinct values, which is the correct output the default should also produce.

### Additional context

**Root cause.** The default sequence-id derivation collapses the thread/block index to its x component only, so it is not unique across a multi-dimensional launch and hands colliding `seq` values to `curand_init`. Concretely, [`tilelang/language/random.py` L28–L33](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/tilelang/language/random.py#L28-L33) builds `id` from only the x components of the bindings (`get_thread_binding()`/`get_block_binding()`/`get_thread_extent()` all default to `dim=0`), dropping `threadIdx.y/z` and `blockIdx.y/z`. That `seq` is forwarded to `curand_init` as arg[1] at [`src/cuda/codegen/codegen_cuda.cc` L4298](https://github.com/tile-ai/tilelang/blob/cd37ed5fc35ae7a60a1277c8eb49028174ac51e6/src/cuda/codegen/codegen_cuda.cc#L4298).

**Suggested fix.** Fold in the remaining dimensions — a row-major flatten over `get_thread_bindings()` / `get_block_bindings()` weighted by the corresponding extents — so the default id is unique across the full launch grid. (Minimally: document that the default `seq` is x-dimension-only and require an explicit `seq` for multi-dimensional launches — but silent correlation is a poor default for an RNG.)

**Provenance.** Added in PR [#1582](https://github.com/tile-ai/tilelang/pull/1582) ("Add more curand operations"); current form last touched by `b939fa01` (#2216, tirx rename). The x-only derivation has been present since the feature landed — not a regression.

**Dedup.** Searched the open and closed tracker (`rng_init`, `curand`, `rng_rand`, `sequence`) and FILED_ISSUES.md — no existing report.

The `seq is None` default branch appears to have no test coverage — the tests/examples pass an explicit `seq`, so the x-only default is never exercised against a multi-dimensional launch.

### LINKED ISSUE #3029: [BUG][Fuzzer][ice-on-valid-code] Binding `T.rng_init()`'s documented state handle to a variable emits `void state = ;` and nvcc rejects the kernel
### Required prerequisites

0.1.13 (reproduced on 0.1.13).

### System information

TileLang 0.1.13; NVIDIA L40S (sm_89), CUDA 12.8. The defect is in the frontend `rng_init` return type + the generic C codegen of a `void`-typed variable binding, so it is not specific to one GPU.

### Problem description

`T.rng_init(...)` documents its return value as the RNG state handle (`Returns — state: PrimExpr — The random number generator state handle`). Capturing it the natural, documented way — `state = T.rng_init(seed)` — makes the kernel fail to compile: the emitted CUDA contains `void state = ;`, which nvcc rejects with `error: incomplete type "void" is not allowed`. The byte-for-byte identical kernel that *discards* the return value compiles and runs correctly (uniform values in `[0,1)`), so the RNG lowering itself is fine — only capturing the documented handle breaks it.

The abort happens the moment the binding exists, even if `state` is never read again and no random number is ever drawn.

### Reproducible example code

```python
N = 256

# CRASHES: bind the rng_init state handle to a variable (the documented "Returns: state" usage)
@T.prim_func
def rng_bind(B: T.Tensor((N,), "float32")):
    with T.Kernel(1, threads=256) as bx:
        tx = T.get_thread_binding()
        state = T.rng_init(1234)                 # docstring: "Returns state: the RNG state handle"
        B[tx] = T.rng_rand_float(32, "uniform")

# CONTROL: identical kernel, rng_init result NOT bound -> compiles & runs, values in [0,1)
@T.prim_func
def rng_nobind(B: T.Tensor((N,), "float32")):
    with T.Kernel(1, threads=256) as bx:
        tx = T.get_thread_binding()
        T.rng_init(1234)                         # same call, return value discarded
        B[tx] = T.rng_rand_float(32, "uniform")

try:
    tilelang.compile(rng_bind, out_idx=[0])()
    print("bind:   compiled")
except Exception as e:
    print("bind:   CRASH", type(e).__name__)          # -> CRASH RuntimeError (nvcc: incomplete type "void")

b = tilelang.compile(rng_nobind, out_idx=[0])().cpu().numpy()
print("nobind: min", b.min(), "max", b.max())          # -> min ~0.004  max ~0.998 (uniform [0,1))
```

### Traceback

No Python-level traceback with a clean message — the failure is an nvcc compile error surfaced as a `RuntimeError` from `tilelang.contrib.nvcc.compile_cuda`.

<details><summary>Emitted CUDA (the malformed line) + nvcc error (0.1.13, sm_89)</summary>

```cpp
curandStatePhilox4_32_10_t __random_generator_state;
curand_init(1234, ((int)threadIdx.x), 0, &__random_generator_state);
void state = ;                                             // <- the bound handle; void-typed, empty init
B[((int)threadIdx.x)] = curand_uniform(&__random_generator_state);
```
```
tvm_kernels.cu(22): error: incomplete type "void" is not allowed
      void state = ;
tvm_kernels.cu(22): error: expected an expression
      void state = ;
2 errors detected in the compilation of "tvm_kernels.cu".
```

`rng_init` lowers to a `void`-typed intrinsic, so binding it produces a `void`-typed variable declaration with no initializer — `void state = ;` — which is doubly illegal (void has no storage, and the RHS is empty). The discard-the-result control never forms a binding, so this line is never emitted.
</details>

### Expected behavior

Either the kernel should compile — capturing the state handle the docstring's `Returns` section describes should be a legal, no-op-if-unused binding — or `rng_init` should not advertise a returnable handle it cannot actually hand back. The RNG codegen (`curand_init`/`curand_uniform`) is already present and correct (the discard control produces valid uniform values), so the value is computable; the defect is that a documented return handle is `void` and binding it emits illegal CUDA. Making `rng_init` return a usable handle type (or the codegen elide a `void` binding) would honor the docstring; alternatively the docstring should stop promising a bindable state handle.

### Additional context

**Root cause.** `rng_init` builds its call with a `void` result dtype, while its docstring promises a returnable `state` handle. At [`tilelang/cuda/language/random.py#L39`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/cuda/language/random.py#L39):

```python
return tirx.call_intrin("void", tirx.op.Op.get("tl.rng_init"), seed, seq, off, generator)
```

but the [docstring (L8-L27)](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/cuda/language/random.py#L8-L27) states `Returns — state: PrimExpr — The random number generator state handle`. When the caller follows that and writes `state = T.rng_init(...)`, the frontend forms a variable binding whose value is this `void`-typed expression; the C codegen prints the binding's declaration as `PrintType(void) name = <value>`, i.e. `void state = ;`, which nvcc rejects. The actual curand state is a separate module-level variable (`__random_generator_state`) the intrinsic writes to as a side effect — so the intrinsic genuinely has nothing to return, and the docstring's "returns a handle" is the inaccurate half.

**Suggested fix.** Make the documented contract and the implementation agree: either return a real handle to the curand state (so binding it is meaningful and codegens as a reference to `__random_generator_state`), or change `rng_init` to not claim a returnable handle (return `None`/no value and update the docstring), so a user cannot form a `void state = ;` binding. A narrower codegen guard — refuse to emit (or elide) a `void`-typed variable binding with a clear frontend error — would at least turn the opaque nvcc failure into an actionable one.

**Provenance.** The RNG API (`rng_init` returning a `void`-typed intrinsic while documenting a state handle) was added by [#1461](https://github.com/tile-ai/tilelang/pull/1461) (merged 2025-12-18). The `void`-return + documented-handle mismatch is present as written at the v0.1.13 tag (`random.py#L39` vs the `Returns` docstring). Reproduced at runtime on 0.1.13 this session (bind → nvcc `void state = ;`; discard control → uniform `[0,1)`). Not a recent regression.

> Note: on 0.1.9 the same binding aborted *earlier*, inside `VerifyParallelLoop` (an `InternalError: A fixed length vector doesn't have a vscale factor.` from feeding the `void` let-value to the arithmetic analyzer). That pass path no longer forms the constraint on 0.1.13, so the crash now surfaces downstream in nvcc codegen; the underlying cause (a `void`-typed `rng_init` handle bound to a variable) is the same.

**Generalization (tested this session, 0.1.13, L40S sm_89).**

<details><summary>Two-level root, 4-axis sweep, and example-run result</summary>

**Root — two levels.**
- **SOURCE-level (fragile implementation).** `rng_init` returns `tirx.call_intrin("void", tl.rng_init, ...)` at [`random.py#L39`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/cuda/language/random.py#L39) while its `Returns` docstring (L23-26) names a `state` handle. Downstream, the generic C codegen prints a let-binding as `PrintType(dtype) name = <value>;`; when `dtype` is `void` and the intrinsic hands back no value, it emits `void state = ;`. The fragile code is the pairing of a *no-value* frontend intrinsic dtype with the let-binding codegen — nothing rejects a valueless capture at the frontend.
- **OPERATOR-level (what correlates).** Any TileLang frontend op that returns `call_intrin("void"/"handle", ...)` (a valueless/side-effect-only intrinsic) and is allowed to be bound to a variable hits the same `PrintType(dtype) name = ;` codegen. `rng_init` is the *only* such op whose own docstring documents a returnable handle, so it is the in-contract representative; the others are out-of-contract siblings.

**4-axis sweep (every cell RUN, result observed):**

| axis | cell tested | observed result | same-root? |
|---|---|---|---|
| the knob | `state = T.rng_init(1234)` with NO subsequent `rng_rand*` | `void state = ;` → nvcc `incomplete type "void"` | yes (binding alone triggers; drawing RNG irrelevant) |
| similar-logic / related-operator | bind void intrinsic `s = T.sync_warp()` | `void s = ;` → nvcc `incomplete type "void" is not allowed` | **same root** |
| related-type (intrinsic dtype) | bind `handle`-typed `h = T.named_barrier_arrive(0,256)` | `void* h = ;` → nvcc `expected an expression` (only) | same mechanism (empty-RHS let-binding), **distinct face**: `void*` is a complete type, so no "incomplete type" error |
| related-source / related-type | bind non-void sibling `r = T.rng_rand()` (uint32, same file) | compiles + runs; samples 546353992, 3507855656 | no — non-void return binds fine (control: confirms root is the `void`/no-value dtype, not "rng ops") |

- **Boundary — kept specific.** Reframed the *class* to "binding a valueless (`void`/`handle`) intrinsic → empty-RHS declaration", but the FILED bug stays the `rng_init` documented-handle case, because `rng_init` is the only op whose docstring invites the capture. `sync_warp`/`named_barrier_arrive` have no documented return (their docstrings show bare-statement use), so binding them is user error — recorded here as **distinct-adjacent, not filed**.

**Example-run result (PART 1).** Both cited RNG sites run clean on 0.1.13 and dodge the bug: `examples/rand/rand_uint.py` compiles + passes its Triton `assert_close` for all 3 default configs; `testing/python/language/test_tilelang_language_rand.py` → `6 passed`. Every shipped `rng_init` call is a bare discard statement (verified: `random.py` example + all 4 test-file call sites), so no shipped code exercises the crash.
</details>

**Dedup.** I searched the open and closed tracker (`rng_init`, `void state`, `VerifyParallelLoop`, `incomplete type void`) and found no existing report of this defect. Two related-but-distinct issues exist:
- **#2625** also concerns `rng_init` but is a different defect (default `seq` giving y/z threads one stream — a silent wrong result, not a compile crash) — different operator behavior, unrelated root.
- **#2667** (closed) shares the *downstream codegen symptom* — a `void`-typed op bound to a variable emits an empty-initializer declaration nvcc rejects (`int prev = ;` there, `void state = ;` here). But the *root differs*: #2667 is an atomic op-selection bug (`atomic_min`/`max` `return_prev=True` wrongly dispatches the void `*_elem_op` when it should return a value), whereas here `rng_init` genuinely has nothing to return and the docstring over-promises a handle. Fixing #2667 (route min/max to a value-returning path) does not fix this (rng has no value); they are the same codegen weakness reached from two different fronts, not the same bug. That #2667 was accepted and fixed confirms this class of empty-binding crash is treated as a real defect.

**Reach.** The trigger is a documented input: `rng_init`'s docstring names its return value the RNG state handle, and capturing a documented return value in a variable is the ordinary way to use an API. Ran the shipped RNG sites verbatim on 0.1.13 this session: `examples/rand/rand_uint.py` (`python rand_uint.py`, all 3 default configs) compiles and passes its Triton `assert_close`; `testing/python/language/test_tilelang_language_rand.py` runs `6 passed`. Every shipped site calls `T.rng_init(...)` as a bare statement and discards the return, so all of them dodge the crash and CI stays green — i.e. the shipped examples do NOT exercise this bug. A user who instead follows the docstring's `Returns` section and binds the handle hits the crash on the first compile.

**Impact.** The trigger is narrow — it needs the specific act of binding `rng_init`'s documented `void`-typed return handle to a variable (a code-shape gate, not a data value). When it fires it is a compile-time abort (an nvcc `void state = ;` error surfaced as a `RuntimeError`), so it is loud and caught on the first build, blocks that one kernel, and cannot silently corrupt output. Fixing it makes a documented `rng_init` usage compile as the docstring implies, removing a docstring/implementation disagreement.

### LINKED ISSUE #3034: [BUG][Fuzzer][ice-on-invalid-code] `T.rng_rand`/`T.rng_rand_float` with no preceding `T.rng_init` emit `curand…(&)` and crash nvcc, instead of a clean "init required" error
### Required prerequisites

0.1.13

### System information

- Installed via pip into a clean venv
- Python 3.13
- `tilelang` 0.1.13, `torch` 2.8.0
- GPU: NVIDIA L40S (sm_89); CUDA 12.x, `nvcc` invoked with `-arch=sm_89`

### Problem description

Calling `T.rng_rand()` (or `T.rng_rand_float()`) in a kernel that does **not** also call `T.rng_init()` produces a kernel that fails to compile with an opaque nvcc error, instead of a clean TileLang-level message that `rng_init` is required first.

The CUDA codegen for `rng_rand` emits `curand(&<state>)`, where `<state>` is the name of the curand state variable. That name is only ever set when an `rng_init` call is codegen'd; with no `rng_init` present it stays the empty string, so the emitted line is literally `curand(&)` — a bare `&` with no operand — and nvcc aborts:

```
/tmp/.../tvm_kernels.cu(20): error: expected an expression
    B[((int)threadIdx.x)] = curand(&);
                                    ^
```

The same kernel *with* an `T.rng_init(...)` compiles and codegens correctly (control below), so the failure is specific to the missing-init case.

### Reproducible example code

```python
import tilelang
import tilelang.language as T


def build(with_init):

    @T.prim_func
    def main(B: T.Tensor((128,), "uint32")):
        with T.Kernel(1, threads=128) as _:
            tx = T.get_thread_binding()
            if with_init:
                T.rng_init(0)        # -> curand_init(...); state var declared
            B[tx] = T.rng_rand()     # -> curand(&<state>)

    return main


# CONTROL: identical kernel WITH T.rng_init -> compiles fine
tilelang.compile(build(True), target="cuda")     # OK
#   emitted: curandStatePhilox4_32_10_t __random_generator_state;
#            curand_init(0, ((int)threadIdx.x), 0, &__random_generator_state);
#            B[((int)threadIdx.x)] = curand(&__random_generator_state);

# TRIGGER: identical kernel WITHOUT T.rng_init -> nvcc "expected an expression"
tilelang.compile(build(False), target="cuda")     # RuntimeError
#   emitted: B[((int)threadIdx.x)] = curand(&);   <-- bare '&', no operand
```

`T.rng_rand_float()` (which emits `curand_uniform(&<state>)` etc.) fails identically for the same reason.

### Traceback

```pytb
  ...
    raise RuntimeError(msg)
RuntimeError: ...
Compilation error:
/tmp/.../tvm_kernels.cu(20): error: expected an expression
    B[((int)threadIdx.x)] = curand(&);
                                    ^
1 error detected in the compilation of "/tmp/.../tvm_kernels.cu".
```

### Expected behavior

A `T.rng_rand` / `T.rng_rand_float` used without a preceding `T.rng_init` reads an uninitialized (in fact, never-declared) curand state, so there is no valid value to produce. The natural expectation is a clear error at build time — e.g. "`rng_rand` used without a preceding `rng_init`" — rather than emitting `curand(&)` and surfacing an opaque `expected an expression` from nvcc. The frontend already asserts other RNG preconditions eagerly (`random.py` asserts the `generator`/`bit`/`dist` values), so an early check here would be consistent with that.

### Additional context

**Root cause.** The CUDA codegen prints the curand state operand from a member string that is only assigned inside the `rng_init` branch, with no check that an `rng_init` was seen for the current function; with no `rng_init` the member stays empty and the emit collapses to `curand(&)`.

<details>
<summary>Where the empty state name comes from (pinned to v0.1.13)</summary>

`rng_rand` emits [`curand(&" << this->curand_random_generator_state << ")"`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4797-L4799) and `rng_rand_float` [does the same](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4801-L4807). `curand_random_generator_state` is a plain [`std::string` member](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.h#L92-L93) that defaults to empty and is set only in the [`rng_init` branch](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4781-L4788). The per-function pre-scan in `AddFunction` that declares the state variables walks the body [only for `rng_init` calls](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L6143-L6158); when the body has `rng_rand` but no `rng_init`, nothing is declared and the member stays empty, so the emit becomes `curand(&)`. Nothing in the [frontend `rng_rand`/`rng_rand_float`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/cuda/language/random.py#L42-L50) requires or checks for a prior `rng_init` either.

</details>

**Impact.** A user who forgets (or mis-scopes) the `rng_init` call gets no TileLang-level guidance — the kernel is accepted through the whole frontend and lowering pipeline, then dies deep in `nvcc` with `expected an expression` on a machine-generated `.cu` line the user never wrote. The failure surfaces far from its cause (a missing frontend call), so the message is actively misleading rather than merely unhelpful; diagnosing it requires reading the emitted CUDA. It is a build-time failure, not a silent miscompile, so no wrong results reach silicon — but it costs debugging time on a precondition the compiler already knows how to detect.

**Suggested fix.** Add a frontend guard that a function using `rng_rand`/`rng_rand_float` contains at least one `rng_init`, raising a clear error otherwise. Alternatively the codegen could `ICHECK(!curand_random_generator_state.empty())` at the `rng_rand`/`rng_rand_float` emit sites (the `rng_init` branch already `ICHECK`s its own pre-scan registration), turning the silent `curand(&)` into an actionable message.

**Provenance.** Introduced with the RNG feature in [#1461](https://github.com/tile-ai/tilelang/pull/1461) ("[Language] Adds a random number generation capability through curand_kernel", merged 2025-12-18): as first written, `rng_rand` emitted `curand(&" << this->curand_philox_state << ")"` where the state member is set only in the `rng_init` branch — the same missing-precondition gap. [#2297](https://github.com/tile-ai/tilelang/pull/2297) (2026-06-02) only relocated the code into a backend namespace and renamed the member to `curand_random_generator_state`; it did not introduce or fix the gap. The pattern has therefore shipped since RNG first landed; not a regression.

**Dedup.** I searched the open and closed tracker and found no existing report of this defect. The nearest reports are a different RNG bug — [#2625](https://github.com/tile-ai/tilelang/issues/2625) (`rng_init`'s default `seq` is x-dim-only; a silent wrong-value bug in `rng_init`, not a compile failure) — and a same-shaped-but-unrelated emit — [#2667](https://github.com/tile-ai/tilelang/issues/2667) (`atomic_min`/`atomic_max` `return_prev=True` emits `int prev = ;`); both have a different root and site.

**Reach.** Triggering requires calling `rng_rand`/`rng_rand_float` with no `rng_init` anywhere in the function. Every RNG site shipped at 0.1.13 pairs them, so no shipped example or test exercises the no-init path (which is why CI does not catch it). **Verified by running the shipped example on 0.1.13 (L40S, sm_89):** `examples/rand/rand_uint.py` — where `T.rng_init` is on L18 and `T.rng_rand()` on L23 of the same kernel — runs to completion (exit 0; all three configs `(1024,42)/(512,123)/(128,0)` pass its torch-vs-triton `assert_close`). It **dodges the bug because it always calls `T.rng_init` before `T.rng_rand`**. The RNG tests likewise precede every producer with a `T.rng_init` in the same kernel: `testing/python/language/test_tilelang_language_rand.py` (including `test_rand_init_in_split_guard`, which puts the init in a runtime guard — the pre-scan still registers it), and `testing/python/language/test_tilelang_cast_rounding.py` (`T.rng_init(42)` L473 before `T.rng_rand()` L475). No shipped site hits the no-init path.

**Generalization / boundary (all cells run on 0.1.13, L40S, sm_89).**

- **SOURCE-level root:** `CodeGenTileLangCUDA::VisitExpr_` in `src/cuda/codegen/codegen_cuda.cc`. The `curand_random_generator_state` `std::string` member is assigned *only* in the `rng_init` branch ([L4787](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4787)) and read unconditionally by the two producer branches ([L4799](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4799), [L4806](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L4806)), with no non-empty guard. The `AddFunction` pre-scan ([L6143-L6158](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/cuda/codegen/codegen_cuda.cc#L6143-L6158)) only registers `rng_init` calls, so absent any init the member stays empty and every producer collapses to `curand…(&)`.
- **OPERATOR-level root:** the curand *producer* intrinsics that consume the shared state member — `tl.rng_rand` and the four `tl.rng_rand_float` shapes.

|---|---|---|---|
| related-operator | `rng_rand_float()` (uint→f32 uniform), no init | `curand_uniform(&);` → nvcc `expected an expression` | **same root** |
| related-operator | `rng_rand_float(bit=64)`, no init | `curand_uniform_double(&);` → same nvcc error | **same root** |
| related-operator | `rng_rand_float(dist="normal")`, no init | `curand_normal(&);` → same nvcc error | **same root** |
| related-operator | `rng_rand_float(bit=64, dist="normal")`, no init | `curand_normal_double(&);` → same nvcc error | **same root** |
| related-source | `rng_init` alone, no producer | compiles OK | n/a (no producer reads the member) |
| related-source (boundary) | `rng_init` inside runtime guard `if d!=0:` then top-level `rng_rand` | compiles OK — pre-scan is `PostOrderVisit`, not flow-sensitive | boundary: crash needs *total absence* of `rng_init`, not scope/branch mismatch |
| similar-logic | any other codegen intrinsic reading a `std::string` member set in a sibling branch | none exist — `curand_random_generator_state[_type]` are the only such members in `codegen_cuda.h` (L92-93) | class confined to RNG |
| related-type | `generator` knob (Philox/MRG32k3a/XORWOW) | does not apply — the generator name is only consumed inside `rng_init`, which is absent in the trigger; the empty-state emit is generator-independent | n/a |

All five producer shapes (`rng_rand` + 4 × `rng_rand_float`) share one root; the fix (frontend "`rng_init` required" guard, or an `ICHECK(!curand_random_generator_state.empty())` at the producer emit sites) covers the whole class. The nearest same-*shape* emit, `atomic_min/max` `return_prev=True` emitting `int prev = ;` ([#2667](https://github.com/tile-ai/tilelang/issues/2667)), is a **distinct-adjacent bug (not filed here)** — different site (atomic branch ~L4839) and a different empty-operand mechanism. No new bug was found while sweeping.

0z5a:
## Summary

Three related defects in the CUDA RNG surface:

- `rng_init(seq=None)` derived its default sequence from `threadIdx.x` and
  `blockIdx.x` only, so every thread that differed only in `y`/`z` shared one
  curand subsequence and produced identical draws.
- `rng_init` is a side-effect-only intrinsic, but binding its result produced
  `void state = ;`, which nvcc rejected as an incomplete type.
- A draw with no `rng_init` anywhere in the function emitted `curand(&)`,
  turning a TileLang semantic error into an nvcc syntax error.

Fixes #2625
Fixes #3029
Fixes #3034

## Changes

`tilelang/language/eager/builder.py`
- reject binding a value-less expression in `bind_immutable`, so `rng_init` and
  any other side-effecting intrinsic is diagnosed by name and by the offending
  expression before lowering.

`tilelang/cuda/language/random.py`
- add `_default_rng_sequence()`: a row-major linear id over every launched
  thread and block dimension. Dimensions that were not launched have extent 1
  and contribute nothing, so the historical 1-D id
  `threadIdx.x + blockIdx.x * blockDim.x` is unchanged.
- correct the `seq` and `Returns` documentation: `rng_init` is a statement and
  its result is not a state handle.

`src/cuda/codegen/codegen_cuda.cc`
- validate RNG state once per function in the existing `AddFunction` pre-scan:
  if the body contains an `rng_rand` / `rng_rand_float` but no `rng_init`,
  fail with an intentional TileLang error. Deliberately *not*
  control-flow-sensitive: an `rng_init` inside a runtime branch still counts.
- clear the per-function curand state name and type, so a draw can never
  reference the state registered by an earlier function of the same module.
- add `VisitStmt_(const BindNode *op)`, rejecting a bound `rng_init` call as a
  backstop for paths that bypass the eager builder (it previously produced
  `void state = ;`). (`Bind` is this tree's name for
  the `LetStmt` node; the override matches codegen_py.cc and
  codegen_cutedsl.cc.)

`src/cuda/codegen/codegen_cuda.h`
- declare the new `BindNode` visitor.

`testing/python/language/test_tilelang_language_rand.py`
- 15 new cases: 2-D thread and 2-D grid default-sequence coverage, 1-D layout
  and explicit-`seq` controls, all five producer shapes with and without
  `rng_init`, and the void-binding rejection. Because the frontend now rejects
  the void bind first, that regression accepts either diagnostic (frontend
  "value-less expression" or codegen "cannot be bound to variable").

## Validation

Environment: NVIDIA L20 (sm_89), CUDA 12.8, PyTorch 2.10.0+cu128,
Python 3.13.14. The branch is rebased onto current `main` (`e40e09e3`); every
result below was re-confirmed on the rebased revision. The before/after
comparison was run against `e688a439`, the `main` at the time of writing. `bash format.sh` reports every hook
passing on the changed files (clang-format, ruff-check, ruff-format, codespell).

### 1. Repo RNG test module

| build | result |
|---|---|
| before (pristine `e688a439`) | **8 failed, 13 passed** |
| after | **21 passed** |

The 8 pre-patch failures are exactly the new regressions:

```text
FAILED test_rand_default_seq_covers_thread_y
FAILED test_rand_default_seq_covers_block_y
FAILED test_rand_producer_without_init_is_rejected[rng_rand]
FAILED test_rand_producer_without_init_is_rejected[rng_rand_float]
FAILED test_rand_producer_without_init_is_rejected[rng_rand_float_bit64]
FAILED test_rand_producer_without_init_is_rejected[rng_rand_float_normal]
FAILED test_rand_producer_without_init_is_rejected[rng_rand_float_bit64_normal]
FAILED test_rand_init_result_cannot_be_bound
```

### 2. End-to-end kernel validation

Real CUDA kernels, observable behaviour only (no source-text assertions).
`e2e_rng.py`, before vs after:

| check | before | after | detail |
|---|---|---|---|
| `e2e_2d_threads_streams` | FAIL | PASS | 256/256 distinct threads own distinct streams |
| `e2e_2d_grid_values` | FAIL | PASS | 512/512 distinct |
| `e2e_uniform_quality` | PASS | PASS | mean 0.50034, var 0.08333, chi2 60.46 (63 dof), max CDF dev 7.4e-4 |
| `e2e_seed_sensitivity` | PASS | PASS | 0 equal values across 2^20 samples |
| `e2e_no_init_diagnostic` | FAIL | PASS | TileLang diagnostic |
| `e2e_void_bind_diagnostic` | FAIL | PASS | TileLang diagnostic |
| `e2e_bare_init_valid` | PASS | PASS | 8/8 distinct |
| **total** | **3/7** | **7/7** | |

`e2e_2d_threads_streams` launches a `(32, 8)` thread block where each thread
draws 16 values. Before the patch the eight threads sharing a `threadIdx.x`
produced byte-identical 16-value sequences; after the patch all 256 streams are
distinct. `e2e_uniform_quality` confirms the change does not disturb the
distribution (chi-square consistent with uniform, KS-style CDF deviation two
orders of magnitude inside the 5% critical value).

### 3. Generated `seq`

```text
before: curand_init(1234, ((int)threadIdx.x), 0, &__random_generator_state);
after:  curand_init(1234, ((((int)threadIdx.y) * 2) + ((int)threadIdx.x)), 0, &__random_generator_state);
```

4 logical threads, distinct values 2 -> 4, row collisions `[True, True]` ->
`[False, False]`. A 1-D launch is byte-identical to the explicit
`seq=threadIdx.x + blockIdx.x * blockDim.x` control, and an explicit `seq=` is
forwarded unchanged.

### 4. Error surface

```text
#3029 before: nvcc "incomplete type" on the generated `void _state = ;`
#3029 after:  frontend "Cannot bind `_state` to a value-less expression: ... has void type"
              (codegen keeps a backstop: "The result of `T.rng_init(...)` cannot be
              bound to variable ...")

#3034 before: nvcc syntax error inside generated `curand(&)`
#3034 after:  TileLang "`T.rng_rand()`/`T.rng_rand_float()` requires a preceding
              `T.rng_init(...)` in the same kernel ..."
```

### 5. Broader suite

`python -m pytest -q testing/python/language`

```text
6 failed, 1264 passed, 186 skipped
```

All six failures are pre-existing. Running the same five affected modules
against the pristine and the patched build gives an **identical** failure set
(6 failed, 29 passed on both), so this patch changes no test outcome outside
the RNG module:

```text
test_tilelang_language_func_attrs.py::test_out_idx_via_attr_lazy
test_tilelang_language_get_warp_info.py::test_shuffle_elect_default
test_tilelang_language_get_warp_info.py::test_shuffle_elect_block_leader
test_tilelang_language_let.py::test_producer_bind_kernel
test_tilelang_language_ptr.py::test_matmul
test_tilelang_language_tma_1d.py::test_tilelang_language_tma_1d
```

## Scope

This PR intentionally does not add flow-sensitive RNG
definite-initialization analysis, and does not introduce an explicit public RNG
state-handle API.

## Summary

- Treats `T.rng_init(...)` as a `void` statement and rejects invalid bindings.
- Reports clear build-time errors when RNG producers lack prior initialization.
- Clears CURAND state between functions.
- Derives default `seq` values across all thread and block dimensions while preserving explicit sequences and one-dimensional behavior.
- Adds regression tests for multidimensional launches, initialization errors, valid initialization, explicit sequences, and invalid bindings.

## C++ style / lint notes

- C++ code changed, but `docs/developer_guide/cpp_style.md` was not modified.
- The C++ API Style Audit is warning-only.
- No supplied findings indicate correctness or build issues.


0z5a:
Superseded by three focused PRs, which split this change back apart for review:

- #3242 — derive the default `rng_init` sequence from the full launch (#2625)
- #3243 — reject binding the void `rng_init` result (#3029)
- #3244 — diagnose `rng_rand`/`rng_rand_float` without `rng_init` (#3034)

Each is rebased on current `main` and carries its own before/after test table.
Closing this combined PR in favour of those.

Issue #3240, opened 2026-09-17
[BUG][CuTeDSL][FP4/FP8] DeepSeek V4 act quant: FP4 conversion/store broken against CUTLASS DSL 4.7, FP8 store rejected by libNVVM


0z5a:
### Required prerequisites

- [x] I have read the documentation <https://tilelang.com>.
- [x] I searched the issue tracker for an existing dedicated report.

### What version of TileLang are you using?

Current `main` at `e688a439a8f79d69497d02c4fad7c75fada954ad`.

### System information

- GPU: NVIDIA L20 (sm_89)
- CUDA / nvcc: 12.8.61 (`/usr/local/cuda-12.8`)
- PyTorch: 2.10.0+cu128
- TileLang: editable build of `main` (`0.1.14+cuda.gite688a439`)
- CuTeDSL / CUTLASS DSL: `nvidia-cutlass-dsl==4.7.0` with `nvidia-cutlass-dsl-libs-cu12==4.7.0`

### How the CuTeDSL backend is really selected

`examples/conftest.py` xfails `deepseek_v4/test_tilelang_example_deepseek_v4.py::test_example_act_quant`
when `TILELANG_TARGET=cutedsl`, and `.github/workflows/ci.yml` runs the examples with
`TILELANG_TARGET: cutedsl`. But nothing in the library reads `TILELANG_TARGET` — only
`examples/conftest.py` and two example tests do. Verified on current `main`:

```text
$ TILELANG_TARGET=cutedsl python -c "...compile a trivial kernel, print adapter/target..."
TILELANG_TARGET= cutedsl
adapter: TVMFFIKernelAdapter
target: {"kind":"cuda", ..., "arch":"sm_89"}
```

So that env var only toggles the xfail markers; the examples still run on the default
backend. The backend is actually selected with `TILELANG_DEFAULT_TARGET=cutedsl`
(or `tilelang.compile(..., target="cutedsl")`). Worth fixing in its own right, because it
makes the CuTeDSL CI job and every "reproduce under CuTeDSL" recipe a no-op.

### Problem description

With the backend actually selected, the DeepSeek V4 activation-quantization example fails
in two independent ways: the FP4 conversion/store helpers call MLIR ops that no longer
exist in CUTLASS DSL 4.7, and the FP8 store path is rejected by libNVVM.

```text
[...]

Issue #1454, opened 2025-12-17
[RoadMap] CuteDSL Backend


LeiWang1999:
@cherichy and @lucifer1004 have kindly agreed to help review the related PRs. :)

- [ ] Initial Commit: Python Codegen / Runtime (#1421 )
- [x] TVM FFI Support
- [ ] GEMM V2 Support
- [ ] More Examples

...

PR #3241, opened 2026-09-17
[BugFix][CuTeDSL] Repair FP4 conversion/store lowering and packed FP4 output shapes

0z5a:
**Incremental review:** [Files changed](https://github.com/tile-ai/tilelang/pull/3241/files) against `main`.

## Summary

Repairs the CuTeDSL conversion/store path for sub-byte float types. With these changes the
**FP4** activation-quantization path compiles and passes under `target="cutedsl"`, and the FP8
destination-type mismatch is fixed as well.

The combined `test_example_act_quant` still fails, because its FP8 stage now reaches a
separate `libNVVM` compilation failure (see the verification section at the end and
#3240), so it cannot be removed from `CUTEDSL_KNOWN_FAILURES` yet.

Refs #3240
Related: #1454

### 1. FP4 conversion helpers use the CUTLASS DSL 4.7 MLIR API

`tilelang/contrib/cutedsl/utils.py` converted between `float4_e2m1fn` and float vectors with
`vector.extractelement` / `vector.insertelement`. Those ops no longer exist in the `vector`
dialect shipped with `nvidia-cutlass-dsl` 4.7 (they are `vector.extract` / `vector.insert`
now), so every FP4 cast raised

```text
AttributeError: module 'cutlass._mlir.dialects.vector' has no attribute 'extractelement'
  at tilelang/contrib/cutedsl/utils.py:256 in _float32_to_f4e2m1_tensor
```

The helpers now use `vector.extract(v, [], [idx])` / `vector.insert(v, dst, [], [idx])`, which
also removes the now-unneeded dynamic index constants.

### 2. Sub-byte outputs are allocated packed

`CuTeDSLKernelAdapter._convert_torch_func` materialized outputs as
`torch.empty(*param_shapes[i], dtype=param_dtypes[i])` from the *logical* TIR shape. For a
4-bit dtype the torch storage type (`float4_e2m1fn_x2`) packs two elements per byte, so the
last dimension has to be divided by the packing factor — exactly what the CUDA adapters
already do (`stride_scale` in `tilelang/jit/adapter/tvm_ffi.py`). Without this,
`fp4_act_quant` returned `(M, N)` instead of the packed `(M, N/2)`:

```text
AssertionError: Shape mismatch: torch.Size([64, 256]) vs torch.Size([64, 128])
```

Measured A/B with the fix toggled (same script, same GPU):

| | returned shape |
|---|---|
| without this hunk | `(256,)` |
| with this hunk | `(128,)` |

### 3. The wrapper's FP8 dtype key matches the codegen

`TLCuTeDSLSourceWrapper._TYPE_MAP` mapped the TIR dtype string `"float8_e4m3"` to
`cutlass.Float8E4M3`, while `src/cuda/codegen/codegen_cutedsl.cc` only emits
`Float8E4M3FN` ("Only Float8E4M3FN is supported at the moment"). Since the TIR string for
`float8_e4m3fn` is `float8_e4m3`, generated fake tensors got `Float8E4M3` and every fp8
store attempted the unsupported conversion

```text
TypeError: Unsupported destination float type: Float8E4M3
  at cutlass/base_dsl/typing.py:1123 in Numeric.to
```

Both keys now map to `Float8E4M3FN`.

### 4. Focused regression test

`testing/python/jit/test_tilelang_jit_cutedsl_fp4_convert.py` covers the packed output shape
and decodes every stored nibble back to a representable FP4 E2M1 value, so the lowering is
not only exercised through the full DeepSeek example.

## Validation

NVIDIA L20 (sm_89), CUDA 12.8.61, PyTorch 2.10.0+cu128, `nvidia-cutlass-dsl==4.7.0`, current
`main` at `e688a439a8f79d69497d02c4fad7c75fada954ad`.

Note that `TILELANG_TARGET=cutedsl` does not actually select this backend (only
`examples/conftest.py` reads it); the results below use `TILELANG_DEFAULT_TARGET=cutedsl` /
`target="cutedsl"`. That mismatch is reported in #3240.

| target | `test_fp8_act_quant` | `test_fp4_act_quant` | `test_round_trip_error` |
|---|---|---|---|
| `cuda` before/after | pass | pass | pass |
| `cutedsl` before | `TypeError: Unsupported destination float type` | `AttributeError: ... extractelement` | same as FP8 |
| `cutedsl` after | libNVVM failure (unchanged, see below) | **pass** | libNVVM failure |

`test_fp4_act_quant` under CuTeDSL now reports `85/8192` packed bytes differing from the
PyTorch reference, all at FP4 rounding tie-breaks (max dequantized difference `0.5`), which
is the same behavior the CUDA backend documents.

```text
$ pytest -q testing/python/jit/test_tilelang_jit_cutedsl_fp4_convert.py
2 passed
```

## What this does not fix

The FP8 half of the same example still fails inside libNVVM
(`NVVM_ERROR_COMPILATION` on a minimal `bf16 -> float8_e4m3fn` elementwise store, with an
empty backend log), so `test_example_act_quant` remains in `CUTEDSL_KNOWN_FAILURES`. That
blocker, its minimal reproducer and the `TILELANG_TARGET` finding are tracked in #3240.

## Summary

- Updated FP4 conversion lowering to use CUTLASS DSL 4.7 `vector.extract` and `vector.insert`.
- Packed sub-byte output shapes by dividing the final dimension by the packing factor.
- Mapped `float8_e4m3` and `float8_e4m3fn` to `cutlass.Float8E4M3FN`.
- Added regression tests for packed FP4 output and decoded FP4 values.

## Validation

- FP4 activation quantization passes in the reported environment.
- 85 of 8192 packed bytes differ only at rounding tie-breaks.
- FP8 validation remains blocked by a libNVVM compilation failure and is tracked in issue `#3240`.
- Review finding counts are unavailable from the supplied evidence.

---

## Independent verification

Reproduced on `e40e09e3` (current `main`) with `nvidia-cutlass-dsl==4.7.0`, NVIDIA L20
(sm_89), CUDA 12.8. Backend selected with `TILELANG_DEFAULT_TARGET=cutedsl` — note that
`TILELANG_TARGET=cutedsl` does **not** select a backend (only `examples/conftest.py` reads
it, to attach xfails), so it cannot be used to check this PR.

`test_example_act_quant` was split into its three sub-paths, each run in isolation, with the
tilelang kernel cache cleared between arms (the cache key is the git commit, so the patched
and unpatched builds would otherwise share entries).

| sub-path | unpatched `main` | with this PR |
|---|---|---|
| `test_fp4_act_quant` | FAIL `module 'cutlass._mlir.dialects.vector' has no attribute 'extractelement'` | **PASS** |
| `test_fp8_act_quant` | FAIL `TypeError: Unsupported destination float type: Float8E4M3` | FAIL `CompilerDiagnosticError: NVVM backend compilation failed` |
| `test_round_trip_error` | FAIL (FP8) | FAIL (NVVM) |
| combined `test_example_act_quant` | FAIL | **FAIL** |
| this PR's focused regression `test_tilelang_jit_cutedsl_fp4_convert.py` | n/a | **PASS** (2 passed) |

So both code fixes in this PR are real and verified: the FP4 conversion helpers now use the
CUTLASS DSL 4.7 `vector.extract`/`vector.insert` API, and the `Float8E4M3` -> `Float8E4M3FN`
type-map fix removes the `Unsupported destination float type` error. What remains is the
`libNVVM` failure on the FP8 store path, which is a different problem already tracked in
#3240.


0z5a:
Independent verification on `e40e09e3` (current `main`), with the tilelang kernel cache
cleared between arms, splitting `test_example_act_quant` into its three sub-paths:

| sub-path | unpatched `main` | with this PR |
|---|---|---|
| `test_fp4_act_quant` | FAIL `... has no attribute 'extractelement'` | **PASS** |
| `test_fp8_act_quant` | FAIL `TypeError: Unsupported destination float type: Float8E4M3` | FAIL `NVVM backend compilation failed` |
| `test_round_trip_error` | FAIL (FP8) | FAIL (NVVM) |
| `test_example_act_quant` | FAIL | **FAIL** |
| focused regression in this PR | n/a | **PASS** (2 passed) |

Both fixes here are confirmed working: FP4 now compiles and passes, and the FP8 dtype-map
fix removes the `Float8E4M3` error. The example test still fails because the FP8 stage now
reaches the `libNVVM` failure tracked in #3240, so it can't be removed from
`CUTEDSL_KNOWN_FAILURES` yet.

One process note: `TILELANG_TARGET=cutedsl` does not select a backend — only
`examples/conftest.py` reads it, to attach xfails. Use `TILELANG_DEFAULT_TARGET=cutedsl`
(or `target="cutedsl"`) to actually exercise CuTeDSL; with only the former set the test runs
on the default CUDA backend and passes.

PR #3245, opened 2026-09-17
Reject unsupported GEMM dtype combos before CUDA codegen

### LINKED ISSUE #3004: [BUG][Fuzzer][ice-on-valid-code] `T.gemm` accepts bf16/int8/tf32 input with `float16` accum (no MMA specialization) then aborts nvcc with a template `static_assert`
### Required prerequisites

0.1.13 (latest release)

### System information

Linux; TileLang 0.1.13; CUDA 12.8 (nvcc), PyTorch 2.8; reproduced on an NVIDIA L40S (sm_89), classic MMA path.

### Problem description

`T.gemm` does not validate the accumulator dtype against the input dtypes. Any A/B/accumulator dtype triple whose dtypes each individually pass the frontend's global dtype gate is accepted, generates otherwise well-formed CUDA (correct `m16n8k16` shape), and then — if the `MmaDispatcher` template has no specialization for that exact triple — aborts `nvcc` deep in template instantiation with `static_assert failed "tl::mma_sync: unsupported configuration"`. The compile dies instead of either computing the GEMM or reporting a clean frontend error.

The class covers every input dtype whose *only* dispatcher accumulator is wider than `float16` — `bfloat16`, `int8`, and `tf32` inputs each have only a `→float32`/`→int32` MMA atom in the table — paired with a `float16` accumulator (which passes the global dtype gate). Verified crashes (§17): `bfloat16`/`float16`, `int8`/`float16`, `float32`(tf32)/`float16`. Inputs whose narrow-accumulator entry *does* exist escape: `float16`/`float16` and `float8_e4m3`/`float16` both compile.

The accumulator dtype is a normal, documented knob — `example_gemm_sp.py` exposes `--accum_dtype` with `choices=[T.float, T.float16]`. That example pairs the `float16` accumulator with `float16` inputs, which has a dispatcher specialization and compiles+runs (verified, §17), so it dodges the bug.

<details><summary>Trigger boundary — which A/B × accumulator dtype combinations crash vs compile (tested, sm_89)</summary>

| A/B input dtype | accumulator dtype | result |
|---|---|---|
| `float16` | `float16` | compiles, correct (`maxdiff 2.6e-4`) |
| `float16` | `float32` | compiles, correct |
| `bfloat16` | `float32` | compiles, correct (`maxdiff 6e-8`) |
| `float8_e4m3` | `float16` | compiles (dispatcher entry exists) |
| `bfloat16` | `float16` | **nvcc `static_assert` abort** |
| `int8` | `float16` | **nvcc `static_assert` abort** (same root) |
| `float32` (tf32) | `float16` | **nvcc `static_assert` abort** (same root) |

The crash fires whenever the input dtype has no `float16`-accumulator MMA atom in the dispatcher table (`bfloat16`, `int8`, `tf32`), yet the `float16` accumulator passes the frontend's global dtype gate. Inputs whose `float16`-accumulator entry *does* exist (`float16`, `float8_e4m3`) compile. All of the above reproduced on sm_89 (classic MMA path); the WGMMA path (`wgmma.h`) has the same missing-`bfloat16`→`float16` structure but is not runnable on sm_89 (see §17).

</details>

### Reproducible example code

```python

# REPRO: bfloat16 GEMM accumulating into a float16 accumulator aborts nvcc.
M = N = K = 64
@T.prim_func
def main(A: T.Tensor((M, K), "bfloat16"),
         B: T.Tensor((K, N), "bfloat16"),
         C: T.Tensor((M, N), "float16")):
    with T.Kernel(1, 1, threads=32):
        As = T.alloc_shared((M, K), "bfloat16")
        Bs = T.alloc_shared((K, N), "bfloat16")
        Cl = T.alloc_fragment((M, N), "float16")
        T.clear(Cl)
        T.copy(A, As); T.copy(B, Bs)
        T.gemm(As, Bs, Cl)          # -> nvcc: static_assert "tl::mma_sync: unsupported configuration"
        T.copy(Cl, C)

k = tilelang.compile(main, out_idx=[2], target="cuda")   # aborts here

# CONTROL (ORACLE: PASS): identical kernel, float16 inputs instead of bfloat16.
# A float16 accumulator is a valid target -> compiles and is correct.
@T.prim_func
def ctrl(A: T.Tensor((M, K), "float16"),
         C: T.Tensor((M, N), "float16")):
    with T.Kernel(1, 1, threads=32):
        As = T.alloc_shared((M, K), "float16")
        Bs = T.alloc_shared((K, N), "float16")
        Cl = T.alloc_fragment((M, N), "float16")
        T.clear(Cl)
        T.copy(A, As); T.copy(B, Bs)
        T.gemm(As, Bs, Cl)
        T.copy(Cl, C)

kc = tilelang.compile(ctrl, out_idx=[2], target="cuda")
A = torch.randn(M, K, device="cuda", dtype=torch.float16) * 0.1
B = torch.randn(K, N, device="cuda", dtype=torch.float16) * 0.1
out = kc(A, B).cpu().float(); ref = (A.float() @ B.float()).cpu()
print("maxdiff", (out - ref).abs().max().item())          # -> ~2.6e-4
print("OK" if torch.allclose(out, ref, atol=2e-2, rtol=2e-2) else "BAD")   # -> OK
```

### Traceback

```
src/tl_templates/cuda/instruction/mma.h(282): error: static assertion failed with "tl::mma_sync: unsupported configuration"
    static_assert(!std::is_void_v<typename Dispatcher::CRegType>, ...
      detected during instantiation of "void tl::mma_sync<...>(...)"
      [with AType=tl::DataType::kBFloat16, BType=tl::DataType::kBFloat16,
       CType=tl::DataType::kFloat16, M=16, N=8, K=16, TransA=false, TransB=true, Saturate=false]

src/tl_templates/cuda/instruction/mma.h(167): error: static assertion failed with "tl::mma_sync: unsupported configuration"
    static_assert(always_false_v<std::integral_constant<int, M>>, ...
2 errors detected in the compilation of ".../tvm_kernels.cu".
```

### Expected behavior

Either compile and compute the GEMM (accumulating in a supported width and narrowing to the requested `float16` accumulator, the way the `float16`-input path already produces a correct `float16` result), or reject the unsupported dtype combination in the frontend with a clear message — rather than emitting CUDA that a `static_assert` rejects after the frontend has already accepted it.

### Additional context

**Root cause.** The GEMM frontend accepts an accumulator dtype the MMA dispatcher has no atom for, so codegen builds an `mma_sync` template that has no matching specialization and detonates a `static_assert` at nvcc time. (Two levels.)

*Source-level:* the GEMM frontend does not validate the accumulator dtype against the input dtypes: [`GemmBase.accum_dtype`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/tileop/gemm/gemm_base.py#L104-L105) is simply `self.C.dtype`, so any `C` dtype that passes the global dtype gate is passed straight through to codegen; there is no accum-vs-input dtype check anywhere in the gemm lowering (`gemm_base.py` is the only per-op source at v0.1.13; `src/op/gemm.cc` has no dtype validation). The generated `mma_sync<Ain, Bin, kFloat16, 16, 8, K, ...>` then falls through to the primary-template [`MmaDispatcher` `static_assert(always_false_v<…>)`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/tl_templates/cuda/instruction/mma.h#L167) (and the paired [`is_void_v<CRegType>` assert](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/tl_templates/cuda/instruction/mma.h#L282)) whenever the input/accumulator triple has no entry in the [dispatcher specialization table](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/tl_templates/cuda/instruction/mma.h#L194-L217). The table defines `kFloat16→kFloat16` (L194) and `kFloat16→kFloat32` (L196), but for `kBFloat16` inputs only `kBFloat16→kFloat32` (L205), for `kInt8` only `→kInt32` (L209), and for `kTensorFloat32` only `→kFloat32` (L256). *Operator-level:* the fragile mechanism is the "frontend accepts, dispatcher table decides" split shared by every dtype pair routed through `mma_sync` — so the same crash reproduces across input dtypes (bf16 / int8 / tf32) and across the sibling WGMMA dispatcher (`wgmma.h`), not just the single bf16/f16 case.

</details>

**Suggested fix.** The two Expected-behavior options are not equivalent, and which is right depends on intent: PTX `mma` has a native `f16.f16.f16` accumulate atom (so the `float16`-input path can accumulate directly in `float16`), but `bf16` inputs only have an `f32`-accumulate atom — so a `bfloat16`-input / `float16`-accumulator GEMM would have to accumulate in `float32` and narrow, not map to a native instruction. Adding that specialization vs. rejecting the combination in the frontend is a design choice; the frontend-reject path at minimum turns the late `static_assert` into an early, legible error.

**Provenance.** Reproduced on 0.1.13; the `accum_dtype = self.C.dtype` passthrough (`gemm_base.py`) and the `MmaDispatcher` table (`mma.h`) are both present at the v0.1.9 tag and unchanged since. Not verified against earlier versions, so not characterized as a regression.

**Dedup.** I searched the open and closed tracker (`MmaDispatcher`, `mma_sync unsupported configuration`, gemm accumulator-dtype terms) and found no existing report of this defect. The only keyword hit, #1441, is an unrelated closed report about compiling a specific bf16 FlashAttention kernel.

**Impact.** When it fires the failure is a compile-time `nvcc` `static_assert` abort — loud and caught immediately, after the frontend has already accepted the kernel; it corrupts no values and cannot reach a running workload silently, but it blocks that kernel from building. The trigger is not a single dtype triple but a *family*: any input dtype with no `float16`-accumulator MMA atom (`bfloat16`, `int8`, `tf32`) paired with a `float16` accumulator (§17). No shipped example pairs those. Fixing it either adds the accumulate-in-`f32`-and-narrow specializations or moves the rejection into the frontend, turning a late template-instantiation abort into an early legible error.

**Generalization.** 4-axis sweep, all cells run on 0.1.13 / L40S sm_89 unless noted. Two-level root as under **Root cause** above: (a) source = the `GemmBase.accum_dtype` passthrough with no accum-vs-input validation; (b) operator = the "frontend accepts / dispatcher table decides" split shared by every dtype pair routed through `mma_sync`.

| axis | cell tested (input → accum) | observed result | same root? |
|---|---|---|---|
| related-type (wider-only-atom neighbor) | `int8`/`int8` → `float16` | nvcc `static_assert "unsupported configuration"`, `mma_sync<kInt8,kInt8,kFloat16,16,8,32,...>` | **same root** |
| related-type (fp32/tf32 input) | `float32`/`float32` → `float16` | nvcc `static_assert`, `mma_sync<kFloat32,kFloat32,kFloat16,16,8,8,...>` | **same root** |
| related-type (covered-atom contrast) | `float8_e4m3`/`float8_e4m3` → `float16` | **compiles OK** — table has `kFloat8_e4m3→kFloat16` (mma.h L237) | same mechanism, cell covered → no crash (boundary) |
| related-type (narrow-int accum) | `int8`/`int8` → `int16` | frontend `ValueError: Unsupported dtype: int16` (caught early by a different, global dtype gate) | **distinct** — rejected before codegen, not a dispatcher fall-through |
| related-source (sibling dispatcher) | WGMMA `wgmma.h` `bfloat16` → `float16` | source-confirmed: `wgmma.h` has `F16_F16_F16` + `F16_F16_F32` + `BF16_BF16_F32` (L274-282) but **no** `BF16_BF16_F16`, same `always_false_v` primary template (L95/L106) | **same root** (Hopper WGMMA path; not runnable on sm_89) |

- **similar-logic / related-operator:** the identical "construct-then-narrow, validate late" shape recurs across every `mma_sync` input dtype — the int8, tf32, and bf16 cells above are the meaningful operator siblings and all share the root.
- **related-type (covered):** `float16` and `float8_e4m3` inputs escape only because their `→float16` entries exist; this pins the root to the *specific missing table cell*, not to fp16 accumulators in general.
- **Example run (Part 1):** `examples/gemm_sp/example_gemm_sp.py --accum_dtype float16` (sparse gemm, fp16 inputs) compiled and ran on 0.1.13 → dodges the bug (fp16/fp16 has a dispatcher entry); the numeric mismatch it prints is fp16-accumulate tolerance, not a compile failure. `--accum_dtype float32` passed cleanly.
- **Distinct-adjacent (not filed):** the `int16`-accumulator rejection (`ValueError: Unsupported dtype: int16`) is a *different* mechanism (global dtype gate, fires before codegen) — not this bug.
- **No new bug** was found while sweeping; every crashing cell reduces to the same missing-`MmaDispatcher`-entry root.

**Reach.** (Verified by running the shipped example.) A `float16` accumulator is a documented, user-facing config: `examples/gemm_sp/example_gemm_sp.py` exposes `--accum_dtype` with `choices=[T.float, T.float16]`. I ran that shipped example verbatim on 0.1.13/sm_89: with `--accum_dtype float16` it compiled and ran (only a fp16-accumulate tolerance miss, not a compile error); with `--accum_dtype float32` it passed cleanly. It dodges this bug because it is the *sparse* gemm path with `float16` inputs (`torch.half`) — the `float16`/`float16` pairing has a dispatcher entry. The other cited sites also use `float16` (not `bfloat16`) inputs (`example_gemm_intrinsics.py` even asserts `in_dtype in [T.float16, ...]`), so none pairs a `float16` accumulator with `bfloat16`/`int8`/`tf32` inputs, and CI stays green. The crash needs one of those non-`float16` inputs combined with the `float16` accumulator; no shipped example uses that pairing.

hywosdpl:
Fixes #3004. Some T.gemm operand dtype combos (e.g. bf16 x bf16 -> fp16) were accepted in the frontend and only failed later inside nvcc with an opaque static_assert. This PR adds an explicit dtype whitelist for the Ampere/Ada mma.sync path, enforced both in the frontend (tilelang/tileop/gemm/__init__.py) and at C++ instruction selection (src/cuda/op/gemm.cc), so unsupported combos are rejected early with a clear message while all supported combos and non-Ampere targets are unaffected. Adds regression tests (1 negative + 3 positive) in testing/python/cuda/test_cuda_gemm_mma_sync_dispatch.py.

## Summary

- Added a C++ dtype whitelist for Ampere/Ada `mma.sync` GEMM selection.
- Rejected unsupported combinations before CUDA code generation.
- Preserved supported combinations and non-Ampere behavior.
- Removed duplicate Python-side validation.
- Added one negative and three positive regression tests. Tests pin whitelist coverage to `sm_89`.

## C++ style / lint notes

- The PR changes C++ instruction-selection code.
- It does not change `docs/developer_guide/cpp_style.md`.
- The **C++ API Style Audit (warning only)** is relevant.
- No review findings or test execution results were supplied.


src/cuda/op/gemm.cc
```diff
@@ -371,6 +420,9 @@ struct Gemm {
     if (TargetIsTuring(target) && !AllowTuringMma(op)) {
       return kCudaFMA;
     }
+    if (TargetIsAmpere(target) && !AllowMmaSync(op, target)) {
```
LeiWang1999: What about `AllowAmpereMMA`?

hywosdpl: Thanks for the review. I checked: there is no existing AllowAmpereMMA — AllowVoltaMma/AllowTuringMma only cover the SM70/SM75 FMA fallback, and Ampere has no fallback, so I fatal here (that's exactly why #3004 hit a hard nvcc error). I've renamed the gate to AllowAmpereMMA as suggested.

LeiWang1999:
**💬 Comment** — re-reviewed at `50d5c1e5`. The TF32 regression is fixed the right way (adds `tfloat32`, keeps `float32`/`float64`, ignores the two incorrect CodeRabbit suggestions). Verified on H800: 15 dtype×arch combos (sm_80/89/90) produce `kernel_source` byte-identical to main; `bf16→f16` on sm_80/89 now fails at lowering instead of in nvcc. No longer blocking.

Remaining:

1. ⚠️ **The positive tests never run the whitelist.** `test_cuda_gemm_mma_sync_dispatch.py:55` uses `target="cuda"`; CI is not Ampere, so `TargetIsAmpere` is false. That is how the TF32 regression got through, and the fix still has no test. Pin to `arch: sm_89` (compile only) and add `tfloat32→f32`, `f32→f32`, `f64→f64` rows.
2. 💡 **Keep one whitelist, not two.** The C++ `AllowAmpereMMA` is unreachable in the normal pipeline (Python raises first). Suggest dropping the Python check and keeping only `SelectInst`.
3. 💡 If the Python check stays: `_is_mma_sync_dtype_supported(node, …)` and `_raise_mma_sync_dtype_unsupported(self, …)` are module-level functions only ever called with `self` from `Gemm` — make them one `Gemm._check_mma_sync_dtype(self, target)` method, raising `ValueError` rather than `InternalError`.
4. 💡 Guard is gated on Ampere, but sm_90/100/120 fall back to the same `kCudaMMA` path and still hit the raw nvcc static_assert. Fine as a follow-up.

*Automated review by Hermes Agent (tile-ai/tilelang reviewer bot); verified on H800.*

testing/python/cuda/test_cuda_gemm_mma_sync_dispatch.py
```diff
@@ -0,0 +1,63 @@
...
+)
+@tilelang.testing.requires_cuda
+@tilelang.testing.requires_cuda_compute_version(8)
+def test_gemm_mma_sync_supported_dtype_combo_runs(dtype, accum_dtype):
+    import torch
+
+    kernel = tilelang.compile(
+        _make_gemm_kernel(dtype, accum_dtype),
+        out_idx=[2],
+        target="cuda",
```
LeiWang1999, line 56: ⚠️ With `target="cuda"` this resolves to the runner's own arch; CI is not Ampere, so `TargetIsAmpere` is false and the whitelist is never executed by these positive cases. That is how the TF32 regression got through in the earlier revisions.

Suggest pinning like the negative test does (compile-only, no execution) and adding the whitelist edge rows:

```python
@pytest.mark.parametrize("dtype,accum_dtype", [
    ("float16", "float16"), ("float16", "float32"), ("bfloat16", "float32"),
    (T.tfloat32, "float32"), ("float32", "float32"), ("float64", "float64"),
])
def test_gemm_mma_sync_supported_dtype_combo_lowers_on_ampere(dtype, accum_dtype):
    tilelang.compile(_make_gemm_kernel(dtype, accum_dtype), out_idx=[2],
                     target={"kind": "cuda", "arch": "sm_89"})
```

tilelang/tileop/gemm/__init__.py
```diff
@@ -9,6 +9,48 @@
 from tilelang import _ffi_api
 
 
+def _is_mma_sync_dtype_supported(node, target) -> bool:
```
LeiWang1999, line 12: 💡 This function and `_raise_mma_sync_dtype_unsupported` below are only ever called from `Gemm.infer_layout` / `Gemm.lower` with `self` as the first argument (the second one even names the parameter `self`). They are `Gemm` methods living outside the class.

If the Python-side check stays, fold both into one private method:

```python
def _check_mma_sync_dtype(self, target: Target) -> None:
    if not _ffi_api.TargetIsAmpere(target):
        return
    ...
    raise ValueError(f"T.gemm: unsupported mma.sync dtype combination ...")
```

and call `self._check_mma_sync_dtype(target)` from both entry points. `ValueError` rather than `tvm.error.InternalError` — this is user-input validation, not a compiler bug (cf. `tilelang/language/dtypes.py:337`).

src/cuda/op/gemm.cc
```diff
@@ -143,6 +143,33 @@ bool AllowTuringMma(const GemmNode &op) {
   return false;
 }
 
+bool AllowAmpereMMA(const GemmNode &op, Target target) {
```
LeiWang1999, line 146: 💡 In the normal `tilelang.lower` pipeline this function is unreachable: `GemmNode::InferLayout` / `GemmNode::Lower` call the Python hooks first, and `_is_mma_sync_dtype_supported` raises before `GetGemmInstructionKey` is ever called. It only fires for direct callers of `tl.GemmGetGemmInstructionKey`.

Two hand-synced copies of the same table is how `tfloat32` got missed in both last time. I'd keep just this one and drop the Python pre-check.

src/cuda/op/gemm.cc
```diff
@@ -385,6 +423,9 @@ struct Gemm {
     if (TargetIsTuring(target) && !AllowTuringMma(op)) {
       return kCudaFMA;
     }
+    if (TargetIsAmpere(target) && !AllowAmpereMMA(op, target)) {
```
LeiWang1999, line 426: 💡 The guard is keyed on `TargetIsAmpere`, but the failure it prevents is keyed on the *path*: Hopper without WGMMA eligibility (M<64, warps % 4 != 0, `kDisableWGMMA`), Blackwell without tmem, and SM120 non-block-scaled all fall through to `kCudaMMA` and the same `MmaDispatcher` table. Verified: `bf16→f16` on sm_90 still dies in nvcc's static_assert on this branch.

Moving the check to where `kCudaMMA` is finally selected (arch-independent) would cover them too. Fine as a follow-up.

LeiWang1999 (commented):
Inline notes for the four items in my comment above (verified on H800 at `50d5c1e5`).

PR #3246, opened 2026-09-18
[Fix]Clamp T.copy/T.async_copy coalesced_width to achievable vector size instead of LOG(FATAL)

### LINKED ISSUE #3007: [BUG][Fuzzer][diagnostic] `T.copy(coalesced_width=N)` `LOG(FATAL)`s when N doesn't divide the geometry-derived vector size, instead of clamping-with-warning or a clean diagnostic
### Required prerequisites

0.1.13 (latest release)

### System information

TileLang 0.1.13 / PyTorch 2.8.0 / CUDA 12.8 / NVIDIA L40S (sm_89). The defect is in an architecture-independent layout-planning pass; also present on `main` (see Root cause / Provenance).

### Problem description

`coalesced_width` is a documented, `int`-typed public keyword of [`T.copy`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/copy_op.py#L58-L70) ("Width for coalesced memory access"). When the copy's auto-derived per-thread vector size is *smaller* than the requested width, the compiler aborts with an internal `LOG(FATAL)` instead of using the value as a hint:

```
tvm.error.InternalError: Vector size 1 is not divisible by coalesced width 2
```

The value is legal and the copy itself compiles fine — removing only the `coalesced_width` kwarg makes the identical kernel compile and produce correct results. The abort fires whenever the copy region's geometry forces the auto-derived vector size below the requested width (here a ragged 1-column tail block makes the SIMT loop non-vectorizable, so the derived width is 1 and `1 % 2 != 0`). The same `coalesced_width=2` compiles for tiles whose geometry admits a ≥2-wide vector, so it is the hint-vs-geometry interaction, not the value, that crashes. Not a regression — see Provenance.

### Reproducible example code

```python

M, N, BM, BN = 32, 33, 32, 32   # N=33 => one ragged tail column beyond the 32-wide tile

def build(cw):                  # cw=None -> control; cw=2 -> trigger
    @T.prim_func
    def main(A: T.Tensor((M, N), "float32"), B: T.Tensor((M, N), "float32")):
            sh = T.alloc_shared((BM, BN), "float32")
            if cw is None:
                T.copy(A[by*BM, bx*BN], sh)
                T.copy(sh, B[by*BM, bx*BN])
            else:
                T.copy(A[by*BM, bx*BN], sh, coalesced_width=cw)   # documented int kwarg -> FATAL
                T.copy(sh, B[by*BM, bx*BN], coalesced_width=cw)
    return main

def run(cw):
    try:
        k = tilelang.compile(build(cw), out_idx=[1])
        a = torch.randn(M, N, device="cuda"); b = k(a)
        return "PASS" if torch.allclose(b, a, atol=1e-4) else "MISMATCH"
    except Exception as e:
        return "CRASH: " + str(e).split("Fatal:")[-1].strip().splitlines()[0][:60]

print("control  no coalesced_width :", run(None))   # -> PASS
print("trigger  coalesced_width=2  :", run(2))      # -> CRASH: Vector size 1 is not divisible by coalesced width 2
```

### Traceback

```
[TileLang] : Fatal: Vector size 1 is not divisible by coalesced width 2
  File "<unknown>", line 0, in tvm::runtime::detail::LogFatal::~LogFatal()
tvm.error.InternalError: Vector size 1 is not divisible by coalesced width 2
```

### Expected behavior

A documented public `int` hint should not reach an internal `LOG(FATAL)`. When the requested width cannot be met by the copy's geometry, the natural behavior is to clamp/decline the hint (use the largest achievable width) or reject with a user-facing diagnostic — not abort with an internal assert. The pass already does exactly this graceful clamp for the *auto-derived* width one branch earlier: it halves `vector_size` until it divides the thread/loop geometry ([parallel.cc:783-787](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L783-L787)). Applying the same clamp when the user supplies `coalesced_width` would keep the hint honored where possible and non-fatal where not.

### Additional context

**Root cause.** The parallel-loop layout planner treats `coalesced_width` as a hard divisor of the auto-derived vector size and aborts when it doesn't divide, rather than clamping it. In [`ParallelOpNode::ComputePlanCandidate`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L798-L808) the derived `vector_size` (already reduced to fit the thread bounds by the loop at [:783-787](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L783-L787)) is checked with `vector_size % expected != 0` → `LOG(FATAL)`. For a ragged region whose loop is non-vectorizable, `vector_size == 1`, so any `coalesced_width > 1` fails the modulo and aborts. The check assumes the user width is always ≤ and a divisor of the achievable width; when the geometry forces a smaller vector, that assumption breaks with no fallback.

<details>
<summary>boundary: the crash is width-vs-geometry, and the value is honorable on a fitting tile</summary>

Same `coalesced_width` value, different tile geometry (float32):

```
tile 128x128, threads 128  (derived vector_size 4):  cw=1 PASS, cw=2 PASS, cw=4 PASS, cw=8 CRASH ("Vector size 4 not divisible by coalesced width 8")
tensor 32x33, tile 32x32, threads 128 (ragged tail -> derived vector_size 1):  cw=2/4/8 CRASH ("Vector size 1 ...")
tensor 32x32, tile 32x32, threads 128 (derived vector_size >=2):  cw=2 PASS
```

So `coalesced_width=2/4` are values the planner can honor (they compile and return correct results on a tile whose derived vector size is a multiple); the abort is specific to the case where the region geometry drops the derived width below the request.

Re-verified this session (0.1.13, L40S), each a separate compile: `128×128` tile → `cw=4` PASS, `cw=8` CRASH ("Vector size 4 is not divisible by coalesced width 8"); `32×32` exact tile → `cw=2` PASS; `32×33` ragged tail → `cw=2` CRASH ("Vector size 1 ..."). This is a **tight-guard** defect: the planner treats the user width as a hard divisor of the (already geometry-reduced) derived vector size instead of clamping it. It is a *distinct* branch from the sibling `T.Parallel` bare-int abort ("`coalesced_width should be an IntImmNode`", line 808): that is a type-coercion gap on a raw Python int; this is the divisibility check on line 804 — same function, adjacent code, different root, reported separately.
</details>

**Suggested fix.** `coalesced_width` is documented as a *hint* ("Width for coalesced memory access"), so an unmeetable value should degrade, not abort. In [parallel.cc:798-808](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L798-L808), when the requested width doesn't divide the geometry-derived `vector_size`, clamp the effective width to `gcd(vector_size, expected)` (or `min` when it divides) **and emit a warning** that the hint was reduced (naming the requested vs achievable width) — reusing the graceful-degradation policy the auto path at [:783-787](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L783-L787) already applies to the derived width. If instead the project intends `coalesced_width` to be a hard requirement rather than a hint, then at minimum replace the internal `LOG(FATAL)` with a clean user-facing diagnostic naming the kwarg and the achievable width — the current internal assert is the wrong failure mode either way. (Fix not built end-to-end this session.)

**Provenance.** Present at least since the block was last rewritten by [#1559](https://github.com/tile-ai/tilelang/pull/1559) ("[Parallel][Infer] Free-mode chooses minimal replication…", merged 2025-12-30) — blame attributes lines 798-808 to that commit, which was a layout-inference refactor that relocated/rewrote the surrounding code rather than introducing the check; the `coalesced_width` divisibility abort predates it (it exists unchanged in the 0.1.13 wheel I ran). Origin before #1559 is unverified. Not a regression. The permalinks above pin v0.1.13 (`8001cc4`).

**Dedup.** I searched the open and closed tracker (issues and PRs) for `coalesced_width`, `"not divisible by coalesced width"`, and `ComputePlanCandidate` and found no existing report of this defect. (#1802 concerns a stray `coalesced_width` kwarg passed to a Python test helper — a `TypeError`, unrelated to this compiler check; [#1424](https://github.com/tile-ai/tilelang/issues/1424) is a vectorization feature request.)

**Reach.** `coalesced_width` is documented on [`T.copy`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/copy_op.py#L58-L70) and used in shipped examples: `git grep -n coalesced_width v0.1.13 -- examples/ testing/` finds it in the AMD flash-attention examples (`examples/amd/example_amd_flash_attn_fwd.py:183/192/193`, `example_amd_flash_attn_bwd.py`) which pass `coalesced_width=8` (qk) / `4` (v) into `T.copy`, and in `testing/python/amd/test_tilelang_gfx950_copy_async.py` (cw=4/8 into `T.copy`/`T.async_copy`).

I verified the dodge by RUNNING the load-bearing copy geometry from the fwd example verbatim on 0.1.13 (L40S), isolated from the AMD/RDNA-only kernel body: `Q_shared[block_M, dim=128]` fp16 copies with `coalesced_width=8`, and `V` copies with `coalesced_width=4`, for the non-RDNA configs (`block_M∈{64,128,256}`, `dim=128`, `threads∈{128,256}`):

```
Q-copy block_M=64  dim=128 threads=128 cw=8 : PASS
V-copy block_M=64  dim=128 threads=128 cw=4 : PASS
Q-copy block_M=128 dim=128 threads=256 cw=8 : PASS
```

All PASS: the head-dim is a full contiguous 128-wide fp16 span, so the auto-derived vector size is ≥ the requested width and it divides cleanly — the abort is dodged, which is why CI is green. (The full `example_amd_flash_attn_fwd.py` targets ROCm/RDNA — `IsRDNA()` gates the kernel body and its config grid — so end-to-end it autotunes but is not the correct target on an NVIDIA sm_89; the standalone copy-geometry run above is the faithful, runnable slice of its `coalesced_width` usage.) The crash needs the intersection of (a) a user-supplied `coalesced_width` (rare — only the AMD examples/tests do it in-repo) with (b) a copy region whose geometry forces the derived vector size below the requested width — a ragged/partial tile at the array edge, or simply a `coalesced_width` larger than or not dividing the achievable vector (e.g. `cw=8` on a tile whose derived vector is 4). No shipped test exercises `coalesced_width` on such a region, so the branch has no coverage.

**Impact.** Firing needs the narrow intersection of a user-supplied `coalesced_width` hint and a copy region whose geometry forces the derived vector size below it (a ragged edge tile). When it fires it is a compile-time `LOG(FATAL)` — loud and caught immediately at build, blocking that one kernel from compiling but corrupting no output and unable to reach a running workload silently. Fixing it removes the internal-assert failure on a legal documented kwarg (clamping the hint like the auto path already does), turning a crash on a valid input into a compiled kernel or a user-facing diagnostic.

**Generalization (root + 4-axis sweep).**

Two-level root:
- **SOURCE-level.** [`ParallelOpNode::ComputePlanCandidate`](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L798-L810) consumes the `kCoalescedWidth` annotation as a **hard `%`-divisor** of the already-geometry-reduced `vector_size`: [line 802-804](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L802-L804) `LOG(FATAL)`s when `vector_size % expected != 0`, and [line 808](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L808) `LOG(FATAL)`s when the annotation is not an `IntImmNode`. Neither clamps, unlike the auto-derived path just above ([:783-787](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L783-L787) halves until it fits). Note [:789](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L789): when `coalesced_width` is set, the `SelectMinPaddingVectorSize` softening is *skipped*, so the user value meets the raw derived vector head-on.
- **OPERATOR-level.** Any frontend op that attaches the `coalesced_width` annotation and lowers through this planner: `T.copy` / `T.async_copy` (both variants in [copy_op.py](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/copy_op.py)) and the raw `T.Parallel` loop ([loop.py:79-80](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/loop.py#L79-L80)).

**4-axis sweep** (all cells RUN on 0.1.13 / L40S, one fresh process each):

| axis | cell tested | observed result | same root? |
|---|---|---|---|
| (repro) | `T.copy` global↔shared fp32, N=33 ragged tail, cw=2 | `CRASH: Vector size 1 is not divisible by coalesced width 2` | — |
| similar-logic | `T.copy` full 128×128 tile, derived vector=4, **cw=4** | PASS | same (divides) |
| similar-logic | `T.copy` full 128×128 tile, **cw=8** (exceeds derived 4) | `CRASH: Vector size 4 is not divisible by coalesced width 8` | **YES** — not ragged-only; any cw>derived |
| similar-logic | `T.copy` full 128×128 tile, **cw=3** (non-divisor of 4) | `CRASH: Vector size 4 is not divisible by coalesced width 3` | **YES** — strict `%`-divisor, not just `>` |
| related-operator | `T.async_copy` global→shared, N=33 ragged, cw=2 | `CRASH: Vector size 1 is not divisible by coalesced width 2` | **YES** — same line 804 |
| related-operator | `T.Parallel(M,N,coalesced_width=2)`, N=33 ragged | `CRASH: coalesced_width should be an IntImmNode.` | **distinct** — line 808, fires *before* 804 |
| related-operator | `T.Parallel(M,N,coalesced_width=2)`, N=32 clean/vectorizable | `CRASH: coalesced_width should be an IntImmNode.` | **distinct** — geometry-independent, always fires |
| related-type | `T.copy` **fp16** N=33 ragged, cw=2 | `CRASH: Vector size 1 is not divisible by coalesced width 2` | **YES** — dtype-independent (check is on element count) |
| related-type | `T.copy` **int8** N=33 ragged, cw=2 | `CRASH: Vector size 1 is not divisible by coalesced width 2` | **YES** — dtype-independent |
| related-type (scope) | `T.copy` global↔**fragment** (register) fp32, N=33 ragged, cw=2 | `CRASH: Vector size 1 is not divisible by coalesced width 2` | **YES** — fragment-access branch ([:782](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L782)) reaches the same 804 |

**What the sweep establishes.**
- The class is **broader than "ragged tail"**: the divisibility abort fires for *any* `coalesced_width` that is not a divisor of the geometry-derived vector size — a too-large width (`cw=8` on derived-4) and a non-power-of-2 width (`cw=3`) crash on a fully-rectangular tile too. The draft title/Problem are accordingly framed around the width-vs-derived-vector divisibility guard, with the ragged tail as the most common way to force the derived vector to 1. Kept the ragged-tail example as the minimal repro but the boundary note (`<details>`) already covers the full-tile cw=8 face; both are now RUN-confirmed this session.
- It is **operator-general** across `T.copy` and `T.async_copy` (both hit line 804) and **dtype/scope-general** (fp32/fp16/int8, shared and fragment destinations) — one root, reachable through every data-movement op that annotates `coalesced_width`.
- **Distinct-adjacent bug (not filed here).** `T.Parallel(..., coalesced_width=<python int>)` hits the *other* `LOG(FATAL)` at [line 808](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/src/op/parallel.cc#L808) ("coalesced_width should be an IntImmNode") **unconditionally, on any geometry** — the raw Python int stored by [loop.py:80](https://github.com/tile-ai/tilelang/blob/8001cc4ccf6149382d2019654a19f59c1d4d0482/tilelang/language/loop.py#L80) is never wrapped to an `IntImm`, so it never reaches the divisibility check. This is a *separate* root (missing type coercion on the `T.Parallel` frontend path, not the divisibility guard) and would be reported separately; noted here as the adjacent sibling this sweep confirmed. `T.copy`/`T.async_copy` do not hit it because their op machinery wraps the int into an `IntImm` before it reaches the planner.
- **Example-run result (PART 1).** The only in-repo `coalesced_width` consumers are the AMD flash-attention examples/tests. Their `T.copy(... coalesced_width=8/4)` on full contiguous head-dim-wide (128-elem fp16) tiles was run standalone on 0.1.13 and all PASS (see Reach) — they **dodge** the abort because their derived vector ≥ requested width. No shipped example/test exercises `coalesced_width` on a ragged or under-vectorizable region, so the branch is uncovered. No NEW bug surfaced while sweeping (the `T.Parallel` non-IntImm abort is the pre-noted distinct sibling).

edragain2nd:
fixes #3007
Treat coalesced_width as a hint during layout inference and clamp it to the achievable vector size instead of aborting when the requested width is not divisible by the geometry-derived width. Add regression coverage for ragged and oversized-width copy cases.

## Summary

- Treat `coalesced_width` as a hint in `ParallelOpNode::ComputePlanCandidate`.
- Clamp incompatible widths with `std::gcd` instead of triggering `LOG(FATAL)`.
- Warn when the requested width is reduced.
- Reject non-positive widths.
- Add regression tests for ragged and oversized `T.copy` cases.

## C++ style / lint notes

- The PR does not modify rules in `docs/developer_guide/cpp_style.md`.
- The **C++ API Style Audit (warning only)** CI step is relevant.
- No audit findings or CI results are provided.
- Warning-only style findings are not correctness or build failures.

## Testing

- Tests lower kernels without device compilation.
- Coverage includes ragged and full-tile copies with incompatible widths.
- Test pass status is not provided.


LeiWang1999 (commented):
Thanks for the fix — treating `coalesced_width` as a hint is the right call, and the ragged-tile regression test covers the original report well.

I pushed a small follow-up commit (3f0ef3e0) to this branch rather than leaving line comments, since it's easier to show than describe:

- **Clamp rule.** `std::gcd(vector_size, expected)` collapses any request that shares no factor with the geometry width down to 1: with a `float4`-capable copy, `coalesced_width=257` (your second test) produced a fully scalar copy, and `3` did too, while `6` landed on `float2`. Since every divisor of `vector_size` inherits the alignment and contiguity already proven for `vector_size`, the natural reading of the hint is an upper bound: take the widest achievable width not exceeding the request. That gives `8`/`257`/`6 → 4` and `3 → 2`. For power-of-two requests the two rules agree, so this is about making the fallback predictable rather than a behaviour change for typical inputs.
- **Test.** The oversized case now runs for both `8` (the realistic scenario, e.g. a hint of 8 on a geometry that only aligns to 4) and `257`, asserts `using 4 instead`, and checks that `float4` is actually emitted — previously `"using " in warnings` would have passed with the scalar fallback.

Everything else in the PR is unchanged. Local runs of this file plus the layout-inference and reducer tests pass; we can merge once CI is green.

PR #3247, opened 2026-09-18
[TileIR] Add a CUDA Tile IR execution backend

0xtaruhi:
## Summary

Add a `tileir` execution backend that lowers TileLang programs to NVIDIA CUDA Tile IR and launches the assembled cubin through the cuTile runtime. This includes JIT compilation, caching, autotuning, typed IR lowering, and source-language semantic checks before TileIR launch materialization.

- Target CUDA Tile IR 13.4 bindings / tileiras 13.4 / cuTile 1.5, with installation documentation and a dedicated CI job that builds the pinned bindings.
- Keep TileIR launch and copy hints in `tilelang.tileir.language`, compatible with the current common language APIs.
- Include the tuned DeepSeek V4 sparse-attention kernel and its FP16/BF16 regression tests. Pro defaults use H128/BN64/CTA2; Flash H64/D512/topk512 uses H64/CTA1/eight worker warps. The example documents both configurations and the cuTile stride limit.
- Preserve attention sinks, masked padding, and zero output for all-padding queries. Stage and reuse TopK indices, with TMA disabled for the index copy.
- Cover packed tensor dispatch, loop-carried state, conditional load tokens, and explicit unsupported-example diagnostics.

## Validation

Validated on NVIDIA B200 with Python 3.12, CUDA Tile IR v13.4.0 bindings, tileiras 13.4.92, and cuTile 1.5.0. The native library was rebuilt using the compute node's CUDA 13.5 toolkit; generated CUDA kernels and TileIR used the 13.4.92 tools.

- Sparse-attention correctness: **14 passed**, including Pro/Flash schedules, FP16/BF16, padding, and all-padding rows.
- TileIR, TileIR JIT/cache/runtime/autotuning, atomic-save, backend registration, and constant-rebind regression coverage: **758 passed, 2 expected failures**.
- Eager frontend test file: **7 passed** after installing the missing CCCL headers in the test environment.
- Pro and Flash command-line examples passed sampled reference checks; measured **2.957 ms** and **0.119 ms**, respectively, with the documented B200 configurations.
- `bash format.sh`, `git diff --check`, and Python 3.10 syntax checks for all 128 changed Python files passed.


Nicholas-Malone:
As an outside observer, my two cents:

Functionally, the addition of this backend is understandable, and it validates TileLang's multi-backend support capability — the JIT/cache/autotuning integration and the test coverage look solid.

However, I have a concern about where TileIR sits in the abstraction landscape. If we take TileLang as the origin of the coordinate system, then TileIR occupies a position similar to Triton's — that is, on the opposite side from CUDA C++ and CuTe DSL. The CUDA C++ / CuTe DSL backends lower TileLang toward targets that preserve its explicit tile-level scheduling decisions, whereas TileIR is itself a compiler IR that re-derives much of the scheduling internally. Lowering TileLang to TileIR therefore moves in the opposite direction — in a sense, it is a regression, because the explicit scheduling control that TileLang provides is handed back to another compiler.

PR #3264, opened 2026-09-21
[BugFix][NVRTC] Preserve host scalar launch arguments

### LINKED ISSUE #3263: [BUG] NVRTC wrapper omits host-computed scalar kernel arguments
### Required prerequisites

`0.1.14+cuda.git6ba187e2` (source checkout).

Initially reproduced at `eab74a4a`; the affected wrapper is unchanged in `6ba187e2`.

### System information

- Installation: source build with CUDA enabled, CMake/Ninja.
- OS: Windows x64 (`win32`).
- Python: 3.12.13, 64-bit.
- PyTorch: `2.11.0+cu130`.
- NVRTC: 13.0.

### Problem description

The NVRTC wrapper can omit scalar kernel arguments computed on the host after host/device splitting.

For valid host IR such as:

```python
p = n * 3 + 1
q = p * 2 + 5
T.call_packed("main_kernel", A, B, p, q, 128)
```

the device signature is `main_kernel(A, B, p, q)`, but the generated NVRTC dispatcher contains only:

```python
arg_values = A.data_ptr(), B.data_ptr()
arg_types = ctypes.c_void_p, ctypes.c_void_p
```

`SplitHostDevice` correctly captures the prepared values. The wrapper subsequently matches CUDA parameter names against the original function's arguments, where `p` and `q` do not exist. The existing TVM-FFI path executes the same host preparation correctly.

This may have gone unnoticed because ordinary compilation inlines this example's arithmetic into the kernel before `SplitHostDevice`: `A[i] + p + q` becomes `A[i] + 9 * n + 8`, leaving only the original scalar parameter `n` to forward. This happens before the NVRTC adapter; writing the bindings outside `T.Kernel` does not prevent it.

The reproducer preserves the host bindings to exercise the adapter boundary. It does not demonstrate a failure through the default compilation pipeline. Launch-time hoisting such as #3261 could expose this gap if the computed values remain on the host and become additional kernel arguments.

### Reproducible example code

Requires a CUDA-enabled TileLang source build. This inspects generated source without launching the incomplete argument list.

```python
import tilelang as tl
import tilelang.language as T
from tilelang import tvm
from tilelang.backend.module import create_backend_context
from tilelang.engine.lower import device_codegen_without_compile, get_device_call, get_host_call
from tilelang.jit.adapter.nvrtc.wrapper import TLNVRTCSourceWrapper


@T.prim_func
def main(A: T.Tensor((128,), "int32"), B: T.Tensor((128,), "int32"), n: T.int32):
    p = T.bind(n * 3 + 1)
    q = T.bind(p * 2 + 5)
        i = T.get_thread_binding()
        B[i] = A[i] + p + q


ctx = create_backend_context({"kind": "cuda", "arch": "sm_80"}, "c", "nvrtc")
original = tvm.IRModule({"main": main})
# Preserve host bindings to exercise the adapter boundary.
mod = tvm.transform.Sequential(
    [
        tvm.tirx.transform.BindTarget(ctx.target),
        tl.transform.MaterializeKernelLaunch(),
        tl.transform.LowerOpaqueBlock(),
        tl.transform.AnnotateDeviceRegions(),
        tl.transform.SplitHostDevice(),
        tvm.tirx.transform.AnnotateEntryFunc(),
        tl.transform.MakePackedAPI(),
        tl.transform.LowerDeviceKernelLaunch(),
    ]
)(original)
host = tvm.tirx.transform.Filter(get_host_call())(mod)
device = tvm.tirx.transform.Filter(get_device_call())(mod)
source = device_codegen_without_compile(device, ctx).inspect_source()
wrapper = TLNVRTCSourceWrapper(original, source, ctx.target, device_mod=device, host_mod=host)
print(source)
print(wrapper.host_func)
```

### Traceback

No traceback is produced by the source-generation reproducer. The malformed dispatcher is inspected, not executed with an incomplete CUDA argument list.

### Expected behavior

Resolve launch operands from the actual host call, preserve the required scalar bindings and their execution scope, and marshal every device argument in signature order. Unsupported host computations should produce an explicit diagnostic rather than an incomplete CUDA argument list.

### Additional context

The existing TVM-FFI path and a separate IR-driven NVRTC prototype passed changing-input and guarded-launch checks on the same host/device program.

This differs from #2755/#2756, which concerned forwarding original scalar parameters and dynamic strides. The missing values here are introduced by host preparation and are not original function parameters. Related preparation work is discussed in #3261.

sepcnt:
Fixes #3263.

The NVRTC dispatcher currently matches CUDA parameter names against original function arguments, dropping scalars introduced by host preparation. Resolve operands from host call sites and pair them with device parameters in signature order instead.

- Emit required integer bindings with dtype-preserving conversions.
- Preserve launch conditions and repeated call sites; remap launch geometry to host operands.
- Retain TMA descriptor handling and reject unsupported host expressions or looped launches explicitly.
- Add DSL-based regressions for host scalar preparation, guarded launches, integer conversions and TMA arguments.

## Summary

- Resolve NVRTC launch operands from host IR call sites.
- Preserve host-computed scalar bindings, integer dtypes, launch conditions, and repeated launches.
- Match operands to device parameters in signature order.
- Retain TMA descriptor handling.
- Reject unsupported expressions, missing launch sites, argument-count mismatches, and looped launches with diagnostics.
- Add end-to-end and IR tests for scalar preparation, conditional launches, integer division, repeated launches, loop rejection, and TMA descriptors.

## Testing

- Added CUDA/NVRTC-gated execution tests and host-launch collection tests.
- Test execution results were not provided.


LeiWang1999:
Thanks for investigating this and adding the regression coverage. We’re closing this PR in favor of integrating NVRTC device compilation with the existing TVM-FFI host execution path. The bug reported in #3263 remains valid and should stay open until that replacement is implemented and verified.

The diagnosis here is correct: matching CUDA parameter names against the original function arguments cannot handle scalars introduced by host preparation. The actual host call operands, in device-signature order, are the correct source of truth.

The concern is how we preserve the rest of the host program. `collect_host_launches` and `HostScalarEmitter` introduce another implementation of host IR semantics inside the NVRTC adapter: tracking bindings and scopes, reproducing dtype conversions, evaluating expressions, and reconstructing conditional execution. Supporting more host preparation would require extending this implementation to cover loops, floating-point operations, external helper calls, and their execution order. These responsibilities already belong to the host lowering and code generation path used by TVM-FFI.

We found a concrete semantic mismatch during review. For host parameters declared as `float32`, consider:

```python
if a == b:
    # launch the kernel
```

With Python inputs `a = 1.0` and `b = 1.0 + 2**-25`, both values become `1.0` when converted to `float32`, so the branch should execute. The current Python launcher compares the original Python floating-point values and skips the launch. This illustrates the maintenance burden of reproducing host semantics separately in each adapter.

The direction we want to pursue is to separate device compilation from host execution:

- NVRTC compiles generated CUDA source into cubin/PTX.
- The existing CUDA runtime module consumes that binary and its kernel metadata.
- The existing host codegen compiles the host IR, and TVM-FFI executes the resulting module with the imported CUDA module.

The necessary composition points already exist. `BuildTileLangCUDA` obtains the device binary through `tilelang_callback_cuda_compile` and constructs a CUDA runtime module. `lower.py` builds the host module and imports the device module. The compile callback currently uses NVCC; consuming its output through TVM-FFI does not inherently require NVCC.

During review, a process-local experiment replaced that callback with NVRTC compilation while retaining `execution_backend="tvm_ffi"` and explicitly disallowing NVCC compilation. The kernel compiled and executed successfully on the GPU, and the float32 conditional example behaved correctly. This establishes the feasibility of the composition; a production integration still needs compiler-option handling, appropriate cache keys, and validation of launch features and export/load behavior.

There is also a toolchain tradeoff to address explicitly: the C host codegen path requires a host C/C++ compiler, while the current Python launcher avoids that compilation step. Windows support and any requirement to run without a host compiler should be considered in the replacement design. An intentionally restricted fallback can be evaluated if that requirement is needed.

Please carry forward the useful regression cases from this PR, especially host-prepared scalar arguments, guarded launches, dtype conversions, repeated launches, and TMA arguments. We should use #3263 to track the NVRTC + TVM-FFI integration and ensure those behaviors are covered through the shared host execution path.

Issue #3261, opened 2026-09-21
[Feature Request] [Perf] LowerMagicDiv: host-precomputed magic division for launch-invariant dynamic divisors


penguin-wwy:
## Summary

`tl.LowerMagicDiv` rewrites integer division/modulo by a **launch-invariant but compile-time-unknown positive int32 divisor** into host-precomputed magic numbers, with each device thread executing only `umulhi + shift`:

```text
q = umulhi(uint32(x), M) >> s
r = x - q * d
```

A plain dynamic integer division typically expands into long reciprocal-estimation sequences, corrections, or helper calls. Magic division moves all `d`-dependent expensive work to the host (computed **once per kernel launch**), so every device thread performs only an unsigned multiply-high and a right shift. This is **not** a floating-point approximate division — within the specified input domain the result is bit-exact with integer division.

## 1. Mathematical Principle

For `x = q * d + r` (dividend `x`, divisor `d`, quotient `q`, remainder `r`) with `d ≥ 2`:

```text
k = ceil(log2(d))
p = 31 + k
M = ceil(2^p / d)
s = p - 32 = k - 1
```

For `0 ≤ x < 2^31`:

```text
floor(x / d) = floor(x * M / 2^p)
             = umulhi(uint32(x), M) >> s
```

where `umulhi(a,b)` returns the high 32 bits of the unsigned 32×32 product, i.e. `umulhi(a,b) = floor(uint64(a) * uint64(b) / 2^32)`; the following right shift by `s = p - 32` completes the divide-by-`2^p` floor.

**Why it is exact.** The key is that `M` is rounded *up*. Writing `M = 2^p/d + ε` with `0 ≤ ε < 1`:

```text
x * M / 2^p = x / d + x * ε / 2^p
```

Since `x < 2^31` and `p = 31 + k`:

```text
0 ≤ x * ε / 2^p < 1 / 2^k ≤ 1 / d
```

With `x = q·d + r` (`0 ≤ r < d`), `x/d` is at least `1/d` away from the next integer; the additive error is strictly smaller than that distance, so it never crosses an integer boundary and the floor still yields the exact `q = floor(x/d)`. This is also why the fast path requires a non-negative dividend below `2^31`.

### Worked example: 100 / 7

Host precomputation (`d = 7`):

```text
k = ceil(log2(7)) = 3,  p = 31 + k = 34
M = ceil(2^34 / 7) = 2454267027 = 0x92492493
s = p - 32 = 2
```

[...]

PR #3265, opened 2026-09-21
[CUDA][Transform] Lower launch-invariant integer arithmetic

sepcnt:
**Near-specialized performance for dynamic shapes, without per-shape recompilation.**

Related issue: https://github.com/tile-ai/tilelang/issues/3261.

Most of the credit for the real-world improvement belongs to @penguin-wwy. Without the core idea from #3267, this work would not have benefited the real-world cases below. The remainder canonicalization and factor matching substantially follow that implementation; the MoE permutation kernel, reference and shapes also come from its benchmark. Thanks also for the original analysis in #3261.

## Real-operator comparison from #3267

All three of @penguin-wwy's MoE scale-permutation cases are retained in `benchmark/launch_plan/bench_permute_scales.py`, with the original dynamic kernel body, reference and shapes, without added assumptions. Shapes `(experts, N, K, group size)` are `(256, 2048, 7168, 32)` for both DSV3 cases and `(8, 4096, 8192, 128)` for Llama4.

RTX 5090, Linux, NVCC 13.2.78. Saved CUDA sources from [#3267 at d53291a4](https://github.com/tile-ai/tilelang/commit/d53291a428a24b550b611ab88eb6e25b77875b4f) and this implementation are compiled with the same options and timed in one process: 21 interleaved rounds, 64 launches per graph, five replays per sample. Times are median GPU microseconds and exclude host preparation. The C host bit-scan update leaves these device sources unchanged. Speedups are medians of per-round paired ratios, rather than ratios of the independently aggregated times. These measurements do not reproduce the H200 results.

| Case | Off (µs) | #3267 (µs) | Dynamic, this PR (µs) | Const (µs) | Off / dynamic | #3267 / dynamic | Dynamic / const |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| DSV3 W4A16 | 1062.118 | 1046.904 | 968.041 | 968.385 | 1.096× | 1.079× | 1.000× |
| DSV3 W4A8 | 1052.271 | 1037.397 | 975.814 | 974.034 | 1.077× | 1.063× | 1.002× |
| Llama4 W4A16 | 12.584 | 11.874 | 10.734 | 10.290 | 1.171× | 1.106× | 1.043× |

Const fixes the shape at compile time with this pass disabled, while retaining the same kernel body, 256-thread layout and reference. It is an approximate specialization ceiling for this schedule, not a strict performance bound. Dynamic latency is within about 0.2% of Const for the two DSV3 cases and 4.3% for Llama4 in this run. The aim is to approach specialization performance with a reusable dynamic kernel instead of compiling a variant for every shape.

All four variants pass exact output comparisons and poisoned-output graph replay checks. `bench_permute_scales.py` measures native TVM-FFI Off/on/Const; `bench_permute_sources.py` reproduces the GPU-only table from their saved sources and the #3267 sources.

## Implementation

TileLang adds opt-in `LowerInvariantArithmetic`, enabled with `tl.enable_invariant_arithmetic`. It prepares integer division/remainder parameters on the host before `SplitHostDevice`, which captures them as flat kernel arguments.

- Supports scalar signed/unsigned 8-, 16-, 32- and 64-bit arithmetic, including widened index expressions.
- Shares multiplier/shift or Barrett parameters between div/rem and equivalent widened uses.
- Selects reciprocal width from dividend magnitude bounds, rather than divisor storage width. Signed floor/truncating div/rem use unsigned magnitude arithmetic; negative and wide inputs no longer select ordinary-division fallbacks.
- Preserves original-width host arithmetic wrap before extracting sign and magnitude: `uint64(h * w)` and `uint64(h) * uint64(w)` are not interchangeable.
- Eliminates redundant coordinate reconstruction and reuses existing quotient prefixes when positive, non-overflowing divisor products are proven.
- Replaces bounded swizzle remainders with one compare/subtract, without reciprocal preparation. Bounds can come from existing conditions or explicit min/max clamps; the pass does not add assumptions.

Existing conditions, assumptions and range information supply compile-time sign proofs to the device intrinsics, without extra launch arguments. Positive factors only establish a positive product when its range excludes overflow. Eligible unsigned computations stay in 32 bits until the result is widened; identity-divisor results are selected after reciprocal arithmetic to preserve common subexpressions.

Preparation is structurally deduplicated within each device region. Device operations remain intrinsics until CUDA codegen, rather than expanding nested guarded formulas during IR simplification. The pass uses TVM-FFI without adapter changes. C host codegen emits scalar 32/64-bit bit-scan helpers for `clz`, replacing the comparison-sum expansion, with explicit zero handling and a portable fallback. CUDA Graph replay retains capture-time parameters; host preparation runs during capture, not replay.

Late materialization reuses pure arithmetic already evaluated by a dominating condition inside its branches. It does not speculate across lazy guards or reuse mutable loads. Remainder normalization is folded into the existing lowering traversal, keeping two passes rather than adding a separate normalization pass. It also recovers remainders split across reassociated swizzle expressions, including missing multiples of an outer power-of-two modulus. This avoids recomputing coordinate quotients without adding assumptions or narrowing indices. Within a single unguarded load address, an existing remainder also replaces its matching quotient product using $qd=x-r$, preserving word width; reuse does not cross lazy expressions or statements.

Launch-extent types provide additional index bounds without preventing device index widening. Potentially wrapping arithmetic blocks mathematical sign/range proofs; remainder and bitwise operations can re-establish facts about their results independently of the input range.

### FastDiv / FastRem

For divisor magnitude $1<d\le2^{31}$ and normalized dividend $0\le x\le2^{31}$, define

$$
s=\lfloor\log_2(d-1)\rfloor,\qquad P=2^{32+s},\qquad
m=\left\lceil\frac{P}{d}\right\rceil.
$$

For this divisor range, $m$ fits uint32. Write $md=P+e$ and $x=qd+r$, with $0\le e<d\le2^{s+1}$ and $0\le r<d$. Since $xe<P$,

$$
\frac{xm}{P}=q+\frac{r+xe/P}{d},\qquad
0\le\frac{r+xe/P}{d}<1.
$$

Thus `mulhi32(x, m) >> s` gives $q=\lfloor x/d\rfloor$ exactly, and $r=x-qd$. Since $qd\le x\le2^{31}$, FastRem can compute its magnitude multiply/subtract in uint32 and widen only the result. The $d=1$ result is selected separately.

### Barrett quotient and remainder

Let $B=2^w$, where $w\in\{32,64\}$, $0\le x<B$, and $d>1$. Prepare $\mu=\lfloor B/d\rfloor$ and compute $\widehat q=\lfloor x\mu/B\rfloor$ by high multiplication. Then

$$
0\le\frac{x}{d}-\frac{x\mu}{B}
=\frac{x(B-\mu d)}{Bd}<1.
$$

Therefore $\widehat q$ is either $q$ or $q-1$. The residual $\widehat r=x-\widehat qd$ fits the dividend word, and one correction suffices:

$$
c=\mathbf{1}[\widehat r\ge d],\qquad
q=\widehat q+c,\qquad r=\widehat r-cd.
$$

This also covers a wider divisor: if $d\ge B$, then $\mu\le1$ and the same bound applies. The correction compares against the full-width divisor. Native 64-bit preparation needs no host int128:

$$
\left\lfloor\frac{2^{64}}d\right\rfloor
=\left\lfloor\frac{2^{64}-1}d\right\rfloor
+\mathbf{1}\!\left[(2^{64}-1)\bmod d=d-1\right].
$$

The $d=1$ case is handled separately.

### Signed floor and truncating arithmetic

Magnitudes $A=|x|$ and $D=|d|$ are computed with unsigned arithmetic, including the signed minimum. For truncation, divide $A$ by $D$ and restore the quotient sign from both operands and the remainder sign from $x$.

For floor division, define

$$
b=\mathbf{1}[(x<0)\ne(d<0)\ \land\ A\ne0],\qquad
A-b=qD+r,\qquad 0\le r<D.
$$

The quotient magnitude is $q+b$. The remainder magnitude is $r$ when $b=0$, otherwise $D-1-r$, with the divisor's sign. This follows from $A=qD+r+1$ in the opposite-sign case. A positive-divisor specialization uses bitwise complement to normalize negative dividends without per-lane fallback branches. Unsigned operands omit sign restoration.

Only a zero divisor retains the original fallback; a nonzero proof removes it. Division by zero and the unrepresentable signed minimum divided by $-1$ are outside the correctness claim.

### Layout decoding and bounded swizzles

For positive $a,b$ with a non-overflowing product, let $q=\lfloor x/a\rfloor$. Then

$$
\left\lfloor\frac{x}{ab}\right\rfloor=\left\lfloor\frac{q}{b}\right\rfloor,\qquad
\left\lfloor\frac{x}{ab}\right\rfloor b+(q\bmod b)=q.
$$

This removes redundant coordinate reconstruction, including for negative $x$. Quotient factorization is applied only when the corresponding prefix already occurs. Products that may wrap are left unfactored.

For $0\le r,m<d$, the swizzled index satisfies $r\mathbin{\mathrm{xor}}m\le r+m<2d$. Its remainder therefore needs only one conditional subtraction. The bound is proved without evaluating an overflowing $2d$.

For reassociated expressions, let $x=t+a$, $q=\lfloor x/d\rfloor$ and $r=x-qd$. If $M=2^k$ and $M\mid a$, then

$$
(t-qd)\bmod M=(r-a)\bmod M=r\bmod M.
$$

This remains valid under $w$-bit wrapping when $k\le w$, since $M\mid2^w$. TileLang restores an omitted term only after proving its divisibility by $M$, without distributing products or crossing casts. Non-power-of-two outer moduli do not use this rule.

## Real-operator host-wall comparison

Native TVM-FFI measurements below include host preparation and synchronization, excluding compilation; 21 interleaved rounds. This is a separate Off/on/Const comparison, not a host-overhead comparison with #3267.

| Case | Wall Off (µs) | Wall dynamic (µs) | Wall Const (µs) | Off / dynamic |
| --- | ---: | ---: | ---: | ---: |
| DSV3 W4A16 | 1066.627 | 968.567 | 970.937 | 1.101× |
| DSV3 W4A8 | 1055.693 | 976.457 | 977.349 | 1.081× |
| Llama4 W4A16 | 13.692 | 12.506 | 12.502 | 1.095× |

### Host preparation ablation

Replacing the C host `clz` expansion leaves device sources unchanged. In two complete old/new ablations (31 interleaved rounds each), positive-parameter gather measures 1.039× wall speedup at 4,096 elements and 1.031–1.044× at 1,048,576 elements. The three real MoE cases show no meaningful end-to-end change from this host-only adjustment.

An isolated scalar FastDiv-preparation benchmark measures 16.20 ns with the comparison sum and 3.74 ns with bit scanning (4.33×); this is parameter-preparation cost, not kernel or end-to-end speedup.

## Representative results

Same GPU and toolchain, TVM-FFI. Medians of 11 interleaved rounds, 256 launches per sample. GPU time uses CUDA graphs; wall time includes host preparation and synchronization, excluding compilation.

**Off** disables the optimization; **Lower** enables host-prepared arithmetic; **Late** additionally enables late device materialization. Lower and Late have identical host code. Speedups compare Off with Late.

### 1,048,576 elements

| Workload | GPU Off (µs) | GPU Lower (µs) | GPU Late (µs) | Speedup | Wall Off (µs) | Wall Lower (µs) | Wall Late (µs) | Speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Swizzled layout decode | 4.811 | 4.607 | 4.552 | 1.06× | 7.398 | 7.130 | 7.170 | 1.03× |
| Swizzled, pitched gather | 13.720 | 12.036 | 7.137 | 1.92× | 14.491 | 14.400 | 8.279 | 1.75× |
| Alignment-checked gather | 8.649 | 8.700 | 8.590 | 1.01× | 10.315 | 10.314 | 10.303 | 1.00× |
| Banked layout | 4.341 | 4.342 | 4.319 | 1.01× | 6.921 | 6.594 | 6.547 | 1.06× |
| Native int64 pitched layout | 4.721 | 4.417 | 4.350 | 1.09× | 7.109 | 6.489 | 6.595 | 1.08× |
| Native int64 remainder | 4.626 | 4.341 | 4.340 | 1.07× | 6.489 | 6.818 | 6.804 | 0.95× |
| Int64 layout, int32 divisor | 4.725 | 4.352 | 4.350 | 1.09× | 6.743 | 6.737 | 6.744 | 1.00× |
| Int64 remainder, int32 divisor | 4.628 | 4.332 | 4.337 | 1.07× | 7.364 | 7.296 | 7.375 | 1.00× |
| Six-stage mixed-sign div/rem chain | 6.082 | 4.597 | 4.598 | 1.32× | 8.257 | 6.648 | 6.689 | 1.23× |
| Gather, NonZero | 13.748 | 9.669 | 6.308 | 2.18× | 14.692 | 10.993 | 8.231 | 1.78× |
| Gather, positive parameters | 8.780 | 4.777 | 4.623 | 1.90× | 10.332 | 8.261 | 8.283 | 1.25× |
| Gather, bounded positive parameters | 8.735 | 4.463 | 4.456 | 1.96× | 10.326 | 7.009 | 7.016 | 1.47× |
| Gather, unsigned parameters | 4.999 | 4.516 | 4.515 | 1.11× | 7.236 | 7.934 | 7.929 | 0.91× |
| Gather, bounded swizzle | 8.723 | 4.456 | 4.446 | 1.96× | 10.324 | 7.016 | 6.926 | 1.49× |
| Gather, clamped parameters (no assumptions) | 8.105 | 4.506 | 4.494 | 1.80× | 10.298 | 7.610 | 7.564 | 1.36× |

The original gather has no added assumptions. The four additional variants use `h,w,c != 0`, `h,w,c > 0`, `0 < h,w,c <= 1024`, and uint32 parameters with nonzero assumptions, respectively. The bounded variant also establishes non-overflowing divisor products. The bounded-swizzle variant additionally assumes `c >= 32`. The clamped variant clamps `h` and `w` to `[1,1024]` and `c` to `[32,1024]`, with no added assumptions. Each variant is compared against its own Off baseline.

`benchmark/launch_plan/bench_three_stage.py` retains all 15 workloads at both sizes and all three modes (90 combinations).

<details>
<summary>4,096-element results, same workloads and measurement protocol</summary>

| Workload | GPU Off (µs) | GPU Lower (µs) | GPU Late (µs) | Speedup | Wall Off (µs) | Wall Lower (µs) | Wall Late (µs) | Speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Swizzled layout decode | 0.801 | 0.944 | 0.902 | 0.89× | 7.191 | 6.628 | 6.781 | 1.06× |
| Swizzled, pitched gather | 1.580 | 1.695 | 1.394 | 1.13× | 7.364 | 7.750 | 7.854 | 0.94× |
| Alignment-checked gather | 1.360 | 1.360 | 1.170 | 1.16× | 7.219 | 7.195 | 7.232 | 1.00× |
| Banked layout | 0.730 | 0.744 | 0.728 | 1.00× | 6.817 | 6.852 | 6.771 | 1.01× |
| Native int64 pitched layout | 0.888 | 0.768 | 0.728 | 1.22× | 6.433 | 6.716 | 6.788 | 0.95× |
| Native int64 remainder | 0.864 | 0.728 | 0.729 | 1.19× | 6.686 | 6.476 | 6.403 | 1.04× |
| Int64 layout, int32 divisor | 0.896 | 0.728 | 0.728 | 1.23× | 7.046 | 6.970 | 6.918 | 1.02× |
| Int64 remainder, int32 divisor | 0.872 | 0.728 | 0.728 | 1.20× | 6.625 | 6.611 | 6.614 | 1.00× |
| Six-stage mixed-sign div/rem chain | 1.032 | 0.928 | 0.928 | 1.11× | 6.738 | 7.012 | 6.958 | 0.97× |
| Gather, NonZero | 1.568 | 1.433 | 1.232 | 1.27× | 7.528 | 7.851 | 7.720 | 0.98× |
| Gather, positive parameters | 1.320 | 1.024 | 0.984 | 1.34× | 7.379 | 7.598 | 7.564 | 0.98× |
| Gather, bounded positive parameters | 1.312 | 0.850 | 0.848 | 1.55× | 7.138 | 6.823 | 6.812 | 1.05× |
| Gather, unsigned parameters | 1.171 | 0.872 | 0.874 | 1.34× | 7.212 | 6.769 | 6.796 | 1.06× |
| Gather, bounded swizzle | 1.312 | 0.848 | 0.842 | 1.56× | 7.200 | 6.897 | 6.853 | 1.05× |
| Gather, clamped parameters (no assumptions) | 1.328 | 0.864 | 0.864 | 1.54× | 7.261 | 6.623 | 6.622 | 1.10× |

</details>

## Summary

- Adds opt-in `LowerInvariantArithmetic` through `tl.enable_invariant_arithmetic`.
- Prepares launch-invariant integer division and remainder parameters on the host before `SplitHostDevice`.
- Adds FastDiv/FastRem, Barrett reduction, and bounded remainder intrinsics.
- Adds CUDA support for signed, unsigned, narrow, and widened integer types.
- Adds launch-plan preparation and late materialization for shared pure arithmetic.
- Adds portable C and CPU support for 32-bit and 64-bit `tirx.clz`.
- Adds CUDA, CPU, benchmark, and CUDA Graph coverage.
- Removes the previously added exact-division path.

## C++ style / lint notes

- The PR changes C++ implementation, code generation, and public C++ declarations.
- The style guide documents rules for public APIs, FFI-visible names, and public API comments. The supplied evidence does not show a specific documented rule change.
- CI includes the “C++ API Style Audit (warning only)” step. Its execution result is unavailable.
- No correctness or build issue is established by the supplied evidence.
- TLCPP003/TLCPP004 findings remain advisory unless they identify a clear API, FFI, or maintainability risk.

## Tests and benchmarks

- Tests cover invariant arithmetic lowering, signed and unsigned widths, wrapping, lazy evaluation, materialization, dynamic shapes, and CUDA Graph capture.
- CPU tests cover `clz` for `int32`, `uint32`, `int64`, and `uint64`.
- Benchmarks cover MoE permutation and representative layout workloads.
- Current test and benchmark execution results are unavailable.


penguin-wwy:
Hi @sepcnt, thanks a lot for your interest in this issue and for putting together an implementation so quickly!
To be transparent: when I filed this issue, I was already organizing my code and preparing a PR of my own — the only reason it didn't go up together with the issue is that I needed some experiments to confirm which parts of the design are actually effective. Unfortunately, this PR conflicts with that design in a few important ways:
* **Validation on the original case.** This issue originated from a real-world case I ran into, and the bar for closing it is that the fix demonstrably works on that specific case. While the PR implements the related logic, it hasn't been verified against the original reproducer, so we can't be sure it actually resolves the problem in practice.
* **A unified foundation for follow-ups.** I have a series of follow-up optimizations designed on top of this change, and they all need to build on one unified, coherent foundation. Landing an implementation that hasn't been aligned with that overall design would make the follow-up work harder rather than easier.
* **Collaboration is very welcome.** I'd genuinely love to have you involved — but I'd prefer to first finish organizing the complete design and share it in the issue, and then we can work out together how to split and implement the pieces. Please stay tuned!

Thanks again for the contribution, and I hope you understand.

sepcnt:
@penguin-wwy I'd be glad if this work could help, and would be happy to collaborate.

Ablation experiments pointed to an important detail: eagerly expanding arithmetic expressions can produce very large nested ASTs and lose sharing. Keeping div/rem as intrinsics through simplification, then materializing repeated expressions in local scopes, recovered substantial performance without additional assumptions. The NVRTC adapter also has a separate gap in handling host-computed scalar bindings from SplitHostDevice; the current implementation uses TVM-FFI.

Host-prepared FastDiv/FastRem and Barrett parameters already give useful gains, though parameter preparation is not free. For swizzle-heavy layout calculations motivated by rendering and diffusion workloads, further gains came from proving divisors nonzero or positive and ruling out product overflow. Those proofs enable redundant coordinate reconstruction removal, quotient reuse, and single-subtraction swizzle remainders—not just faster division.

The table includes explicit-assumption variants and a clamped variant whose bounds are inferred without T.assume. Exploring how much of this information TileLang can derive from existing layout expressions, rather than requiring extra annotations, seems like a useful next step.

sepcnt:
@penguin-wwy TMA-related layout conversions are another direction not yet explored; they might offer more broadly useful applications. More generally, if runtime layout parameters can approach compile-time-specialized performance, fewer parameters would need to be constants, potentially reducing the number of kernel variants and their compilation cost substantially.

sepcnt:
@penguin-wwy The latest comparison with constant-shape versions looks encouraging: dynamic latency is within about 0.2% for the two DSV3 cases and 4.3% for Llama4 on RTX 5090, with the same kernel layout. These cases suggest that runtime layout parameters can approach specialized performance without compiling a separate variant for every shape. TMA-related layout conversions remain unexplored and might extend the benefit to more workloads.

PR #3267, opened 2026-09-21
[Transform] Add tl.LowerMagicDiv: host-precomputed magic division for dynamic-shape divisors

penguin-wwy:
## Summary

Adds optional host-precomputed magic division for dynamic-shape divisors.

- Adds `tl.LowerMagicDiv` and `tl.MagicCallHoist`.
- Adds `tl.magic_div`, `tl.magic_mod`, and `tl.magic_mod_from_quotient`.
- Adds CUDA, ROCm, CPU C, and host code generation support.
- Adds host runtime helpers for magic constants.
- Preserves fallback floor-division and floor-modulo behavior.
- Adds regression examples and compile-only tests.
- Rejects unsupported Cython/NVRTC launch paths with magic-division parameters.

## C++ style / lint notes

The PR changes C++ code and adds exported runtime functions. It does not change `docs/developer_guide/cpp_style.md` or the C++ lint configuration. The repository includes the “C++ API Style Audit (warning only)” CI step. Warning-only findings should not block merging unless they indicate an API, FFI, or maintainability risk.

## Review status

Current review finding counts and test results are unavailable.


LeiWang1999 (commented):
I rebuilt d53291a4 and ran the four added tests, which pass, along with three additional transpose shapes. The inline comments document five reproduced correctness failures: hoisting across memory writes and guards, suppressed index widening, unconditional host evaluation of guarded expressions, an out-of-range host-helper shift, and hoisting outside a loop variable's scope. These need to be addressed before this optimization is safe to merge.

src/transform/lower_magic_div.cc
```diff
@@ -0,0 +1,738 @@
...
+      }
+
+      Stmt VisitStmt_(const AttrStmtNode *op) final {
+        if (GetRef<AttrStmt>(op).same_as(scope_)) {
+          Stmt new_body = VisitStmt(op->body);
+          std::vector<Stmt> seq;
+          for (const auto &[var, value] : binds_) {
+            seq.emplace_back(Bind(var, value));
+          }
+          seq.push_back(new_body);
```
LeiWang1999, line 910: [P1] Preserve memory dependencies and guards when hoisting magic calls

Prepending every binding to the thread scope moves argument loads across both stores and bounds checks. For example, with a dynamic shape `n`:

```python
if idx < n:
    A[idx] = 7 * n
    B[idx] = T.max(A[idx], 0) // n
```

With `A` initially zero and `n=128`, the GPU result is all `7` with this pass disabled and all `0` with it enabled. The generated `tl_magic_q_0` reads `A[idx]` before the store. For a partial block, that same load also executes outside `idx < n`, introducing an out-of-bounds access.

Please keep memory-dependent calls at their original execution point unless dependence and control-flow analysis proves the move safe. Limiting CSE to a valid basic block is a conservative starting point. Add a numerical write-then-read regression and a partial-block test that checks the load stays guarded.

src/transform/common/int64_promoter.h
```diff
@@ -54,6 +55,19 @@ class Int64Promoter : public tirx::IndexDataTypeRewriter {
...
+
+  PrimExpr VisitExpr_(const tirx::CallNode *op) final {
+    // tl.magic_div has a fixed 32-bit contract (uint32 mul-high semantics).
+    // Promoting its operands or result would break the bit-pattern of the
+    // magic multiplier and leave widen/narrow pairs in device code, so the
+    // whole call subtree is left untouched; the surrounding index expression
+    // widens the int32 quotient with an explicit cast if needed.
+    if (op->op.same_as(tl::magic_div()) || op->op.same_as(tl::magic_mod()) ||
+        op->op.same_as(tl::magic_mod_from_quotient())) {
+      return ffi::GetRef<PrimExpr>(op);
```
LeiWang1999: [P1] Preserve index widening before selecting the int32 magic path

Returning the entire subtree unchanged suppresses widening of the dividend's intermediate arithmetic. `LowerMagicDiv` runs before `FlattenBuffer`/`ConfigIndexBitwidth`, and its eligibility check establishes non-negativity without proving that those intermediates fit int32.

I reproduced this with `B[tx] = A[((tx + base) * 4) // n]`, guarded by `n > 0` and `base >= 0`, using `base=2**30`, `n=7`, and `tx=0..31`. With the pass disabled, the address calculation is widened to int64 and reads indices `613566756..613566774`. With it enabled, the int32 dividend wraps to `0..124` and reads indices `0..17`. Filling the two regions with different sentinels gives GPU outputs of all `29` versus all `11`. This test disables safe-memory legalization to isolate widening; both sets of actual accesses are within the input buffer.

Please determine the required arithmetic width before introducing the fixed-width intrinsic, or prove that every relevant intermediate fits its contract and retain native widened division otherwise. A runtime check on an already-overflowed int32 dividend cannot recover the original index. This case should be a numerical regression.

src/transform/lower_magic_div.cc
```diff
@@ -0,0 +1,738 @@
...
+  // constant (multiplier / shift) for CUTLASS FastDivmod
+  // (k = ceil_log2(d), p = 31 + k, M = ceil(2^p / d), s = p - 32; 0 for
+  // d == 1 where the device expansion selects x).
+  Stmt WrapWithHostBinds(const Stmt &body) {
+    ICHECK(!entries_.empty());
+    std::vector<Stmt> seq;
+    for (size_t i = 0; i < entries_.size(); i++) {
+      const MagicDivisorEntry &e = entries_[i];
+      std::string suffix = std::to_string(i);
+      seq.emplace_back(Bind(e.d_var, e.divisor));
```
LeiWang1999, line 286: [P1] Prove divisor expressions are safe to evaluate at the host insertion point

Launch invariance does not imply that an expression can be evaluated unconditionally. Positivity can be established inside a device branch, but this binding is prepended to the whole function:

```python
if p > 0:
    B[tx] = tx // (7 // p + 1)
```

For `p=0`, the original kernel leaves `B` untouched. Enabling the pass emits unconditional host calls whose argument contains `7 // p + 1`, so the host divides by zero before launching the kernel. I reproduced a normal return with the pass disabled and process termination by `SIGFPE` with it enabled on the current head.

Please preserve the necessary execution predicate, or decline the rewrite unless the divisor expression is safe to evaluate at the host insertion point. Keeping the device fallback does not protect this host computation. Add a regression for an inactive branch whose divisor expression would otherwise divide by zero.

src/runtime/magic_div.cc
```diff
@@ -0,0 +1,54 @@
...
+// though the device will not launch any work for shape-derived divisors.
+void HostFastDivmodU32(uint32_t d, uint32_t &mul, uint32_t &shift) {
+  if (d <= 1) {
+    mul = 0;
+    shift = 0;
+    return;
+  }
+  uint32_t k = 0;
+  while ((uint32_t(1) << k) < d) {
+    ++k; // ceil_log2(d), k in [1, 31] for d >= 2
```
LeiWang1999, line 32: [P1] Reject out-of-range uint32 divisors before the shift loop

A negative scalar can reach this helper without any overflowing shape arithmetic. For `if p > 0: A[tx] = tx // p`, the rewriter proves positivity inside the branch, then emits the host helper unconditionally. Calling the kernel with `p=-1` passes `UINT32_MAX` here even though the original division would never execute.

The loop reaches `k=32` and evaluates an invalid 32-bit shift. UBSan reports `shift exponent 32 is too large for 32-bit type 'unsigned int'` at line 29. The full kernel invocation also hangs after compilation; I terminated it after a 25-second timeout.

Please return neutral constants for `d > INT32_MAX` before entering the loop, as for `d <= 1`, and add direct helper boundary tests plus an inactive-branch test with a negative scalar divisor. The device validity guard runs too late to protect the host.

src/transform/lower_magic_div.cc
```diff
@@ -0,0 +1,738 @@
...
+      for (const Call &c : g.mod_calls) {
+        replacer.replace_[c] = g.r;
+      }
+    }
+    for (const Group &g : groups) {
+      // Bind values are built by mutating the args only, so a group nested
+      // inside x refers to the already-bound inner quotient/remainder.
+      Array<PrimExpr> new_args;
+      for (const PrimExpr &arg : g.repr->args) {
+        new_args.push_back(replacer.Apply(arg));
```
LeiWang1999, line 997: [P1] Keep loop-dependent bindings inside the loop that defines their inputs

The chosen `thread_extent` contains all uses, but variables referenced by `new_args` are not necessarily defined at the beginning of its body. A dynamic serial loop is enough to trigger this:

```python
for k in T.serial(n):
    B[k, tx] = k // d
```

Here `d` is a positive dynamic buffer dimension. With the pass disabled, CUDA code generation succeeds. With it enabled, the quotient binding is emitted before the loop and code generation fails with `Find undefined Variable k`.

Please choose a legal insertion point for each group based on the definitions of all its arguments, including loop variables and local bindings, or leave the calls inline when no such point is available. Add a regression with a non-unrolled dynamic loop. Checking variable scope is also needed independently of the memory-dependence checks described in the other hoisting comment.

